PR #8, opened 2024-12-19
Fix bug in from_torch method

abhaybd:
This bug specifically causes an issue in MacOS, since the MPS framework (Apple silicon) doesn't support float64. Since the newly created tensor was being moved to MPS _before_ being casted to the target dtype (float32), the code would crash. Doing both casting and moving devices at the same time fixes this bug.

## To reproduce
On MacOS, simply run `python examples/pbd_liquid.py` and observe the following error:
```
TypeError: Cannot convert a MPS Tensor to float64 dtype as the MPS framework doesn't support float64. Please use float32 instead.
```

After this bugfix, this error no longer occurs.

### DIFF
--- a/genesis/grad/creation_ops.py
+++ b/genesis/grad/creation_ops.py
@@ -89,7 +89,7 @@ def from_torch(torch_tensor, dtype=None, requires_grad=False, detach=True, scene
         )
         requires_grad = True
 
-    gs_tensor = Tensor(torch_tensor.to(gs.device).to(dtype), scene=scene).clone()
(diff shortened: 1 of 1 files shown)


Genesis-Embodied-AI:
Thank you! Would other OS be affected by any chance?

abhaybd:
If you're asking whether this bug affects other platforms, I don't think so beyond temporarily increasing memory usage, since I don't know of any other platforms that support float32 but not float64, beyond MPS.

YilingQiao:
Thanks! Let me find a Mac and check.

zhenjia-xu:
I just tested. The issue on macOS can be solved and performance on Ubuntu won't be affected.
Thanks for the contribution!

Issue #26, opened 2024-12-19
Genesis fails silently when ffmpeg is not installed


RobRoyce:
Great project, really excited to give it a shot!

I am running Genesis on a MacBook Pro M1 Max. The installation process went fine, but I immediately started getting errors. In particular, this try/except in [rasterizer_context.py](https://github.com/Genesis-Embodied-AI/Genesis/blob/6d719541c29da17d35043d5b2bf2fc5ffaad7429/genesis/vis/rasterizer_context.py#L8-L12) is causing issues. 

### Initial Error 
Initially, the error I get is that `trimesh` is missing:

```
~/Developer/Genesis git:[main]
python3.11 scripts/main.py
objc[21098]: Class GLFWHelper is implemented in both /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/taichi/_lib/core/taichi_python.cpython-311-darwin.so (0x15baf4570) and /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/pygel3d/libPyGEL.dylib (0x318dd17c0). One of the two will be used. Which one is undefined.
objc[21098]: Class GLFWApplicationDelegate is implemented in both /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/taichi/_lib/core/taichi_python.cpython-311-darwin.so (0x15baf45c0) and /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/pygel3d/libPyGEL.dylib (0x318dd1810). One of the two will be used. Which one is undefined.
objc[21098]: Class GLFWWindowDelegate is implemented in both /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/taichi/_lib/core/taichi_python.cpython-311-darwin.so (0x15baf45e8) and /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/pygel3d/libPyGEL.dylib (0x318dd1838). One of the two will be used. Which one is undefined.
[...]

PR #27, opened 2024-12-19
Fix silent FFMPEG failure

RobRoyce:
This pull request addresses issue #26 by ensuring an error message is displayed when the `trimesh`, `pyrender`, and/or `JITRenderer` modules fail to import due to a missing `ffmpeg` library, rather than allowing the process to fail silently.

Note that this use of `try/except` is still problematic and should probably be addressed in future updates. But at least this temporary fix will keep people from spending a lot of time trying to figure out the issue.


VincentCCandela:
SOLUTION:

1) install ffmpeg on your system
2) put this in your main runtime to make sure that your code points correctly to your ffmpeg binary

import os
os.environ["IMAGEIO_FFMPEG_EXE"] = "/usr/bin/ffmpeg"

PR #35, opened 2024-12-19
Include time_outs in environment info for PPO

yang-zj1026:
In `go2_env.py`, the returned infos did not include `time_outs`, which led to the value bootstrapping in PPO not being used. As a result, `mean_action_std` kept increasing. This PR adds `time_outs` in the environment's info.


zhenjia-xu:
Thanks for pointing out!

PR #54, opened 2024-12-19
Remove torch import, keep viewer open in render_on_macos.py

do-me:
- `torch` import is not needed
- I guess most users would like to keep the viewer open to do something with the simulation. Maybe it's an idea to add an argument for this behavior where the default is to keep it open.


zhouxian:
Thank you, but this seems not something that will enhance the capability <:(

Issue #91, opened 2024-12-19
ext/pyrender/viewer: NSException when save image


NekoAsakura:
### System Information
Genesis version: 0.2.0
Operating System / Platform: MacOS Sequoia 15.2
Python version: 3.10
Tk version: 8.6.14
Hint: check Tk version by `python -c "from tkinter import Tk; print(Tk().eval('info patchlevel'))"`

### How to reproduce

```
python examples/render_on_macos.py -v
```

When Viewer created, press "s" to save image. Then program crashed with:

```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[NSApplication macOSVersion]: unrecognized selector sent to instance 0x3661c4ab0'
...
libc++abi: terminating due to uncaught exception of type NSException
zsh: abort      python examples/render_on_macos.py -v
```

### Detailed description

Based on information from internet, it seems to be caused by Tkinter.
Few bugs were reported from Python 2.7 to 3.9, which should be fixed in Python 3.10 w/ Tk 8.6.11, according to this [announcement](https://www.python.org/download/mac/tcltk/). No similar bug reports have been noted for Tk 8.6.14.

Ref:
https://core.tcl-lang.org/tk/tktview/85f316beb15108ac43b03fa6c8608e31f3ae5f92
https://bugs.python.org/issue35209

PR #92, opened 2024-12-19
[BUG FIX] ext/pyrender/viewer: resolve macOS crashes when initialising Tk()

NekoAsakura:
This PR should resolve [Issue#91](https://github.com/Genesis-Embodied-AI/Genesis/issues/91).
On macOS, a Tk instance should be initiated before creating the viewer.

### DIFF
--- a/genesis/ext/pyrender/viewer.py
+++ b/genesis/ext/pyrender/viewer.py
@@ -16,6 +16,8 @@ import genesis as gs
 
 from tkinter import Tk
 from tkinter import filedialog
+root = Tk()
+root.withdraw()
 
 import pyglet
 from moviepy.video.io.ffmpeg_writer import FFMPEG_VideoWriter
@@ -961,7 +963,6 @@ class Viewer(pyglet.window.Window):
         }
         filetypes = [file_types[x] for x in file_exts]
         try:
-            root = Tk()
             save_dir = self.viewer_flags["save_directory"]
             if save_dir is None:
                 save_dir = os.getcwd()
@@ -971,7 +972,6 @@ class Viewer(pyglet.window.Window):
         except Exception:
             return None
 
-        root.destroy()
         if filename == ():
             return None
         return filename
--- a/genesis/ext/pyrender/viewer.py
+++ b/genesis/ext/pyrender/viewer.py
@@ -14,15 +14,8 @@ import OpenGL
 
 import genesis as gs
 
-try:
-    from Tkinter import Tk
-    from Tkinter import tkFileDialog as filedialog
-except Exception:
-    try:
-        from tkinter import Tk
-        from tkinter import filedialog as filedialog
-    except Exception:
-        pass
+from tkinter import Tk
+from tkinter import filedialog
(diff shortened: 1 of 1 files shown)


genesis/ext/pyrender/viewer.py
```diff
@@ -14,15 +14,8 @@
...
 
-try:
-    from Tkinter import Tk
-    from Tkinter import tkFileDialog as filedialog
-except Exception:
-    try:
-        from tkinter import Tk
-        from tkinter import filedialog as filedialog
-    except Exception:
-        pass
```
NekoAsakura, line 25: The purpose of this code is to ensure compatibility with Tkinter, supporting both Python 2 and Python 3. Genesis requires Python 3.9 or higher, such compatibility adjustments are unnecessary.

https://github.com/Genesis-Embodied-AI/Genesis/blob/2985aa9c63b4162cea10a52c98161e1270477746/pyproject.toml#L12

genesis/ext/pyrender/viewer.py
```diff
@@ -16,6 +16,8 @@
 
 from tkinter import Tk
 from tkinter import filedialog
+root = Tk()
+root.withdraw()
```
NekoAsakura, line 20: Hide root window as we only use `filedialog` to save image.

zhouxian:
Thank you for this!!!! Could you test and make sure it doesn't interfere with other os? if not we can merge

NekoAsakura:
I have tested viewer on macOS 15.2 and Ubuntu 22.04 w/ cu118, and it worked as expected. It should also function correctly on Windows, but I don't have a testing platform to confirm this.

To test it, run any script containing `genesis.Scene(...).viewer.start()`. When viewer window appears, press "s" to save an image.

The use of `root.withdraw()` can also be found in CPython:
https://github.com/python/cpython/blob/09d15aa9a87f69867e69f00d0e9671d70a4a49c5/Lib/tkinter/__init__.py#L329-L343

So I assume it works well across multiple platforms.

zhouxian:
@zhenjia-xu could you take care of this?

NekoAsakura:
@zhouxian I have made a test on Windows 10 22H2 19045.5247 and it was passed. I believe it's safe to merge, and it shouldn't cause any issues with other supporting os.

zhouxian:
Thank you:)

zhouxian:
Hi, it seems this caused issue on linux via SSH, and pyrender will not be able to be imported (works fine in local terminal, but fails when doing this via ssh)

Could you fix this? it seems a bit urgent.
For now, i'll add a simple try except to the two lines you added

NekoAsakura:
Oops, apologies for any trouble may caused. 
I saw https://github.com/Genesis-Embodied-AI/Genesis/issues/324. This is because tkinter should not be initialised on a headless server unless an X window is configured with a virtual screen.
try-except block should be enough for this, as they are not expected to use a viewer on a headless server.

zhouxian:
Ok thanks!

PR #141, opened 2024-12-20
Add badges and update Chinese README

00make:
Enhance the README with visually appealing badges and update the Chinese version for better accessibility.


zhouxian:
Thank you! looks nicer indeed

PR #179, opened 2024-12-20
 [BUG FIX] Sloved the conflict that both taichi_python and pygel3d libraries has duplicate class names on MacOS (M1) by changing  "PyGEL3D == 0.1.0"

VioletBenin:


zhouxian:
We didn't want to put torch in the pyproject toml, because torch version depends on user's specific system and is quite flexible. What do you think?

Esmebgch (approved):

PR #186, opened 2024-12-21
Fix:Remove ANSI sequences and format issues in RBC.__repr

JonnyDing:
### Problem description
ANSI escape sequences (color and format control characters) used in the `__repr__` method may appear garbled in unsupported environments in debugging or terminal output, affecting the readability of debugger output or logging.

### Problem
- The console of a debugger such as PyCharm or VSCode does not parse ANSI escape sequences correctly.
- Attribute values are displayed inconsistently across multiple lines, with some members being displayed as “gibberish”.
- Some property values may throw exceptions when accessed, resulting in no display.
before

### Changes
Determine whether the current environment is a debug environment, so as to solve the problem of ANSI garbled code in the debug environment.

**Example of modified output**:
```text
<gs.Scene> t=<int>: 0 dt=<float>: 0.02 requires_grad=<bool>: False substeps=<int>: 2 visualizer=<gs.
![after](https://github.com/user-attachments/assets/b8d3a6d9-d9f3-4ccf-988c-7f7e6d607545)

### DIFF
--- a/genesis/datatypes.py
+++ b/genesis/datatypes.py
@@ -42,10 +42,10 @@ class List(list, RBC):
     def _repr_elem_colorized(self, elem, common_length=0):
         content = self._repr_elem(elem, common_length)
         idx = content.find(">")
-        formatted_content = f"{colors.BLUE}{formats.ITALIC}{content[:idx+1]}{formats.RESET}{content[idx+1:]}"
+        formatted_content = f"{colors.BLUE}{formats.ITALIC}{content[:idx + 1]}{formats.RESET}{content[idx + 1:]}"
         idx = formatted_content.find(":")
         if idx >= 0:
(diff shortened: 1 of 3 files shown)


JonnyDing:
### Problem description
ANSI escape sequences (color and format control characters) used in the `__repr__` method may appear garbled in unsupported environments in debugging or terminal output, affecting the readability of debugger output or logging.

### Problem
- The console of a debugger such as PyCharm or VSCode does not parse ANSI escape sequences correctly.
- Attribute values are displayed inconsistently across multiple lines, with some members being displayed as “gibberish”.
- Some property values may throw exceptions when accessed, resulting in no display.
before

### Changes
Determine whether the current environment is a debug environment, so as to solve the problem of ANSI garbled code in the debug environment.

**Example of modified output**:
after

zhouxian:
Looks useful! @zswang666 could you double check and merge?

Issue #181, opened 2024-12-21
Simulation Speed Benchmark likely has significant issues leading to overstated numbers


StoneT2000:
Just some important context, I have went through the Genesis code base and benchmarks and found some issues that likely lead to inflated benchmark speeds.

I've written a full report on it here: https://x.com/stone_tao/status/1870243004730225009?s=46&t=LBFTca4dqDdDCjhzaM56tA

The main issue in question is why the original speed benchmark 
1. uses the fastest physics setting when other tutorial/example code don't
2. takes one action followed by 999 steps of no actions (which likely means the rigid body solver is early exiting or something most of the time leading to faster speed?)
3. disables robot self collisions

![image](https://github.com/user-attachments/assets/06ab5e1a-6618-4448-9826-80658c196453)

Sharing it here so people can see and verify numbers themselves on at least the Genesis only benchmark. I look forward to improvements that address these issues! Genesis still stands as an impressive effort to integrate many physics solvers into one package.

ziyanx02:
Thank you for bringing this up. To show the results for both benchmarking tasks, here is the robot dog result extracted from the comment in your code.

|Original|*+ more sim accuracy*|+ self collision|+ random action|
|--|--|--|--|
|14.5M|8M|5.5M-7.5M|1.2M|\

Allow me to provide some explanation within the context of the locomotion. 
1. *+ more sim accuracy* here is simply setting substeps=2. However, for locomotion tasks such as backflip, we use dt=0.005, substep=1,  decimation=4, which are widely used in other locomotion works using IsaacGym. We successfully deployed the first double backflip policy obtained in the simulation, so there is no fidelity issue (at least) in locomotion setting. From my perspective of view, setting substeps=2 only causes the speed test results to decrease by half. 
2. In the locomotion example, we use substeps=2 because there is no need to introduce decimation because control_dofs_pos in Genesis will update the applied torques in each substep.

By the way, here are my results from a 6000Ada:

|Original|substeps=2|+ self collision|+ random action|
|--|--|--|--|
|16.4M|-|15.9M|6M|
|16.4M|8M|7.8M|3.4M|

It is notable that the decreasing proportion varies significantly, possibly due to differences in system environments. Therefore, it's too early to draw a conclusion at this point.

StoneT2000:
[...]

PR #213, opened 2024-12-21
Simulation Speed Benchmark likely has significant issues leading to overstated numbers #181

sahil3471:
Encapsulation of setup and teardown: The setUpClass method initializes the environment and scene setup, while tearDownClass ensures proper cleanup by calling gs.destroy().

assertEqual and assertLess: Used assertEqual for checking basic sums and assertLess for comparing the maximum difference between computed and reference positions.

Descriptive function and variable names: The test methods are clearly named, and variables have been given more descriptive names where necessary.

Removed hardcoded n_envs = 0: This value could be dynamically set depending on the test's needs.

### DIFF
--- a/tests/test_rigid.py
+++ b/tests/test_rigid.py
@@ -1,18 +1,14 @@
-import subprocess
 import unittest
 import numpy as np
 import genesis as gs
 
 
 class TestRigid(unittest.TestCase):
-
-    def test_simple(self):
-        add = 1 + 1
-        assert add == 2
-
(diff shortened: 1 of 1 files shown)


zhouxian:
Unfortunately this is not the common practice when testing speed... Compilation time should be removed, as other platforms  such as MJX and isaac sim also has a long compilation and loading time :) Also, the unittest is for monitoring performance change when we merge pr, not for benchmarking

PR #247, opened 2024-12-22
[MISC] add CONTRIBUTING.md

ziyanx02:
Add CONTRIBUTING.md.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,75 @@
+# Contributing to Genesis
+
+Thank you for your interest in contributing to Genesis! We welcome contributions from everyone. Please take a moment to review this guide to ensure a smooth collaboration.
+
+- [Reporting Bugs](#reporting-bugs)
+- [Suggesting Features](#suggesting-features)
+- [Submitting Code Changes](#submitting-code-changes)
+- [Reviewing and Merging](#reviewing-and-merging)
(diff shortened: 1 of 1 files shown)


YilingQiao (approved):

YilingQiao:
Hi, could you please test if the pre-commit hook is working as expected? I want to set the line width to 120, but it seems to format everything with a smaller value.

ziyanx02:
> Hi, could you please test if the pre-commit hook is working as expected? I want to set the line width to 120, but it seems to format everything with a smaller value.

I will try it tomorrow

zswang666:
Should we add guide for test/check in CI on local machine?
```
black --check .
black . # to auto-format if the above check failed
python -m unittest discover tests
```

ziyanx02:
Seems those CI tests require >6G gpu memory so I will add local tests as optional.

As for auto-reformat, I got 19 files to be reformatted after running `black --check .` XD. That doesn't seem right, does it?

ziyanx02:
> Hi, could you please test if the pre-commit hook is working as expected? I want to set the line width to 120, but it seems to format everything with a smaller value.

I can't find a counter case. An extremely long string or name won't be truncated even longer than 120. Long numerical expressions will be divided into parts as short as possible I guess?

ziyanx02:
The final version is
- using pre-commit to avoid forgetting to check.
- Suggesting to do local CI tests.

PR #284, opened 2024-12-24
fix asset path

zhenjia-xu:


zhenjia-xu:
Since this is an asset issue and not related to code, I skipped CI and merged directly into the main branch.
I will strictly follow the contribution guidelines next time.

PR #332, opened 2024-12-26
[BUG FIX] Fix for dockerfile error

skurtyyskirts:
Windows-style CRLF line endings in the build script causing compatibility issues in the Linux container. Added dos2unix package installation to handle Windows/Unix line ending conversion.


zhouxian:
@Kashu7100

skurtyyskirts:
I should add, this relates to a current open issue https://github.com/Genesis-Embodied-AI/Genesis/issues/326

Kashu7100:
@skurtyyskirts I am wondering if I run dos2unix outside the Docker on the build_luisa.sh script, can we avoid the installation of dos2unix in the Docker? I don't have any Windows machine available for now so cannot test it quickly.

skurtyyskirts:
Oh that's a good question, if anyone see's this ticket and is available to test it would be much appreciated, I'm away from the home pc for a bit :(

skurtyyskirts:
My apologies for the delay, actually I believe this can be solved with a simpler solution I updated my Git configuration to handle the necessary line endings correctly by running `git config --global core.autocrlf input`. I then deleted my local Genesis repository, and then made another gitclone. After doing this, the Luisa was able to build no problem. This seems to be a windows exclusive issue as it's only the Windows-style CRLF line endings.

Having done this and completed a successful build_luisa.sh, I suggest adding a .gitattributes to the repository that contains
```
 # Force LF line endings for shell scripts
*.sh text eol=lf

# Force LF line endings for configuration files
*.yml text eol=lf
*.yaml text eol=lf
*.json text eol=lf
*.conf text eol=lf

# Force LF for Docker-related files
Dockerfile text eol=lf
*.dockerignore text eol=lf
```

This will ensure no matter what operating system the user is running, the executables and builds will always be successful. Adding this will also make it so the user doesn't have do the process of `git config --global core.autocrlf input` and rebuilding a fresh repository, it'll come configured for all users. I'll create a pull request to add the .gitattributes

skurtyyskirts:
My apologies, I pushed the pull request to this ticket.  I've created a new one for cleaner documentation purposes if needed. https://github.com/Genesis-Embodied-AI/Genesis/pull/532

Kashu7100:
Thank you! Looks good to me! Let's move to #532

PR #359, opened 2024-12-27
[FEATURE] Add Unitree G1 Walking Demo

0nhc:
I think it's cool to have another humanoid RL demo, so I added an RL demo for Unitree G1 walking in this PR. Here's my modification:

* Added Unitree G1's urdf under [assets/urdf](https://github.com/Genesis-Embodied-AI/Genesis/tree/main/genesis/assets/urdf), with original BSD-3 License.

* Added 3 Python scripts (g1_env.py, g1_train.py, g1_eval.py) under [examples/locomotion](https://github.com/Genesis-Embodied-AI/Genesis/tree/main/examples/locomotion).

Demo video after training for 69.5 seconds on my PC:

https://github.com/user-attachments/assets/f8c37f61-1101-407e-81ac-a9b53c969927


zhouxian:
Cool! @ziyanx02 can you review this?

0nhc:
Actually I borrowed some reward functions from [https://github.com/unitreerobotics/unitree_rl_gym](https://github.com/unitreerobotics/unitree_rl_gym), which is under BSD-3 License. So I added another commit to mention the source in each function's docstring.

ziyanx02:
Looks impressive!
I have two questions before merging:
- Have you tried sim-to-sim transfer with the policy you trained?
- The motion appears slightly different from Unitree RL Gym. Could you try to make the upper body more stable?

0nhc:
Thanks!

For the 2 questions.
* I haven't tried sim2sim. Actually I just made this demo work few hours ago.
* To solve this, I have added more reward functions according to [unitree_rl_gym](https://github.com/unitreerobotics/unitree_rl_gym) to make the upper body more stable. I found the functions work indeed. However, the original code in [unitree_rl_gym](https://github.com/unitreerobotics/unitree_rl_gym) trained for 10,000 iterations, but this demo only trained 100 iterations. I think training time is also an important factor.

I have pushed latest code with more reward functions.

0nhc:
Latest video after training for 64 seconds:

https://github.com/user-attachments/assets/f2eb46a5-dd78-428f-b85f-0b953e0d22a1

ziyanx02:
The training code does not include domain randomization or observation noise, which might explain why it requires a much shorter training time. However, the current results show no sign of possibly successful deployment. Improving the motion and trying sim-to-sim transfer or directly deploying it to the real world would be better.

0nhc:
I totally understand your consideration. Because it's hard for me to acquire a G1 hardware, I think the most feasible way for me to verify the policy is Sim2Sim. Also, I'll see how to make it walk better. Thanks.

erwincoumans:
I wonder why it was closed? I have the G1 hardware and am willing to try it out, with additional randomization.

0nhc:
Hi @erwincoumans! I closed this issue because I recently returned to university and was worried I wouldn't have enough time to continue working on it. However, I’ve already added domain randomization and observation noise to my forked repo, as well as another MuJoCo environment to test the trained policy. While the policy works well in Genesis after training, it crashes in MuJoCo for reasons I haven’t been able to identify. The Sim2Sim transfer isn’t functioning as expected.

I’m happy to reopen the issue, and if you’re able to test this pipeline on a real robot, that would be amazing!

miguelalonsojr:
I developed a similar version. I can PR it if anyone is interested. It uses the full dof G1. I also did the same with the H1. Here's a video: https://bsky.app/profile/miguelalonsojr.bsky.social/post/3lez5qcpe5k26

0nhc:
Hi @miguelalonsojr, I think we have the same problem: it is not enough to only make it work in Genesis. We have to prove the policy can be transferred to another simulator (Sim2Sim) or the real robot (Sim2Real). This usually requires some techniques such as domain randomization and observation noise.

miguelalonsojr:
@0nhc I think it's ok not to model realistic observations or add domain randomization to have a humanoid walking example in the repo. The quadruped example, for instance, doesn't model sensor noise nor domain randomization. I think these are just starter examples and not really meant to be completely deployable. I wouldn't expect to be able to deploy a locomotion policy directly from the examples.

NirajPudasaini:
@miguelalonsojr can you please create a PR for the H1 as well ?

0nhc:
This feature has already been solved by [HumanoidVerse](https://github.com/LeCAR-Lab/HumanoidVerse), PR closed.

PR #367, opened 2024-12-28
[FEATURE] Add Reset Scene and Stop Rendering function and fix black formatting

JonnyDing:
### This PR introduces two new key controls:

- Pressing Space toggles the rendering on/off.
- Pressing X resets the scene.
## ScreenShot
image

**when I stop rendering ,the training time will be optimized a lot** 

### Before stop rendering

image

### After stop rendering

image


YilingQiao:
Thanks for your work! I will try this feature soon!

duburcqa:
I'm not sure about this PR. What is the point of being able to reset the scene with the viewer using a keyboard shortcut ? Besides, enabling the interactive viewer during learning sounds like a bad practice in the first place.

neo-alex:
I'm not sure either about the "scene reset", however I definitely see value in the ability to toggle rendering on/off at any time, especially when training (e.g. with RL): it enables high-speed training most of the time (with rendering off) while allowing some sneak peeks at the robot behaviour from time to time to visualize learning progress (with rendering on).

By the way, Isaac Lab has this capability to change [RenderMode](https://isaac-sim.github.io/IsaacLab/main/source/api/lab/isaaclab.sim.html#isaaclab.sim.SimulationContext.RenderMode) during training (via Isaac Lab tab in Isaac Sim UI) between:
- NO_RENDERING (no viewport nor camera renders)
- PARTIAL_RENDERING (just camera renders, e.g. for visual-based RL)
- FULL_RENDERING (viewport & camera renders)

So, please don't discard this nice feature too quickly :)

PR #409, opened 2024-12-30
[MISC] Added README_TR

ituperceptron:


ziyanx02:
Thank you for your contribution! However, we've decided not to accept additional READMEs, as their number would become overwhelming if we provided one for each language.

PR #424, opened 2024-12-31
[FEATURE] merge entity

Kashu7100:
You can combine two rigid entities in a scene (can be a mixture of URDF and MJCF!). 
Use case: attach a gripper to an arm, attach a manipulator to a quadruped, etc.

Example use case (Panda + Figure hand): 
image


ziyanx02:
Well done! This will solve most of the problems! But there are some small issues here:
1. I have to change 
`"genesis/assets/urdf/panda_bullet/panda_nohand.urdf"` to `"urdf/panda_bullet/panda_nohand.urdf"`
`"genesis/assets/urdf/panda_bullet/hand.urdf"` to `"urdf/panda_bullet/hand.urdf"`
to make the example work.
2. Once the link of the child entity assigned to merge is not the root link, changing this link's parent link will result in the child entity breaking up into two pieces. But this could be a quick fix for mobile manipulation stuff.

examples/rigid/merge_entities.py
```diff
@@ -0,0 +1,157 @@
...
+    ########################## entities ##########################
+    plane = scene.add_entity(
+        gs.morphs.Plane(),
+    )
+
+    if args.comb == "urdf2urdf" or args.comb == "urdf2mjcf":
+        gs.logger.info("loading URDF panda arm")
+        franka = scene.add_entity(
+            gs.morphs.URDF(
+                file="genesis/assets/urdf/panda_bullet/panda_nohand.urdf", merge_fixed_links=False, fixed=True
```
ziyanx02: Maybe double-check the relative path. It’s not working locally on my machine.

ziyanx02 (changes requested):

examples/rigid/merge_entities.py
Kashu7100: Oh, nice catch! This is my dev env problem. It's always a good idea to have a review!!

Kashu7100:
Ok, fixed the pointed part. 

1. fixed path to genesis assets 
2. set default of child link to its base and when user assigns non-base link, it will warn in logger

ziyanx02 (approved):

Issue #262, opened 2024-12-23
Adding user-defined IK chains


Hardy-cool:
For a humanoid robot, when I only want to solve the inverse solution of the arm part, the current solver can only solve the inverse solution of the entire robot link chain, or fix the position outside the arm as the current position. Can this part be optimized to make it more convenient to use?

Genesis-Embodied-AI:
Sounds like a very helpful feature! -- if possible, do you think you can help us with this? Implementation should be straightforward, just adding some mask when computing the jacobian udpate. We are really in a shortage of manpower :( If it's not urgent, we will try to add this in the next version bump

Hardy-cool:
sure, I'll try the way you said

PR #430, opened 2025-01-01
[feature] support custom ik chain

zswang666:
In `rigid_entity.inverse_kinematics_multilink`, we can specify multiple targets for IK with arguments `links`, `poss`, and `quats`.  This PR add a new argument `dofs_idx_local` in `rigid_entity.inverse_kinematics` and `rigid_entity.inverse_kinematics_multilink`. This argument allows to specify which dofs are to involved in the IK optimization.

High-level changes:
- (no change) we still compute the jacobian for all dofs since I haven't found a way to cleanly do this -- ti.func (`_func_get_jacobian`) does not take varying length array like `links_idx_by_dofs` (of length `n_links_by_dofs`) otherwise we can do something similar to the init q resampling.
- In the LM algorithm, we do all computations with the specified dofs only.
- When resampling initial q, we only consider the specified dofs.

Related issues: https://github.com/Genesis-Embodied-AI/Genesis/issues/262

### DIFF
--- a/examples/rigid/ik_custom_chain.py
+++ b/examples/rigid/ik_custom_chain.py
@@ -1,0 +1,89 @@
+import argparse
+
+import numpy as np
+
+import genesis as gs
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-v", "--vis", action="store_true", default=False)
+    args = parser.parse_args()
+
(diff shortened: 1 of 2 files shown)


YilingQiao (approved):
I can run the demo and it looks reasonable.

PR #431, opened 2025-01-01
Fix AttributeError in Trackball class

Likhithsai2580:
Fix the AttributeError in the `Trackball` class by initializing `_pdown` and handling its usage correctly.

* Initialize `_pdown` attribute to `None` in the `__init__` method.
* Modify the `drag` method to handle the case where `_pdown` is not set by defaulting to the current point.
* Update the `drag` method to interpret drag as a roll about the camera axis and handle different states (rotate, roll, pan) correctly.

### DIFF
--- a/genesis/ext/pyrender/trackball.py
+++ b/genesis/ext/pyrender/trackball.py
@@ -46,6 +46,8 @@ class Trackball(object):
 
         self._state = Trackball.STATE_ROTATE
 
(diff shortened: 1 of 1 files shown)


YilingQiao:
Hi @Likhithsai2580 , thanks for your work! Could you please provide a test script for this change? What's its use case?

YilingQiao (approved):

PR #449, opened 2025-01-02
Fix Dockerfile to properly execute build_luisa.sh

serg-yalosovetsky:
## Description
This PR fixes the issue with the `Dockerfile` where the `build_luisa.sh` script was not being found during execution. 

### Changes
- Ensured the script is executed explicitly using `/bin/bash`.
- Added debugging steps to verify the script's presence and permissions during the build process.
- Confirmed the correct working directory for script execution.

### Steps to Test
1. Build the Docker image:
   ```bash
   docker build -t genesis -f docker/Dockerfile docker


Kashu7100:
When does the current Dockerfile fails to compile? Also, if there is a path issue, do you need to add the `/workspace` to the `chmod` part as well?

PR #455, opened 2025-01-03
Update evaluation command in example script for backflip environment

takeshi8989:
Updated the evaluation command in the script comment to use the correct file (go2_backflip.py) and argument (-e single) for the backflip environment. The previous command referenced an unrelated script and arguments, which could confuse users. Verified that the updated command works as intended.


ziyanx02:
Thanks for your PR! I've directly updated the main branch.

PR #477, opened 2025-01-05
[BUG FIX] update RigidEntity.get_link() as id no longer exists in RigidLink

ziyanx02:
Previous:
```RigidEntity.get_link(name: str, id: str)``` but link.id no longer exists.

Updated:
```RigidEntity.get_link(name: str, idx: int)``` search link by global idx.


zhouxian:
No it’s uid,  not idx

idx can simply use entity.links[id]

ziyanx02:
Update both `get_joint` and `get_link`

PR #482, opened 2025-01-06
[Bug Fix] avoid changing input link index in rigid_entity

woshialex:
RigidEntiry set_mass_shift and set_COM_shift changed the function parameters link_indicies and we normally don't expect so.

create a temporary list instead so it does not affect the input parameters:

        global_link_indices = [self._link_start + i for i in link_indices]
        self._solver.set_links_COM_shift(com_shift, global_link_indices, envs_idx)

### DIFF
--- a/examples/rigid/domain_randomization.py
+++ b/examples/rigid/domain_randomization.py
@@ -46,15 +46,15 @@ def main():
     ########################## domain randomization ##########################
     robot.set_friction_ratio(
         friction_ratio=0.5 + torch.rand(scene.n_envs, robot.n_links),
-        link_indices=np.arange(0, robot.n_links),
+        ls_idx_local=np.arange(0, robot.n_links),
     )
     robot.set_mass_shift(
         mass_shift=-0.5 + torch.rand(scene.n_envs, robot.n_links),
-        link_indices=np.arange(0, robot.n_links),
+        ls_idx_local=np.arange(0, robot.n_links),
     )
     robot.set_COM_shift(
         com_shift=-0.05 + 0.1 * torch.rand(scene.n_envs, robot.n_links, 3),
-        link_indices=np.arange(0, robot.n_links),
+        ls_idx_local=np.arange(0, robot.n_links),
     )
 
     joint_names = [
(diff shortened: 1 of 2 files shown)


ziyanx02:
Thank you for pointing this out! Could you please also rewrite `set_friction_ratio`, `set_mass_shift`, and `set_COM_shift` similarly to `set_links_inertial_mass`, using `ls_idx_local` as the input and `self._get_ls_idx` to convert local indices to global indices?

woshialex:
Thanks. I updated set_mass_shift and set_COM_shift to use _get_ls_idx. 

set_friction_ratio itself has no problem and it also can't use _get_ls_idx since its offset is based on _geom_start, not link_start. It might worth to change the function name (given there is a set_friction api already).

Kashu7100:
I think we should keep the API the same, i.e., keep `link_indices` instead of `ls_idx_local`.
@ziyanx02 what do you think?

woshialex:
The API's argument naming are not very consistent in the same class. Either call the argument link_indices everywhere or  ls_idx_local in all these APIs. I have no preference (or links_idx_local? :D ), someone pick a conventions and I'll stick to it.

ziyanx02:
> I think we should keep the API the same, i.e., keep `link_indices` instead of `ls_idx_local`. @ziyanx02 what do you think?

It seems other APIs in `RigidEntity` are using `ls_idx_local` so I think it's better to follow it. By the way, in `RigidSolver` it's `links_idx` instead of `link_indices`. Only code for locomotion and domain randomization uses `link_indices` (I can't remember but perhaps that's because locomotion code is adapted from legged_gym in which they use `link_indices`).

Could you also update the API usage in examples/rigid/domain_randomization.py? I really appreciate your help.

Kashu7100:
sounds good to me.

woshialex:
@ziyanx02 Thanks. Updated the example code.  It's my bad, I should have updated all the places using the APIs when making changes to APIs.

ziyanx02:
Thanks, everyone! I will merge this PR once the CI tests are completed.

Issue #507, opened 2025-01-08
[Bug Report] [Deprecated Code] NotImplementedError: `interp2d` has been removed in SciPy 1.14.0.


CloudRipple:
当我运行 `Genesis/examples/rigid/terrain_subterrain.py` 这个文件的时候，出现了一个报错：
When I was running Genesis/examples/rigid/terrain_subterrain.py , an error was raised:

```bash
Traceback (most recent call last):
  File "/home/cloud/Genesis/examples/rigid/terrain_subterrain.py", line 70, in <module>
    main()
  File "/home/cloud/Genesis/examples/rigid/terrain_subterrain.py", line 39, in main
    terrain = scene.add_entity(
              ^^^^^^^^^^^^^^^^^
  File "/home/cloud/Genesis/genesis/utils/misc.py", line 38, in wrapper
    return method(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/cloud/Genesis/genesis/engine/scene.py", line 359, in add_entity
    entity = self._sim._add_entity(morph, material, surface, visualize_contact)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/cloud/Genesis/genesis/engine/simulator.py", line 134, in _add_entity
    entity = self.rigid_solver.add_entity(self.n_entities, material, morph, surface, visualize_contact)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/cloud/Genesis/genesis/engine/solvers/rigid/rigid_solver_decomp.py", line 64, in add_entity
    entity = entity_class(
             ^^^^^^^^^^^^^
  File "/home/cloud/Genesis/genesis/engine/entities/rigid_entity/rigid_entity.py", line 68, in __init__
    self._load_model()
  File "/home/cloud/Genesis/genesis/engine/entities/rigid_entity/rigid_entity.py", line 90, in _load_model
    self._load_terrain(self._morph, self._surface)
  File "/home/cloud/Genesis/genesis/engine/entities/rigid_entity/rigid_entity.py", line 301, in _load_terrain
[...]

PR #508, opened 2025-01-08
[Bug Fix] Modify the deprecated interp2d method into the supported RegularGridInterpolator method

CloudRipple:
将不受新版本`SciPy`支持的方法`interp2d`，修改为受支持的`RegularGridInterpolator`方法
Modify the method interp2d, which is not supported by the new version of SciPy, into the supported RegularGridInterpolator method.

issue #507

具体修改如下：
```python
    # f = interpolate.interp2d(y, x, height_field_downsampled, kind="linear")
    f = interpolate.RegularGridInterpolator((y, x), height_field_downsampled, method="linear")

    x_upsampled = np.linspace(0, terrain.width * terrain.horizontal_scale, terrain.width)
    y_upsampled = np.linspace(0, terrain.length * terrain.horizontal_scale, terrain.length)
    # z_upsampled = np.rint(f(y_upsampled, x_upsampled))
    z_upsampled = np.rint(f((y_upsampled, x_upsampled)))
```

### DIFF
--- a/genesis/ext/isaacgym/terrain_utils.py
+++ b/genesis/ext/isaacgym/terrain_utils.py
@@ -77,11 +77,13 @@ def random_uniform_terrain(
     x = np.linspace(0, terrain.width * terrain.horizontal_scale, height_field_downsampled.shape[0])
     y = np.linspace(0, terrain.length * terrain.horizontal_scale, height_field_downsampled.shape[1])
 
-    f = interpolate.interp2d(y, x, height_field_downsampled, kind="linear")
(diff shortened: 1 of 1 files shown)


zhenjia-xu:
Thanks for fixing this!

Kashu7100:
I think we need to add the scipy requirements to `pyproject.toml`.

Nate711:
Thanks, I needed this change to make the terrain work with scipy 1.15. I also agree the pyproject.toml should be updated. Although it's a separate issue, open3d also needs to be installed.

Kashu7100:
Addressed in #680, thanks!!

PR #518, opened 2025-01-09
[BUG FIX] Fix the dimensional issue in the transformation functions.

zhenjia-xu:
- fix the dimension issue in transform_quat_by_quat and transform_by_quat.
- remove transform_by_quat_yaw, which is not used at all.


Kashu7100:
@zhenjia-xu could you keep `transform_by_quat_yaw` function?

PR #532, opened 2025-01-10
[BUG FIX] Add .gitattributes file for line endings handling

skurtyyskirts:
When shell scripts have Windows style CRLF line endings, Unix systems may not recognize them as valid executible files. The core.autocrlf input setting tells Git to convert CRLF to LF, and keep LF as is. Fixes errors building dockers, and addresses the build_luisa.sh errors.


Kashu7100:
previous context in #332

Issue #567, opened 2025-01-14
[FEATURE] Add Pytorch 2.5+ Intel XPU support


lgleim:
Intel GPUs are capable of accelerating simulation via the (currently broken, cf https://github.com/Genesis-Embodied-AI/Genesis/issues/274) Vulkan backend but the current implementation will allocate PyTorch tensors on the CPU, causing avoidable IO overhead.

Instead, the PyTorch XPU device should be used (instead of CPU) when available for the Vulkan genesis backend.

As of PyTorch 2.5, [support for acceleration on Intel XPUs was added](https://pytorch.org/docs/stable/notes/get_start_xpu.html).

Issue #500, opened 2025-01-07
Assertion failed: (p != nullptr), function bind_pipeline, file metal_device.mm, line 382.


ZeroSaturn:
https://github.com/user-attachments/assets/5705adc1-13f1-48f1-ad2e-82985266911f

MAC OS M1

I have been adjusting the code (CHATGPT) to allow for rendering some of the example files since the viewers not supported on Mac, I got elastic dragon and pbd_liquid to work, i linked my dragon and liquid render above, but the code im attatching below produces the error , its originaly from  pbd_cloth. I've seen this error posted before but I hope that I could possibly give some more insight into the cause of the problem, 

Here is the code I ran to get the error :

```
import argparse
import numpy as np
import genesis as gs


def main():
    parser = argparse.ArgumentParser()
    parser.add_argument("-v", "--vis", action="store_true", default=False)
    parser.add_argument("-c", "--cpu", action="store_true", default=False)
    args = parser.parse_args()

    ########################## init ##########################
    gs.init(backend=gs.cpu if args.cpu else gs.gpu, logging_level="debug")

    ########################## create a scene ##########################

    scene = gs.Scene(
        sim_options=gs.options.SimOptions(
            substeps=10,
            gravity=(0, 0, -9.8),
        ),
        viewer_options=gs.options.ViewerOptions(
            camera_pos=(2, 2, 1.5),
            camera_lookat=(0, 0, 0.5),
            camera_up=(0, 0, 1),
        ),
        show_viewer=args.vis,
    )

    ########################## materials ##########################
    mat_elastic = gs.materials.PBD.Elastic()

    ########################## entities ##########################

    bunny = scene.add_entity(
        material=mat_elastic,
        morph=gs.morphs.Mesh(
            file="meshes/dragon/dragon.obj",
            scale=0.003,
            pos=(0, 0, 0.8),
        ),
        surface=gs.surfaces.Default(
            # vis_mode='recon',
        ),
    )

    ########################## add camera ##########################
    cam = scene.add_camera(
        res=(1280, 720),  # You can adjust the resolution as needed
        pos=(2, 2, 1.5),
        lookat=(0, 0, 0.5),
        fov=60,  # Field of view in degrees
        GUI=False,  # Disable GUI for recording
    )

    ########################## build ##########################
    scene.build()

    ########################## start recording ##########################
    cam.start_recording()

    ########################## run simulation and record ##########################
    horizon = 150
    for i in range(horizon):
[...]

PR #568, opened 2025-01-14
[BUG FIX][FEATURE] Rework backend & device selection logic

lgleim:
Fixes https://github.com/Genesis-Embodied-AI/Genesis/issues/274 - Vulkan backend on non-cuda GPUs 
Resolves https://github.com/Genesis-Embodied-AI/Genesis/issues/567 - Adds Pytorch 2.5+ Intel XPU support

Merges the logic of get_gpu_device() and get_cpu_device() to handle device selection in a single place.

If a specific backend is requested, this request is honored.

The vulkan backend will use intel's pytorch xps device if available, reducing cross-device data transfer.

The `gpu` backend will
- default to `cuda` when available,
- resolve to `mps` on macOS
- or fallback to `vulkan`.

### DIFF
--- a/genesis/utils/misc.py
+++ b/genesis/utils/misc.py
@@ -98,45 +98,44 @@ def get_cpu_name():
 
 
 def get_device(backend: gs_backend):
-    match backend:
-        case gs_backend.cuda:
-            if not torch.cuda.is_available():
-                gs.raise_exception("cuda device not available")
+    if backend == gs_backend.cuda:
+        if not torch.cuda.is_available():
+            gs.raise_exception("cuda device not available")
 
-            device = torch.device("cuda")
-            device_property = torch.cuda.get_device_properties(0)
+        device = torch.device("cuda")
+        device_property = torch.cuda.get_device_properties(0)
+        device_name = device_property.name
+        total_mem = device_property.total_memory / 1024**3
+
+    elif backend == gs_backend.metal:
+        if not torch.backends.mps.is_available():
+            gs.raise_exception("metal device not available")
+
+        # on mac, cpu and gpu are in the same device
+        _, device_name, total_mem, _ = get_device(gs_backend.cpu)
+        device = torch.device("mps")
+
+    elif backend == gs_backend.vulkan:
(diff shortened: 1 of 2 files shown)


genesis/utils/misc.py
```diff
@@ -96,32 +97,48 @@ def get_cpu_name():
...
-
-        device = torch.device("cuda")
-        device_property = torch.cuda.get_device_properties(0)
-        device_name = device_property.name
-        total_mem = device_property.total_memory / 1024**3
-
-    return device, device_name, total_mem
+def get_device(backend: gs_backend):
+    match backend:
+        case gs_backend.cuda:
```
bxtbold: match-case is supported for Python 3.10+, but the genesis requirement is 3.9+. it might cause issue for someone is working with 3.9

lgleim: @bxtbold Good catch. Switched to if/else. Kindly re-review

bxtbold (approved):
Looking okay to me.

But I don't have the push access to this repo. You may wanna add the main contributors as reviewers too.

zhouxian:
@zswang666 could you review and approve this if all look good?

zswang666:
The logic looks good and thus may merge.

Some test results though. Using different backends with `examples/rigid/control_franka.py`:
- On Linux, everything works
- On mac, cpu works; when using metal/gpu encountered error similar to https://github.com/Genesis-Embodied-AI/Genesis/issues/500; haven't setup vulkan on my macos 
- I don't have windows machine.

lgleim:
I tested on Ubuntu 22.04 (CPU, CUDA, Vulkan) and Windows 10 22H2 (CPU, Vulkan). 

The mac device/backend logic remains unchanged.

Issue #500 needs further investigation, likely into TaiChi's metal backend. The metal compiler error `Compiler encountered an internal error (code=3)` seems to indicate that the metal performance shaders emitted by taichi cannot be compiled. Cf. https://github.com/Genesis-Embodied-AI/Genesis/issues/500#issuecomment-2591975112. Either way, this is independent of the backend selection logic.

PR #591, opened 2025-01-17
[BUG FIX] Adding torch as dependency

lindemeier:
When installing via `pip install -e .` in a conda environment, torch had to be installed after to make the examples work.
a manual `pip install torch` worked.

This PR adds torch as dependency in pyproject.toml


Genesis-Embodied-AI:
In the installation tutorial, we explicitly mentioned torch needs to be installed

we decided not to put it as a dependency because torch version might vary across different platforms

PR #610, opened 2025-01-19
[FEATURE] Windows-compatible viewer support

kohya-ss:
This PR adds Windows compatibility to Genesis Viewer by providing a non-JIT implementation of the renderer.

## Background
The original Genesis Viewer uses JIT compilation for OpenGL rendering, which causes issues on Windows machines. This PR provides a non-JIT implementation while maintaining the same functionality.

## Changes
- Implemented non-JIT renderer in `genesis/ext/pyrender/non_jit_renderer.py`
- Modified viewer implementation to support both JIT and non-JIT rendering
- Updated rasterizer context to handle the non-JIT renderer
- Kept the original JIT implementation intact for non-Windows environments

## Testing
To test this implementation:
1. Install Genesis on a Windows machine
2. Run the viewer with a sample scene like: `python examples\rigid\control_franka.py --vis` or `python examples\rendering\moving_camera.py --vis`
3. Verify that the 3D rendering works correctly

The implementation has been tested on Windows 10 and 11, with Python 3.10.

Note: Due to the removal of JIT optimization, there may be a decrease in performance compared to the original implementation on non-Windows systems.

This implementation is focused on Windows compatibility while maintaining the existing functionality. As I am not an OpenGL expert, reviews and suggestions from OpenGL experts would be particularly appreciated.

### DIFF
--- a/genesis/ext/pyrender/non_jit_renderer.py
+++ b/genesis/ext/pyrender/non_jit_renderer.py
@@ -1,0 +1,727 @@
+import numpy as np
+
+from OpenGL.GL import *
+
+from .material import MetallicRoughnessMaterial, SpecularGlossinessMaterial
+from .light import DirectionalLight, PointLight
+from .constants import RenderFlags, MAX_N_LIGHTS
+
+RenderFlags_DEPTH_ONLY = RenderFlags.DEPTH_ONLY
+RenderFlags_SEG = RenderFlags.SEG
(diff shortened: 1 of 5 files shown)


ziyanx02:
@YilingQiao Could you try reformatting the `rasterizer_context.py` if you have time? I met the same problem, pass locally but fail online.

ziyanx02:
Thank you for your contribution!

PR #636, opened 2025-01-25
[FEATURE] Support Equality Constraint and Closed-loop robots

YilingQiao:
A joint work with @zhenjia-xu to support equality constraint so that we can simulate closed-loop robots. 

Try
```
python examples/rigid/closed_loop.py
```

https://github.com/user-attachments/assets/3c462f43-8de2-4934-af52-79532a535c8b


Kashu7100:
Thanks for the awesome work, I have checked `examples/rigid/closed_loop.py` works on my end.

Quick question: how can we use this for a closed link gripper like [robotiq85.zip](https://github.com/user-attachments/files/18557315/robotiq85.zip)? (Are there any restrictions on how `equality` is supposed to be defined in the XML?")

YilingQiao:
I will try your file.

YilingQiao:
> Thanks for the awesome work, I have checked `examples/rigid/closed_loop.py` works on my end.
> 
> Quick question: how can we use this for a closed link gripper like [robotiq85.zip](https://github.com/user-attachments/files/18557315/robotiq85.zip)? (Are there any restrictions on how `equality` is supposed to be defined in the XML?")

I tried this xml. It does not have equality constraint.

Equality should be defined this way (example from [here](https://github.com/google-deepmind/mujoco_menagerie/blob/dec370da8895f4a04c20bae57a52f3bd16148ee6/agility_cassie/cassie.xml#L227))
```
  <equality>
    <connect body1="left-plantar-rod" body2="left-foot" anchor="0.35012 0 0"/>
    <connect body1="left-achilles-rod" body2="left-heel-spring" anchor="0.5012 0 0"/>
    <connect body1="right-plantar-rod" body2="right-foot" anchor="0.35012 0 0"/>
    <connect body1="right-achilles-rod" body2="right-heel-spring" anchor="0.5012 0 0"/>
  </equality>
```

YilingQiao:
There might be some bugs and I'm still working on it.

darren-chiu:
Hello, 

I've imported a previously working MJCF file (tested in mujoco), but I noticed that the gripper collisions and constraints are not respected i.e. the gripper does not correctly close. I've attached my .xml file below. This was checked using the `gs view` command and installing using `pip install genesis-world`  Please let me know if there is anything I can provide. 

[ur5e_gripper.zip](https://github.com/user-attachments/files/18793918/ur5e_gripper.zip)

Kashu7100:
@darren-chiu The `pip install genesis-world` nor the current main includes this equality constraint feature (yet).

YilingQiao:
https://github.com/user-attachments/assets/a9333bde-8ba5-47b9-a06d-94e47899fcb9

Now also works on `mujoco_menagerie/agility_cassie` @zhenjia-xu

Kashu7100:
@YilingQiao Could you check if this works for you?
https://github.com/google-deepmind/mujoco_menagerie/blob/main/robotiq_2f85_v4/mjx_2f85.xml

Kashu7100:
Not very sure but `mujoco_menagerie/agility_cassie` fails on my end...

image

YilingQiao:
Oh, yeah, I recall that my cassie comments out the ball joint. There might be some problem with ball joint and equality constraint. I will fix that in another PR.

kavisrees:
even with using anchoring to connect the kinematic chain there seems to be a good bit of play even when tightening solimp and soref. is anchoring with connect the only way that works? what about welds with site references?

YilingQiao:
@kavisrees  We have `add_weld_constraint` which enfore the position as well and the relative orientation to be constant. Is that what you want? Although I think it's different from the connect constraint which only enfore the postion to be the same.

kavisrees:
> @kavisrees We have `add_weld_constraint` which enfore the position as well and the relative orientation to be constant. Is that what you want? Although I think it's different from the connect constraint which only enfore the postion to be the same.

well seems like the anchor is not fixed at that point, there is a little wobble, in mujoco i'm using 2 sites and can use connect or weld and define body1 and body2 and they stay rigid, i just want to make the joint rigid so there is no wobble/play in the joint

YilingQiao:
can you share some example script/asset/video? @kavisrees

kavisrees:
[Screencast from 2025-07-25 03_44_44 PM.webm](https://github.com/user-attachments/assets/f2ed1691-38a4-42f5-90e4-6628479ab726)

this is what im using to connect the rods to the ankle
`<equality>
   <connect body1="top_rod_link_1" body2="foot_roll_link_1" anchor="0.00469 0 -0.218" solimp="0.99 0.99 1e-05" solref="0.02 1"/>
   <connect body1="bottom_rod_link_1" body2="foot_roll_link_1" anchor="0.00471 0 -0.1394" solimp="0.99 0.99 1e-05" solref="0.02 1"/>
 </equality>`

ax-anoop:
Any information on the above video ?

PR #723, opened 2025-02-13
[FEATURE] Video Recording for Parallel Rendering

ACMLCZH:
1. Added an example of video recording for parallel rendering in `examples/rigid/single_franka_envs.py`. To test it, run the command: 
```
python examples/rigid/single_franka_envs.py --num_env [num_env] --sep --record
```
2. The intrinsics matrix can now be directly specified in `Camera.set_param()`. Please ensure that when both `fov` and `intrinsics` are provided, they are consistent with each other.


YilingQiao:
I have tested it. Looks good!

Issue #287, opened 2024-12-24
which kinds of mjcf file it support? I used xml from mujoco/models but 


milong26:
I used `arm26.xml` from [mujoco/models](https://github.com/google-deepmind/mujoco/blob/main/model/tendon_arm/arm26.xml). It shows well in mujoco.

![Screenshot from 2024-12-24 17-26-16](https://github.com/user-attachments/assets/4e6ee68b-9d53-4489-a4cb-450246346795)

But when I try to use it in genesis,it only shows a simple plane,lack the entity i added.

My whole code about entity is:
```python
plane = scene.add_entity(
    gs.morphs.Plane(),
)
franka = scene.add_entity(
    gs.morphs.MJCF(
        file='robot/arm26.xml',
    ),
)

```

![Screenshot from 2024-12-24 17-22-24](https://github.com/user-attachments/assets/ea641224-dcc8-43f5-af1a-8b37d3c248b5)

genesis version latest.

Kashu7100:
I will take a look

lupinjia:
same problem, can't visualize it

milong26:
xD I tested many exmaples and find somewhat may casue this

contype="0" conaffinity="0"  in xml file in <geom> label.

Wish this may help u

Kashu7100:
For MJCF, we currently support following types (PR for adding more types are welcome: check `genesis/utils/mjcf.py`):
- `mjGEOM_PLANE`
- `mjGEOM_SPHERE`
- `mjGEOM_CAPSULE`
- `mjGEOM_CYLINDER`
- `mjGEOM_MESH`

I think `contype` and `conaffinity` decides the mesh type of geom (line 238 of `genesis/utils/mjcf.py`) as:
 
```python
is_col = bool(mj.geom_conaffinity[i_g] or mj.geom_contype[i_g])
```

milong26:
> For MJCF, we currently support following types (PR for adding more types are welcome: check `genesis/utils/mjcf.py`):
> 
> * `mjGEOM_PLANE`
> * `mjGEOM_SPHERE`
> * `mjGEOM_CAPSULE`
> * `mjGEOM_CYLINDER`
> * `mjGEOM_MESH`
> 
> I think `contype` and `conaffinity` decides the mesh type of geom (line 238 of `genesis/utils/mjcf.py`) as:
> 
> ```python
> is_col = bool(mj.geom_conaffinity[i_g] or mj.geom_contype[i_g])
[...]

PR #734, opened 2025-02-15
[MISC] Add warning when importing MJCF Box w/collision checking

alexanderswerdlow:
## Description
Collision checking currently doesn't work with an MJCF Box. This PR adds a warning. 

## Related Issue
See: https://github.com/Genesis-Embodied-AI/Genesis/issues/287

Latest main commit, Linux Kernel 5.14.0

## Checklist:

### DIFF
--- a/genesis/utils/mjcf.py
+++ b/genesis/utils/mjcf.py
@@ -284,8 +284,10 @@ def parse_geom(mj, i_g, scale, convexify, surface, xml_path):
 
     elif mj_type == mujoco.mjtGeom.mjGEOM_BOX:
         if is_col:
-            gs.logger.warning(f"Genesis does not currently support MJCF geom type: \
-                              {mj_type} with collisions enabled (conaffinity or contype).")
+            gs.logger.warning(
+                f"Genesis does not currently support MJCF geom type: \
+                              {mj_type} with collisions enabled (conaffinity or contype)."
+            )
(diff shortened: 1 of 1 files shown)


YilingQiao (approved):

YilingQiao:
can you run?
```
black --line-length 120 genesis/utils/mjcf.py
```

alexanderswerdlow:
Done!

YilingQiao:
oh, wait, let me double check

YilingQiao:
We support box. It's just not visualized because by default we did not show collision mesh, just visual meshes. Didn't realize it has caused so many confusions. Thanks for bring this up to me.

YilingQiao:
Do you think it makes sense? @alexanderswerdlow

alexanderswerdlow:
@YilingQiao Yes, that makes perfect sense, thanks!

I think the main confusing part is just that the default behavior differs from mujoco (when importing their format) so it might trip up other people, but I have no idea how common this is (and personally my sim knowledge is a bit rusty and didn't realize that I might need to add a separate collision primitive). I'll leave it up to you, thanks again!

YilingQiao:
Thank you! We might have followed URDF's convention here and it differs from MJCF. At least we could through a warning for MJCF collision meshes. Appreciate it!

PR #735, opened 2025-02-15
[BUG] MJCF Parsing

YilingQiao:
## Description
Apply Zhenjia's change to MJCF parsing and add a warning for mjcf collision mesh's visualization.

## Related Issue
Add a warning for #287 
```
import genesis as gs

gs.init(backend=gs.gpu)

scene = gs.Scene(
    show_viewer=True,
)

plane = scene.add_entity(
    gs.morphs.MJCF(
        file="./arm26.xml",
        # from https://github.com/google-deepmind/mujoco/blob/main/model/tendon_arm/arm26.xml
    ),
    vis_mode="collision",
)

scene.build()
for i in range(1000):
    scene.step()


```

## Checklist:

### DIFF
--- a/genesis/utils/mjcf.py
+++ b/genesis/utils/mjcf.py
@@ -1,5 +1,4 @@
 import os
-import xml.etree.ElementTree as ET
 
 import mujoco
 import numpy as np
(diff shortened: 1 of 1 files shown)


Kashu7100:
Sorry for the delay. This is better than the current MJCF loading. However, I found one test case where the texture is not correctly loaded. This asset -> [dex_cube.zip](https://github.com/user-attachments/files/18832366/dex_cube.zip)

image

should look like this:
image

YilingQiao:
thank you, I will debug it!

YilingQiao:
I see. Without the change there would be an assertion error right?

YilingQiao:
We should merge the equality constraint one. Anyway, let's close that for now and use another PR to solve Kashu's case.

PR #752, opened 2025-02-19
[FEATURE] Modify n_rendered_envs to rendered_envs_idx

lupinjia:
## Description
modify `n_rendered_envs` attribute in vis_options to `rendered_envs_idx`, to support specification of which env to be rendered. 

## Motivation and Context
When training, sometimes I need to render specific env to see what happened to that env, so this specification can be useful

- modified `examples/tutorials/visualization.py` to test different idx rendered, run `visualization.py` to see the result.
- CI tests have been conducted locally, and there's no problem.
  
  ```txt
  True island 	| 1 cubes 	| <CG: 0> 	| 10,347,075.60 fps 	| 8192 envs
  False island 	| 1 cubes 	| <CG: 0> 	| 21,736,861.82 fps 	| 8192 envs
  True island 	| 10 cubes 	| <CG: 0> 	| 1,093,222.22 fps 	| 8192 envs
  False island 	| 10 cubes 	| <CG: 0> 	| 816,686.37 fps 	| 8192 envs
  True island 	| 1 cubes 	| <Newton: 1> 	| 8,947,203.02 fps 	| 8192 envs
  False island 	| 1 cubes 	| <Newton: 1> 	| 17,418,196.78 fps 	| 8192 envs
  True island 	| 10 cubes 	| <Newton: 1> 	| 925,635.75 fps 	| 8192 envs
  False island 	| 10 cubes 	| <Newton: 1> 	| 317,976.76 fps 	| 8192 envs
  random 	| <CG: 0> 	| 1,243,238.90 fps 	| 30000 envs
  random 	| <Newton: 1> 	| 1,823,401.43 fps 	| 30000 envs
  anymal 	| <CG: 0> 	| 4,958,232.78 fps 	| 30000 envs
  anymal 	| <Newton: 1> 	| 4,623,612.63 fps 	| 30000 envs
  frank 	| <CG: 0> 	| 10,531,791.39 fps 	| 30000 envs
  frank 	| <Newton: 1> 	| 11,374,890.36 fps 	| 30000 envs
  ```

render first 5 envs:
first_5_envs

render envs with idx 10 to 14:
11_to_15_envs

## Checklist:


examples/tutorials/visualization.py
```diff
@@ -18,6 +18,8 @@
         show_cameras=False,
         plane_reflection=True,
         ambient_light=(0.1, 0.1, 0.1),
+        rendered_envs_idx=[10, 11, 12, 13, 14],  # render the 11th to 15th environments
```
Kashu7100: I think it might be better to have a separate example for specific env rendering.

Kashu7100:
@YilingQiao do you think we can merge this after we modified add the examples using `n_rendered_envs`?

genesis/vis/rasterizer_context.py
```diff
@@ -285,7 +285,7 @@ def on_rigid(self):
                         mesh = geom.get_sdf_trimesh()
                     else:
                         mesh = geom.get_trimesh()
-                    geom_T = geoms_T[geom.idx][: self.n_rendered_envs]
+                    geom_T = geoms_T[geom.idx][self.rendered_envs_idx]
```
duburcqa, line 280: Beware this would trigger a copy since ‘self.rendered_envs_idx’ is not a slice but a list. No sure about the overall performance penalty it induces.

Kashu7100: I see. We should bench first...

examples/tutorials/visualization.py
lupinjia: I'll write a separate example for it ASAP :)

YilingQiao:
> @YilingQiao do you think we can merge this after we modified add the examples using `n_rendered_envs`?

Yeah, I think we can merge this after incorporating the suggested changes.

examples/tutorials/visualization.py
lupinjia: I have added a new example "selecting_rendered_envs"

genesis/vis/rasterizer_context.py
lupinjia: For this problem, I added a new variable called `rendered_envs_idx_slice` to avoid copy list to slice in the loop

genesis/vis/rasterizer_context.py
```diff
@@ -64,6 +64,7 @@ def build(self, scene):
 
         if self.rendered_envs_idx is None:
             self.rendered_envs_idx = list(range(self.sim._B))
+        self.rendered_envs_idx_slice = slice(self.rendered_envs_idx[0], self.rendered_envs_idx[-1] + 1)
```
Kashu7100: I don't think the assumption of consecutive `rendered_envs_idx` is good. What we should do instead is to measure the overhead introduced by list indexing and verify it won't add more overhead compared to the current version.

genesis/vis/raytracer.py
```diff
@@ -208,7 +208,8 @@ def build(self, scene):
         self.sim = scene.sim
         self.visualizer = scene.visualizer
         if self.rendered_envs_idx is None:
-            self.rendered_envs = list(range(self.sim._B))
+            self.rendered_envs_idx = list(range(self.sim._B))
+        self.rendered_envs_idx_slice = slice(self.rendered_envs_idx[0], self.rendered_envs_idx[-1] + 1)
```
Kashu7100: I don't think the assumption of consecutive rendered_envs_idx is good.

genesis/vis/rasterizer_context.py
```diff
@@ -285,7 +286,7 @@ def on_rigid(self):
                         mesh = geom.get_sdf_trimesh()
                     else:
                         mesh = geom.get_trimesh()
-                    geom_T = geoms_T[geom.idx][self.rendered_envs_idx]
+                    geom_T = geoms_T[geom.idx][self.rendered_envs_idx_slice]
```
Kashu7100: should keep `self.rendered_envs_idx` but make sure this won't affect the performance.

genesis/vis/rasterizer_context.py
```diff
@@ -312,7 +313,7 @@ def update_rigid(self, buffer_updates):
                     geoms_T = self.sim.rigid_solver._geoms_render_T
 
                 for geom in geoms:
-                    geom_T = geoms_T[geom.idx][self.rendered_envs_idx]
+                    geom_T = geoms_T[geom.idx][self.rendered_envs_idx_slice]
```
Kashu7100: should keep self.rendered_envs_idx but make sure this won't affect the performance.

genesis/vis/rasterizer_context.py
duburcqa: Definitely

lupinjia: I think it's not consecutive, because this line is in `build()`, and `build()` will only be called once, so it's basically to convert list to slice once built, and use that slice for indexing afterwards.

lupinjia:
I simply tested the performance of rendered_envs_idx, for list version and slice version. Slice version is the newer one, using slice for indexing; and list version is the older one. 
The result is as follows:
- slice version, build 200 envs, render 50 envs
    

https://github.com/user-attachments/assets/5fca13c0-443f-48c6-b2e2-393d283ddcf6

- list version, build 200  envs, render 50 envs
   

https://github.com/user-attachments/assets/97c9f3b1-e779-4e56-b8fe-4f4e0499db8b

In both situations, the FPS per env can reach 60 FPS, as the terminal outputs.

Kashu7100:
If both performs the same. I think we should use the  list version

lupinjia:
> If both performs the same. I think we should use the list version

ok, I'll change it back to list version

Kashu7100 (approved):

PR #792, opened 2025-02-27
[BUG FIX] Parse mujoco ball joint

YilingQiao:
## Description

mujoco's qpos0 has 4 components (quaternion) for ball joints (3 dofs). Need to convert that quat to euler angle.

Now it can run cassie without error.

## Related Issue

```
import genesis as gs


########################## init ##########################
gs.init()

########################## create a scene ##########################
viewer_options = gs.options.ViewerOptions(
    camera_pos=(0, -3.5, 2.5),
    camera_lookat=(0.0, 0.0, 0.5),
    camera_fov=40,
    max_FPS=200,
)
scene = gs.Scene(
    viewer_options=viewer_options,
        dt=0.001,
        contact_resolve_time=0.02,
    ),
    show_viewer=True,
)

########################## entities ##########################
plane = scene.add_entity(
    morph=gs.morphs.Plane(),
)

path_robot = "mujoco_menagerie/agility_cassie/cassie.xml"
robot = scene.add_entity(
    morph=gs.morphs.MJCF(
        file=path_robot,
    ),
)

########################## build ##########################
scene.build()

for i in range(1000):
    scene.step()

```
## Checklist:

### DIFF
--- a/genesis/utils/mjcf.py
+++ b/genesis/utils/mjcf.py
@@ -66,18 +66,16 @@ def parse_link(mj, i_l, q_offset, dof_offset, qpos0_offset, scale):
 
     def add_more_joint_info(j_info, jnt_offset=0):
         d_off = dof_offset + jnt_offset
-        q_off = q_offset + jnt_offset
         qpos0_off = qpos0_offset + jnt_offset
 
         j_info["dofs_damping"] = np.array(mj.dof_damping[d_off : d_off + j_info["n_dofs"]])
         j_info["dofs_invweight"] = np.array(mj.dof_invweight0[d_off : d_off + j_info["n_dofs"]])
         j_info["dofs_armature"] = np.array(mj.dof_armature[d_off : d_off + j_info["n_dofs"]])
-
-        if j_info["type"] == gs.JOINT_TYPE.SPHERICAL:
-            init_qpos = gu.quat_to_xyz(np.array(mj.qpos0[qpos0_off : qpos0_off + 4]))
+        if j_info["n_qpos0"] == 4 and j_info["type"] == gs.JOINT_TYPE.SPHERICAL:
+            # this is a real mujoco ball joint
+            j_info["init_qpos"] = gu.quat_to_xyz(mj.qpos0[qpos0_off : qpos0_off + 4])
         else:
-            init_qpos = np.array(mj.qpos0[qpos0_off : qpos0_off + j_info["n_qs"]])
-        j_info["init_qpos"] = init_qpos
+            j_info["init_qpos"] = np.array(mj.qpos0[qpos0_off : qpos0_off + j_info["n_qpos0"]])
 
         # apply scale
         j_info["pos"] *= scale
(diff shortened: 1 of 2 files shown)


YilingQiao:
wait a minute, let me test more examples

YilingQiao:
Ok, it should be good now. Can you please take a look @Kashu7100 ? It fix that Cassie issue.

Kashu7100:
Yes, I will

genesis/utils/mjcf.py
```diff
@@ -196,6 +202,13 @@ def add_more_joint_info(j_info, jnt_offset=0):
         elif j_info["n_dofs"] == 6:
             j_info["type"] = gs.JOINT_TYPE.FREE
 
+        j_info["n_qpos0"] = j_info["n_qs"]
+        if j_info["type"] == gs.JOINT_TYPE.SPHERICAL and len(j_info_list) == 1:
```
duburcqa, line 206: I would rather keep track of the individual `n_qpos0` and sum them just like `n_qs`. This way it is completely generic and you do not need this kind of edge-case handling.

genesis/utils/mjcf.py
```diff
@@ -66,12 +66,16 @@ def add_actuator(j_info, i_j=None):
...
     def add_more_joint_info(j_info, jnt_offset=0):
         d_off = dof_offset + jnt_offset
-        q_off = q_offset + jnt_offset
+        qpos0_off = qpos0_offset + jnt_offset
 
         j_info["dofs_damping"] = np.array(mj.dof_damping[d_off : d_off + j_info["n_dofs"]])
         j_info["dofs_invweight"] = np.array(mj.dof_invweight0[d_off : d_off + j_info["n_dofs"]])
         j_info["dofs_armature"] = np.array(mj.dof_armature[d_off : d_off + j_info["n_dofs"]])
-        j_info["init_qpos"] = np.array(mj.qpos0[q_off : q_off + j_info["n_qs"]])
+        if j_info["n_qpos0"] == 4 and j_info["type"] == gs.JOINT_TYPE.SPHERICAL:
```
duburcqa, line 74: What about computing `init_qpos` for each joint individually and just stacking them at the end as many of the other fields ? It would avoid this kind of "ugly" edge-case handling completely.

Kashu7100:
@YilingQiao overall this PR is very nice.
I think we need to update the `gs view` as well. Currently we are using `dof_pos_limits = torch.stack(entity.get_dofs_limit(motor_dofs), dim=1).numpy()` for joint limit calculation, which raise error for ball joint. 
I think we can address this in a separate PR (this PR works fine for my assets). 

Also one more thing I noticed is that when I use `set_dofs_position` instead of `control_dofs_position`, it will break the closed links - not sure this is expected or bug.

genesis/utils/mjcf.py
Kashu7100: I agree. But should we generalize in separate PR? I think fixing the ball joint issue is the priority?

YilingQiao:
Thank you both for those comments! I'll probably have time to resolve some of them tomorrow.

genesis/utils/mjcf.py
duburcqa: Yes! I will take care of it.

duburcqa:
You can merge this PR if you want. I will take into account my own comments in a following PR!

Issue #762, opened 2025-02-20
[Bug]: TypeError: not a sequence when loading MJCF


HaoxiangYou:
### Bug Description

I want to implement several RL benchmark with Genesis. 

However, when I try to open(via gs view) xml from [deepmind mujoco](https://github.com/google-deepmind/mujoco/blob/main/model/humanoid/humanoid.xml) or [brax](https://github.com/google/brax/blob/main/brax/envs/assets/humanoid.xml),
it report  
```
 File "/media/haoxiangyou/6c569641-7e1c-4595-b053-7021a7dbe092/Documents/Yale/Research/Genesis/genesis/engine/entities/rigid_entity/rigid_entity.py", line 1825, in _get_dofs_idx
    return self._get_dofs_idx_local(dofs_idx_local) + self._dof_start
  File "/media/haoxiangyou/6c569641-7e1c-4595-b053-7021a7dbe092/Documents/Yale/Research/Genesis/genesis/engine/entities/rigid_entity/rigid_entity.py", line 1819, in _get_dofs_idx_local
    dofs_idx_local = torch.as_tensor(dofs_idx_local, dtype=gs.tc_int, device=gs.device)
TypeError: not a sequence
```

### Steps to Reproduce

If possible, provide a script triggering the bug, e.g.
```python
# code snippet triggering the bug
import genesis as gs
gs.init()
scene = gs.Scene(show_viewer=False)
broken = scene.add_entity(
    gs.morphs.MJCF(file='attachement.xml'),
)
scene.build()
```
Make sure to attached any needed assets (here `attachement.xml`)!

### Expected Behavior

```
gs.init(backend=gs.cpu)

scene = gs.Scene(sim_options=gs.options.SimOptions(dt=dt))

humanoid = scene.add_entity(
    gs.morphs.MJCF(file=os.path.join(dir,"asset/humanoid.xml")),
)

dofs_names = [
    # "abdomen_y",
    "abdomen_z",
    "abdomen_x"
]

dofs_idx = [humanoid.get_joint(name).dof_idx_local for name in dofs_names]

print(dofs_idx)
```
I saw the dofs_idx = [[6,7], 12].
I also cannot retrieve joint "abdomen_y" via get_joint(name) althrough it is in xml file

### Relevant log output

```shell

```

### Environment

- OS: Ubuntu 22.04

### Release version or Commit ID

commit ID: e7f98d9360e6ab0d60f6cffcb902ae583f31b72c

[...]

PR #793, opened 2025-02-27
[Bug Fix] A temporary solution for merged joints

YilingQiao:
## Description

For MJCF, we will merge joints under the same link. For this situation, we will output a warning and return the merged joint when user call entity.get_joint()

## Related Issue

#762 

```
import numpy as np

gs.init(backend=gs.gpu)

scene = gs.Scene(
    show_viewer=True,
)

robot = scene.add_entity(
    gs.morphs.MJCF(
        file="./humanoid/humanoid.xml", # https://github.com/google-deepmind/mujoco/tree/main/model/humanoid
    ),
    vis_mode="collision",
)

scene.build()

for joint in robot.joints:
    print(joint.name)

print(robot.get_joint("abdomen_z"))
```
output:
```
root
head_joint
abdomen_z,abdomen_y
shoulder1_right,shoulder2_right
shoulder1_left,shoulder2_left
abdomen_x
elbow_right
elbow_left
hip_x_right,hip_z_right,hip_y_right
hip_x_left,hip_z_left,hip_y_left
hand_right_joint
hand_left_joint
knee_right
knee_left
ankle_y_right,ankle_x_right
ankle_y_left,ankle_x_left
world_joint
[Genesis] [23:41:47] [WARNING] we only found abdomen_z,abdomen_y. It's probably due to we merge joints under the same link.

```

## Checklist:

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -1529,6 +1529,12 @@ class RigidEntity(Entity):
             for joint in self._joints:
                 if joint.name == name:
                     return joint
+            for joint in self._joints:
+                if f"{name}," in joint.name or f",{name}" in joint.name:
+                    gs.logger.warning(
+                        f"we only found {joint.name}. It's probably due to we merge joints under the same link."
+                    )
+                    return joint
             gs.raise_exception(f"Joint not found for name: {name}.")
 
         elif uid is not None:
--- a/genesis/utils/mjcf.py
+++ b/genesis/utils/mjcf.py
@@ -198,7 +198,7 @@ def parse_link(mj, i_l, q_offset, dof_offset, scale):
 
         j_info["quat"] = j_info_list[0]["quat"]
         j_info["pos"] = j_info_list[0]["pos"]
-        j_info["name"] = j_info_list[0]["name"]
+        j_info["name"] = ",".join([j["name"] for j in j_info_list])
 
         final_joint_list.append(j_info)
 


genesis/utils/mjcf.py
```diff
@@ -198,7 +198,7 @@ def add_more_joint_info(j_info, jnt_offset=0):
 
         j_info["quat"] = j_info_list[0]["quat"]
         j_info["pos"] = j_info_list[0]["pos"]
-        j_info["name"] = j_info_list[0]["name"]
+        j_info["name"] = ",".join([j["name"] for j in j_info_list])
```
duburcqa, line 201: This is definitely much better!

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1529,6 +1529,12 @@ def get_joint(self, name=None, uid=None):
             for joint in self._joints:
                 if joint.name == name:
                     return joint
+            for joint in self._joints:
+                if f"{name}," in joint.name or f",{name}" in joint.name:
+                    gs.logger.warning(
```
duburcqa, line 1534: I do not think it is a good idea to just report a warning and return the merged joint silently. I would advice to raise an exception by default, and add an extra optional argument to specify whether merged joint should be considered.

Besides, the logic to detect whether it is a merged joint or not is brittle, and it is not possible to split them back into individual joints. It would advise to modify `RigidJoint` to add a new field `original_joints` or `true_joints`, that keeps track of the original joints, without using them for computation (each of them being some `VirtualJoint` instances that would only contain information can be recovered). This way, it would be possible to return a joint object that would be helpful for the end-user systematically.

duburcqa:
You can merge this PR if you want. I will take into account my own comments in a following PR!

Issue #788, opened 2025-02-26
[Bug]: Build Fails for Python < 3.12


bxtbold:
### Bug Description

The build fails for Python<3.12 in `genesis/ext/fast_simplification/_simplify.cpp` with errors like `PyDataType_ELSIZE` was not declared. This seems to be a numpy API mismatch in the cython extensions, possibly from changes in #711

### Steps to Reproduce

Try building from the source.

### Relevant log output
```shell
      Compiling genesis/ext/fast_simplification/_simplify.pyx because it depends on /tmp/pip-build-env-yhwkmbad/overlay/lib/python3.10/site-packages/Cython/Includes/libc/string.pxd.
      [1/1] Cythonizing genesis/ext/fast_simplification/_simplify.pyx
      building 'genesis.ext.fast_simplification._simplify' extension
      x86_64-linux-gnu-g++ -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O2 -Wall -g -fstack-protector-strong -Wformat -Werror=format-security -g -fwrapv -O2 -fPIC -DIS64BITPLATFORM -I/home/bat/ws/venvs/genesis_test/.venv/include -I/usr/include/python3.10 -I/tmp/pip-build-env-yhwkmbad/overlay/lib/python3.10/site-packages/numpy/_core/include -c genesis/ext/fast_simplification/_simplify.cpp -o build/temp.linux-x86_64-cpython-310/genesis/ext/fast_simplification/_simplify.o -std=gnu++11 -O3 -w
      genesis/ext/fast_simplification/_simplify.cpp: In function ‘npy_intp __pyx_f_5numpy_5dtype_8itemsize_itemsize(PyArray_Descr*)’:
      genesis/ext/fast_simplification/_simplify.cpp:18234:13: error: ‘PyDataType_ELSIZE’ was not declared in this scope; did you mean ‘PyDataType_ISUNSIZED’?
      18234 |   __pyx_r = PyDataType_ELSIZE(__pyx_v_self);
            |             ^~~~~~~~~~~~~~~~~
            |             PyDataType_ISUNSIZED
      genesis/ext/fast_simplification/_simplify.cpp: In function ‘npy_intp __pyx_f_5numpy_5dtype_9alignment_alignment(PyArray_Descr*)’:
[...]

PR #794, opened 2025-02-27
[BUG] update setup.py to fix build fails

bxtbold:
## Description
Fix build fails for Python < 3.12 resulted by cython extensions that use numpy. After making genesis compatible with numpy 2.0, building failed for lower versions of Python.

## Related Issue
Resolves Genesis-Embodied-AI/Genesis#788

## Motivation and Context
In numpy>2.0, some C APIs (e.g. PyDataType_ELSIZE, PyDataType_FIELDS) are moved to new headers and required explicit setups (which wasn't required in numpy<2.0). Adding `include_dirs=[np.get_include()]` explicitly links to the updated numpy headers.

Source:
- [NumPy 2.0 migration guide](https://numpy.org/doc/stable/numpy_2_0_migration_guide.html)
- [Cython for NumPy users](https://cython.readthedocs.io/en/latest/src/userguide/numpy_tutorial.html#adding-numpy-headers)

1. have a clean python env (using venv, conda etc)
2. clone the branch
3. pip install -e .

## Checklist:

edit 1: update motivation and context for a better explanation of the solution
edit 2: update source (cython for numpy users)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,7 @@ import builtins
 from io import open as io_open
 import os
 import sys
+import numpy as np
 from setuptools import Extension, setup
 from setuptools.command.build_ext import build_ext as _build_ext
 
@@ -53,6 +54,7 @@ setup(
         Extension(
             "genesis.ext.fast_simplification._simplify",
             ["genesis/ext/fast_simplification/_simplify.pyx"],
+            include_dirs=[np.get_include()],
             language="c++",
             extra_compile_args=extra_compile_args,
             define_macros=macros,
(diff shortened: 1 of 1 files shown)


duburcqa:
Any clue or official documentation reference that would explain why this change is necessary with numpy 2.X but not before ?

bxtbold:
Thanks for a review! I have updated the motivation and context part above. Lemme know if there is any clarifications! @duburcqa

duburcqa:
From what I read in the documentation, nothing is mentioning explicitly that numpy headers must now be explicitly included manually. But I would rather say I'm surprised it was working fine previously since Cython documentation states that Numpy headers should be included...

duburcqa:
I guess it is rather related to "official" numpy accessors now being mandatory to access fields of numpy arrays, will before it was possible to manipulate the C data structure directly. Because of this, the helpers are now involved in the generated c++ code, which in turns requires including the numpy headers. So it was only working before because it uses of Numpy API was very limited, probably only a container for array, without any calling any Numpy operators.

PR #815, opened 2025-03-03
[FEATURE] Enable visualizing path

bxtbold:
## Description

Enable visualizing a path for rigid entities using FK.

## Related Issue

## Motivation and Context

Opening this PR to improve the convenience of visualizing a path.

I have modified `examples/tutorials/IK_motion_planning_grasp.py` to visualize the path. The example script can be used to test. (If requested, I can add a separate example script to demonstrate this feature.)

image

## Checklist:

### DIFF
--- a/genesis/engine/scene.py
+++ b/genesis/engine/scene.py
@@ -2,6 +2,7 @@ import numpy as np
 import torch
 
 import genesis as gs
+import genesis.utils.geom as gu
 from genesis.engine.entities import Emitter
 from genesis.engine.simulator import Simulator
 from genesis.options import (
@@ -970,11 +971,13 @@ class Scene(RBC):
             The created debug object.
         """
         with self._visualizer.viewer_lock:
-            poss = torch.zeros(len(qposs), 3)
+            Ts = torch.zeros((len(qposs), 4, 4))
             for i, qpos in enumerate(qposs):
-                pos, _ = entity.forward_kinematics(qpos)
-                poss[i] = pos[-1]
-            return self._visualizer.context.draw_debug_points(poss, colors)
+                pos, quat = entity.forward_kinematics(qpos)
+                T = gu.quat_to_T(quat[-1])
+                T[:3, 3] = pos[-1]
+                Ts[i] = T
+            return self._visualizer.context.draw_debug_frames(Ts)
 
     @gs.assert_built
     def clear_debug_object(self, object):
--- a/genesis/engine/scene.py
+++ b/genesis/engine/scene.py
@@ -950,7 +950,7 @@ class Scene(RBC):
             return self._visualizer.context.draw_debug_points(poss, colors)
 
     @gs.assert_built
-    def draw_debug_path(self, qposs, entity, density=0.3, frame_scaling=1.0):
+    def draw_debug_path(self, qposs, entity, link_idx=-1, density=0.3, frame_scaling=1.0):
         """
         Draws a planned joint trajectory in the scene for visualization.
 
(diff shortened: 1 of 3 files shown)


duburcqa:
What about displaying small "frames" along the way (x, y, z axes), since the orientation of the end-effector is also planned? Besides, there should be a way to draw an actual line instead of a sequence of points. Displaying both a line and frames (not all of them, something like `[::10]`) would be great!

bxtbold:
Thanks for a feedback!! @duburcqa 

> What about displaying small "frames" along the way (x, y, z axes), since the orientation of the end-effector is also planned? Besides, there should be a way to draw an actual line instead of a sequence of points.

I agree on drawing frames since orientations are also planned, but I don't think it's necessary to draw lines because the planned path is not continuous path but rather segments. So, frames will be exact representation/visualization of the path.

For now, there are not functions for drawing a batch of frames. I can work on that in `genesis.vis.RasterizerContext` to implement the above. 

> not all of them, something like `[::10]`

How about adding an argument to let users choose the density (between 0 and 1)? It could be something like this:

```python
def draw_debug_path(self, qposs, entity, density=1.0, colors=(1.0, 0.0, 0.0, 0.5)):
    ...
```

duburcqa:
>  because the planned path is not continuous path but rather segments

Sure, but then I think it is "important" to print those segments. It would make it very obvious that the 3D frames are only a few "keypoints" being rendered of the underlying trajectory.

> How about adding an argument to let users choose the density (between 0 and 1)?

Do as you please. Both is fine for me.

Kashu7100:
One minor comment is that if you could change the color to more vivid one, that will be nicer.

genesis/engine/scene.py
```diff
@@ -924,6 +949,52 @@ def draw_debug_points(self, poss, colors=(1.0, 0.0, 0.0, 0.5)):
...
+            N = len(qposs)
+            density = np.clip(density, 0.0, 1.0)
+            N_new = int(N * density)
+            indices = torch.linspace(0, N - 2, N_new, dtype=int)
+
+            Ts = np.zeros((N_new, 4, 4))
+
+            for i in range(N_new):
+                pos, quat = entity.forward_kinematics(qposs[indices[i]])
+                T = gu.quat_to_T(quat[-1])
```
duburcqa: `Ts[i] = gu.ti_trans_quat_to_T(pos[-1], quat[-1])`

bxtbold: Thx for a review! Updated it using `trans_quat_to_T` instead of taichi implementation

bxtbold:
Sorry for some backlogs. The end result looks like this under diff args (the same example I used in the PR description):

density 1.0, frame_scaling 1.0


density 0.5, frame_scaling 1.0


density 0.3, frame_scaling 1.0


density 0.3, frame_scaling 0.5


density 0.3, frame_scaling 0.3

bxtbold:
> One minor comment is that if you could change the color to more vivid one, that will be nicer.

I have changed from points to frames, so it should be more vivid/visible rn @Kashu7100

Kashu7100:
beautiful

duburcqa (approved):

duburcqa:
It would be great to increase transparency or color fading over time, but this is already to a good start!

bxtbold:
Thanks for an approve!

> It would be great to increase transparency or color fading over time, but this is already to a good start!

I am already aware of transparency because meshes cast shades too as alpha is 1 by default.

I will open a new PR to make the frame creating function pass a color vector (or only alpha). For colors, I will think about some ideas ;)

PR #839, opened 2025-03-07
[FEATURE] add support for pyglet<2 so that genesis can be used with trimesh show

Kashu7100:
## Description

Add support for pyglet (1.5.31) so that trimesh show can run on the same environment. (also applies to the genesis.ext.trimesh)

## Motivation and Context

trimesh viewer does not support pyglet>2, which is the current default for genesis. In case trimesh viewer is required, this adds support for 1.5.31 version. (we can keep the recommended version as >2)

## Checklist:

### DIFF
--- a/genesis/vis/visualizer.py
+++ b/genesis/vis/visualizer.py
@@ -1,4 +1,9 @@
 import pyglet
+
+BACKWARD_COMPATIBLE = False
+if int(pyglet.version.split(".")[0]) < 2:
+    BACKWARD_COMPATIBLE = True
+    import screeninfo
 import genesis as gs
 from genesis.repr_base import RBC
 
@@ -34,8 +39,11 @@ class Visualizer(RBC):
 
         # try to connect to display
         try:
-            display = pyglet.display.get_display()
-            screen = display.get_default_screen()
+            if BACKWARD_COMPATIBLE:
+                screen = screeninfo.get_monitors()[0]
+            else:
+                display = pyglet.display.get_display()
+                screen = display.get_default_screen()
             self._connected_to_display = True
         except Exception:
             self._connected_to_display = False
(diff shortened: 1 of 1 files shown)


duburcqa:
‘screeninfo’ is broken and there is no point of introducing an additional dependency. There must be a good reason to add dependencies as it makes the project harder to maintain. I would rather suggest relying on the old API of pyglet to get the screen size. This is nothing new and it was already possible before 2.0.

Kashu7100:
Thanks for your feedback. I will double check!!

Kashu7100:
@duburcqa I removed the dependencies for `screeninfo`. Can you review?

genesis/vis/visualizer.py
```diff
@@ -34,20 +41,24 @@ def __init__(self, scene, show_viewer, vis_options, viewer_options, renderer):
...
+            gs.logger.warning(e)
             self._connected_to_display = False
 
         if show_viewer:
             if not self._connected_to_display:
                 gs.raise_exception("No display detected. Use `show_viewer=False` for headless mode.")
 
             if viewer_options.res is None:
-                viewer_size_ratio = screen.get_scale() * 0.5
+                viewer_size_ratio = 0.5
```
duburcqa: You do need to take into account the scaling factor, as it is 2.0 on macos

genesis/vis/visualizer.py
```diff
@@ -34,20 +41,24 @@ def __init__(self, scene, show_viewer, vis_options, viewer_options, renderer):
...
                 viewer_options.res = (
-                    int(screen.height * viewer_size_ratio / 0.75),
+                    int(screen.width * viewer_size_ratio),
```
duburcqa: This does not have to be changed.

genesis/vis/visualizer.py
Kashu7100: I will check alternative method for older versions.

Kashu7100: couldn't find a straightforward way for the older version so setting it as constant for older versions.

Kashu7100:
@duburcqa I reflected your reviews.

PR #842, opened 2025-03-10
[MISC]  Remove unused files

YilingQiao:
## Description

Remove unused files. We are now using `*_decomp.py`.

## Related Issue

## Checklist:


genesis/engine/solvers/avatar_solver.py
```diff
@@ -5,7 +5,7 @@
 from genesis.engine.states.solvers import AvatarSolverState
 
 from .base_solver import Solver
-from .rigid.rigid_solver import RigidSolver
+from .rigid.rigid_solver_decomp import RigidSolver
```
duburcqa, line 8: what about renaming `rigid_solver_decomp` in `rigid_solver` in the meantime?

Kashu7100: for me keeping `*_decomp` is preferred (but this is just my preference)

YilingQiao:
Decomp means that we have decomposed a large simulation kernel into several sub-kernels. Ideally, the decomposed version will have much faster compilation but slightly slower runtime performance. We might revisit this later to determine whether a large kernel or the decomposed version is better.

Kashu7100:
should we also remove sdf.py??

PR #853, opened 2025-03-12
[FEATURE] Add support of compound joints

duburcqa:
## Description

This PR gets rid of joint fusing completely in order to keep the original joint composition. First, this was confusing end-user that were trying to access data related to the original joints for their application. Second, it was impossible to support specifying different joint pose, as they cannot be fixed into one time-independent joint.

## Related Issue

Related to Genesis-Embodied-AI/Genesis/issues/461
Resolves Genesis-Embodied-AI/Genesis/issues/762
Supersedes Genesis-Embodied-AI/Genesis/pull/792
Resolves Genesis-Embodied-AI/Genesis/discussions/837

## Motivation and Context

Chaining multiple basic 1DoF joints is the standard approach in Mujoco for implementing complex N-DoF joints, including spherical joints. Currently, only chaining multiple DoFs is supported in Genesis, which is not the same as chaining multiple joints as it is not possible to specify individual pose offset between them. Anyway, chaining multiple DoF is not strictly equivalent to a single N-DoF joint in the general case.

Going through the various assets that are shipping with Genesis.

## Checklist:

### DIFF
--- a/genesis/engine/solvers/rigid/rigid_solver_decomp.py
+++ b/genesis/engine/solvers/rigid/rigid_solver_decomp.py
@@ -1,3 +1,5 @@
+from itertools import chain
+
 import numpy as np
 import taichi as ti
 import torch
@@ -280,8 +282,8 @@ class RigidSolver(Solver):
         for i in range(self.n_links):
             j = i
             while j > -1:
-                for i_d in range(self.joints[i].dof_start, self.joints[i].dof_end):
-                    for j_d in range(self.joints[j].dof_start, self.joints[j].dof_end):
+                for i_d in range(self.links[i].dof_start, self.links[i].dof_end):
+                    for j_d in range(self.links[j].dof_start, self.links[j].dof_end):
                         mass_parent_mask[i_d, j_d] = 1.0
                 j = self.links[j].parent_idx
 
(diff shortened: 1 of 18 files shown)


tests/test_rigid.py
```diff
@@ -1,77 +1,295 @@
...
+        self.gs_scene = gs.Scene(
+            viewer_options=gs.options.ViewerOptions(
+                camera_pos=(3, -1, 1.5),
+                camera_lookat=(0.0, 0.0, 0.5),
+                camera_fov=30,
+                res=(960, 640),
+                max_FPS=60,
+            ),
+            sim_options=gs.options.SimOptions(
+                dt=self.mj_model.opt.timestep,
```
zswang666: should `self.mj_model.opt.timestep` be `self.gs_scene.dt / self.gs_scene.substeps`? by default, `gs_cene.substeps` is 1 and thus if directly working with `mj_model` (which implies `substeps` in mujoco is 1), it may be fine?!

duburcqa: What do you mean ? If `substeps` is basically 1 for Mujoco, we have no other choice than enforcing `self.gs_scene.substeps = 1` in Genesis unit tests.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -4686,146 +4840,115 @@ def _kernel_set_geoms_friction(
...
-            return len(self.links)
+        return len(self.links)
 
     @property
     def n_joints(self):
         if self.is_built:
             return self._n_joints
-        else:
-            return len(self.joints)
+        return sum(map(len, self.joints))
```
zswang666, line 4873: This causes error, e.g., `sum(map(len, [1,2]))`

PR #884, opened 2025-03-18
[BUG FIX] Only add world link/joint if it has at least one geom.

duburcqa:
## Related Issue

Resolves Genesis-Embodied-AI/Genesis/issues/883

## Checklist:

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -341,11 +341,14 @@ class RigidEntity(Entity):
 
         # Parse all bodies (links and joints)
         (l_world_info, *l_infos), (j_world_info, *j_infos) = mju.parse_links(mj, morph.scale)
-
         l_infos, j_infos, links_g_info, ordered_links_idx = uu._order_links(l_infos, j_infos, links_g_info)
-        for l_info, j_info, link_g_info in zip(
-            (*l_infos, l_world_info), (*j_infos, j_world_info), (*links_g_info, world_g_info)
-        ):
+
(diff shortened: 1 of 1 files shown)


genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -341,11 +341,14 @@ def _load_MJCF(self, morph, surface):
...
         l_infos, j_infos, links_g_info, ordered_links_idx = uu._order_links(l_infos, j_infos, links_g_info)
-        for l_info, j_info, link_g_info in zip(
-            (*l_infos, l_world_info), (*j_infos, j_world_info), (*links_g_info, world_g_info)
-        ):
+
+        # Add all bodies to this entity
+        all_infos = list(zip(l_infos, j_infos, links_g_info))
+        if world_g_info:
+            # Only take into account the world if it features at least one geometry
+            all_infos.append((l_world_info, j_world_info, world_g_info))
```
bxtbold, line 350: Can you actually move world_infos at the beginning of the list? So the links and joints lists would be in order by starting from world.

Maybe sthg like this should work:
```python
        all_infos = []
        if world_g_info:
            # Only take into account the world if it features at least one geometry
            all_infos.append((l_world_info, j_world_info, world_g_info))

        # Add the other infos after world_info
        all_infos.extend(list(zip(l_infos, j_infos, links_g_info)))
```

duburcqa: No you cannot, otherwise the simulation would go crazy. Unfortunately, I would not be able to explain why precisely. By the way, it was not the case before already.

Issue #905, opened 2025-03-21
[Bug]: gs view not working in macOS - if joint.type == gs.JOINT_TYPE.FREE:        ^^^^^^^^^^ AttributeError: 'List' object has no attribute 'type'


antragoudaras:
### Bug Description

# Description

When running `gs view` cli on macOS, I encountered an error related to joints iteration in the `get_movable_dofs` function. The issue stems from a platform-specific difference in how joint data is structured:

### The Bug

- On macOS, the `robot.joints` property returns a nested list structure (a List of Lists), while on other platforms it appears to return a flat list.
- When the code tries to iterate through `robot.joints` and access `joint.type` directly, it fails on macOS because it's trying to access a property on a List object rather than a joint object.

After fixing this problem, the procedure hangs, with the error: This method can only be called from the thread that started the viewer. The Genesis viewer appears to be thread-bound, meaning it can only be updated from the thread that created it. This is a common limitation in GUI frameworks and visualization libraries.

### Steps to Reproduce

For example you can try to visualize a mesh `.obj`, with this command:  gs view -s 0.01 genesis/assets/meshes/Airplane/airplane.obj, you will get the error: AttributeError: 'List' object has no attribute 'type'

### Expected Behavior

The gs view cli should instead render the required mesh.

### Screenshots/Videos

<img width="705" alt="Image" src="https://github.com/user-attachments/assets/3e936000-a988-41f3-9673-3e75fad36b41" />

### Relevant log output

```shell

```

### Environment

- OS: macOS
- GPU/CPU : M4 pro
- GPU-driver version : metal Apple
- CUDA / CUDA-toolkit version: N/A

### Release version or Commit ID

SHA:9622eaa20a1980d4ca94f01e29d18854acc71565

duburcqa:
That is an actual bug indeed. I will fix it.

PR #906, opened 2025-03-21
[BUG_FIX]: gs view cli fix for macOS

antragoudaras:
## Description

Fixes gs view cli bug.

When running `genesis view` on macOS, I encountered an error related to joints iteration in the `get_movable_dofs` function. The issue stems from a platform-specific difference in how joint data is structured:

- On macOS, the `robot.joints` property returns a nested list structure (a List of Lists), while on other platforms it appears to return a flat list.
- When the code tries to iterate through `robot.joints` and access `joint.type` directly, it fails on macOS because it's trying to access a property on a List object rather than a joint object.

- I've added special handling for macOS that flattens the nested list structure before iterating through the joints.
- For other platforms, the existing approach is maintained for backward compatibility.
- This platform-specific approach ensures the code works correctly regardless of the environment.

Then a new error arise related to : The Genesis viewer appears to be thread-bound, meaning it can only be updated from the thread that created it. This is a common limitation in GUI frameworks and visualization libraries.

The issue can be fixed by separating the responsibilities:
1. Use a separate thread for entity state updates (positions, rotations)
2. Keep all viewer updates in the main thread

I've implemented a solution that:
- Uses `run_in_another_thread` for calculations and entity state updates
- Keeps the viewer refresh cycle in the main thread
- Properly handles cleanup when the viewer is closed

## Related Issue
Resolves Genesis-Embodied-AI/Genesis#905

## Motivation and Context

If you need to test any mesh, URDF files you can finally use the gs view to render it correctly.

``` bash
#on macOS after installing the environment from root
gs view genesis/assets/meshes/Airplane/airplane.obj
```
**Before fix:**
Screenshot 2025-03-21 at 19 57 18

**After fix:**
Screenshot 2025-03-21 at 19 15 51

## Checklist:

- [ ✅] I read the **CONTRIBUTING** document.
- [✅ ] I followed the `Submitting Code Changes` section of **CONTRIBUTING** document.
- [✅ ] I tagged the title correctly (including BUG FIX/FEATURE/MISC/BREAKING)
- [✅ ] I updated the [documentation](https://github.com/Genesis-Embodied-AI/genesis-doc) accordingly or no change is needed.
- [✅ ] I tested my changes and added instructions on how to test it for reviewers.

### DIFF
--- a/genesis/_main.py
+++ b/genesis/_main.py
@@ -78,7 +78,7 @@ class JointControlGUI:
 def get_movable_dofs(robot):
     motor_dofs = []
     motor_dof_names = []
-    
+
     if gs.platform == "macOS":
         # Flatten the nested list structure for macOS
         all_joints = []
@@ -87,9 +87,8 @@ def get_movable_dofs(robot):
                 all_joints.append(joint)
         joints_to_process = all_joints
     else:
-        # For other platforms, assume the structure is already flat
         joints_to_process = robot.joints
-    
+
     for joint in joints_to_process:
         if joint.type == gs.JOINT_TYPE.FREE:
             continue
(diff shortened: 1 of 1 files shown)


duburcqa:
See https://github.com/Genesis-Embodied-AI/Genesis/pull/912.

PR #911, opened 2025-03-22
[MISC] Get rid of joint world pos/quat getters.

duburcqa:
## Description

This PR gets rid of the buggy and confusing joint pose getters.

## Related Issue

Resolves Genesis-Embodied-AI/Genesis/issues/772

## Motivation and Context

Exposing the cartesian pose of a joint in world is a bad idea because it is poorly defined (i.e. before or after applying transform corresponding to the general coordinate of the joint). Moreover, it has weak physical meaning and is rarely what the user would expect. Removing these getters (which are currently buggy after merging compound joints anyway) would encourage the user to rely on link poses instead.

## Checklist:


YilingQiao (approved):

genesis/engine/entities/rigid_entity/rigid_joint.py
```diff
@@ -73,86 +73,6 @@ def __init__(
             self._dofs_damping = np.zeros_like(self._dofs_damping)
             self._dofs_armature = np.zeros_like(self._dofs_armature)
 
-    # ------------------------------------------------------------------------------------
-    # -------------------------------- real-time state -----------------------------------
-    # ------------------------------------------------------------------------------------
-
-    @gs.assert_built
-    def get_pos(self):
```
Kashu7100, line 81: should we add `gs.raise_exception('this interface is removed')` to let users know this function was deleted by intention?

Kashu7100 (approved):

genesis/engine/entities/rigid_entity/rigid_joint.py
duburcqa: Good idea! Will do

PR #912, opened 2025-03-22
 [BUG FIX] Fix Genesis standalone viewer tool.

duburcqa:
## Description

This PR enforces the MacOS-friendly viewer management approach on all platforms whenever it is necessary. This means that all GUI processes are either managed by the main thread or completely separated processes instead of using child threads.

## Related Issue

Resolves Genesis-Embodied-AI/Genesis#905
Supersedes Genesis-Embodied-AI/Genesis/pull/906

## Motivation and Context

Having to maintain branching between MacOS and other platforms is problematic and must be avoided as much as possible.

Running `gs view genesis/assets/xml/franka_emika_panda/panda.xml`

## Checklist:

### DIFF
--- a/examples/drone/fly.py
+++ b/examples/drone/fly.py
@@ -1,4 +1,6 @@
 import argparse
+import threading
+
 import numpy as np
 
 import genesis as gs
@@ -7,7 +9,6 @@ import genesis as gs
 def main():
     parser = argparse.ArgumentParser()
     parser.add_argument("-v", "--vis", action="store_true", default=False)
-    parser.add_argument("-m", "--mac", action="store_true", default=False)
     args = parser.parse_args()
 
     ########################## init ##########################
@@ -43,15 +44,6 @@ def main():
     ########################## build ##########################
     scene.build()
 
-    if args.mac:
-        gs.tools.run_in_another_thread(fn=run_sim, args=(scene, drone, args.vis))
-        if args.vis:
-            scene.viewer.start()
-    else:
-        run_sim(scene, drone, args.vis)
-
-
-def run_sim(scene, drone, enable_vis):
     traj = np.array(
         [
             [1.0, 1.0, 0.98824805, 1.0],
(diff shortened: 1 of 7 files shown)


YilingQiao (approved):
Works fine on my Linux machine

gaford99:
`[WARNING] Non-Linux system detected. In order to use the interactive viewer, you need to manually run simulation in a separate thread and then start viewer. See `examples/render_on_macos.py` `

Can you add documentation on doing this? I am thinking of just opening on a VM, but would rather be able to open from my mac.

duburcqa:
Do you really need to get around this limitation ? I’m using macOS on a daily basis and since the interactive viewer is working fine as long as the simulation is running, everything is fine for me. If I specifically need the play with the viewer and analyze the final state of the simulation, then I just do `scene.viewer.run()`.

duburcqa:
But it seems that the warning you mention refers to an old version of Genesis (basically the one available on PyPI). I would recommend installing the latest version following the official README instructions.

gaford99:
> Do you really need to get around this limitation ? I’m using macOS on a daily basis and since the interactive viewer is working fine as long as the simulation is running, everything is fine for me. If I specifically need the play with the viewer and analyze the final state of the simulation, then I just do `scene.viewer.run()`.

Hm, I see. I was hoping more for documentation on how to use the interactive viewer... It does not appear for me. When I try scene.viewer.run(), I get an error `AttributeError: 'Viewer' object has no attribute 'run'`. I think it is just a matter of not knowing how to run sim in another thread and then start the viewer, it is a skill issue. 

FYI: Updated following the latest instructions, and I get the same error.

duburcqa:
There is definitely something wrong with your install. Your are still using the old version 0.2.1 for some reason.

duburcqa:
For the record, here is the example to run the simulation in another thread https://github.com/Genesis-Embodied-AI/Genesis/blob/29e1ed57b6c8ed1a4c14bf320a261826f33ed899/examples/render_async.py#L43

Issue #129, opened 2024-12-20
Mimic Support for URDF & Equality Support for MJCF


COMoER:
Thanks for releasing such an awesome simulator for robotics researchers. For simulation of the gripper or dexterous hand, the complex linkage structures are often simplified as underactuated joints, which are defined as the mimic attribute or equality attribute in URDF or MJCF respectively. Is there any support for mimic joints in Genesis? I would appreciate it if you could reply.

Richal13Yu:
Same question here

RaneWouters:
Same question here.

zhouxian:
Yes this could be a helpful feature and should be relatively simple to implement

YilingQiao:
We now support MJCF's equality constraint and will enhance its compatibility with ball joints in a future PR.
https://github.com/Genesis-Embodied-AI/Genesis/pull/636

PR #919, opened 2025-03-24
[FEATURE] Add equality joint constraint

YilingQiao:
## Description

Add support for joint equality constraint in MJCF. 

## Related Issue

#129 

Will support Mimic constraint in URDF and WELD constraint in MJCF in following PRs.

## Checklist:


genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -1174,13 +1174,19 @@ def _init_equality_fields(self):
         )
         if self.n_equalities > 0:
             equalities = self.equalities
+
+            equalities_sol_params = np.array([equality.sol_params for equality in equalities], dtype=gs.np_float)
+            equalities_sol_params = _sanitize_sol_params(
+                equalities_sol_params, self._substep_dt, self._sol_contact_resolve_time
```
duburcqa: You should not specify `self._sol_contact_resolve_time` because equality constraints are NOT contact. Just leave this field empty: `_sanitize_sol_params(equalities_sol_params, self._substep_dt)`

genesis/assets/xml/test_equality.xml
```diff
@@ -0,0 +1,23 @@
+<mujoco model="equality_constraint_test">
```
duburcqa: Could you generate this file procedurally instead?

genesis/assets/xml/test_equality.xml
```diff
@@ -0,0 +1,23 @@
...
+    <compiler angle="degree"/>
+    <option timestep="0.002"/>
```
duburcqa: The timestet seems to be very small

genesis/assets/xml/test_equality.xml
duburcqa:
```python
def mimic_hinges(asset_tmp_path):
    mjcf = ET.Element("mujoco", model="mimic_hinges")
    ET.SubElement(mjcf, "compiler", angle="degree")
    ET.SubElement(mjcf, "option", timestep="0.002")
    worldbody = ET.SubElement(mjcf, "worldbody")
    parent = ET.SubElement(worldbody, "body", name="parent", pos="0 0 1.0")
    child1 = ET.SubElement(parent, "body", name="child1", pos="0.5 0 0")
    ET.SubElement(child1, "geom", type="capsule", size="0.05 0.2", rgba="0.9 0.1 0.1 1")
    ET.SubElement(child1, "joint", type="hinge", name="joint1", axis="0 1 0", range="-45 45")
    child2 = ET.SubElement(parent, "body", name="child2", pos="0 0.5 0")
    ET.SubElement(child2, "geom", type="capsule", size="0.05 0.2", rgba="0.1 0.1 0.9 1")
    ET.SubElement(child2, "joint", type="hinge", name="joint2", axis="0 1 0", range="-45 45")
    equality = ET.SubElement(mjcf, "equality")
    ET.SubElement(equality, "joint", name="joint_equality", joint1="joint1", joint2="joint2")
    return mjcf
```

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -351,20 +351,14 @@ def _load_MJCF(self, morph, surface):
 
         for i_e in range(mj.neq):
             e_info = mju.parse_equality(mj, i_e, morph.scale, ordered_links_idx)
-            if e_info["type"] == gs.EQUALITY_TYPE.CONNECT:  # only this type is supported right now
```
duburcqa, line 354: Missing error for `WELD` equality type

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -573,20 +567,23 @@ def _add_by_info(self, l_info, j_infos, g_infos, morph, surface):
...
-        self, name, type, link1_idx, link2_idx, anchor1_pos, anchor2_pos, rel_pose, torque_scale, sol_params
-    ):
+    def _add_equality(self, name, type, eq_obj1id, eq_obj2id, eq_data, sol_params):
+        if type == gs.EQUALITY_TYPE.CONNECT:
+            eq_obj1id += self._link_start
+            eq_obj2id += self._link_start
+        elif type == gs.EQUALITY_TYPE.JOINT:
+            eq_obj1id += self._joint_start
+            eq_obj2id += self._joint_start
+        else:
```
duburcqa, line 579: Why just pass ?

genesis/utils/mjcf.py
```diff
@@ -209,6 +209,7 @@ def parse_link(mj, i_l, scale):
     l_info["invweight"] /= scale**3
     for j_info in j_infos:
         j_info["pos"] *= scale
+    j_infos = [j_info for j_info in j_infos if j_info["n_dofs"] > 0]
```
duburcqa, line 213: Add comment to explain what you are doing and why this is important.

tests/test_rigid_physics.py
```diff
@@ -430,3 +431,20 @@ def test_nonconvex_collision(show_viewer):
...
         scene.viewer.stop()
+
+
+@pytest.mark.parametrize("xml_path", ["xml/test_equality.xml"])
+@pytest.mark.parametrize("gs_solver", [gs.constraint_solver.CG])
+@pytest.mark.parametrize("gs_integrator", [gs.integrator.Euler])
+@pytest.mark.parametrize("backend", [gs.cpu], indirect=True)
+def test_equality_joint(gs_sim, mj_sim):
+    # there is an equality constraint
+    np.testing.assert_allclose(gs_sim.rigid_solver.n_equalities, 1, 0.1)
```
duburcqa: `assert gs_sim.rigid_solver.n_equalities == 1` would do fine in this case, or `assert (gs_sim.rigid_solver.n_equalities == 1).all()` in case it is a vector.

tests/test_rigid_physics.py
```diff
@@ -430,3 +431,20 @@ def test_nonconvex_collision(show_viewer):
...
+
+    qpos = np.array((0.0, -1.0))
+    qvel = np.array((1, -0.3))
+    simulate_and_check_mujoco_consistency(gs_sim, mj_sim, qpos, qvel, num_steps=150)
+
+    # check if the two joints are equal
+    gs_qpos = gs_sim.rigid_solver.qpos.to_numpy()[:, 0]
+    np.testing.assert_allclose(gs_qpos[0], gs_qpos[1], atol=1e-3)
```
duburcqa: You cannot do better than that in terms of precision ? This is pretty bad.

duburcqa:
Could you pin existing issues related to this PR if any ? This is important for tracking.

genesis/engine/entities/rigid_entity/rigid_entity.py
YilingQiao: yeah, I want to add WELD in the next PR.

genesis/engine/entities/rigid_entity/rigid_entity.py
YilingQiao: Will implement for WELD in next PR.

genesis/engine/entities/rigid_entity/rigid_entity.py
duburcqa: Yes, so the check should be kept for now 😄

tests/test_rigid_physics.py
YilingQiao: max difference is like 2.54518715e-05. Since this is not a hard constraint, we should not expect it to be very small.

duburcqa: I see. But it is supposed to hard even if approximated as soft in practice. That is what the user would expect. Does increasing the simulation help to converge? I think it should be the case. If so, then increase the tolerance. I think 1e-5 would be good enough.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
YilingQiao: add a warning "contact_resolve_time is deprecated. Please use constraint_resolve_time instead."

tests/test_rigid_physics.py
YilingQiao: Sure. Increase the simulation step and it can reach 1e-9.

genesis/options/solvers.py
```diff
@@ -163,7 +165,9 @@ class RigidOptions(Options):
     sparse_solve : bool, optional
         Whether to exploit sparsity in the constraint system. Defaults to False.
     contact_resolve_time : float, optional
-        Time to resolve a contact. The smaller the value, the more stiff the constraint. Defaults to 0.02. (called timeconst in https://mujoco.readthedocs.io/en/latest/modeling.html#solver-parameters)
+        Please note that this argument will be deprecated in a future version. Use constraint_resolve_time instead. Time to resolve a contact. The smaller the value, the more stiff the constraint. Defaults to 0.02. (called timeconst in https://mujoco.readthedocs.io/en/latest/modeling.html#solver-parameters)
```
duburcqa: For the next PR, remove the original doc. It is already explaining that one must refer to constraint_resolve_time instead.

genesis/utils/mjcf.py
duburcqa: Mujoco removes all fixed joints ? Ok but do you get their pose ? This may be convenient for the user. What about rather filtering the fixed joints in the unit tests instead ?

Could you check if this is true for the head of the stickman in the unit tests (aka "humanoid")?

tests/test_rigid_physics.py
```diff
@@ -39,6 +39,24 @@ def box_plan():
     return mjcf
 
 
+@pytest.fixture(scope="session")
+def mimic_hinges(asset_tmp_path):
```
duburcqa: remove unused input argument

genesis/engine/entities/rigid_entity/rigid_entity.py
YilingQiao:
```
            # only those two types of equality are supported
            if e_info["type"] == gs.EQUALITY_TYPE.CONNECT or e_info["type"] == gs.EQUALITY_TYPE.JOINT: 
```

genesis/utils/mjcf.py
YilingQiao: yeah, mujoco does not fixed joint.

genesis/options/solvers.py
YilingQiao: removed

tests/test_rigid_physics.py
YilingQiao: done

duburcqa (approved):

PR #946, opened 2025-03-27
Update pyproject.toml Add torch to dependencies.

ccoulombe:
Fixes import error as torch is not installed.

## Description
Fixes
```
  File "/tmp/coulombc/27836/lib/python3.11/site-packages/genesis/__init__.py", line 17, in <module>
    import torch
ModuleNotFoundError: No module named 'torch'
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,6 +11,7 @@ description = "A universal and generative physics engine"
 readme = "README.md"
 requires-python = ">=3.10,<3.13"
 dependencies = [
+    "torch",
     "psutil",
     "scikit-image",
     "taichi >= 1.7.2",


duburcqa:
torch must be installed independently for now, as explained in README file or online documentation.

ccoulombe:
Having it as a dependency solves exactly that purpose in my opinion.

Otherwise, for the user, a proper error message should be raised when torch is imported. 
For instance, short and simple : `torch was not found. Please first install it following <url>`.

duburcqa:
> Having it as a dependency solves exactly that purpose in my opinion.

No it doesn't. There are many different versions that are not available on PyPI depending on your setup. As a result, torch cannot be added in requirement otherwise the wrong version would be installed:

image

> Otherwise, for the user, a proper error message should be raised when torch is imported.

Yes, this is a good idea.

ccoulombe:
Oui ;) since the requirements is `torch` it means any versions, not a specific version. Plus, all torch wheels are named `torch`, so no matter which one the users decides to install, it would fulfill the need of having torch installed. 
If one decides to install the cpu variant for example, it is still named `torch` and has a local version (`+cpu`), and so pip still sees the requirements as fulfilled.

duburcqa:
Yes, but any package specified in requirements will be installed automatically if not already installed. This is what we want to avoid. There is no way to add a requirement that is only used to make sure all dependencies are there without installing them if some are missing.

ccoulombe:
I've made a PR for an error message

PR #947, opened 2025-03-27
Raise exception with message when torch is missing

ccoulombe:
## Description
Raise informative message on missing dependency

## Related Issue

See #946 for discussion.

### DIFF
--- a/genesis/__init__.py
+++ b/genesis/__init__.py
@@ -20,7 +20,10 @@ def fake_print(*args, **kwargs):
 with patch("builtins.print", fake_print):
     import taichi as ti
 
-import torch
+try:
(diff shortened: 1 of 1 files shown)


genesis/__init__.py
```diff
@@ -20,7 +20,10 @@ def fake_print(*args, **kwargs):
 with patch("builtins.print", fake_print):
     import taichi as ti
 
-import torch
+try:
+    import torch
+except 
+    raise Exception("torch module was not found. Please install pytorch: https://pytorch.org/get-started/locally/")
```
duburcqa, line 26: `gs.raise_exception`

Issue #625, opened 2025-01-22
[Bug]: Unitree Go2 gets “launched away” when spawned on complex terrain generated by the terrain generator


Adwaver4157:
### Bug Description

Hello, I’m currently encountering an issue with the Unitree Go2 robot model when using a terrain generator(https://github.com/leggedrobotics/terrain-generator) to create complex terrains. As soon as I spawn the Go2 on this generated mesh, the robot instantly flies off to some distant location. Calling the reset function afterwards does not resolve the problem—once it’s launched away, the robot cannot be properly reset.

Context / Steps to Reproduce:

I have a trained model (using go2_eval.py) that was originally trained on a flat plane.
In go2_env.py, I load a mesh generated by the terrain generator to create a more complex terrain.
Immediately after launching the simulator in go2_eval.py, the robot gets flung away, and calling reset() does not fix the situation.
Additional Details:

The training on a flat plane works without issues.
The problem only arises when using the newly generated terrain mesh.
Once the robot is “launched” upon spawn, it’s impossible to reset it properly.
Any guidance or suggestions on how to prevent this unexpected behavior would be greatly appreciated. Thank you!

[Launched Unitree Go2](https://drive.google.com/file/d/1zVtNEZcH-OG30PuajSKPUHY4hXA61qHz/view?usp=sharing)
[Mesh I used](https://drive.google.com/file/d/1vHEbWtCdI7oYplFvtRH2S5yP0vQgPus7/view?usp=sharing)

### Steps to Reproduce

```python
# I only modified go2_env.py to add complex terrain mesh.
import torch
import math
import genesis as gs
from genesis.utils.geom import quat_to_xyz, transform_by_quat, inv_quat, transform_quat_by_quat


def gs_rand_float(lower, upper, shape, device):
    return (upper - lower) * torch.rand(size=shape, device=device) + lower


class Go2Env:
    def __init__(self, num_envs, env_cfg, obs_cfg, reward_cfg, command_cfg, show_viewer=False, device="cuda"):
        self.device = torch.device(device)

        self.num_envs = num_envs
        self.num_obs = obs_cfg["num_obs"]
        self.num_privileged_obs = None
        self.num_actions = env_cfg["num_actions"]
        self.num_commands = command_cfg["num_commands"]

        self.simulate_action_latency = True  # there is a 1 step latency on real robot
        self.dt = 0.02  # control frequency on real robot is 50hz
[...]

PR #1033, opened 2025-04-16
[FEATURE] Add a function to convert terrain mesh to height field

YilingQiao:
## Description

Sometimes user might want to import a non-convex terrain represented by a mesh. The best practice is to convert that mesh to a height field for more accurate collision detection. 
A function based on ray-casting is added for this purpose.
Try 
```bash
python examples/rigid/terrain_from_mesh.py -v
```

## Related Issue

Resolves Genesis-Embodied-AI/Genesis/issues/790 
Resolves Genesis-Embodied-AI/Genesis/issues/491 

Related to Genesis-Embodied-AI/Genesis/issues/690 
Related to Genesis-Embodied-AI/Genesis/issues/625 

## Checklist:

### DIFF
--- a/examples/rigid/terrain_from_mesh.py
+++ b/examples/rigid/terrain_from_mesh.py
@@ -24,7 +24,7 @@ def main():
 
     horizontal_scale = 2.0
     path_terrain = "./genesis/assets/meshes/terrain_45.obj"
-    hf_terrain, xs, ys = mesh_to_heightfield(path_terrain, spacing=horizontal_scale, oversample=3)
+    hf_terrain, xs, ys = mesh_to_heightfield(path_terrain, spacing=horizontal_scale, oversample=1)
     print("hf_terrain", hf_terrain.shape, np.max(hf_terrain))
 
     # default heightfield starts at 0, 0, 0
--- a/examples/rigid/terrain_from_mesh.py
+++ b/examples/rigid/terrain_from_mesh.py
@@ -1,7 +1,8 @@
-import genesis as gs
-import numpy as np
 import argparse
 import os
+
+import genesis as gs
+import numpy as np
 from genesis.utils.terrain import mesh_to_heightfield
 
 
@@ -16,11 +17,11 @@ def main():
 
     ########################## create a scene ##########################
     scene = gs.Scene(
-        show_viewer=args.vis,
         viewer_options=gs.options.ViewerOptions(
             camera_pos=(0, -50, 0),
             camera_lookat=(0, 0, 0),
         ),
+        show_viewer=args.vis,
     )
 
     horizontal_scale = 2.0
@@ -41,14 +42,6 @@ def main():
             pos=translation,
         ),
         vis_mode="collision",
-    )
-
-    terrain_ori = scene.add_entity(
-        morph=gs.morphs.Mesh(
-            file=path_terrain,
-            fixed=True,
-            convexify=False,
-        ),
     )
 
     ball = scene.add_entity(
--- a/examples/rigid/terrain_from_mesh.py
+++ b/examples/rigid/terrain_from_mesh.py
@@ -16,10 +16,7 @@ def main():
 
     ########################## create a scene ##########################
     scene = gs.Scene(
-        show_viewer=True,
-        sim_options=gs.options.SimOptions(
-            gravity=(2, 0, -2),
-        ),
+        show_viewer=args.vis,
         viewer_options=gs.options.ViewerOptions(
             camera_pos=(0, -50, 0),
             camera_lookat=(0, 0, 0),
@@ -29,9 +26,8 @@ def main():
     horizontal_scale = 2.0
     gs_root = os.path.dirname(os.path.abspath(gs.__file__))
     path_terrain = os.path.join(gs_root, "assets", "meshes", "terrain_45.obj")
-    print("path_terrain", path_terrain)
     hf_terrain, xs, ys = mesh_to_heightfield(path_terrain, spacing=horizontal_scale, oversample=1)
-    print("hf_terrain", hf_terrain.shape, np.max(hf_terrain))
+    print("hf_terrain", path_terrain, hf_terrain.shape, np.max(hf_terrain))
 
     # default heightfield starts at 0, 0, 0
     # translate to the center of the mesh
@@ -65,15 +61,8 @@ def main():
 
     scene.build()
 
-    from IPython import embed
-
-    embed()
     for i in range(2000):
         scene.step()
-
-        qvel = scene.sim.rigid_solver.dofs_state.vel.to_numpy()[:, 0]
-        qvel_norm = np.linalg.norm(qvel, axis=-1)
-        print("qvel_norm", i, qvel_norm)
 
 
 if __name__ == "__main__":
--- a/examples/rigid/terrain_from_mesh.py
+++ b/examples/rigid/terrain_from_mesh.py
@@ -1,6 +1,7 @@
 import genesis as gs
 import numpy as np
 import argparse
+import os
 from genesis.utils.terrain import mesh_to_heightfield
 
 
@@ -16,6 +17,9 @@ def main():
     ########################## create a scene ##########################
     scene = gs.Scene(
         show_viewer=True,
+        sim_options=gs.options.SimOptions(
+            gravity=(2, 0, -2),
+        ),
         viewer_options=gs.options.ViewerOptions(
             camera_pos=(0, -50, 0),
             camera_lookat=(0, 0, 0),
@@ -23,7 +27,9 @@ def main():
     )
 
     horizontal_scale = 2.0
-    path_terrain = "./genesis/assets/meshes/terrain_45.obj"
+    gs_root = os.path.dirname(os.path.abspath(gs.__file__))
+    path_terrain = os.path.join(gs_root, "assets", "meshes", "terrain_45.obj")
+    print("path_terrain", path_terrain)
     hf_terrain, xs, ys = mesh_to_heightfield(path_terrain, spacing=horizontal_scale, oversample=1)
     print("hf_terrain", hf_terrain.shape, np.max(hf_terrain))
 
@@ -51,16 +57,23 @@ def main():
 
     ball = scene.add_entity(
         gs.morphs.Sphere(
-            pos=(10, 15, 18),
-            radius=2,
+            pos=(10, 15, 10),
+            radius=1,
         ),
         vis_mode="collision",
     )
 
     scene.build()
 
+    from IPython import embed
+
+    embed()
     for i in range(2000):
         scene.step()
+
+        qvel = scene.sim.rigid_solver.dofs_state.vel.to_numpy()[:, 0]
+        qvel_norm = np.linalg.norm(qvel, axis=-1)
+        print("qvel_norm", i, qvel_norm)
 
 
 if __name__ == "__main__":
--- a/examples/rigid/terrain_from_mesh.py
+++ b/examples/rigid/terrain_from_mesh.py
@@ -1,0 +1,67 @@
+import genesis as gs
+import numpy as np
+import argparse
+from genesis.utils.terrain import mesh_to_heightfield
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-v", "--vis", action="store_true", default=False)
+    parser.add_argument("-c", "--cpu", action="store_true", default=False)
+    args = parser.parse_args()
+
+    ########################## init ##########################
+    gs.init(backend=gs.cpu if args.cpu else gs.gpu)
+
+    ########################## create a scene ##########################
+    scene = gs.Scene(
+        show_viewer=True,
+        viewer_options=gs.options.ViewerOptions(
+            camera_pos=(0, -50, 0),
+            camera_lookat=(0, 0, 0),
+        ),
+    )
+
+    horizontal_scale = 2.0
+    path_terrain = "./genesis/assets/meshes/terrain_45.obj"
+    hf_terrain, xs, ys = mesh_to_heightfield(path_terrain, spacing=horizontal_scale, oversample=3)
+    print("hf_terrain", hf_terrain.shape, np.max(hf_terrain))
+
+    # default heightfield starts at 0, 0, 0
+    # translate to the center of the mesh
+    translation = np.array([np.nanmin(xs), np.nanmin(ys), 0])
+
+    terrain_heightfield = scene.add_entity(
+        morph=gs.morphs.Terrain(
+            horizontal_scale=horizontal_scale,
+            vertical_scale=1.0,
+            height_field=hf_terrain,
+            pos=translation,
+        ),
+        vis_mode="collision",
+    )
+
+    terrain_ori = scene.add_entity(
+        morph=gs.morphs.Mesh(
+            file=path_terrain,
+            fixed=True,
+            convexify=False,
+        ),
+    )
+
+    ball = scene.add_entity(
+        gs.morphs.Sphere(
+            pos=(10, 15, 18),
+            radius=2,
+        ),
+        vis_mode="collision",
+    )
+
+    scene.build()
+
+    for i in range(2000):
+        scene.step()
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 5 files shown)


duburcqa:
Just missing a unit test :p

YilingQiao:
> Just missing a unit test :p

done

YilingQiao:
Unit test added @duburcqa

examples/rigid/terrain_from_mesh.py
```diff
@@ -0,0 +1,69 @@
...
+            pos=translation,
+        ),
+        vis_mode="collision",
+    )
+
+    terrain_ori = scene.add_entity(
+        morph=gs.morphs.Mesh(
+            file=path_terrain,
+            fixed=True,
+            convexify=False,
```
duburcqa: Why `convexify=False`? Since it is a plane that as no volume, it should not be convexified anyway. I think it is better to not specify arguments that are doing nothing.

genesis/utils/terrain.py
```diff
@@ -1,5 +1,5 @@
 import os
-
+import math
 import numpy as np
 import pickle
 import trimesh
```
duburcqa, line 6: Keep system deps seperated from common deps:
```
import os
import math
import pickle

import numpy as np
import trimesh
```

genesis/utils/terrain.py
```diff
@@ -352,3 +352,83 @@ def convert_heightfield_to_watertight_trimesh(height_field_raw, horizontal_scale
...
+        Highest elevation per coarse cell (NaN if no hit).
+    xs      : (nx,)     float32 ndarray
+    ys      : (ny,)     float32 ndarray
+        Coordinates of grid lines (cell centres).
+
+    Notes
+    -----
+    • Requires `pyembree` or `rtree` for fast ray‑tracing.
+    • Memory cost grows as  oversample².  For very large terrains keep
+      `oversample` ≤ 4 unless you have ample RAM.
```
duburcqa: "unless you have ample RAM" if vague. Do you have any rough estimate? (>64Gb, >512Gb...)

genesis/utils/terrain.py
```diff
@@ -352,3 +352,83 @@ def convert_heightfield_to_watertight_trimesh(height_field_raw, horizontal_scale
...
+    # -------------------------------- axis handling ---------------------------
+    if up_axis.lower() == "y":
+        # rotate so Z becomes up (‑90° around X)
+        T = trimesh.transformations.rotation_matrix(np.deg2rad(-90), [1, 0, 0])
+        mesh.apply_transform(T)
+
+    (minx, miny, _), (maxx, maxy, maxz) = mesh.bounds
+
+    # -------------------------------- coarse grid ----------------------------
+    dx, dy = (spacing, spacing) if isinstance(spacing, (int, float)) else spacing
```
duburcqa: It would be nice to avoid duplicating every lines for x and y:
```python
from collections.abc import ValuesView, Sequence

if np.isscalar(spacing):
    spacing = (spacing, spacing)

n, s, f, f_lin = [None, None], [None, None], [None, None], [None, None]
for i, (vmin, vmax, delta) in range(2):
    # -------------------------------- coarse grid ----------------------------
    vmin, vmax = mesh.bounds[0][i], mesh.bounds[1][i]
    n[i] = math.ceil((vmax - vmin) / spacing[i]) + 1
    s[i] = np.linspace(vmin, vmax, n[i], dtype=np.float32)

    # -------------------------------- fine grid ------------------------------
    f[i] =  n[i] * oversample
    f_lin[i] = np.linspace(vmin, vmax, f[i], dtype=np.float32)

f_grid = np.meshgrid(*f_lin, indexing="ij")
```

genesis/utils/terrain.py
```diff
@@ -352,3 +352,83 @@ def convert_heightfield_to_watertight_trimesh(height_field_raw, horizontal_scale
...
+    ys = np.linspace(miny, maxy, ny, dtype=np.float32)
+
+    # -------------------------------- fine grid ------------------------------
+    fx = nx * oversample
+    fy = ny * oversample
+    fx_lin = np.linspace(minx, maxx, fx, dtype=np.float32)
+    fy_lin = np.linspace(miny, maxy, fy, dtype=np.float32)
+    fxx, fyy = np.meshgrid(fx_lin, fy_lin, indexing="ij")
+
+    origins = np.c_[fxx.ravel(), fyy.ravel(), np.full(fxx.size, maxz + 1.0)]
```
duburcqa: I  would recommend `np.stack((fxx.ravel(), fyy.ravel(), np.full(fxx.size, maxz + 1.0)), axis=-1)` which is more readable and widley used.

genesis/utils/terrain.py
```diff
@@ -352,3 +352,83 @@ def convert_heightfield_to_watertight_trimesh(height_field_raw, horizontal_scale
...
+    h_fine = np.full(fxx.size, np.nan, dtype=np.float32)
+    h_fine[hit_ids] = locs[:, 2]
+    h_fine = h_fine.reshape((fx, fy))  # (fx, fy) = (nx*k, ny*k)
+
+    # -------------------------------- down‑sample ----------------------------
+    # reshape to (nx, k, ny, k) then take max over the 2 fine axes
+    h_coarse = np.nanmax(h_fine.reshape(nx, oversample, ny, oversample).swapaxes(1, 2), axis=(2, 3))
+
+    # change nan to min
+    h_coarse = np.where(np.isnan(h_coarse), np.nanmin(h_coarse), h_coarse)
```
duburcqa: why not `mesh.bounds[0][2]` ?

genesis/utils/terrain.py
```diff
@@ -352,3 +352,83 @@ def convert_heightfield_to_watertight_trimesh(height_field_raw, horizontal_scale
...
+    -------
+    heights : (nx, ny)  float32 ndarray
+        Highest elevation per coarse cell (NaN if no hit).
+    xs      : (nx,)     float32 ndarray
+    ys      : (ny,)     float32 ndarray
+        Coordinates of grid lines (cell centres).
+
+    Notes
+    -----
+    • Requires `pyembree` or `rtree` for fast ray‑tracing.
```
duburcqa: I would remove this comment entirely. `rtree` is already installed automatically as part of the requirements.

tests/test_rigid_physics.py
```diff
@@ -4,6 +4,7 @@
 import pytest
 import trimesh
```
duburcqa, line 6: Could you split dependencies as mentioned above?

tests/test_rigid_physics.py
```diff
@@ -1100,3 +1101,57 @@ def test_equality_weld(gs_sim, mj_sim):
     qpos = gs_sim.rigid_solver.dofs_state.pos.to_numpy()[:, 0]
     qpos[0], qpos[1], qpos[2] = 0.1, 0.1, 0.1
     simulate_and_check_mujoco_consistency(gs_sim, mj_sim, qpos, qvel, num_steps=300, atol=atol)
+
+
+@pytest.mark.parametrize("backend", [gs.cpu])
+def test_mesh_to_heightfield(show_viewer):
+    from genesis.utils.terrain import mesh_to_heightfield
```
duburcqa: Just import this on top if it cannot fail.

tests/test_rigid_physics.py
```diff
@@ -1100,3 +1101,57 @@ def test_equality_weld(gs_sim, mj_sim):
...
+        vis_mode="collision",
+    )
+
+    scene.build()
+
+    for i in range(1000):
+        scene.step()
+
+    # speed is around 0
+    qvel = scene.sim.rigid_solver.dofs_state.vel.to_numpy()[:, 0]
```
duburcqa: `qvel = ball.get_dofs_velocity().cpu()`

tests/test_rigid_physics.py
```diff
@@ -1100,3 +1101,57 @@ def test_equality_weld(gs_sim, mj_sim):
...
+    qvel_norm = np.linalg.norm(qvel, axis=-1)
```
duburcqa: Remove the norm, it is useless

tests/test_rigid_physics.py
```diff
@@ -1100,3 +1101,57 @@ def test_equality_weld(gs_sim, mj_sim):
...
+    np.testing.assert_allclose(qvel_norm, 0, atol=1e-2)
```
duburcqa: How can this condition be met if the ground is not horizontal ?!

tests/test_rigid_physics.py
```diff
@@ -1100,3 +1101,57 @@ def test_equality_weld(gs_sim, mj_sim):
...
+        viewer_options=gs.options.ViewerOptions(
+            camera_pos=(0, -50, 0),
+            camera_lookat=(0, 0, 0),
+        ),
+    )
+
+    horizontal_scale = 2.0
+    gs_root = os.path.dirname(os.path.abspath(gs.__file__))
+    path_terrain = os.path.join(gs_root, "assets", "meshes", "terrain_45.obj")
+    hf_terrain, xs, ys = mesh_to_heightfield(path_terrain, spacing=horizontal_scale, oversample=1)
```
duburcqa: Ideally, we could like to automate this process, as for `convexify` but for terrains. So that mesh terrain are automatically converted to heightfield unless specified otherwise. That would be a neat API!

tests/test_rigid_physics.py
duburcqa: For this basic unit test, I would recommend check if the simulation gives the same result with or without heightmap conversion. Since the mesh is very basic, is should work.

examples/rigid/terrain_from_mesh.py
YilingQiao: I added this plane just to check whether it overlaps with the terrain I built. Feel free to remove it if you think it looks better without it.

genesis/utils/terrain.py
YilingQiao: done

genesis/utils/terrain.py
YilingQiao: I think explicitly writing x and y is better for readability.

examples/rigid/terrain_from_mesh.py
duburcqa: No I think it is fine, I was just wondering about the argument `convexify=False`.

genesis/utils/terrain.py
duburcqa: As you wish! 

Still, I think it is better to pre-process input arguments that can have multiple types on top of the function, so that everywhere within the body of the function the type is unique and well-known. Moreover, checking if int or float is not very robust, for example is would fail with numpy scalar types. Basically, just this part is important:
```python
if np.isscalar(spacing):
    spacing = (spacing, spacing)
```

tests/test_rigid_physics.py
YilingQiao: The gravity is `(2, 0, -2)`, orthogonal to the terrain

tests/test_rigid_physics.py
YilingQiao: I want to compute the magnitude of the velocity, because it's expected to 0.

tests/test_rigid_physics.py
duburcqa: Nice trick 👍

tests/test_rigid_physics.py
duburcqa: If all components are zero, the norm would be zero.

duburcqa:
Although the feature you are adding is definitely worth it, It would be nice to spend more time investigating what is happening with SDF non-convex collision detection. It should not explode as it does. If it does not work better, then one should consider disabling this option for terrain completely.

YilingQiao:
> Although the feature you are adding is definitely worth it, It would be nice to spend more time investigating what is happening with SDF non-convex collision detection. It should not explode as it does. If it does not work better, then one should consider disabling this option for terrain completely.

yeah, will check!

genesis/utils/terrain.py
YilingQiao: cool! I added 
```
if np.isscalar(spacing):
    spacing = (spacing, spacing)
```

genesis/utils/terrain.py
YilingQiao: done

genesis/utils/terrain.py
YilingQiao: this might be more explicit?

tests/test_rigid_physics.py
YilingQiao: The components are not all zero. What would be the potential problem for 

```
    qvel_norm = np.linalg.norm(qvel, axis=-1)
    np.testing.assert_allclose(qvel_norm, 0, atol=1e-2)
```

tests/test_rigid_physics.py
YilingQiao: will probably do that in the next PR.

tests/test_rigid_physics.py
duburcqa: The components are all zero if the following assert succeeds (up to some tolerance of course by this does not matters) ! That is the all point !

Mathematically, all norms are said to be "equivalent", which means that it exists some constants k1, k2 such as norm1(x) * k1 < norm2(x) < norm1(x) * k1. This implies that enforcing norm equals 0 is the same for all of them. `np.testing.assert_allclose` can be viewed as a norm itself, ie `L_{inf}-norm`. As a result, it is the same than checking `L_2-norm` as you are doing.

examples/rigid/terrain_from_mesh.py
```diff
@@ -0,0 +1,69 @@
+import genesis as gs
+import numpy as np
```
duburcqa: Fix ordering of the import.

genesis/utils/terrain.py
duburcqa: This would not only be more explicit (if you call it 'minz' of course), but also more robust (what if there are only nan values?).

examples/rigid/terrain_from_mesh.py
duburcqa: As I said, if `convexify=False` argument does nothing, just remove it.

tests/test_rigid_physics.py
duburcqa: `import os` is a system dependency unlike `pytest`, which goes with `trimesh`.

tests/test_rigid_physics.py
```diff
@@ -1100,3 +1101,56 @@ def test_equality_weld(gs_sim, mj_sim):
...
     qpos[0], qpos[1], qpos[2] = 0.1, 0.1, 0.1
     simulate_and_check_mujoco_consistency(gs_sim, mj_sim, qpos, qvel, num_steps=300, atol=atol)
+
+
+@pytest.mark.parametrize("backend", [gs.cpu])
+def test_mesh_to_heightfield(show_viewer):
+
+    ########################## create a scene ##########################
+    scene = gs.Scene(
+        show_viewer=show_viewer,
```
duburcqa, line 1111: for consistency with all the other tests, keep the various options in the same order: sim opt -> rigid opt -> others opt -> viewer opt -> show_*

tests/test_rigid_physics.py
YilingQiao: done

tests/test_rigid_physics.py
YilingQiao: why is `trimesh` more "system" than `pytest`?

duburcqa: It is not the case, that is why they go together x)

duburcqa: "system" means built-in, aka standard libraries.

There are at least 3 groups: standard libs, external dependencies, internal dependencies. Then external dependencies are usually split in 2 groups: common dependencies (numpy, torch...), exotic dependencies (taichi, cpuinfo, psutil, mujoco). In this case it is rather subjective and it is up to the contributor to do what feels right for him. The point is to make sure that the notable dependencies stands out clearly at first glance (whether the category).

duburcqa (approved):

Nate711:
@duburcqa @YilingQiao Is this issue https://github.com/Genesis-Embodied-AI/Genesis/issues/625 caused by genesis using SDFs for these nonconvex mesh terrains? I don't think isaacsim uses SDF for the same mesh terrain, probably trimesh.

PR #1045, opened 2025-04-20
[FEATURE] expose ref interface for get_vel in rigid_entiry

Kashu7100:
update doc string

## Description

Expose the `ref: Literal["link_origin", "link_com", "entity_com"] = "link_origin"` options in the RigidEntity's get_vel() method.

## Related Issue

## Checklist:

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -1681,7 +1681,9 @@ class RigidEntity(Entity):
         return self._solver.get_links_quat(self._base_links_idx, envs_idx, unsafe=unsafe).squeeze(-2)
 
     @gs.assert_built
-    def get_vel(self, envs_idx=None, *, unsafe=False):
+    def get_vel(
+        self, envs_idx=None, *, ref: Literal["link_origin", "link_com", "entity_com"] = "link_origin", unsafe=False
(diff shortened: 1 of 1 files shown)


duburcqa:
What is the actually use case of this? From my experience, link origin is what you need 99% of the time. Basically, anything else does not make much sense. But I’m ok to expose the reference point anyway.

Kashu7100:
For RL, I think sometimes it is useful to use the entity_com

duburcqa:
No I don’t think so. Unless it is not robotic related, because this quantity has no physical meaning and cannot be computed at lot on a real robot.

Kashu7100:
I think not always the link origin is a good coordinate frame to use (COM is sometimes more convenient). 
Also for a links connected with fixed joint and when we don't merge the fixed joints, I think entity's CoM can make sense.

duburcqa:
> I think not always the link origin is a good coordinate frame to use (COM is sometimes more convenient).

Do you have any use case to support this assertion? 

> Also for a links connected with fixed joint and when we don't merge the fixed joints, I think entity's CoM can make sense.

I'm not convinced. If you are so much interested in splitting a body in multiple separated parts connected by fixed joints, which requires extra effort in the first place, why would you like to virtually get them back together by computing quantities related to their aggregated CoM?

PR #886, opened 2025-03-18
[Feature] Luisa Render upgrade to support aarch64

ACMLCZH:
1. I merged the latest branch of LuisaRender into our project, aiming to provide `aarch64` support for Raytracing as discussed in issue #754.
2. Fixed some bugs in GLB mesh parsing.
3. Added multiple FEM object support for Raytracing.
4. Enabled some command line argument controls for OpenVDB surface reconstruction.

PR #971, opened 2025-04-02
Yizhou/usd

yizhouzhao:
# Add Support for Parsing Geometries from USD format.
 
This function supports reading geometries from the default prim of the USD stage.
 
Especially, for NVIDIA's : [SimReady USD dataset](https://developer.nvidia.com/omniverse/simready-assets)

# Examples 
[Chair](https://omniverse-content-production.s3.us-west-2.amazonaws.com/Assets/simready_content/common_assets/props/chair_array/chair_array.usd)
[CornerRail](https://omniverse-content-production.s3.us-west-2.amazonaws.com/Assets/simready_content/common_assets/props/cornerrail_a1/cornerrail_a1.usd)
[Fruit](https://omniverse-content-production.s3.us-west-2.amazonaws.com/Assets/simready_content/common_assets/props/pomegranate01/pomegranate01.usd)

Screenshot 2025-04-02 125025

# References:
[USD in Action](https://www.manning.com/books/universal-scene-description-in-action)
[USD in Action Github](https://github.com/linfanz/OpenUSDInAction)

PR #1051, opened 2025-04-20
[Feature] Add USD parsing

ACMLCZH:
## Description
1. Reorganized `utils/mesh.py` and separated `utils/gltf.py` for improved structure and maintainability.
2. Introduced `utils/usda.py` for `.usd` mesh file parsing. This initial implementation primarily supports `gltf` and `UsdPreviewSurface` materials, with future enhancements planned for broader compatibility.
3. Resolved an issue discussed [here](https://github.com/Genesis-Embodied-AI/Genesis/pull/886#issuecomment-2817070066).


duburcqa:
How does this PR relates to https://github.com/Genesis-Embodied-AI/Genesis/pull/971 ?

ACMLCZH:
This PR aims to serve as a more general parsing tool that supports the most common USD materials, specifically `UsdPreviewSurface` and `gltf` materials. However, both my PR and #971 currently do not fully handle MDL file parsing due to its complexity. A potential solution could involve baking materials into `UsdPreviewSurface` before parsing them, which is planned for future implementation. In this PR, my primary focus is on reorganizing `utils/mesh.py` for better structure and introducing a basic `usda` file parsing approach as an initial placeholder.

duburcqa:
OK I see. In the end, are you implementing anything redundant or both are complementary?

ACMLCZH:
It is expected to cover #971 now.

duburcqa:
Your PR is great but doing to many stuff that are not directly related to USD support. Would it be possible to split it in 2 or even 3 different PRs ? Moreover, could you add some unit test to make sure USD parsing keep working? At least not raising any exception when loading a file, and some basic checks about number of geoms, joints stuffs like this (not physics related not kinematic tree).

ACMLCZH:
I plan to create two separate PRs—one focusing on USD support and the other covering the remaining changes. Where do you think the USD parsing example should go? Should it be in the `tests` folder or the `examples` folder? Many usage examples are currently located in the `examples` folder.

duburcqa:
> Where do you think the USD parsing example should go? Should it be in the tests folder or the examples folder? Many usage examples are currently located in the examples folder.

Are you sure an example for loading USD is necessary? If it is actually an example (like the one we already have), then it should definitely go in `examples` folder. However, it does not supplement a unit test, that is mandatory, and checking that the result of loading the USD file is valid (what examples would not do as they document usage not validity).

ACMLCZH:
And if I want to add some testing mesh files, where should I put them? In the `assets` folder or in the `tests` folder?

duburcqa:
Sorry I just saw your message, I did not get any notification for some reason. For now, let's say in `genesis/assets` folder.

.gitignore
```diff
@@ -45,6 +45,7 @@ MANIFEST
 
 # debug
 examples/debug/*
+genesis/assets/usd/*
```
YilingQiao: why do you want to ignore those assets?

genesis/options/textures.py
```diff
@@ -26,6 +26,9 @@ def __init__(self, **data):
     def check_dim(self, dim):
         raise NotImplementedError
 
+    def apply_cutoff(self, cutoff):
```
YilingQiao, line 32: what's this for?

YilingQiao:
Can you add an example script and usd asset ?

genesis/utils/usda.py
```diff
@@ -0,0 +1,453 @@
...
+                assert len(prim.GetPrimStack()) == 2, f"Invalid instance {prim.GetPath()} in usd file {path}."
+                if prim.GetPrimStack()[0].hasReferences:
+                    matrix = np.array(xform_cache.GetLocalToWorldTransform(prim))
+                    instance_spec = prim.GetPrimStack()[-1]
+                    instance_list.append((matrix.T, instance_spec.layer.identifier))
+
+    return instance_list
+
+
+if __name__ == "__main__":
```
YilingQiao: Can we delete this?

tests/test_mesh.py
```diff
@@ -137,11 +262,57 @@ def check_texture(tm_color, tm_texture, gs_texture, dim, material_name, texture_
...
+                0.0,
                 3,
                 material_name,
                 "emissive",
             )
+
+
+@pytest.mark.parametrize("usd_filename", ["tests/sneaker_airforce", "tests/RoughnessTest"])
+def test_usd_parse(usd_filename):
+    glb_file = os.path.join(mu.get_assets_dir(), f"{usd_filename}.glb")
```
YilingQiao: let's put them into hugging face

doc
duburcqa, line 1: Why are you updating the documentation?

genesis/assets/tests/RoughnessTest.glb
duburcqa, line 1: Move the assets on hugging face repo.

genesis/options/surfaces.py
```diff
@@ -254,10 +254,11 @@ def get_rgba(self):
             ):
                 rgba_array = np.dstack((texture.image_array, opacity_texture.image_array))
                 rgba_scale = (*texture.image_color, *opacity_texture.image_color)
+
```
duburcqa, line 257: Do not add extra space after code block.

genesis/options/textures.py
```diff
@@ -26,28 +26,41 @@ def __init__(self, **data):
...
     """
-    A texture with a single color.
+    A texture that consists of a single color.
 
     Parameters
     ----------
-    color : tuple, shape (3,)
-        RGB color value.
+    color : list of float
+        A list of color values, stored as tuple, supporting any number of channels within the range [0.0, 1.0]. Default is (1.0, 1.0, 1.0).
```
duburcqa: Do not exceed 120 characters per line.

genesis/options/textures.py
```diff
@@ -26,28 +26,41 @@ def __init__(self, **data):
...
     """
 
-    color: tuple = (1.0, 1.0, 1.0)
+    color: Union[float, List[float]] = (1.0, 1.0, 1.0)
```
duburcqa, line 47: What is supposed to be a "float" color? Besides, do not use `List[float]` but `typing.Sequence[float]`.

genesis/options/textures.py
```diff
@@ -26,28 +26,41 @@ def __init__(self, **data):
...
+
+    def __init__(self, **data):
+        super().__init__(**data)
+        if isinstance(self.color, float):
+            self.color = (self.color,)
+        else:
+            self.color = tuple(self.color)  # Use tuple to store image color since it is more efficient
```
duburcqa, line 54: More efficient than what? A list ? Does this really make a difference for handling options?

genesis/utils/gltf.py
```diff
@@ -389,7 +389,6 @@ def parse_mesh_glb(path, group_by_material, scale, surface):
                 mesh_info.set_property(
                     surface=material, metadata={"path": path, "name": material_name if group_by_material else mesh_name}
                 )
-
```
duburcqa, line 392: Avoid this kind of changes that are just adding "noise" if possible.

tests/test_mesh.py
```diff
@@ -6,6 +6,171 @@
...
+        np.testing.assert_allclose(
+            gs_texture1.color,
+            gs_color2,
+            rtol=0,
+            atol=1e-06,
+            err_msg=f"Color mismatch for material {material_name} in {texture_name}.",
+        )
+    elif isinstance(gs_texture1, gs.textures.ImageTexture):
+        assert isinstance(gs_texture2, gs.textures.ImageTexture)
+        np.testing.assert_allclose(
```
duburcqa: Using `assert_allclose` that is defined in `utils.py`. Update this helper to support `err_msg`.

tests/test_mesh.py
```diff
@@ -6,6 +6,171 @@
...
+            gs_texture1.image_color,
+            gs_texture2.image_color,
+            rtol=0,
+            atol=1e-06,
+            err_msg=f"Color mismatch for material {material_name} in {texture_name}.",
+        )
+        np.testing.assert_array_equal(
```
duburcqa: `assert_allclose` with 0 tol should do fine.

tests/test_mesh.py
```diff
@@ -6,6 +6,171 @@
...
+    if gs_texture2 is not None:
+        gs_texture2 = gs_texture2.check_simplify()
+
+    if isinstance(gs_texture1, gs.textures.ColorTexture):
+        gs_color2 = (default_value,) * len(gs_texture1.color) if gs_texture2 is None else gs_texture2.color
+        np.testing.assert_allclose(
+            gs_texture1.color,
+            gs_color2,
+            rtol=0,
+            atol=1e-06,
```
duburcqa: Move hard-coded constant in argument or define it on top of this file using uppercase naming convention. Add a comment to justify the rational behind this value. Why not 1e-9 or even gs.EPS ?

tests/test_mesh.py
```diff
@@ -6,6 +6,171 @@
...
+            vertices2[:, 0],
+        ]
+    )
+    vertices2 = vertices2[indices2]
+    normals2 = normals2[indices2]
+    uvs2 = uvs2[indices2]
+    invdices2 = np.argsort(indices2)
+    faces2 = invdices2[gs_mesh2.trimesh.faces]
+
+    np.testing.assert_allclose(
```
duburcqa: Keep all the arguments on the same line for readability.

genesis/options/textures.py
ACMLCZH: Like roughness and metallic

genesis/options/textures.py
ACMLCZH: Than np.ndarray.

duburcqa: First, `np.ndarray` is not a supported type based on type annotation so you should update them. Second, I'm not convinced speed is a real concern here. But making sure that ultimately `self.color` has a unique type is a good reason for converting the array already.

genesis/assets/tests/RoughnessTest.glb
ACMLCZH: Can you move them for me?

genesis/options/textures.py
ACMLCZH: If we restrict docs of functions to 120 characters, [some docs](https://github.com/Genesis-Embodied-AI/Genesis/blob/a99d893228c3ccc44914d9fc93fdd28cff1dcf82/genesis/options/morphs.py#L57) in other files also need to be modified I think?

duburcqa: Good point. The answer is yes but actually no. Doing so would create a large diff that would make cherry-picking, rebase ... and any other kind of git tree manipulation more painful. We want to avoid this at the moment because Genesis is not very stable at the point and there are several regressions that must be investigated. So we will do this later when everything is fine but not now. Still, it is not a reason for not following the guidelines when pushing new code. Here is the rational. This holds true not only for doc but for the whole codebase in general.

genesis/assets/tests/RoughnessTest.glb
duburcqa: Of course ! Give me a second.

duburcqa: Done! They are under `./usd/` folder.

EDIT: Upload failed actually. Let me check.

EDIT2: It is fine now!

genesis/options/textures.py
ACMLCZH: Changing `color` to `ndarray` may introduce a lot of changes involving other codebases, I think persisting tuple is OK here, and maybe I will change it in other PRs. Also, `Sequence[float]` does not accept `np.ndarray` but `List[float]` does. If we want to be general to accept color here, using `List[float]` is OK I think?

tests/test_mesh.py
ACMLCZH: "Update this helper to support `err_msg`." Do you mean I should update the  `assert_allclose` defined in `utils.py` to support `err_msg` since it is currently not supported?

genesis/assets/tests/RoughnessTest.glb
ACMLCZH: Actually these files contain `.usd` and `.glb` files. Is it OK to put them all in the `./usd` folder? Or we could create another folder named `./glb` to include `glb` files for testing? If so, should files like [chopper.glb](https://github.com/Genesis-Embodied-AI/Genesis/blob/a99d893228c3ccc44914d9fc93fdd28cff1dcf82/genesis/assets/tests/chopper.glb) also be moved?

genesis/options/textures.py
duburcqa: Something is wrong here: `List[float]` should not even accept tuple in the first place, let alone `np.ndaar`. Unless I’m missing something this is the whole point of `Sequence`.

tests/test_mesh.py
duburcqa: Yes, since you are using this feature.

genesis/assets/tests/RoughnessTest.glb
duburcqa: So you want, for now this repo is just a huge mess honestly.

genesis/options/textures.py
ACMLCZH: I just want to find a type that can generally accept `tuple`, `list` and `np.adarray` as the type of the argument `color` for user convenience. And during my attempts, I found that `List[float]` can accept all three types without runtime error and yield correct results, but `Sequence[float]` produced runtime error when it accepted `ndarray`. That is why I am using `List[float]` here. `List` and `Tuple` in the `typing` module seem a little different from `list` and `tuple` of python type in concepts I think?

duburcqa: Typing cannot cause runtime error. You must have done something wrong.

ACMLCZH: That error raises from pydantic type validation?

duburcqa: Ah yes probably. This is strange but let’s move on, no big deal.

tests/test_mesh.py
```diff
@@ -53,57 +60,56 @@ def check_gs_meshes(gs_mesh1, gs_mesh2, mesh_name):
     invdices2 = np.argsort(indices2)
     faces2 = invdices2[gs_mesh2.trimesh.faces]
 
-    np.testing.assert_allclose(
+    assert_allclose(
```
duburcqa: Keep everything on one line if possible, it is important for readability. This comment holds for all the other assertions.

```suggestion
    assert_allclose(vertices1, vertices2, atol=VERTICES_TOL, err_msg=f"Vertices match failed in mesh {mesh_name}.")
```

tests/test_mesh.py
```diff
@@ -2,11 +2,18 @@
...
 import os
+from huggingface_hub import snapshot_download
 
 import genesis as gs
 import genesis.utils.mesh as mu
 import genesis.utils.gltf as gltf_utils
 import genesis.utils.usda as usda_utils
+from .utils import assert_allclose, assert_array_equal
+
+VERTICES_TOL = 1e-05
```
duburcqa: Add comment to explain how you chose these tolerances.

tests/test_mesh.py
```diff
@@ -2,10 +2,177 @@
...
+NORMALS_TOL = 5e-02
+UV_TOL = 1e-07
+COLOR_TOL = 1e-07
+
+
+def check_gs_meshes(gs_mesh1, gs_mesh2, mesh_name):
+    """Check if two gs.Mesh objects are equal."""
+    vertices1 = gs_mesh1.trimesh.vertices
```
duburcqa: Prefer for-loop (or a "private" helper method) over duplicated code.

genesis/utils/usda.py
```diff
@@ -0,0 +1,453 @@
...
+        return {
+            "color_texture": color_texture,
+            "opacity_texture": opacity_texture,
+            "roughness_texture": roughness_texture,
+            "metallic_texture": metallic_texture,
+            "emissive_texture": emissive_texture,
+        }, "st"
+
+    elif shader_subid == "gltf_texture_lookup":
+        # offset = shader.GetInput("offset").Get()
```
duburcqa: What is this?

genesis/utils/usda.py
```diff
@@ -0,0 +1,453 @@
...
+            faces_vertex_counts = np.array(mesh_usd.GetFaceVertexCountsAttr().Get())
+            points_faces_varying = False
+
+            # parse normals
+            normals = None
+            normal_attr = mesh_usd.GetNormalsAttr()
+            if normal_attr.HasValue():
+                normals = np.array(normal_attr.Get(), dtype=np.float32)
+                if normals.shape[0] != points.shape[0]:
+                    if normals.shape[0] == faces.shape[0]:  # face varying
```
duburcqa: What does `# face varying` means ? Would be nice to be more specific.

genesis/utils/usda.py
duburcqa: Why this script is runnable ? You should rather expect `genesis/main.py` (aka `gs view` CLI utility) if necessary rather than making modules executable.

tests/test_mesh.py
ACMLCZH: If doing so, some lines will exceed 120 characters?

duburcqa: Yes, by "if possible" I mean: shrink a bit the message if possible and it still makes sense. If not, then split on multiple lines.

ACMLCZH:
This unittest failure is weird. Why the snapshot downloaded files are not found in Windows and Macos?

duburcqa:
Yes, you are missing decorators on top of each tests relying on Hugging Face:
```python
@pytest.mark.xdist_group(name="huggingface_hub")
```

By the way, you should also add the decorator `@pytest.mark.required` for tests that are fast and you consider essential.

ACMLCZH:
I added `@pytest.mark.xdist_group(name="huggingface_hub")` but it still failed. It seemed to fail when downloading `drone_sus` and `image_0000_segmented.glb`, which is so weird.
```
        drones = []
        for offset, merge_fixed_links in ((-0.3, False), (0.3, True)):
            drone = scene.add_entity(
>               morph=gs.morphs.Drone(
                    file=f"{asset_path}/drone_sus/drone_sus.urdf",
                    merge_fixed_links=merge_fixed_links,
                    pos=(0.0, offset, 1.5),
                ),
                vis_mode="collision",
                visualize_contact=True,
            )

tests/test_rigid_physics.py:1674: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
genesis/options/morphs.py:806: in __init__
    super().__init__(**data)
genesis/options/morphs.py:446: in __init__
    gs.raise_exception(f"File not found in either current directory or assets directory: '{self.file}'.")
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

msg = "File not found in either current directory or assets directory: '/Users/runner/.cache/huggingface/hub/datasets--Genesis-Intelligence--assets/snapshots/77c72b5e2544c7348d12b36dcb11b86575b7adae/drone_sus/drone_sus.urdf'."

    def raise_exception(msg="Something went wrong."):
>       raise gs.GenesisException(msg)
E       genesis.GenesisException: File not found in either current directory or assets directory: '/Users/runner/.cache/huggingface/hub/datasets--Genesis-Intelligence--assets/snapshots/77c72b5e2544c7348d12b36dcb11b86575b7adae/drone_sus/drone_sus.urdf'.
```

genesis/options/textures.py
ACMLCZH: It is a method should be overloaded by the derived classes of `Texture`. USD parsing applies `cutoff` to the `ColorTexture` and `ImageTexture`, so I added this checking.

doc
duburcqa, line 1: Are you updating the doc on purpose? I don't remember if I already asked before sorry.

tests/test_mesh.py
```diff
@@ -2,12 +2,149 @@
 import trimesh
 import numpy as np
 import os
+from huggingface_hub import snapshot_download
 
 import genesis as gs
 import genesis.utils.mesh as mu
 import genesis.utils.gltf as gltf_utils
+import genesis.utils.usda as usda_utils
+from .utils import assert_allclose, assert_array_equal
```
duburcqa: Skip line.

tests/test_mesh.py
```diff
@@ -2,12 +2,149 @@
...
 
+VERTICES_TOL = 1e-05  # Transformation loses a little precision in vertices
+NORMALS_TOL = 5e-02  # Conversion from .usd to .glb loses a little precision in normals
+UV_TOL = 1e-07  # Default tolerance for np.testing.assert_allclose
```
duburcqa: This is not a good excuse. If you should match at machine precision, use machine precision: `gs.EPS`

tests/test_mesh.py
```diff
@@ -2,12 +2,149 @@
...
+from huggingface_hub import snapshot_download
 
 import genesis as gs
 import genesis.utils.mesh as mu
 import genesis.utils.gltf as gltf_utils
+import genesis.utils.usda as usda_utils
+from .utils import assert_allclose, assert_array_equal
 
+VERTICES_TOL = 1e-05  # Transformation loses a little precision in vertices
+NORMALS_TOL = 5e-02  # Conversion from .usd to .glb loses a little precision in normals
```
duburcqa: Actually `5e-2` seems to be very large. For a distance it does not sound acceptable.

duburcqa:
>  It seemed to fail when downloading drone_sus and image_0000_segmented.glb, which is so weird.

This error is strange. Maybe something went wrong with Hugging Face assets download.

doc
YilingQiao: fixed. I reverted those doc-related change

tests/test_mesh.py
YilingQiao: Changed it to gs.EPS

tests/test_mesh.py
YilingQiao: 1e-02 is the best we can do. @ACMLCZH mentioned the normal is changed a bit when converting usd to glb using blender.

tests/test_mesh.py
YilingQiao: done

YilingQiao:
@duburcqa do you think it's good to merge?

YilingQiao (dismissed):

duburcqa (approved):

tests/test_mesh.py
```diff
@@ -137,11 +234,72 @@ def check_texture(tm_color, tm_texture, gs_texture, dim, material_name, texture_
...
                 material_name,
                 "emissive",
             )
+
+
+@pytest.mark.required
+@pytest.mark.xdist_group(name="huggingface_hub")
+@pytest.mark.parametrize("usd_filename", ["usd/sneaker_airforce", "usd/RoughnessTest"])
+def test_usd_parse(usd_filename):
+    glb_folder = snapshot_download(
```
duburcqa: We do not use `snapshot_download` and `@pytest.mark.xdist_group(name="huggingface_hub")` anymore

duburcqa (dismissed):

duburcqa (approved):

Issue #1087, opened 2025-04-29
[Feature]: Add maxvolume option for tetrahedralization


kosuke1701:
### What feature or enhancement are you proposing?

Adds `maxvolume` option to `genesis.options.Mesh` class and use it for tetrahedralization of FEM entities.

### Motivation

I want to use the FEM feature of Genesis to simulate soft objects, and I need finer shape deformations for more detailed simulations, which requires fine-grained FEM elements. However, the current Mesh options for TetGen lack direct control over the element density.

### Potential Benefit

By supporting the `maxvolume` option provided by TetGen, we can directly control the fineness of the generated elements without making major modifications to the current processing that uses TetGen.

### What is the expected outcome of the implementation work?

- [ ] Users can control element density via `maxvolume` in the Mesh configuration

duburcqa:
Could this be tuned automatically without user intervention instead of exposing more options and making everything more complex to configure, at risk of being overwhelming at some point? You guess you know better than me. This is parameter has to be tunable by the user then let’s expose it.

PR #1088, opened 2025-04-29
[FEATURE][BREAKING] Add maxvolume support for TetGen-based tetrahedralization, unify TetGen config handling, and add tests

kosuke1701:
## Description
This PR introduces a `maxvolume` option that lets users cap the maximum tetrahedron volume generated by TetGen.
While adding this feature, the TetGen pipeline was refactored as follows:

* **Using switch string** – A new `utils.mesh.make_tetgen_switches()` helper was introduced to convert a config dict into the canonical TetGen switch string. This addresses an issue where the existing TetGen wrapper could silently ignore options like `maxvolume`.

* **Entity simplification** – Previously, FEM and PBD entities had separate implementations for generating TetGen configurations from a morph instance. These were unified to use a single helper (`utils.mesh.generate_tetgen_config_from_morph`), making behaviour consistent across solvers.

* **Public API change (BREAKING)** – `Mesh.tetrahedralize()` used in `PBD3DEntity` was updated to match FEM by taking a single `tet_cfg` dict instead of a long list of positional kwargs. All in-tree callers were updated accordingly.

* **Tests** –  
  * `tests/test_fem.py::test_maxvolume`  
  * `tests/test_pbd.py::test_maxvolume`  
  verify that tightening `maxvolume` indeed produces a denser mesh.
  * A temporary cube OBJ fixture (`box_obj_path`) was added so that it can be reusable across other tests that requires an arbitrary `Mesh` instance.

## Related Issue
Resolves https://github.com/Genesis-Embodied-AI/Genesis/issues/1087

## Motivation and Context
* Users could previously control mesh quality but not density; large objects were often under-tessellated, hurting simulation accuracy.

* New pytest cases (see above) ensure that the element count increases when small `maxvolume` is set.
* I have manually inspected tetrahedrons introduced with `maxvolume=0.01` and without it (nolimit) as shown in the following screenshot.

finer_tetra

* Edge length of the both cube is 1.
* Left cube is tetrahedralized without `maxvolume` option.
* Right cube is tetrahedralized with `maxvolume=0.01` option.

## Checklist:

### DIFF
--- a/genesis/engine/mesh.py
+++ b/genesis/engine/mesh.py
@@ -144,7 +144,7 @@ class Mesh(RBC):
         )
         self.clear_visuals()
 
-    def tetrahedralize(self, order, mindihedral, minratio, nobisect, quality, verbose):
+    def tetrahedralize(self, tet_cfg):
         """
         Tetrahedralize the mesh.
         """
@@ -152,9 +152,8 @@ class Mesh(RBC):
             self.verts, np.concatenate([np.full((self.faces.shape[0], 1), self.faces.shape[1]), self.faces], axis=1)
         )
         tet = tetgen.TetGen(pv_obj)
-        verts, elems = tet.tetrahedralize(
-            order=order, mindihedral=mindihedral, minratio=minratio, nobisect=nobisect, quality=quality, verbose=verbose
-        )
+        switches = mu.make_tetgen_switches(tet_cfg)
+        verts, elems = tet.tetrahedralize(switches=switches)
         # visualize_tet(tet, pv_obj, show_surface=False, plot_cell_qual=False)
         return verts, elems
 
--- a/genesis/utils/mesh.py
+++ b/genesis/utils/mesh.py
@@ -816,7 +816,7 @@ def create_plane(size=1e3, color=None, normal=(0, 0, 1)):
 
 
 def generate_tetgen_config_from_morph(morph):
-    if not isinstance(morph, gs.options.morphs.Morph):
+    if not isinstance(morph, gs.options.morphs.TetGenMixin):
         raise TypeError(f"Expected instance of Morph, got {type(morph).__name__}")
     return dict(
         order=morph.order,
--- a/genesis/utils/mesh.py
+++ b/genesis/utils/mesh.py
@@ -817,7 +817,9 @@ def create_plane(size=1e3, color=None, normal=(0, 0, 1)):
 
 def generate_tetgen_config_from_morph(morph):
     if not isinstance(morph, gs.options.morphs.TetGenMixin):
-        raise TypeError(f"Expected instance of Morph, got {type(morph).__name__}")
+        raise TypeError(
+            f"Expected an instance of a class that inherits from TetGenMixin, but got an instance of {type(morph).name}."
+        )
     return dict(
         order=morph.order,
         mindihedral=morph.mindihedral,
--- a/genesis/utils/mesh.py
+++ b/genesis/utils/mesh.py
@@ -816,6 +816,7 @@ def create_plane(size=1e3, color=None, normal=(0, 0, 1)):
 
 
 def make_tetgen_switches(cfg):
+    """Build a TetGen switches string from a config dict."""
     flags = ["p"]
 
     if cfg.get("quality", True):
@@ -846,14 +847,11 @@ def tetrahedralize_mesh(mesh, tet_cfg):
         mesh.vertices, np.concatenate([np.full((mesh.faces.shape[0], 1), mesh.faces.shape[1]), mesh.faces], axis=1)
     )
     tet = tetgen.TetGen(pv_obj)
-    # NOTE: Use the 'switches' string to pass options directly,
-    # because the tetgen wrapper may ignore some parameters (e.g., maxvolume).
-    # See: https://github.com/pyvista/tetgen/issues/24
-    # verts, elems = tet.tetrahedralize(**tet_cfg)
+    # Build and apply the switches string directly, since
+    # the Python wrapper sometimes ignores certain kwargs
+    # (e.g. maxvolume). See: https://github.com/pyvista/tetgen/issues/24
     switches = make_tetgen_switches(tet_cfg)
-    print(switches)
     verts, elems = tet.tetrahedralize(switches=switches)
-    print(verts.shape, elems.shape)
     # visualize_tet(tet, pv_obj, show_surface=False, plot_cell_qual=False)
     return verts, elems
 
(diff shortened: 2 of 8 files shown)


genesis/utils/mesh.py
```diff
@@ -815,12 +815,55 @@ def create_plane(size=1e3, color=None, normal=(0, 0, 1)):
     return mesh
 
 
+def generate_tetgen_config_from_morph(morph):
+    return dict(
+        order=getattr(morph, "order", 1),
```
duburcqa: `getattr` is not a good approach. `morph` must have these attributes, already set with the right default value. Add some `isinstance` check for the morph and raise an exception if its type is not appropriate.

tests/conftest.py
```diff
@@ -260,3 +261,42 @@ def gs_sim(
     scene.build()
 
     return gs_sim
+
+
+@pytest.fixture
```
duburcqa: I think the scope should be "session" as you expect the same output for all tests? Once this is done, you do not need `uuid` anymore, just call it as you want, e.g. `cube.obj`. It will be created only once anyway so it is perfectly fine.

tests/test_fem.py
```diff
@@ -235,3 +235,30 @@ def _make_triangle_set(verts, faces, tol=4):
         "FEM entity surface triangles and visualizer mesh triangles do not match.\n"
         f"Differences: {set(entity_tris) ^ set(viz_tris)}"
     )
+
+
+@pytest.mark.parametrize("backend", [gs.cpu])
+def test_maxvolume(fem_material, show_viewer, box_obj_path):
+    """Test that imposing a maximum element volume constraint produces a finer mesh (i.e., more elements)."""
+    scene = gs.Scene(
+        sim_options=gs.options.SimOptions(),
```
duburcqa: Remove empty options, it is useless.

tests/conftest.py
```diff
@@ -260,3 +261,42 @@ def gs_sim(
...
+def box_obj_path(tmp_path):
```
duburcqa: I think you could reuse this fixture in tests that were merged in some of your previous PRs no? To avoid code duplication.

duburcqa:
Thank you for your contribution once again! And caring about unit test, this is very important 👍

genesis/utils/mesh.py
kosuke1701: It seems that only `Mesh` has TetGen-specific attributes such as `order`, while other morph classes (e.g., `Sphere`) do not. I also feel that it is not a good approach. Perhaps, we could move the definition of TetGen-related attributes to the `Morph` root class, so that all morphs has necessary attributes. What do you think?

tests/conftest.py
kosuke1701: I set the scope to "function" because the remeshing and tetrahedralization processes seem to have a caching mechanism, and I did not want it to affect the tests. However, I realized that I was mistaken. Since the scope of `tmp_path` is also "function", using a UUID to create different OBJ files for each test is unnecessary. I think we could just remove `uuid` and keep the scope to "function". What do you think?

tests/conftest.py
kosuke1701: The test in my previous PR (`test_interior_tetrahedralized_vertex`) cannot simply use `box_obj_path` because the test requires special shape that introduces interior vertices on tetrahedralization. However, if it is OK to use new `maxvolume` option in the test, we can reuse `box_obj_path` because small `maxvolume` also introduce interior vertices!

duburcqa: I see… do as you think is best!

genesis/utils/mesh.py
duburcqa: If these options are also meaningful for primitive geometries, then yes go for it and move these options in base class! But make sure it does not expose the options for morph for which it does not make sense.

tests/conftest.py
duburcqa: Yes, just do this ! Could you document your findings recording the caching mechanism and why we cannot use "session" scope ? It would avoid changing this scope by mistake.

kosuke1701: When I double-checked the implementation, I found that the caching mechanism uses vertex coordinates and other information to generate the cache filenames. Therefore, creating independent OBJ files doesn't really make sense.
```
rm_file_path = mu.get_remesh_path(self.verts, self.faces, edge_len_abs, edge_len_ratio, fix)
tet_file_path = mu.get_tet_path(mesh.vertices, mesh.faces, tet_cfg)
```
I'll update the scope to "session" as you initially suggested!

genesis/utils/mesh.py
kosuke1701: FEMEntity only supports Sphere, Box, Cylinder, and Mesh, while PBD3DEntity does not specify which morph class it supports. I'm not sure exactly which morph should be supported, but it seems those morphs supported by FEMEntity are enough to me. Other morphs are either 2D or for RigidEntity and DroneEntity.

I'll create a mixin to add options only to Sphere, Box, Cylinder, and Mesh.

duburcqa: Sounds good!

tests/conftest.py
```diff
@@ -260,3 +260,49 @@ def gs_sim(
...
+    return verts, faces
+
+
+@pytest.fixture(scope="session")
+def box_obj_path(tmp_path_factory, cube_verts_and_faces):
+    """Fixture that generates a temporary cube .obj file"""
+    verts, faces = cube_verts_and_faces
+
+    tmp_path = tmp_path_factory.mktemp("obj_fixtures")
+    filename = str(tmp_path / f"fixture_box_obj_path.obj")
```
duburcqa: Do you really need the factory for this? Instead of just `tmp_path`?

kosuke1701: tmp_path seems to be function-scoped and cannot be used in session scope

duburcqa: you should use `asset_tmp_path`

kosuke1701: Oh, I missed the fixture. Thank you for pointing it out!

genesis/engine/mesh.py
```diff
@@ -144,17 +144,16 @@ def remesh(self, edge_len_abs=None, edge_len_ratio=0.01, fix=True):
         )
         self.clear_visuals()
 
-    def tetrahedralize(self, order, mindihedral, minratio, nobisect, quality, verbose):
+    def tetrahedralize(self, tet_cfg):
```
duburcqa, line 147: Did you make sure the new API for `tetrahedralize` is used everywhere in the codebase?

kosuke1701: Yes. Actually, it is used only in PBD3DEntity.

duburcqa:
Let's merge once the CI passes.

PR #1097, opened 2025-05-02
[BUG FIX] Fix mujoco vs genesis discrepancies.

duburcqa:
## Description

* Add support to rotational invweight0, which is involved in weld constraint.
* Fix compilation on MacOS.
* Fix 'draw_contact_arrow' method ignoring user-specified color.
* Replace mass matrix LU factorization by LTDL and use linear system solving instead of explicit matrix inversion.
* Fix constraint solver search vector norm computation.
* Fix handling of first-order correction terms related to implicit integration.
* Add Mujoco compatibility mode for unit test and debugging purposes.

## Motivation and Context

Being able to match mujoco results as closely as possible is important to make sure that Genesis' implementation is valid without having to write a myriade of unit tests.

Running all the unit tests successfully.

## Checklist:

### DIFF
--- a/genesis/__init__.py
+++ b/genesis/__init__.py
@@ -208,6 +208,7 @@ def init(
 
     # Update torch default device
     torch.set_default_device(device)
+    torch.set_default_dtype(tc_float)
 
     logger.info(
         f"Running on ~~<[{device_name}]>~~ with backend ~~<{backend}>~~. Device memory: ~~<{total_mem:.2f}>~~ GB."
--- a/genesis/__init__.py
+++ b/genesis/__init__.py
@@ -45,7 +45,7 @@ def init(
     seed=None,
     precision="32",
     debug=False,
-    eps=1e-12,
+    eps=1e-15,
     logging_level=None,
     backend=None,
     theme="dark",
--- a/genesis/engine/solvers/rigid/constraint_solver_decomp.py
+++ b/genesis/engine/solvers/rigid/constraint_solver_decomp.py
@@ -1207,13 +1207,7 @@ class ConstraintSolver:
             )
 
         if ti.static(self._solver_type == gs.constraint_solver.CG):
-            for i_e in range(self._solver.n_entities):
-                e_info = self._solver.entities_info[i_e]
-                for i_d1 in range(e_info.dof_start, e_info.dof_end):
-                    Mgrad = gs.ti_float(0.0)
-                    for i_d2 in range(e_info.dof_start, e_info.dof_end):
-                        Mgrad += self._solver.mass_mat_inv[i_d1, i_d2, i_b] * self.grad[i_d2, i_b]
-                    self.Mgrad[i_d1, i_b] = Mgrad
+            self._solver._func_solve_mass(self.grad, self.Mgrad)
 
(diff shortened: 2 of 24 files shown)


genesis/__init__.py
```diff
@@ -45,7 +45,7 @@ def init(
     seed=None,
     precision="32",
     debug=False,
-    eps=1e-12,
+    eps=1e-15,
```
YilingQiao, line 48: why change this? Just curious

genesis/engine/solvers/rigid/constraint_solver_decomp.py
```diff
@@ -121,9 +121,10 @@ def add_collision_constraints(self):
 
                 d1, d2 = gu.orthogonals(impact.normal)
 
-                t = self._solver.links_info[link_a_maybe_batch].invweight + self._solver.links_info[
-                    link_b_maybe_batch
-                ].invweight * (link_b > -1)
+                invweight = self._solver.links_info[link_a_maybe_batch].invweight[0]
+                if link_b > -1:
+                    invweight += self._solver.links_info[link_b_maybe_batch].invweight[0]
```
YilingQiao: we might want to avoid += since it might be a slower atomic_add.

genesis/engine/solvers/rigid/constraint_solver_decomp.py
```diff
@@ -1123,7 +1112,7 @@ def update_bracket(self, p_alpha, p_cost, p_deriv_0, p_deriv_1, i_b):
     def _func_solve_body(self, i_b):
         alpha = self._func_linesearch(i_b)
 
-        if ti.abs(alpha) < gs.EPS:
+        if alpha == 0:
```
YilingQiao: Is this good? alpha is a float

YilingQiao (approved):

genesis/__init__.py
duburcqa: Good remark ! It is necessary to match Mujoco's behaviour that is using this specific threshold for machine epsilon.

genesis/engine/solvers/rigid/constraint_solver_decomp.py
duburcqa: You are right

genesis/engine/solvers/rigid/constraint_solver_decomp.py
duburcqa: Yes I don't like it either, but mujoco is doing this check. I could use the newly introduced mujoco compat flag to keep our version by default.

YilingQiao: if we keep using `if ti.abs(alpha) < gs.EPS` , would there be a mismatch?

YilingQiao: if there is no mismatch, can we avoid `alpha == 0`?

duburcqa: On theory there may be a mismatch, i think we could craft a script that specifically trigger this discrepancy. But i think the impact would not be large (within tolerance of our unit tests) so i can revert it.

PR #1102, opened 2025-05-05
[BUG FIX] Unify convex-convex and plane-convex code paths.

duburcqa:
## Description

* Unify convex-convex and plane-convex codepaths except for plane-box.
* Add plane-convex collision detection unit test.
* Improve robustness of SDF collision detection.
* Use 'exact' collision perturbation compensation instead of first-order.
* Increase rotational perturbation to improve robustness of collision detection.

## Checklist:


YilingQiao:
I remember that plan_convex is faster and more robust than convex_convex in some cases. Let’s merge this PR for now, verify that the performance remains unchanged, and revisit how best to leverage the plane in the future.

duburcqa:
I think it was definitely the case before, but I don’t think it is the case anymore now that convex-convex is very robust. Moreover, I’m still relying on the specialized contact point detection method under the hood. It is not fully unified. So I don’t think it would be any faster either.

YilingQiao:
plane_convex uses neighbor information to detect multiple contact points, which could be faster than the perturbation method. But if their performance is similar, a unified logic is definitely better. Let's see.

duburcqa:
> plane_convex uses neighbor information to detect multiple contact points, which could be faster than the perturbation method.

Indeed, the neighbor method should be faster than the perturbation method, but it should be marginal since the new method is only calling `self._mpr.support_driver`. I would even argue that the new method is faster for analytic shape. Anyway, in both cases it should be negligible compared to solving the constraints.

And all of that is without mentioning all the advantages of the unified code path, that is to say: "analytical" contact information for primitive shapes, more robust additional contact point detection (at least for the unit test), and consistent implementation for all convex shapes that would be easier to maintain and less confusing for the end-user.

PR #1107, opened 2025-05-06
[BUG FIX] fix urdf with scale parameter

Kashu7100:
## Description

Fixes the prismatic joint limit issue with the URDF with `scale` parameters (example robot attached).
 
[trossen_wx250s.zip](https://github.com/user-attachments/files/20052801/trossen_wx250s.zip)

## Related Issue

Current Genesis fails to load the attached robot.

Load the asset by adding the URDF entity.

## Checklist:

### DIFF
--- a/genesis/utils/urdf.py
+++ b/genesis/utils/urdf.py
@@ -238,14 +238,23 @@ def parse_urdf(morph, surface):
         elif joint.joint_type == "prismatic":
             j_info["dofs_motion_ang"] = np.zeros((1, 3))
             j_info["dofs_motion_vel"] = np.array([joint.axis])
-            j_info["dofs_limit"] = np.array(
-                [
+            if isinstance(scale, np.ndarray):
+                # NOTE: deal with URDF with scale option. e.g., scale="0.001 0.001 0.001"
+                j_info["dofs_limit"] = np.array(
                     [
                         scale * joint.limit.lower if joint.limit.lower is not None else -np.inf,
                         scale * joint.limit.upper if joint.limit.upper is not None else np.inf,
                     ]
-                ]
-            )
+                ).T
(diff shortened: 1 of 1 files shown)


duburcqa:
None scalar scaling is not supported and will not be supported for URDF and SDF. It breaks other stuffs. I will give you more details in a moment.

duburcqa:
What you want to do with scaling is kinda "zoom" the world from the perspective of the entity, i.e. scale the geometric properties of an entity wrt its root pose. In the general case, ie for a 3D vector scale, (x, y, z) dimensions are scaled independently along (x, y, z) world axes respectively. With this definition, it is an intrinsic uniquely-defined geometric property of the entity, and as such, it does not depend on its current configuration (aka. position vector).

Note that scaling geometries in their local frame (i.e. wrt parent link pose rather than the root pose of the whole entity) does not make sense in the first place, because changing the pose of the parent frame would result in the different scaling, breaking of the idea of intrinsic uniquely-defined geometric property. More specifically, applying a transform on the geometry pose and its inverse of the link pose would result in the same world pose but a different scaling.

For rigid non-articulated objects, this is all good and dimension-wise scaling makes sense, but it is no longer the case for poly-articulated robot. Following the same reasoning as before, the position of each geometry in world frame depends on their parent link poses, which themselves depends on the current configuration of the system. This is problematic as it means that the effect of scaling would depends on the initial configuration of the robot rather then being a intrinsic uniquely-defined geometric property. There is no another way to avoid this inconsistency than limiting scaling to a scalar factor. In this case, scaling between anisotropic and does not depends on the orientation of each geometry anymore, and therefore is independent of the configuration of the entity, which is precisely the property that we want to enforce.

PR #1159, opened 2025-05-19
[BUG FIX] Unified scene loading for MJCF and URDF files.

duburcqa:
## Description

This PR unifies parsing of MJCF and URDF files to rely systematically on Mujoco internal parser if possible. This may fail because some mesh format for visual geometries are not supported my Mujoco, notably Collada `.dae`. In this case, falling back to the now legacy custom URDF parser. Besides, links and joints ordering should now better align with Mujoco, notably world would be in front of the kinematic string if any.

* Improve 'libigl' API workaround.
* Add utils to properly redirect libc stderr.
* Unify MJCF and URDF loading to avoid discrepancies if possible.
* Fix grapsing bottle example.

## Related Issue

Related to Genesis-Embodied-AI/Genesis/issues/1135

## Motivation and Context

Many fields are missing in URDF files to uniquely initialise the physics. Currently, Genesis is relying on default values that are different from Mujoco and arguably somewhat unphysical. This behaviour is both undesirable and confusing for the end-user. It would be better to rely on Mujoco heuristic for loading URDF file, since this approach has already been proven reliable for MJCF files.

## Checklist:

### DIFF
--- a/examples/rigid/grasp_bottle.py
+++ b/examples/rigid/grasp_bottle.py
@@ -78,6 +78,8 @@ def main():
     for waypoint in path:
         franka.control_dofs_position(waypoint)
         scene.step()
+    for i in range(30):
+        scene.step()
 
     # reach
     qpos = franka.inverse_kinematics(
--- a/genesis/__init__.py
+++ b/genesis/__init__.py
@@ -2,7 +2,6 @@
 import os
 import sys
 import site
-import ctypes
 import atexit
 import logging as _logging
 import traceback
@@ -32,7 +31,7 @@ from .constants import GS_ARCH, TI_ARCH
 from .constants import backend as gs_backend
 from .logging import Logger
 from .version import __version__
-from .utils import set_random_seed, get_platform, get_device
+from .utils import redirect_libc_stderr, set_random_seed, get_platform, get_device
 
 _initialized = False
 backend = None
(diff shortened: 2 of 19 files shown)


duburcqa:
> with this PR, I got an error when running

Could you check again, it works fine on my side.

YilingQiao (dismissed):

examples/rigid/grasp_bottle.py
```diff
@@ -78,6 +78,8 @@ def main():
     for waypoint in path:
         franka.control_dofs_position(waypoint)
         scene.step()
+    for i in range(30):
```
YilingQiao, line 81: why?

YilingQiao:
need to have minimal mujoco version 3.3.0

duburcqa:
> need to have minimal mujoco version 3.3.0

Good to know! This is already enforced so it should be fine :)

YilingQiao (approved):

examples/rigid/grasp_bottle.py
duburcqa: The velocity gain of the PD controller is large, and the target position is changing fast whereas the target velocity is kept equal to zero. Because of this discrepancy, tracking is quite poor and it is necessary to give more time to the PD controller to converge to the final target position before starting the grasping motion.

PR #1163, opened 2025-05-19
[MIS] Reduce compile time

YilingQiao:
## Description

Since @ti.func is actually inlined and unrolled during compilation, we can reduce kernel length by reusing these functions. We could also try using @ti.real_func in the future for cleaner code.

The compile time for `_func_narrow_phase` is reduced from 138s to 68s on my side.

## Related Issue

## Checklist:

### DIFF
--- a/genesis/engine/solvers/rigid/collider_decomp.py
+++ b/genesis/engine/solvers/rigid/collider_decomp.py
@@ -1294,6 +1294,7 @@ class Collider:
                         else:
                             # Try using MPR before anything else
                             is_mpr_updated = False
+                            is_mpr_guess_direction_available = True
                             normal_ws = self.contact_cache[i_ga, i_gb, i_b].normal
                             for i_mpr in range(2):
                                 if i_mpr == 1:
@@ -1325,17 +1326,27 @@ class Collider:
                                     # Because of this, it is necessary to run it twice and take the contact information
                                     # associated with the point of deepest penetration.
                                     for i_sdf in range(2):
-                                        is_col_tmp, normal_tmp, penetration_tmp, contact_pos_tmp = self._func_contact_vertex_sdf(
-                                            i_ga if i_sdf == 0 else i_gb, i_gb if i_sdf == 0 else i_ga, i_b
+                                        is_col_tmp, normal_tmp, penetration_tmp, contact_pos_tmp = (
+                                            self._func_contact_vertex_sdf(
+                                                i_ga if i_sdf == 0 else i_gb, i_gb if i_sdf == 0 else i_ga, i_b
+                                            )
                                         )
                                         if i_sdf == 0:
                                             if is_col_tmp:
-                                                normal, penetration, contact_pos = normal_tmp, penetration_tmp, contact_pos_tmp
+                                                normal, penetration, contact_pos = (
+                                                    normal_tmp,
+                                                    penetration_tmp,
+                                                    contact_pos_tmp,
+                                                )
                                             else:
                                                 is_col = False
                                         else:
                                             if is_col_tmp and (not is_col or penetration_tmp >= penetration):
-                                                normal, penetration, contact_pos = normal_tmp, penetration_tmp, contact_pos_tmp
+                                                normal, penetration, contact_pos = (
+                                                    normal_tmp,
+                                                    penetration_tmp,
+                                                    contact_pos_tmp,
+                                                )
                                     is_col = True
 
                 if i_detection == 0:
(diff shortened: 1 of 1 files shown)


genesis/engine/solvers/rigid/collider_decomp.py
```diff
@@ -1320,12 +1321,41 @@ def _func_mpr(self, i_ga, i_gb, i_b):
                                     # Note that SDF may detect different collision points depending on geometry ordering.
                                     # Because of this, it is necessary to run it twice and take the contact information
                                     # associated with the point of deepest penetration.
-                                    is_col_a, normal_a, penetration_a, contact_pos_a = self._func_contact_vertex_sdf(
-                                        i_ga, i_gb, i_b
+                                    is_col_a, normal_a, penetration_a, contact_pos_a = (
```
duburcqa: This is not the right pattern in this case. You don't need to pre-allocate `_a` and `_b`. Just use your `_tmp` and adapt the logic right after, i.e. line 1359:
```python
if i_sdf == 0:
    if is_col_tmp:
        normal, penetration, contact_pos = normal_tmp, penetration_tmp, contact_pos_tmp
else:
    if is_col_tmp and (not is_col or penetration_tmp >= penetration):
        normal, penetration, contact_pos = normal_tmp, penetration_tmp, contact_pos_tmp
```

duburcqa: Actually this is buggy because `is_col` is not updated. Here is the right implementation:
```python
for i_sdf in range(2):
    is_col_tmp, normal_tmp, penetration_tmp, contact_pos_tmp = self._func_contact_vertex_sdf(
        i_ga if i_sdf == 0 else i_gb, i_gb if i_sdf == 0 else i_ga, i_b
    )
    if i_sdf == 0:
        if is_col_tmp:
            normal, penetration, contact_pos = normal_tmp, penetration_tmp, contact_pos_tmp
        else:
            is_col = False
    else:
        if is_col_tmp and (not is_col or penetration_tmp >= penetration):
            normal, penetration, contact_pos = normal_tmp, penetration_tmp, contact_pos_tmp
is_col = True
```

genesis/engine/solvers/rigid/collider_decomp.py
```diff
@@ -1292,21 +1292,22 @@ def _func_mpr(self, i_ga, i_gb, i_b):
...
-                            is_mpr_guess_direction_available = (
-                                ti.abs(self.contact_cache[i_ga, i_gb, i_b].normal) > gs.EPS
-                            ).any()
-                            if (i_detection == 0) and not is_col and is_mpr_guess_direction_available:
-                                is_col, normal, penetration, contact_pos = self._mpr.func_mpr_contact(
-                                    i_ga, i_gb, i_b, ti.Vector.zero(gs.ti_float, 3)
-                                )
+                            is_mpr_guess_direction_available = True
+                            for i_mpr in range(2):
+                                if i_mpr == 0 or (
```
duburcqa: This is buggy. Just define some `normal_ws` variable that equal to `self.contact_cache[i_ga, i_gb, i_b].normal` by default and overwritten to `ti.Vector.zero(gs.ti_float, 3)` for the second trial.

duburcqa: Besides, Move the comment at the right place. Makes no sense on to of the for loop.

duburcqa:
```python
                            # Try using MPR before anything else
                            is_mpr_updated = False
                            normal_ws = self.contact_cache[i_ga, i_gb, i_b].normal
                            for i_mpr in range(2):
                                if i_mpr == 1:
                                    # Try without warm-start if no contact was detected using it.
                                    # When penetration depth is very shallow, MPR may wrongly classify two geometries as not in
                                    # contact while they actually are. This helps to improve contact persistence without increasing
                                    # much the overall computational cost since the fallback should not be triggered very often.
                                    is_mpr_guess_direction_available = (ti.abs(normal_ws) > gs.EPS).any()
                                    if (i_detection == 0) and not is_col and is_mpr_guess_direction_available:
                                        normal_ws = ti.Vector.zero(gs.ti_float, 3)
                                        is_mpr_updated = False

                                if not is_mpr_updated:
                                    is_col, normal, penetration, contact_pos = self._mpr.func_mpr_contact(
                                        i_ga, i_gb, i_b, normal_ws
                                    )
                                    is_mpr_updated = True
```

duburcqa (approved):

PR #1207, opened 2025-05-28
[BUG FIX] Apple Metal only support float32 for FEM

Libero0809:
## Description
Make FEM compatible with Apple Metal Platform by converting arrays to np.float32

## Motivation and Context
Improve Compatibility

```python
import genesis as gs
import time
gs.init(backend=gs.gpu, precision="32", logging_level="debug")

scene = gs.Scene(
        dt=0.001,
        substeps=10,
    ),
    show_viewer=True,
)

scene.add_entity(
    material=gs.materials.FEM.Elastic(),
    morph=gs.morphs.Mesh(
        file="./cube8.obj",
        scale=0.1,
        pos=(0, 0, 1.0),
    ),
)
scene.build()

for i in range(100):
    scene.step()
```

[cube8.zip](https://github.com/user-attachments/files/20470941/cube8.zip)
@YilingQiao

### DIFF
--- a/genesis/engine/entities/fem_entity.py
+++ b/genesis/engine/entities/fem_entity.py
@@ -307,6 +307,10 @@ class FEMEntity(Entity):
         verts_COM = verts.mean(0)
         init_positions = (verts - verts_COM) @ R.T + verts_COM
 
+        if gs.backend == gs.metal:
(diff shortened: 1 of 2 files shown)


Libero0809:
Thank you!

PR #1223, opened 2025-06-02
[MISC] Generate text-based coverage report on CI.

duburcqa:


pyproject.toml
```diff
@@ -105,6 +106,17 @@ markers = [
 ]
 log_cli_level = "WARNING"
 
+[tool.coverage.run]
+branch = true
+source = ["genesis"]
+# parallel = true
```
YilingQiao, line 112: I assume you are going to delete them / add more comments in another commit?

duburcqa: Nope, I wanted to let them here. I'm not sure whether we want to enable them or not in the future. But I can add a # FIXME comment to clarify this.

YilingQiao (approved):

duburcqa:
Let's not merge this for now, it is slowing down the CI too much unfortunately.

davidhommie:
How's it slowing down

duburcqa:
Closing this PR because it does not add enough value. Indeed, the proposed infrastructure does not allow for reporting metrics associated with taichi code, which is arguably what we care most about, and it makes the summary kind of unreadable. To make this report somewhat useful, we should at least remove everything related to taichi from the summary.

PR #1228, opened 2025-06-03
[BUG FIX] Fix links classical linear acceleration computation.

duburcqa:
## Related Issue

## Checklist:

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -1958,9 +1958,10 @@ class RigidEntity(Entity):
         return self._solver.get_links_ang(links_idx, envs_idx, unsafe=unsafe)
 
     @gs.assert_built
-    def get_links_acc(self, links_idx_local=None, envs_idx=None, *, unsafe=False):
+    def get_links_accelerometer_data(self, links_idx_local=None, envs_idx=None, *, imu=False, unsafe=False):
         """
-        Returns linear acceleration of the specified entity's links. (Mimicking accelerometer)
+        Returns the accelerometer data that would be measured by a IMU rigidly attached to the specified entity's links,
+        i.e. the true linear acceleration of the links expressed at their respective origin in local frame coordinates.
 
         Parameters
         ----------
@@ -1972,10 +1973,52 @@ class RigidEntity(Entity):
         Returns
         -------
         acc : torch.Tensor, shape (n_links, 3) or (n_envs, n_links, 3)
-            The linear acceleration of the specified entity's links.
+            The accelerometer data of IMUs rigidly attached of the specified entity's links.
         """
         links_idx = self._get_idx(links_idx_local, self.n_links, self._link_start, unsafe=True)
-        return self._solver.get_links_acc(links_idx, envs_idx, unsafe=unsafe)
+        return self._solver.get_links_acc(links_idx, envs_idx, mimick_imu=True, unsafe=unsafe)
+
+    @gs.assert_built
+    def get_links_acc(self, links_idx_local=None, envs_idx=None, *, unsafe=False):
+        """
+        Returns true linear acceleration (aka. "classical acceleration") of the specified entity's links expressed at
+        their respective origin in world coordinates.
+
+        Parameters
+        ----------
+        links_idx_local : None | array_like
+            The indices of the links. Defaults to None.
+        envs_idx : None | array_like, optional
+            The indices of the environments. If None, all environments will be considered. Defaults to None.
+
+        Returns
+        -------
+        acc : torch.Tensor, shape (n_links, 3) or (n_envs, n_links, 3)
+            The linear classical acceleration of the specified entity's links.
+        """
+        links_idx = self._get_idx(links_idx_local, self.n_links, self._link_start, unsafe=True)
+        return self._solver.get_links_acc(links_idx, envs_idx, mimick_imu=False, unsafe=unsafe)
+
+    @gs.assert_built
+    def get_links_acc_ang(self, links_idx_local=None, envs_idx=None, *, unsafe=False):
+        """
+        Returns angular acceleration of the specified entity's links expressed at their respective origin in world
+        coordinates.
+
+        Parameters
+        ----------
+        links_idx_local : None | array_like
+            The indices of the links. Defaults to None.
+        envs_idx : None | array_like, optional
+            The indices of the environments. If None, all environments will be considered. Defaults to None.
+
+        Returns
+        -------
+        acc : torch.Tensor, shape (n_links, 3) or (n_envs, n_links, 3)
+            The linear classical acceleration of the specified entity's links.
+        """
+        links_idx = self._get_idx(links_idx_local, self.n_links, self._link_start, unsafe=True)
+        return self._solver.get_links_acc_ang(links_idx, envs_idx, unsafe=unsafe)
 
     @gs.assert_built
     def get_links_inertial_mass(self, links_idx_local=None, envs_idx=None, *, unsafe=False):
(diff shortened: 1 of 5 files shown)


genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -2983,7 +2828,7 @@ def _func_torque_and_passive_force(self):
                         )
 
     @ti.func
-    def _func_system_update_acc(self, for_sensor):
+    def _func_system_update_acc(self, update_cacc: ti.template()):
```
YilingQiao: ti.template() is usually for kernel. Why do use ti.template() for ti.func? 
```
By using ti.template() as an argument type hint, a Taichi field or a python object can be passed into a kernel. 
```

Why don't we use bool?

YilingQiao:
Can we add a unit test for it?

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -1359,17 +1361,7 @@ def _init_constraint_solver(self):
...
-        Velocity of a certain point on a rigid link.
-        """
-        link_state = self.links_state[link_idx, i_b]
-        vel_rot = link_state.ang.cross(pos_world - link_state.COM)
-        vel_lin = link_state.vel
-        return vel_rot + vel_lin
-
-    @ti.func
-    def _func_compute_mass_matrix(self, implicit_damping):
+    def _func_compute_mass_matrix(self, implicit_damping: ti.template()):
```
YilingQiao, line 1374: I forgot if I asked before. Why do we use `ti.template()` instead of bool? Does taichi allow `bool`?

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -2523,8 +2360,6 @@ def _func_forward_velocity_entity(self, i_e, i_b):
 
             self.links_state[i_l, i_b].cd_vel = cvel_vel
             self.links_state[i_l, i_b].cd_ang = cvel_ang
-            self.links_state[i_l, i_b].vel = cvel_vel
```
YilingQiao, line 2526: This was indeed wrong. Do we have another place to compute link velocity somewhere?

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -558,10 +560,13 @@ def _init_link_fields(self):
             COM=gs.ti_vec3,
             mass_shift=gs.ti_float,
             i_pos_shift=gs.ti_vec3,
-            # cfrc_flat
-            cfrc_flat_ang=gs.ti_vec3,
-            cfrc_flat_vel=gs.ti_vec3,
-            # COM-based external force
+            # COM-based link's spatial acceleration
+            cacc_ang=gs.ti_vec3,
```
YilingQiao, line 564: what is the difference between cacc and cdd?

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -2992,96 +2827,93 @@ def _func_system_update_acc(self, for_sensor):
...
+                                self.links_state[i_l, i_b].cacc_lin = ti.Vector.zero(gs.ti_float, 3)
+                                self.links_state[i_l, i_b].cacc_ang = ti.Vector.zero(gs.ti_float, 3)
                         else:
                             self.links_state[i_l, i_b].cdd_vel = self.links_state[i_p, i_b].cdd_vel
                             self.links_state[i_l, i_b].cdd_ang = self.links_state[i_p, i_b].cdd_ang
-                            # dimension
-
-                        map_sum_vel = ti.Vector.zero(gs.ti_float, 3)
-                        map_sum_ang = ti.Vector.zero(gs.ti_float, 3)
+                            if ti.static(update_cacc):
```
YilingQiao, line 2851: why do we want to use ti.static?

genesis/engine/solvers/rigid/rigid_solver_decomp.py
duburcqa: This enables compiling 2 versions of this method, one for `update_cacc=True` and one for `update_cacc=False`. This would improves performance a bit, and since this method is very fast to compile, I think we can afford it.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
duburcqa: This is not "wrong" actually. But having 2 fields storing the exact same quantity is useless. (It was checked in the unit test that they were the same the all time).

genesis/engine/solvers/rigid/rigid_solver_decomp.py
duburcqa: `cdd` is the not the true acceleration but the one resulting from the kinematics only (Coriolis effects...etc). It is computed assuming zero acceleration for all bodies. On the contrary, `cacc` is the real acceleration of the links.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
duburcqa: Why not, this would improve performance a bit by removing this branch. I discovered that branches are actually not free of cost. But I can revert in favor of a runtime branching if you want.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
duburcqa: Yes, see my other comment, closing this one.

tests/test_rigid_physics.py
```diff
@@ -793,7 +1012,7 @@ def test_pd_control(show_viewer):
     robot.set_dofs_armature(robot.get_dofs_armature(envs_idx=1) + MOTORS_KD * scene.sim._substep_dt, envs_idx=1)
 
     for i in range(1000):
-        dofs_pos = robot.get_dofs_position(envs_idx=1)
+        dofs_pos = robot.get_qpos(envs_idx=1)
```
YilingQiao, line 1015: Note that get_dofs_position returns the Euler angles for rotational DOFs, while qpos is a quaternion. Want to make sure this change is intentional?

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1957,10 +1957,53 @@ def get_links_ang(self, links_idx_local=None, envs_idx=None, *, unsafe=False):
         links_idx = self._get_idx(links_idx_local, self.n_links, self._link_start, unsafe=True)
         return self._solver.get_links_ang(links_idx, envs_idx, unsafe=unsafe)
 
+    @gs.assert_built
+    def get_links_accelerometer_data(self, links_idx_local=None, envs_idx=None, *, imu=False, unsafe=False):
```
YilingQiao, line 1961: It would be nice to have an example about `get_links_accelerometer_data` and show its difference from `get_links_acc `

tests/test_rigid_physics.py
duburcqa: Yes it is ! This change is necessary because qpos gets updated at after integration but dof vel no longer does :)

duburcqa: (Not that it is better this way, the less is updated post-integration, the better. For consistency of the data.

genesis/engine/entities/rigid_entity/rigid_entity.py
duburcqa: I see. Do you have anything in mind ? What about gripper orientation estimation using accelerometer data ?

duburcqa: I would rather merge this first and create a PR for the example next week if you are ok. Easier to manage with all the apes in the pipe  :)

PR #1229, opened 2025-06-03
[FEATURE] add exclude_self_contact option

Kashu7100:
## Description

Enable to return contacts with other entities only (excluding self collision)

## Related Issue

## Checklist:

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -2595,28 +2595,17 @@ class RigidEntity(Entity):
         scene_contact_info = self._solver.collider.contact_data.to_torch(gs.device)
         n_contacts = self._solver.collider.n_contacts.to_torch(gs.device)
 
-        if exclude_self_contact:
-            valid_mask = torch.logical_xor(
-                torch.logical_and(
-                    scene_contact_info["geom_a"] >= self.geom_start,
-                    scene_contact_info["geom_a"] < self.geom_end,
-                ),
-                torch.logical_and(
-                    scene_contact_info["geom_b"] >= self.geom_start,
-                    scene_contact_info["geom_b"] < self.geom_end,
-                ),
-            )
-        else:
-            valid_mask = torch.logical_or(
-                torch.logical_and(
-                    scene_contact_info["geom_a"] >= self.geom_start,
-                    scene_contact_info["geom_a"] < self.geom_end,
-                ),
-                torch.logical_and(
-                    scene_contact_info["geom_b"] >= self.geom_start,
-                    scene_contact_info["geom_b"] < self.geom_end,
-                ),
-            )
+        logical_operation = torch.logical_xor if exclude_self_contact else torch.logical_or
+        valid_mask = logical_operation(
+            torch.logical_and(
+                scene_contact_info["geom_a"] >= self.geom_start,
+                scene_contact_info["geom_a"] < self.geom_end,
+            ),
+            torch.logical_and(
+                scene_contact_info["geom_b"] >= self.geom_start,
+                scene_contact_info["geom_b"] < self.geom_end,
+            ),
+        )
         if with_entity is not None:
             if self.idx == with_entity.idx:
                 gs.raise_exception("`with_entity` cannot be the same as the caller entity.")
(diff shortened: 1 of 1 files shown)


genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -2593,16 +2595,28 @@ def get_contacts(self, with_entity=None):
...
+            valid_mask = torch.logical_or(
+                torch.logical_and(
+                    scene_contact_info["geom_a"] >= self.geom_start,
+                    scene_contact_info["geom_a"] < self.geom_end,
+                ),
+                torch.logical_and(
+                    scene_contact_info["geom_b"] >= self.geom_start,
+                    scene_contact_info["geom_b"] < self.geom_end,
+                ),
+            )
```
YilingQiao: to reduce repeated code, can you do?

```
logical_operation = torch.logical_xor if exclude_self_contact else torch.logical_or
valid_mask = logical_operation(...)
```

duburcqa:
Could you refactor the whole function when you are here ? The `with_entity` part to be refactored easily.

duburcqa:
Here it is:

```python
        if with_entity is not None:
            logical_operation = torch.logical_and
        elif exclude_self_contact:
            logical_operation = torch.logical_xor
        else:
            logical_operation = torch.logical_or
        if with_entity is not None:
            with_entity = self
        else:
            if self.idx == with_entity.idx:
                gs.raise_exception("`with_entity` cannot be the same as the caller entity.")
            
        valid_mask = logical_operation(
            torch.logical_and(
                scene_contact_info["geom_a"] >= self.geom_start,
                scene_contact_info["geom_a"] < self.geom_end,
            ),
            torch.logical_and(
                scene_contact_info["geom_b"] >= with_entity.geom_start,
                scene_contact_info["geom_b"] < with_entity.geom_end,
            ),
        )
```

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -2561,7 +2561,7 @@ def detect_collision(self, env_idx=0):
         return collision_pairs
 
     @gs.assert_built
-    def get_contacts(self, with_entity=None):
+    def get_contacts(self, with_entity=None, exclude_self_contact=False):
         """
         Returns contact information computed during the most recent `scene.step()`.
         If `with_entity` is provided, only returns contact information involving the caller entity and the specified `with_entity`. Otherwise, returns all contact information involving the caller entity.
```
duburcqa: When you are here, skip line to fit 120 chars max per line.

duburcqa:
By the way, it would be great if you could document at least one relevant possible use case. All "new" features must be motivated by an actual need.

Kashu7100:
> Here it is:
> 
> ```python
>         if with_entity is not None:
>             logical_operation = torch.logical_and
>         elif exclude_self_contact:
>             logical_operation = torch.logical_xor
>         else:
>             logical_operation = torch.logical_or
>         if with_entity is not None:
>             with_entity = self
>         else:
>             if self.idx == with_entity.idx:
>                 gs.raise_exception("`with_entity` cannot be the same as the caller entity.")
>             
>         valid_mask = logical_operation(
>             torch.logical_and(
>                 scene_contact_info["geom_a"] >= self.geom_start,
>                 scene_contact_info["geom_a"] < self.geom_end,
>             ),
>             torch.logical_and(
>                 scene_contact_info["geom_b"] >= with_entity.geom_start,
>                 scene_contact_info["geom_b"] < with_entity.geom_end,
>             ),
>         )
> ```

I think `with_entity` we also need to check the case where `geom_b` is self and `geom_a` is other. (so looks kinda similar to the original one)

duburcqa:
> I think with_entity we also need to check the case where geom_b is self and geom_a is other. (so looks kinda similar to the original one)

I see, so I was wrong here.

Kashu7100:
But I think when with entity is self, we can use logical_and to only extract the self collision. I can add this.

genesis/engine/entities/rigid_entity/rigid_entity.py
Kashu7100: This line is 74 chars

duburcqa: Not it is not. 204 chars actually.

duburcqa: I mean the whole documentation of the method, not just the line you added x)

PR #1238, opened 2025-06-06
[BUG FIX] Make first few FPSs output by FPSTracker more self-consistent

hughperkins:
## Description

The FPS Tracker,  before this change:
- initializes the dt expoential moving average to 0, meaning it takes a while to settle down
    - this pr intiializes it with the first value of dt
- has an initial value of dt also near zero, because:
    - when the calss is instantiated, last_time is set to current time
    - then the first time step is called, dt is calculatd from the last_time set in intializtion
        - these are often very close together
    - this pr doesn't calculate a value for dt until the second time step() is called
 
Before this change, first few fps look like:
Screenshot 2025-06-05 at 5 11 40 PM

After this change, looks like:
Screenshot 2025-06-05 at 5 15 51 PM

(much more consistent)

In addtion, this PR removes compensate_logging_cost, which is only valid on CPU, not on GPU (I suppose we could keep it, but make it default to False, but I dont feel we will want to benchmark on CPU anyway?)

## Related Issue

## Motivation and Context

See description above

By running `python examples/rigid/single_franka_envs.py -n 2048`

See description above

## Checklist:

### DIFF
--- a/genesis/utils/tools.py
+++ b/genesis/utils/tools.py
@@ -171,18 +171,26 @@ class Rate:
 
 
 class FPSTracker:
-    def __init__(self, n_envs, alpha=0.95, compensate_logging_cost=True):
-        self.last_time = time.perf_counter()
+    def __init__(self, n_envs, alpha=0.95):
+        self.last_time = None
         self.n_envs = n_envs
-        self.dt = 0
+        self.dt_ema = None
         self.alpha = alpha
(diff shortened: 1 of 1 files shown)


YilingQiao (approved):

duburcqa:
Apparently @YilingQiao is ok with it, so let's merge.

PR #1240, opened 2025-06-06
[FEATURE] Configurable fps averaging for single franka envs

hughperkins:
## Description

Adds --fps-tracker-alpha to single franka envs, to configure the timeframe of averaging of fps counter

eg with --fps-tracker-alpha 0, output for one configuration is:

Screenshot 2025-06-05 at 5 47 42 PM

## Related Issue

By running eg: python examples/rigid/single_franka_envs.py -n 2048 --horizon 100 --fps-tracker-alpha 0

## Checklist:

### DIFF
--- a/tests/test_rigid_physics.py
+++ b/tests/test_rigid_physics.py
@@ -1394,7 +1394,7 @@ def test_mesh_repair(convexify, show_viewer):
 
 @pytest.mark.required
 @pytest.mark.xdist_group(name="huggingface_hub")
-@pytest.mark.parametrize("euler", [(90, 0, 90), (74, 15, 90)])
+@pytest.mark.parametrize("euler", [(90, 0, 90), (75, 15, 90)])
 @pytest.mark.parametrize("backend", [gs.cpu, gs.gpu])
 def test_convexify(euler, backend, show_viewer):
     OBJ_OFFSET_X = 0.0  # 0.02
@@ -1474,12 +1474,15 @@ def test_convexify(euler, backend, show_viewer):
 
     # Check resting conditions repeateadly rather not just once, for numerical robustness
     # cam.start_recording()
+    qvel_norminf_all = []
     for i in range(1700):
         scene.step()
         # cam.render()
         if i > 1600:
             qvel = gs_sim.rigid_solver.get_dofs_velocity().cpu()
-            assert_allclose(qvel, 0, atol=0.75)
+            qvel_norminf = torch.linalg.norm(qvel, ord=math.inf)
+            qvel_norminf_all.append(qvel_norminf)
+    np.testing.assert_array_less(torch.stack(qvel_norminf_all, dim=0).mean(), 0.1)
     # cam.stop_recording(save_to_filename="video.mp4", fps=60)
 
     for obj in objs:
@@ -1623,7 +1626,7 @@ def test_terrain_generation(show_viewer):
     )
     scene.build(n_envs=225)
 
-    ball.set_pos(torch.cartesian_prod(*(torch.linspace(0.3, 9.2, 15),) * 2, torch.tensor((0.6,))))
+    ball.set_pos(torch.cartesian_prod(*(torch.linspace(1.0, 10.0, 15),) * 2, torch.tensor((0.6,))))
     for _ in range(400):
         scene.step()
 


genesis/engine/scene.py
```diff
@@ -181,6 +182,7 @@ def __init__(
         self._t = 0
         self._is_built = False
         self._show_FPS = show_FPS
+        self._FPS_tracker_alpha = FPS_tracker_alpha
```
YilingQiao, line 185: it's not used?

hughperkins: Here:
Screenshot 2025-06-05 at 6 43 23 PM

YilingQiao: yeah, I mean there is no moving average yet

hughperkins: Here: https://github.com/Genesis-Embodied-AI/Genesis/blob/1867f657e8627d5b04a477519d50d236946afb87/genesis/utils/tools.py#L184

YilingQiao: oh, I see, this is already implemented

hughperkins: Yes. This PR just makes it configurable.

examples/rigid/single_franka_envs.py
```diff
@@ -11,6 +11,12 @@ def main():
     parser.add_argument("-s", "--sep", action="store_true", default=False)
     parser.add_argument("-r", "--record", action="store_true", default=False)
     parser.add_argument("-n", "--num_env", type=int, default=1)
+    parser.add_argument(
+        "--fps-tracker-alpha",
+        type=float,
+        default=0.95,
+        help="lambda value for exponential moving average of fps. 0 means no averaging, 1 is infinite averaging",
```
YilingQiao: a little bit weird it's called alpha but a lambda values.

`1 is infinite averaging` → 1 is not updating at all

hughperkins: "a little bit weird it's called alpha but a lambda values." Yeah, I was pondering that. I should probalby make them consistent. I feel like "lambda over" is a standard verb, but "alpha over" is not something I've heard? Thoughts?

YilingQiao: can we call it something like "exponential decay momentum for moving average"

hughperkins: Sure :)

hughperkins: addressed in 5fc8b92

YilingQiao (dismissed):

YilingQiao (dismissed):

duburcqa:
Why you this be something we would like to expose to the user ? Physics is already complex enough not to add concerns about whether the FPS are properly measured I think.

hughperkins:
> Why you this be something we would like to expose to the user ? Physics is already complex enough not to add concerns about whether the FPS are properly measured I think.

That sounds like a Product question. Not sure who represents Product for us 🤔 Maybe @YilingQiao ?

YilingQiao:
> > Why you this be something we would like to expose to the user ? Physics is already complex enough not to add concerns about whether the FPS are properly measured I think.
> 
> That sounds like a Product question. Not sure who represents Product for us 🤔 Maybe @YilingQiao ?

Thank you for your suggestion yesterday that we should wait for Alexis's review on this PR before merging. I think it’s definitely quite helpful to have both your and Alexis’s input when discussing on product questions. Let’s make prodcut decisions together based on logic.

After thinking it over, I feel that we indeed do not want to add so many miscellaneous flags to the top-level Scene API.

```
        sim_options=None,
        coupler_options=None,
        tool_options=None,
        rigid_options=None,
        avatar_options=None,
        mpm_options=None,
        sph_options=None,
        fem_options=None,
        sf_options=None,
        pbd_options=None,
        vis_options=None,
        viewer_options=None,
        renderer=None,
        show_viewer=True,
        show_FPS=True,
        FPS_tracker_alpha: float = 0.95,
```

I think a more proper and organized way might be to have a `profiling_options` since we might also add more features (like whether to profile the memory, GPU, etc.) to the profiler in the future?

hughperkins:
> > > Why you this be something we would like to expose to the user ? Physics is already complex enough not to add concerns about whether the FPS are properly measured I think.
> > 
> > 
> > That sounds like a Product question. Not sure who represents Product for us 🤔 Maybe @YilingQiao ?
> 
> Thank you for your suggestion yesterday that we should wait for Alexis's review before merging. I think it’s definitely quite helpful to have both your and Alexis’s input when discussing on product questions. Let’s make prodcut decisions together based on logic.
> 
> After thinking it over, I feel that we indeed do not want to add so many miscellaneous flags to the top-level Scene API.
> 
> ```
>         sim_options=None,
>         coupler_options=None,
>         tool_options=None,
>         rigid_options=None,
>         avatar_options=None,
>         mpm_options=None,
>         sph_options=None,
>         fem_options=None,
>         sf_options=None,
>         pbd_options=None,
>         vis_options=None,
>         viewer_options=None,
>         renderer=None,
>         show_viewer=True,
>         show_FPS=True,
>         FPS_tracker_alpha: float = 0.95,
> ```
> 
> I think a more proper and organized way might be to have a `profiling_options` since we might also add more features (like whether to profile the memory, GPU, etc.) to the profiler?

Makes sense. To what extent could you consider my creating a new `profiling_options` now, and we will start by adding:
```
         show_FPS=True,
         FPS_tracker_alpha: float = 0.95
```
into the iniitial set of profiling_options?

YilingQiao:
Yeah, in the meantime, can we keep the `show_FPS` flag for a while for compatibility, since many users are already using it? We can add a deprecation warning if needed. What do you think @duburcqa ?

hughperkins:
> Yeah, in the meantime, can we keep the `show_FPS` flag for a while for compatibility, since many users are already using it? We can add a deprecation warning if needed. What do you think @duburcqa ?

Yeah, it should be theroetically possible to update the new flag based on the old flag if present 👍  (I say that, without having looked at the code yet, to check, but seems plausible)

duburcqa:
I think it makes sense. You would find example of deprecation warnings at several other places already.

By the way, initially I was just saying that we should not add the flag in the example script because it just disturb the attention of the reader from the real objective. But the idea of a new optional profiling group is kinda relevant and better than adding a flag on top level.

YilingQiao:
Yeah, this `examples/rigid/single_franka_envs.py` script is a very basic 101 example for people trying out Genesis, to show them how to run batch simulations (n_env > 1). Maybe FPS_tracker_alpha is too advanced compared to the purpose of this tutorial, and we can hide it.

hughperkins:
> Yeah, this `examples/rigid/single_franka_envs.py` script is a very basic 101 example for people trying out Genesis, to show them how to run batch simulations (n_env > 1). Maybe FPS_tracker_alpha is too advanced compared to the purpose of this tutorial, and we can hide it.

Oh I see, we literally only have 5 options on this script. We arent using eg OmegaConf, to make it possibel to modify any configuraiotn 🤔 

So basically, I can add. profiling_options, but hten basically I would need to either implement OmegaConf (dont think I'm going to unilaterally do that), or else copy and paste single_franka_envs.py into my own repo, and customize the copy?

hughperkins:
Maybe we should revert --horizon too then? 🤔

hughperkins:
replaced with https://github.com/Genesis-Embodied-AI/Genesis/pull/1247

PR #1244, opened 2025-06-06
[BUG FIX] Fix flaky unit test.

duburcqa:


tests/test_deformable_physics.py
```diff
@@ -4,6 +4,11 @@
 import genesis as gs
 
 
+@pytest.fixture(scope="function", autouse=True)
+def auto_initialize_genesis(initialize_genesis):
```
YilingQiao: what is this for?

duburcqa: To start Genesis automatically but locally within each test module instead of systematically as it was the case before. Now that we have more unit tests, some are just testing utilities and starting genesis for them is irrelevant.

duburcqa: Remove this part for now.

tests/test_rigid_physics.py
```diff
@@ -1394,7 +1394,7 @@ def test_mesh_repair(convexify, show_viewer):
 
 @pytest.mark.required
 @pytest.mark.xdist_group(name="huggingface_hub")
-@pytest.mark.parametrize("euler", [(90, 0, 90), (74, 15, 90)])
+@pytest.mark.parametrize("euler", [(90, 0, 90), (75, 15, 90)])
```
YilingQiao, line 1397: This parameter is this sensitive?

duburcqa: It was, but now it should be way more robust :)

PR #1258, opened 2025-06-09
[FEATURE] adding batch indices for collider

Kashu7100:
## Description

Adds `envs_idx` for detect_collision kernels. 

## Related Issue

## Motivation and Context

This will be useful for using this function for scene placement in a specific batch or batched path planning. 

## Checklist:

### DIFF
--- a/.github/workflows/generic-cpu.yml
+++ b/.github/workflows/generic-cpu.yml
@@ -15,6 +15,7 @@ jobs:
         PYTHON_VERSION: ["3.10", "3.11", "3.12"]
 
     env:
+      HF_HUB_DOWNLOAD_TIMEOUT: 60
       GS_CACHE_FILE_PATH: ".cache/genesis"
       TI_OFFLINE_CACHE: "1"
       TI_OFFLINE_CACHE_CLEANING_POLICY: "never"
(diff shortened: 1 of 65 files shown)


duburcqa:
Are you sure we want to take this path? The future is not just batching everything systematically? Loop over env is going to be slower than we can afford in practice no?

Kashu7100:
I think it was also looping over the entire batch in the original implementation as well (e.g., `for i_b in range(self._solver._B):`).

YilingQiao:
passing `envs_idx` from python to kernel is slow. We do not need to do this.

YilingQiao:
What's the motivation of this change? What do you want to achieve?

Kashu7100:
Is there faster way to do the same?
I just want to get the collision state for a specific env in batch and no need to update all batch.

YilingQiao:
`get the collision state for a specific env in batch and no need to update all batch.` Why? can you check all envs and take the envs you want in the python scope?

YilingQiao:
Checking 2024 vs checking 2 environments makes no big difference, as the GPU should have more threads than that. But passing the env_idx from CPU to GPU could be more expensive.

Kashu7100:
ok make sense, closing.

duburcqa:
> I just want to get the collision state for a specific env in batch and no need to update all batch.

Could you benchmark this just out of curiosity? Would be interesting to have some numbers to back our assertion. 

> I just want to get the collision state for a specific env in batch and no need to update all batch.

Can you elaborate? In which cases this is relevant?

PR #1262, opened 2025-06-10
[BUG FIX] More robust MJCF / URDF parsing.

duburcqa:
## Related Issue

## Checklist:


genesis/options/morphs.py
```diff
@@ -708,13 +708,16 @@ class URDF(FileMorph):
         Whether to merge links connected via a fixed joint. Defaults to True.
     links_to_keep : list of str, optional
         A list of link names that should not be skipped during link merging. Defaults to [].
+    armature : float, optional
```
YilingQiao: As you said, maybe`default_amature` is a more accurate name

duburcqa: As you prefer :) will update accordingly in a minute

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -41,7 +41,7 @@ def _sanitize_sol_params(
     timeconst, dampratio, dmin, dmax, width, mid, power = sol_params.reshape((-1, 7)).T
     if global_timeconst is not None:
         timeconst[:] = global_timeconst
-    if (timeconst < gs.EPS).any():
+    if (timeconst <= gs.EPS).any():
```
YilingQiao: It might be unstable / hacky to rely on the logic of checking timeconst == gs.EPS, but I don't have a very strong opinion here.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -4537,7 +4535,7 @@ def get_links_acc_ang(self, links_idx=None, envs_idx=None, *, unsafe=False):
         tensor = ti_field_to_torch(self.links_state.cacc_ang, envs_idx, links_idx, transpose=True, unsafe=unsafe)
         return tensor.squeeze(0) if self.n_envs == 0 else tensor
 
-    def get_links_COM(self, links_idx=None, envs_idx=None, *, unsafe=False):
```
YilingQiao, line 4540: Maybe we should add an API for real links_COM? In a next PR

genesis/engine/solvers/rigid/rigid_solver_decomp.py
duburcqa: Yes it is. The good thing is that it is only about print a warning a not, but we should do better I agree. Since the tests are not passing I will do better

genesis/engine/solvers/rigid/rigid_solver_decomp.py
duburcqa: Yes, would be great adding this in the next PR :)

Issue #1266, opened 2025-06-11
[Feature]: using gs.logger with backoff


Icefoxzhx:
### What feature or enhancement are you proposing?

Add a `log` method to the Logger class wrapped by Genesis, so it could be used with other packages like `backoff`. Would love to use the beautiful Genesis logging theme consistently across other packages needed for projects developed on Genesis.

### Motivation

Want to use backoff with gs.logger.

### Potential Benefit

gs.logger could be leveraged by other packages and provide a beautiful and consistent logging theme

### What is the expected outcome of the implementation work?

- [ ] backoff with gs.logger could work

PR #1267, opened 2025-06-11
[FEATURE] Add 'log' method in Logger to support backoff.

Icefoxzhx:
## Description

Add `log` method to the Logger class wrapped by Genesis, so it could be used with `backoff`. 

## Related Issue

Resolves Genesis-Embodied-AI/Genesis#1266

## Motivation and Context

Add a log method to the Logger class wrapped by Genesis, so it could be used with other packages like backoff. Would love to use the beautiful Genesis logging theme consistently across other packages needed for projects developed on Genesis.

```python
import genesis as gs
import backoff

gs.init(backend=gs.gpu, logging_level='debug')

@backoff.on_exception(backoff.expo, Exception, max_tries=5, logger=gs.logger)
def foo():
    print("Attempting to execute foo")
    raise Exception("Simulated failure")

foo()
```

Before: 

[Genesis] [17:36:32] [ERROR] AttributeError: 'Logger' object has no attribute 'log'

After:

Attempting to execute foo
[Genesis] [17:36:02] [INFO] Backing off foo(...) for 0.7s (Exception: Simulated failure)
Attempting to execute foo
[Genesis] [17:36:02] [INFO] Backing off foo(...) for 0.7s (Exception: Simulated failure)
Attempting to execute foo
[Genesis] [17:36:03] [INFO] Backing off foo(...) for 1.5s (Exception: Simulated failure)
Attempting to execute foo
[Genesis] [17:36:05] [INFO] Backing off foo(...) for 0.7s (Exception: Simulated failure)
Attempting to execute foo
[Genesis] [17:36:05] [ERROR] Giving up foo(...) after 5 tries (Exception: Simulated failure)

## Checklist:

### DIFF
--- a/genesis/logging/logger.py
+++ b/genesis/logging/logger.py
@@ -114,6 +114,10 @@ class Logger:
         finally:
             self.timer_lock.release()
 
+    def log(self, level, msg, *args, **kwargs):
+        with self.log_wrapper():
+            self._logger.log(level, msg, *args, **kwargs)
+
     def debug(self, message):
         with self.log_wrapper():
             self._logger.debug(message)
--- a/genesis/logging/logger.py
+++ b/genesis/logging/logger.py
@@ -98,8 +98,9 @@ class Logger:
         self.timer_lock.acquire()
 
         # swap with timer output
-        if not self._is_new_line:
+        if not self._is_new_line and not self._stream.closed:
             self._stream.write("\r")
+            self._stream.flush()
         try:
             yield
         finally:
--- a/genesis/logging/logger.py
+++ b/genesis/logging/logger.py
@@ -100,7 +100,6 @@ class Logger:
         # swap with timer output
         if not self._is_new_line and not self._stream.closed:
             self._stream.write("\r")
-            self._stream.flush()
         try:
             yield
         finally:


duburcqa:
@Icefoxzhx CI is not passing.

Icefoxzhx:
turns out the change surfaces another issue of unsafe stdout access in log_wrapper. added check of `sys.stdout.closed` to prevent accidental writes to closed streams.

genesis/logging/logger.py
```diff
@@ -98,8 +98,9 @@ def log_wrapper(self):
         self.timer_lock.acquire()
 
         # swap with timer output
-        if not self._is_new_line:
+        if not self._is_new_line and not self._stream.closed:
             self._stream.write("\r")
+            self._stream.flush()
```
duburcqa: Are you sure about systematic flushing ?

duburcqa:
> turns out the change surfaces another issue of unsafe stdout access in log_wrapper.

Great find!

genesis/logging/logger.py
Icefoxzhx: Without` .flush()`, that `\r` might stay buffered and not appear immediately. If performance is a concern, we could also remove this flushing, since the original implementation didn't flush anyway.

duburcqa: I would suggest not flushing for now if it is not necessary for your use-case. If it is, it would be necessary to do some profiling.

duburcqa: Ready to merge after that !

Icefoxzhx: Removed the flushing for now! thanks

duburcqa (approved):

PR #1290, opened 2025-06-16
[BUG FIX] Preserve external forces across sub-steps.

### LINKED ISSUE #822: [Bug]: Inconsistent drone behavior when changing `substeps`
### Bug Description

Using the same propeller RPM values to hover a drone (`gs.morphs.Drone`) results in different behavior depending on the `substeps` value.

The total force required to stabilize the drone should be the same since the drone's weight is constant (0.26487 N). However, changing the `substeps` value appears to alter the total force required for hovering.

- Base RPM ≈ 14468 → 0.2646 N (from `RigidSolver._kernel_set_drone_rpm()`):
  - substeps = 1: Drone hovers.

    https://github.com/user-attachments/assets/999c9202-28f8-4be6-b509-a1349573d531

  - substeps ≥ 2: Drone falls.

    https://github.com/user-attachments/assets/cea6178d-a729-4443-90ac-5a937dd86456

- Base RPM = 20440 → 0.528092 N (expected to always ascend):
  - substeps = 1: Drone ascends.

    https://github.com/user-attachments/assets/63d4f5b4-2796-4cab-8445-de79c762cb8d

  - substeps = 2: Drone hovers.

    https://github.com/user-attachments/assets/a502ccb7-ada4-40e5-8282-218057159fce

  - substeps ≥ 3: Drone falls.

    https://github.com/user-attachments/assets/eac3cbf0-1d6d-4935-b86a-46278014ecb3

Is this the expected behavior? Shouldn't the effect of a constant applied thrust be independent of the `substeps` value?

If I’m misunderstanding how `substeps` interact with the simulation, I’d really appreciate any clarification.

### Steps to Reproduce

1. Run the script with `substeps = 1` → Drone hovers.
2. Increase `substeps` to 2 or higher → Drone falls.
3. Increase RPM (`base_rpm = 20440`):
    - `substeps = 1`: Drone ascends.
    - `substeps = 2`: Drone hovers.
    - `substeps ≥ 3`: Drone falls.

```python
import genesis as gs

base_rpm = 14468.429183500699  # Expected hover RPM
# base_rpm = 20440  # Expected ascend RPM

def main():
    gs.init(backend=gs.gpu)

    ##### scene #####
    scene = gs.Scene(show_viewer=True, sim_options=gs.options.SimOptions(dt=0.01, substeps=1))
    # scene = gs.Scene(show_viewer=True, sim_options=gs.options.SimOptions(dt=0.01, substeps=2))

    ##### entities #####
    plane = scene.add_entity(morph=gs.morphs.Plane())

    drone = scene.add_entity(morph=gs.morphs.Drone(file="urdf/drones/cf2x.urdf", pos=(0, 0, 1.0)))

    ##### build #####
    scene.build()

    ##### simulate #####
    for _ in range(100):
        drone.set_propellels_rpm([base_rpm, base_rpm, base_rpm, base_rpm])
        scene.step()

if __name__ == "__main__":
    main()
```

### Expected Behavior

The effect of a constant applied thrust should be independent of the `substeps` value. Since the total thrust and weight remain the same, the drone should always hover.

### Screenshots/Videos

_No response_

### Relevant log output

```shell

```

### Environment

- OS: Ubuntu 24.04
- GPU/CPU: RTX 3060
- GPU-driver version: 535.183.01
- CUDA / CUDA-toolkit version: 12.2

98b29b4a4e56bdaa55ca1d8a359c5935eb24f26c

### Additional Context

_No response_

LeonLiu4:
### What it does
* Keeps previously-set external forces/torques (e.g., drone thrust) alive for every physics micro-step.
* Prevents loss of lift when `substeps > 1`.

### How
1. Call `RigidSolver.clear_external_forces()` **once at the start of each outer frame**.
2. Move `process_input()` before the micro-step loop so propeller RPMs / scripted forces are written only once per frame.
3. Run the physics solver `self._substeps` times, now reading the same forces each time.

### Why
Without this, drones (and any entity driven by external forces) lose thrust during the extra micro-steps and quickly sink or behave unpredictably (see #822).

### Linked issue
Fixes #822


genesis/ext/isaacgym/terrain_utils.py
```diff
@@ -160,9 +160,9 @@ def discrete_obstacles_terrain(terrain, max_height, min_size, max_size, num_rect
     """
     # switch parameters to discrete units
     max_height = int(max_height / terrain.vertical_scale)
-    min_size = int(min_size / terrain.horizontal_scale)
-    max_size = int(max_size / terrain.horizontal_scale)
-    platform_size = int(platform_size / terrain.horizontal_scale)
+    min_size = int(min_size * pattern_scale / terrain.horizontal_scale)
```
YilingQiao: You mixed in the changes from another PR.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -1699,11 +1701,11 @@ def substep(self):
         timer = create_timer("rigid", level=1, ti_sync=True, skip_first_call=True)
         self._kernel_step_1()
         timer.stamp("kernel_step_1")
-
+
```
YilingQiao: Those changes are not relevant. 

Can you run
```
black --line-length 120 .
```

You can also install a git pre-commit hook to help you do it automatically before every commit.
```
pre-commit install 
```

duburcqa (commented):
It is mandatory to add a new unit test every time a bug is fixed. In this case, I would recommend something based on drone, checking that you get the same resulting for dt=0.01 and substeps=5 vs dt=0.002 and substeps=1

duburcqa:
I don't care if you split your PR in two or not (maybe @YilingQiao has a stronger opinion), but update the PR description accordingly, and add one unit test per issue being addressed.

LeonLiu4:
> I don't care if you split your PR in two or not (maybe @YilingQiao has a stronger opinion), but update the PR description accordingly, and add one unit test per issue being addressed.

I added a unit test for the drone thrust bug. I'm not entirely sure how to check for the terrain any suggestions?

genesis/ext/isaacgym/terrain_utils.py
```diff
@@ -245,9 +245,9 @@ def pyramid_stairs_terrain(terrain, step_width, step_height, platform_size=1.0):
         terrain (SubTerrain): update terrain
     """
     # switch parameters to discrete units
-    step_width = int(step_width / terrain.horizontal_scale)
-    step_height = int(step_height / terrain.vertical_scale)
-    platform_size = int(platform_size / terrain.horizontal_scale)
+    step_width = int(step_width * pattern_scale / terrain.horizontal_scale)
```
YilingQiao: can you remove those changes? They don't belong to this PR.

duburcqa:
You may have a look to the existing unit test for heightmap. There is a low-level buffer that you could access for that I think. Then just add a single pattern and check if the low-level buffer is actually scaled by a given factor. If it is too much work let me know.

tests/test_drone_substeps.py
```diff
@@ -0,0 +1,28 @@
...
+    drone_test = scene_test.add_entity(morph=gs.morphs.Drone(file="urdf/drones/cf2x.urdf", pos=(0, 0, 1.0)))
+    scene_test.build()
+
+    for _ in range(100):
+        drone_test.set_propellels_rpm([base_rpm, base_rpm, base_rpm, base_rpm])
+        scene_test.step()
+
+    z_test = float(drone_test.get_dofs_position()[2])
+
+    assert abs(z_ref - z_test) < TOL
```
duburcqa: Use our own `assert_allclose` implementation.

tests/test_drone_substeps.py
```diff
@@ -0,0 +1,28 @@
...
+
+    scene_test = gs.Scene(show_viewer=True, sim_options=gs.options.SimOptions(dt=0.01, substeps=5))
+    drone_test = scene_test.add_entity(morph=gs.morphs.Drone(file="urdf/drones/cf2x.urdf", pos=(0, 0, 1.0)))
+    scene_test.build()
+
+    for _ in range(100):
+        drone_test.set_propellels_rpm([base_rpm, base_rpm, base_rpm, base_rpm])
+        scene_test.step()
+
+    z_test = float(drone_test.get_dofs_position()[2])
```
duburcqa: Casting to float is unnecessary. Comparing only z is also not enough. Just compare full position.

tests/test_drone_substeps.py
```diff
@@ -0,0 +1,28 @@
+import pytest
+import genesis as gs
+
+
+def test_drone_hover_same_with_and_without_substeps():
+    TOL = 0.02
```
duburcqa: This tolerance is quite low. You should be able to get better matching for a simulation lasting 1s.

tests/test_drone_substeps.py
```diff
@@ -0,0 +1,28 @@
...
+    base_rpm = 14468.429183500699
```
duburcqa: What is this magic number? Just using 15000.

tests/test_drone_substeps.py
```diff
@@ -0,0 +1,28 @@
...
+    scene_ref = gs.Scene(show_viewer=True, sim_options=gs.options.SimOptions(dt=0.002, substeps=1))
+    drone_ref = scene_ref.add_entity(morph=gs.morphs.Drone(file="urdf/drones/cf2x.urdf", pos=(0, 0, 1.0)))
+    scene_ref.build()
+
+    for _ in range(100):
```
duburcqa: You should do 500 steps instead of 100 to compensate for the timestep difference

tests/test_drone_substeps.py
```diff
@@ -0,0 +1,28 @@
...
+        drone_ref.set_propellels_rpm([base_rpm, base_rpm, base_rpm, base_rpm])
+        scene_ref.step()
+
+    z_ref = float(drone_ref.get_dofs_position()[2])
+
+    scene_test = gs.Scene(show_viewer=True, sim_options=gs.options.SimOptions(dt=0.01, substeps=5))
+    drone_test = scene_test.add_entity(morph=gs.morphs.Drone(file="urdf/drones/cf2x.urdf", pos=(0, 0, 1.0)))
+    scene_test.build()
+
+    for _ in range(100):
```
duburcqa: 100 steps is not enough. You should use something between 200 and 500.

YilingQiao:
For this bug, maybe a more appropriate fix should be following @duburcqa comments
```
        # FIXME: External forces should be cleared at the end of the step, not during substeps.
        self._func_clear_external_force()
```

But we should check it will cause performance regression since we have to add one more kernel each step.

YilingQiao:
Changing the code this way will fix the issue, and I tested that the performance is roughly the same.

In `simulator.py`

```
    def step(self, in_backward=False):
        if self._rigid_only:  # "Only Advance!" --Thomas Wade :P
            for _ in range(self._substeps):
                self.rigid_solver.substep()
                self._cur_substep_global += 1

        else:
            self.process_input(in_backward=in_backward)
            for _ in range(self._substeps):
                self.substep(self.cur_substep_local)

                self._cur_substep_global += 1
                if self.cur_substep_local == 0 and not in_backward:
                    self.save_ckpt()

        if self.rigid_solver.is_active():
            self.rigid_solver._kernel_clear_external_force()
```

in `rigid_solver_decomp.py`

```
    @ti.func
    def _func_forward_dynamics(self):
        self._func_compute_mass_matrix(
            implicit_damping=ti.static(self._integrator == gs.integrator.approximate_implicitfast)
        )
        self._func_factor_mass(implicit_damping=False)
        self._func_torque_and_passive_force()
        self._func_update_acc(update_cacc=False)
        self._func_update_force()
        # self._func_actuation()
        self._func_bias_force()
        self._func_compute_qacc()

    @ti.kernel
    def _kernel_clear_external_force(self):
        self._func_clear_external_force()
```

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -1690,9 +1690,18 @@ def _func_forward_dynamics(self):
         # self._func_actuation()
         self._func_bias_force()
         self._func_compute_qacc()
-        # FIXME: External forces should be cleared at the end of the step, not during substeps.
+
+    @ti.kernel
+    def clear_external_forces(self):
         self._func_clear_external_force()
 
+    def set_substep_dt(self, new_dt: float):
```
duburcqa: I don’t think you can change the substep dt like this, because it should also affects all the other solvers in case of hybrid simulation.

tests/test_drone_substeps.py
duburcqa: This needs to run 5 times more steps for the first simulation compared to the second one

tests/test_drone_substeps.py
duburcqa: Still need to decrease the tolerance

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -1704,13 +1706,6 @@ def substep(self):
         self._func_constraint_force()
         timer.stamp("constraint_force")
 
-        # Position, Velocity and Acceleration data must be consistent when computing links acceleration, otherwise it
-        # would not corresponds to anyting physical. There is no other way than doing this right before integration,
-        # because the acceleration at the end of the step is unknown for now as it may change discontinuous between
-        # before and after integration under the effect of external forces and constraints. This means that
-        # acceleration data will be shifted one timestep in the past, but there isn't really any way around.
-        self._kernel_update_acc()
```
YilingQiao: why do you delete this function?

tests/test_drone_substeps.py
LeonLiu4: fixed

tests/test_drone_substeps.py
LeonLiu4: fixed

tests/test_drone_substeps.py
LeonLiu4: fixed

tests/test_drone_substeps.py
LeonLiu4: fixed

tests/test_drone_substeps.py
LeonLiu4: fixed

tests/test_drone_substeps.py
LeonLiu4: fixed

genesis/engine/solvers/rigid/rigid_solver_decomp.py
LeonLiu4: fixed

tests/test_drone_substeps.py
```diff
@@ -0,0 +1,36 @@
+import pytest
+import genesis as gs
+
+from .utils import assert_allclose
+
+
+def test_drone_hover_same_with_and_without_substeps():
```
duburcqa: Do not create a special file for this, just add this in rigid physics right before `test_drone_advanced`.

And use this signature:
`def test_drone_hover_same_with_and_without_substeps(show_viewer, tol):`

It should not be necessary to define your own TOL because `tol` will be defined automatically based on the genesis precision. 

Add these decorators on top:
```python
@pytest.mark.required
```

tests/test_drone_substeps.py
```diff
@@ -0,0 +1,36 @@
...
+    base_rpm = 15000
+    scene_ref = gs.Scene(show_viewer=True, sim_options=gs.options.SimOptions(dt=0.002, substeps=1))
+    drone_ref = scene_ref.add_entity(morph=gs.morphs.Drone(file="urdf/drones/cf2x.urdf", pos=(0, 0, 1.0)))
+    scene_ref.build()
+
+    for _ in range(2500):
+        drone_ref.set_propellels_rpm([base_rpm, base_rpm, base_rpm, base_rpm])
+        scene_ref.step()
+
+    x_ref = drone_ref.get_dofs_position()[0]
```
duburcqa: `pos_ref = drone_ref.get_dofs_position()`

tests/test_drone_substeps.py
```diff
@@ -0,0 +1,36 @@
...
+
+    scene_test = gs.Scene(show_viewer=True, sim_options=gs.options.SimOptions(dt=0.01, substeps=5))
+    drone_test = scene_test.add_entity(morph=gs.morphs.Drone(file="urdf/drones/cf2x.urdf", pos=(0, 0, 1.0)))
+    scene_test.build()
+
+    for _ in range(500):
+        drone_test.set_propellels_rpm([base_rpm, base_rpm, base_rpm, base_rpm])
+        scene_test.step()
+
+    x_test = drone_test.get_dofs_position()[0]
```
duburcqa: `pos_test = drone_test.get_dofs_position()`

tests/test_drone_substeps.py
```diff
@@ -0,0 +1,36 @@
...
+    y_test = drone_test.get_dofs_position()[1]
+    z_test = drone_test.get_dofs_position()[2]
+
+    assert_allclose(x_test, x_ref, tol=TOL)
```
duburcqa: `assert_allclose(pos_test, pos_ref, tol=tol)`

tests/test_drone_substeps.py
```diff
@@ -0,0 +1,36 @@
...
+
+    for _ in range(2500):
+        drone_ref.set_propellels_rpm([base_rpm, base_rpm, base_rpm, base_rpm])
+        scene_ref.step()
+
+    x_ref = drone_ref.get_dofs_position()[0]
+    y_ref = drone_ref.get_dofs_position()[1]
+    z_ref = drone_ref.get_dofs_position()[2]
+
+    scene_test = gs.Scene(show_viewer=True, sim_options=gs.options.SimOptions(dt=0.01, substeps=5))
```
duburcqa: `show_viewer= show_viewer `

tests/test_drone_substeps.py
```diff
@@ -0,0 +1,36 @@
+import pytest
+import genesis as gs
+
+from .utils import assert_allclose
+
+
+def test_drone_hover_same_with_and_without_substeps():
+    TOL = 1e-7
+    base_rpm = 15000
+    scene_ref = gs.Scene(show_viewer=True, sim_options=gs.options.SimOptions(dt=0.002, substeps=1))
```
duburcqa: Use same multi-line formatting as for the other tests.

tests/test_drone_substeps.py
duburcqa: Plus use same multi-line "formatting" as for the other tests

tests/test_drone_substeps.py
LeonLiu4: fixed

tests/test_drone_substeps.py
LeonLiu4: fixed

tests/test_drone_substeps.py
LeonLiu4: fixed

tests/test_drone_substeps.py
LeonLiu4: fixed

tests/test_drone_substeps.py
LeonLiu4: fixed

tests/test_drone_substeps.py
LeonLiu4: done

YilingQiao:
@LeonLiu4 you did not format `test_rigid_physics.py`
```
would reformat /home/runner/work/Genesis/Genesis/tests/test_rigid_physics.py
```

Milotrince:
was the inclusion of `python-3.12.3-macos11.pkg` intentional?

YilingQiao:
oh, it's unintentional

YilingQiao:
@Milotrince thanks for noticing this. We should revert and clean it from the commit history.

duburcqa:
Fixed

PR #1295, opened 2025-06-18
[FEATURE] SAP coupler and Hydroelastic contact model.

Libero0809:
## Description

* new coupler for implicit fem solver based on the SAP solver of DRAKE
* new contact model that support the contact between tetrahedron and floor based on DRAKE's hydroelastic contact.
* Update the code for modifying the position and velocity. Explicit method should only change the next step position at last, while keep using the previous step position during the computation. The old version set the new step position to be the same as old step position and used in the code, leading to confusion. This change will not affect the output but make the code clearer.
* BUG FIX: calculation of m in linesearch is negated

## Motivation and Context

The current contact model is simple and only handles the contact between the vertices on the deformable object. The new contact model can simulate more accurate contact.

The new coupler is more stable and robust than the original explicit coupler.

The test is updated in `tests/test_fem.py` where a sphere collide with the ground.

https://github.com/user-attachments/assets/c3b0cadf-4bcb-4327-948d-da15ae2d8e92

## Checklist:


genesis/engine/coupler.py
```diff
@@ -581,3 +587,905 @@ def couple_grad(self, f):
...
+
+    def linesearch(self, f: ti.i32):
+        """
+        Note
+        ------
+        https://en.wikipedia.org/wiki/Backtracking_line_search#Algorithm
+        """
+        self.init_linesearch(f)
+        for i in range(self._n_linesearch_iterations):
+            self.one_linesearch_iter(f)
```
YilingQiao, line 1492: I wonder if we can use a single kernel launch for this, i.e. put the iteration loop inside the kernel.

genesis/engine/coupler.py
```diff
@@ -581,3 +587,905 @@ def couple_grad(self, f):
...
+
+            x = pairs[i_c].G @ x
+
+            for i in ti.static(range(4)):
+                i_v = fem_solver.elements_i[i_e].el2v[i]
+                self.pcg_state_v[i_b, i_v].Ap += pairs[i_c].barycentric[i] * x
+
+    @ti.kernel
+    def init_pcg_solve(self):
+        fem_solver = self.fem_solver
```
YilingQiao, line 1289: why?

genesis/engine/coupler.py
```diff
@@ -581,3 +587,905 @@ def couple_grad(self, f):
...
+                self.pcg_state_v[i_b, i_v].diag += pairs[i_c].barycentric[i] ** 2 * pairs[i_c].G
+
+        for i_b, i_v in ti.ndrange(fem_solver._B, fem_solver.n_vertices):
+            if not self.batch_active[i_b]:
+                continue
+            self.pcg_state_v[i_b, i_v].prec = self.pcg_state_v[i_b, i_v].diag.inverse()
+
+    @ti.func
+    def compute_contact_energy(self, f: ti.i32):
+        pairs = ti.static(self.fem_floor_contact_pairs)
```
YilingQiao, line 1154: why?

genesis/engine/coupler.py
```diff
@@ -581,3 +587,905 @@ def couple_grad(self, f):
...
+            Rt=gs.ti_float,  # Regularitaion for tangential
+            vn_hat=gs.ti_float,  # Stablization for normal velocity
+            mu=gs.ti_float,  # friction coefficient
+            mu_hat=gs.ti_float,  # friction coefficient regularized
+            mu_factor=gs.ti_float,  # friction coefficient factor, 1/(1+mu_tilde**2)
+            energy=gs.ti_float,  # energy
+            G=gs.ti_mat3,  # Hessian matrix
+        )
+        self.max_fem_floor_contact_pairs = fem_solver.n_surfaces * fem_solver._B
+        self.n_fem_floor_contact_pairs = ti.field(gs.ti_int, shape=())
```
YilingQiao, line 699: no batch dimension?

genesis/engine/entities/fem_entity.py
```diff
@@ -382,6 +387,38 @@ def _add_to_solver(self, in_backward=False):
         )
         self.active = True
 
+    def compute_pressure_field(self):
+        """
+        Compute the pressure field for the FEM entity based on its tetrahedral elements.
+        For hydroelastic contact: https://drake.mit.edu/doxygen_cxx/group__hydroelastic__user__guide.html
```
duburcqa, line 396: Skip one line between short description and details.

genesis/engine/entities/fem_entity.py
```diff
@@ -382,6 +387,38 @@ def _add_to_solver(self, in_backward=False):
...
+
+        Notes
+        -----
+        https://github.com/RobotLocomotion/drake/blob/master/geometry/proximity/make_mesh_field.cc
+        TODO: Add margin support
+        Drake's implementation of margin seems buggy.
+        """
+        init_positions = self.init_positions.cpu().numpy()
+        self.pressure_field_np, _, _, _ = signed_distance(init_positions, init_positions, self._surface_tri_np)
```
duburcqa: Prefer `self.pressure_field_np, *_ = ` to be more robust to API changes.

genesis/engine/entities/fem_entity.py
```diff
@@ -382,6 +387,38 @@ def _add_to_solver(self, in_backward=False):
...
+        margin = 0
```
duburcqa: Remove margin if not used.

genesis/engine/entities/fem_entity.py
```diff
@@ -382,6 +387,38 @@ def _add_to_solver(self, in_backward=False):
...
+        self.pressure_field_np = np.abs(self.pressure_field_np) - margin
+        max_distance = np.max(self.pressure_field_np) + gs.EPS
```
duburcqa: I do not see any reason to add `gs.EPS` to `max_distance` since you raise an exception if `max_distance` is too small. Moreover, it makes sense non-zero comparison a bit irrelevant, only catching exact zero.

genesis/engine/materials/FEM/elastic.py
```diff
@@ -39,23 +43,26 @@ def __init__(
...
+            self.compute_energy_gradient = self.compute_energy_gradient_linear
             self.compute_energy = self.compute_energy_linear
-        elif model == "stable_neohookean":
-            self.update_stress = self.update_stress_stable_neohookean
-            self.compute_energy_gradient_hessian = self.compute_energy_gradient_hessian_stable_neohookean
-            self.compute_energy = self.compute_energy_stable_neohookean
-        elif model == "stable_neohooken":
-            gs.logger.warning("The 'stable_neohooken' model is deprecated. Use 'stable_neohookean' instead.")
+            self.hessian_static = True
+        elif model == "stable_neohookean" or model == "stable_neohooken":
```
duburcqa: `model in ("stable_neohookean", "stable_neohooken")`

genesis/engine/materials/FEM/elastic.py
```diff
@@ -147,6 +154,49 @@ def compute_energy_gradient_hessian_linear(self, mu, lam, J, F, actu, m_dir, i_e
...
+            The computed energy.
+        gradient: ti.Matrix
+            The gradient of the energy with respect to the deformation gradient F.
+
+        Notes
+        -------
+        This implementation assumes small deformations and linear stress-strain relationship.
+        It is adapted from the HOBAKv1 implementation for linear elasticity:
+        https://github.com/theodorekim/HOBAKv1/blob/main/src/Hyperelastic/Volume/LINEAR.cpp
+
```
duburcqa: Remove extra line.

genesis/engine/materials/FEM/elastic.py
```diff
@@ -147,6 +154,49 @@ def compute_energy_gradient_hessian_linear(self, mu, lam, J, F, actu, m_dir, i_e
...
+        energy: float
+            The computed energy.
+        gradient: ti.Matrix
+            The gradient of the energy with respect to the deformation gradient F.
+
+        Notes
+        -------
+        This implementation assumes small deformations and linear stress-strain relationship.
+        It is adapted from the HOBAKv1 implementation for linear elasticity:
+        https://github.com/theodorekim/HOBAKv1/blob/main/src/Hyperelastic/Volume/LINEAR.cpp
```
duburcqa: Use permalink instead of temporary links: https://github.com/theodorekim/HOBAKv1/blob/8420c51b795735d8fb912e0f8810f935d96fb636/src/Hyperelastic/Volume/LINEAR.cpp

image

genesis/engine/materials/FEM/elastic.py
```diff
@@ -228,6 +278,45 @@ def compute_energy_gradient_hessian_stable_neohookean(self, mu, lam, J, F, actu,
...
+
+        Raises
+        -------
+        NotImplementedError
+            This implementation does not compute the Gradient for the stable Neo-Hookean model.
+
+        Notes
+        -------
+        This implementation is adapted from the HOBAKv1 stable Neo-Hookean model:
+        https://github.com/theodorekim/HOBAKv1/blob/main/src/Hyperelastic/Volume/SNH.cpp
```
duburcqa: Same.

genesis/engine/solvers/fem_solver.py
```diff
@@ -336,10 +342,10 @@ def init_pos_and_vel(self, f: ti.i32):
     def compute_vel(self, f: ti.i32):
         for i_e, i_b in ti.ndrange(self.n_elements, self._B):
             i_v0, i_v1, i_v2, i_v3 = self.elements_i[i_e].el2v
-            pos_v0 = self.elements_v[f + 1, i_v0, i_b].pos
-            pos_v1 = self.elements_v[f + 1, i_v1, i_b].pos
-            pos_v2 = self.elements_v[f + 1, i_v2, i_b].pos
-            pos_v3 = self.elements_v[f + 1, i_v3, i_b].pos
+            pos_v0 = self.elements_v[f, i_v0, i_b].pos
```
duburcqa, line 345: What is this ? A bug fix? Create a dedicated PR of this, or at the very least update the PR description to explicitly mention any bug fix and clearly as possible. This will be added to the commit description later on.

genesis/engine/solvers/fem_solver.py
```diff
@@ -386,7 +392,9 @@ def apply_uniform_force(self, f: ti.i32):
     def compute_pos(self, f: ti.i32):
         for i_v, i_b in ti.ndrange(self.n_vertices, self._B):
             dt = self.substep_dt
-            self.elements_v[f + 1, i_v, i_b].pos += dt * self.elements_v[f + 1, i_v, i_b].vel
+            self.elements_v[f + 1, i_v, i_b].pos = (
+                dt * self.elements_v[f + 1, i_v, i_b].vel + self.elements_v[f, i_v, i_b].pos
```
duburcqa, line 396: Same.

genesis/engine/solvers/fem_solver.py
```diff
@@ -425,25 +434,40 @@ def compute_ele_hessian_gradient(self, f: ti.i32):
 
             for mat_idx in ti.static(self._mats_idx):
```
duburcqa, line 434: Not sure if it is a a great idea to, statically loop over a variable that depends on the entities. This means that different entities will trigger recompilation more often. This is going in the opposite direction of what we are trying to achieve. What do you think @YilingQiao? (I know that this change was merged before your PR.)

genesis/engine/solvers/fem_solver.py
```diff
@@ -425,25 +434,40 @@ def compute_ele_hessian_gradient(self, f: ti.i32):
...
+                            i_e=i_e,
+                            i_b=i_b,
+                            hessian_field=self.elements_el_hessian,
+                        )
 
     @ti.func
     def compute_ele_energy(self, f: ti.i32):
         """
-        Compute the energy for each element in the batch. Only for the ones used in linesearch.
+        Compute the energy for each element in the batch. Should only be used in linesearch.
```
duburcqa, line 469: Maybe declare this method "private" to avoid confusing, similarly to what we do in the rigid solver:
`def _func_compute_ele_energy:`

genesis/engine/solvers/fem_solver.py
```diff
@@ -462,84 +486,149 @@ def compute_ele_energy(self, f: ti.i32):
                         m_dir=self.elements_i[i_e].muscle_direction,
                     )
 
+            # add linearized damping energy
+            if self._damping_beta > 0.0:
```
duburcqa: `if self._damping_beta > gs.EPS:` ?

genesis/engine/solvers/fem_solver.py
```diff
@@ -462,84 +486,149 @@ def compute_ele_energy(self, f: ti.i32):
...
+                    x_diff[i * 3 : i * 3 + 3] = (
+                        self.elements_v[f + 1, i_v[i], i_b].pos - self.elements_v[f, i_v[i], i_b].pos
+                    )
+                St_x_diff = ti.Vector.zero(gs.ti_float, 9)
+                for i in ti.static(range(3)):
+                    for j in ti.static(range(4)):
+                        St_x_diff[i * 3 : i * 3 + 3] += S[j, i] * x_diff[j * 3 : j * 3 + 3]
+
+                H_St_x_diff = ti.Vector.zero(gs.ti_float, 9)
+                for i in ti.static(range(3)):
```
duburcqa: `for i, j in ti.static(ti.ndrange(3, 3)):`

genesis/engine/solvers/fem_solver.py
```diff
@@ -462,84 +486,149 @@ def compute_ele_energy(self, f: ti.i32):
...
```
duburcqa: `for i, j in ti.static(ti.ndrange(3, 3)):`

genesis/engine/solvers/fem_solver.py
```diff
@@ -462,84 +486,149 @@ def compute_ele_energy(self, f: ti.i32):
...
             S[3, :] = -B[0, :] - B[1, :] - B[2, :]
+
+            if self._damping_beta > 0.0:
+                x_diff = ti.Vector.zero(gs.ti_float, 12)
+                for i in ti.static(range(4)):
+                    x_diff[i * 3 : i * 3 + 3] = (
+                        self.elements_v[f + 1, i_v[i], i_b].pos - self.elements_v[f, i_v[i], i_b].pos
+                    )
+                St_x_diff = ti.Vector.zero(gs.ti_float, 9)
+                for i in ti.static(range(3)):
```
duburcqa: for i, j in ti.static(ti.ndrange(3, 4)):

genesis/engine/solvers/fem_solver.py
```diff
@@ -462,84 +486,149 @@ def compute_ele_energy(self, f: ti.i32):
...
+                        St_x_diff[i * 3 : i * 3 + 3] += S[j, i] * x_diff[j * 3 : j * 3 + 3]
+
+                H_St_x_diff = ti.Vector.zero(gs.ti_float, 9)
+                for i in ti.static(range(3)):
+                    for j in ti.static(range(3)):
+                        H_St_x_diff[i * 3 : i * 3 + 3] += (
+                            self.elements_el_hessian[i_b, i, j, i_e] @ St_x_diff[j * 3 : j * 3 + 3]
+                        )
+                S_H_St_x_diff = ti.Vector.zero(gs.ti_float, 12)
+                for i in ti.static(range(4)):
```
duburcqa, line 571: `for i, j in ti.static(ti.ndrange(4, 3)):`

genesis/engine/solvers/fem_solver.py
```diff
@@ -462,84 +486,149 @@ def compute_ele_energy(self, f: ti.i32):
...
+                    for j in ti.static(range(3)):
+                        S_H_St_x_diff[i * 3 : i * 3 + 3] += S[i, j] * H_St_x_diff[j * 3 : j * 3 + 3]
+
+                for i in ti.static(range(4)):
+                    self.elements_v_energy[i_b, i_v[0]].force += (
+                        -damping_beta_over_dt * S_H_St_x_diff[i * 3 : i * 3 + 3]
+                    )
+
+            # diagonal of hessian
             for i in ti.static(range(3)):
```
duburcqa: `for i, j in ti.static(ti.ndrange(3, 3)):`

genesis/engine/solvers/fem_solver.py
```diff
@@ -462,84 +486,149 @@ def compute_ele_energy(self, f: ti.i32):
...
             if not self.batch_active[i_b]:
                 continue
-            self.elements_v_energy[i_b, i_v].force = -self.elements_v_info[i_v].mass_scaled_over_dt2 * (
-                self.elements_v[f + 1, i_v, i_b].pos - self.elements_v_energy[i_b, i_v].inertia
+            self.elements_v_energy[i_b, i_v].force = -self.elements_v_info[i_v].mass_over_dt2 * (
+                (self.elements_v[f + 1, i_v, i_b].pos - self.elements_v_energy[i_b, i_v].inertia)
+                + (self.elements_v[f + 1, i_v, i_b].pos - self.elements_v[f, i_v, i_b].pos) * damping_alpha_dt
             )
-            self.pcg_state_v[i_b, i_v].prec = ti.Matrix.zero(gs.ti_float, 3, 3)
+            self.pcg_state_v[i_b, i_v].diag = ti.Matrix.zero(gs.ti_float, 3, 3)
```
duburcqa: `diag` is not clear enough.

genesis/engine/solvers/fem_solver.py
```diff
@@ -562,15 +651,17 @@ def compute_Ap(self):
 
             # atomic
             self.pcg_state_v[i_b, i_v0].Ap += (
-                B[0, 0] * new_p9[0:3] + B[0, 1] * new_p9[3:6] + B[0, 2] * new_p9[6:9]
-            ) * V_scaled
+                (B[0, 0] * new_p9[0:3] + B[0, 1] * new_p9[3:6] + B[0, 2] * new_p9[6:9]) * V * damping_beta_factor
```
duburcqa, line 640: There are extra parentheses there.

genesis/engine/solvers/fem_solver.py
```diff
@@ -562,15 +651,17 @@ def compute_Ap(self):
...
+            )
             self.pcg_state_v[i_b, i_v1].Ap += (
-                B[1, 0] * new_p9[0:3] + B[1, 1] * new_p9[3:6] + B[1, 2] * new_p9[6:9]
-            ) * V_scaled
+                (B[1, 0] * new_p9[0:3] + B[1, 1] * new_p9[3:6] + B[1, 2] * new_p9[6:9]) * V * damping_beta_factor
+            )
             self.pcg_state_v[i_b, i_v2].Ap += (
-                B[2, 0] * new_p9[0:3] + B[2, 1] * new_p9[3:6] + B[2, 2] * new_p9[6:9]
```
duburcqa, line 571: same.

genesis/engine/solvers/fem_solver.py
```diff
@@ -562,15 +651,17 @@ def compute_Ap(self):
...
-            ) * V_scaled
-            self.pcg_state_v[i_b, i_v3].Ap += (s[0] * new_p9[0:3] + s[1] * new_p9[3:6] + s[2] * new_p9[6:9]) * V_scaled
+                (B[2, 0] * new_p9[0:3] + B[2, 1] * new_p9[3:6] + B[2, 2] * new_p9[6:9]) * V * damping_beta_factor
+            )
+            self.pcg_state_v[i_b, i_v3].Ap += (
+                (s[0] * new_p9[0:3] + s[1] * new_p9[3:6] + s[2] * new_p9[6:9]) * V * damping_beta_factor
```
duburcqa, line 649: same.

genesis/engine/solvers/fem_solver.py
```diff
@@ -666,19 +757,14 @@ def init_linesearch(self, f: ti.i32):
...
                 continue
             diff = self.elements_v[f + 1, i_v, i_b].pos - self.elements_v_energy[i_b, i_v].inertia
-            self.linesearch_state[i_b].prev_energy += (
-                0.5 * self.elements_v_info[i_v].mass_scaled_over_dt2 * diff.dot(diff)
-            )
+            self.linesearch_state[i_b].prev_energy += 0.5 * self.elements_v_info[i_v].mass_over_dt2 * diff.dot(diff)
             self.linesearch_state_v[i_b, i_v].x_prev = self.elements_v[f + 1, i_v, i_b].pos
-            self.linesearch_state[i_b].m += self.pcg_state_v[i_b, i_v].x.dot(self.elements_v_energy[i_b, i_v].force)
-
+            self.linesearch_state[i_b].m += -self.pcg_state_v[i_b, i_v].x.dot(self.elements_v_energy[i_b, i_v].force)
```
duburcqa: just use `-=` directly.

This is was a bug, update PR description accordingly.

genesis/engine/solvers/fem_solver.py
```diff
@@ -737,6 +828,11 @@ def batch_solve(self, f: ti.i32):
             # compute element energy and gradient
             self.compute_ele_hessian_gradient(f)
 
+            # If the hessian is static, we only need to compute it once
+            for mat_idx in self._mats_idx:
+                if self._mats[mat_idx].hessian_static:
```
duburcqa: `invariant` would be more appropriate that `static` I think, but as you want.

genesis/engine/solvers/fem_solver.py
```diff
@@ -931,9 +1027,10 @@ def _kernel_add_elements(
             for j in ti.static(range(3)):
                 self.elements_v[f, i_global, i_b].pos[j] = verts[i_v, j]
             self.elements_v[f, i_global, i_b].vel = ti.Vector.zero(gs.ti_float, 3)
-            self.elements_v_info[i_global].mass_scaled_over_dt2 = 0
+            self.elements_v_info[i_global].mass = 0
+            self.elements_v_info[i_global].mass_over_dt2 = 0
 
-        one_over_dt2 = 1.0 / (self.substep_dt * self.substep_dt)
+        one_over_dt2 = 1.0 / (ti.static(self.substep_dt) ** 2)
```
duburcqa: `self.substep_dt` should definitely NOT be considered static. Recompiling systematically every time the simulation tilmestep changes is not acceptable.

genesis/options/solvers.py
```diff
@@ -100,6 +104,38 @@ class CouplerOptions(Options):
     fem_sph: bool = True
 
 
+class SAPCouplerOptions(CouplerOptions):
+    """
+    Options configuring the inter-solver coupling for SAP.
```
duburcqa: `SAP` is not clear, use Complete name here. If you could also point to some reference to explain what it is, that would be great.

genesis/options/solvers.py
```diff
@@ -100,6 +104,38 @@ class CouplerOptions(Options):
...
+        Number of iterations for the line search solver. Defaults to 10.
+    sap_threshold : float, optional
+        Threshold for the SAP solver. Defaults to 1e-6.
+    pcg_threshold : float, optional
+        Threshold for the PCG solver. Defaults to 1e-6.
+    linesearch_c : float, optional
+        Line search sufficient decrease parameter. Defaults to 1e-4.
+    linesearch_tau : float, optional
+        Line search step size reduction factor. Defaults to 0.8.
+
```
duburcqa, line 127: Do not skip line here.

genesis/options/solvers.py
```diff
@@ -100,6 +104,38 @@ class CouplerOptions(Options):
...
+class SAPCouplerOptions(CouplerOptions):
+    """
+    Options configuring the inter-solver coupling for SAP.
+
+    Parameters
+    ----------
+    n_sap_iterations : int, optional
+        Number of iterations for the SAP solver. Defaults to 5.
+    n_pcg_iterations : int, optional
+        Number of iterations for the PCG solver. Defaults to 100.
```
duburcqa: Use full name in documentation, not PCG.

genesis/options/solvers.py
```diff
@@ -527,6 +563,16 @@ class FEMOptions(Options):
...
+    damping_alpha : float, optional
+        Rayleigh Damping factor for the implicit solver. Defaults to 0.0. Only used when `use_implicit_solver` is True.
+    damping_beta : float, optional
+        Rayleigh Damping factor for the implicit solver. Defaults to 0.0. Only used when `use_implicit_solver` is True.
+
+    Note
+    ----
+    damping coefficients are used to control the damping effect in the simulation.
+    They are used in the Rayleigh Damping model, which is a common damping model in FEM simulations.
+    Might move it to material parameters in the future instead of solver options.
```
duburcqa: Add todo in front of this line.

genesis/options/solvers.py
```diff
@@ -527,6 +563,16 @@ class FEMOptions(Options):
...
     linesearch_tau : float, optional
         Line search step size reduction factor. Defaults to 0.5. Only used when `use_implicit_solver` is True.
+    damping_alpha : float, optional
+        Rayleigh Damping factor for the implicit solver. Defaults to 0.0. Only used when `use_implicit_solver` is True.
+    damping_beta : float, optional
+        Rayleigh Damping factor for the implicit solver. Defaults to 0.0. Only used when `use_implicit_solver` is True.
+
+    Note
+    ----
+    damping coefficients are used to control the damping effect in the simulation.
```
duburcqa: Separate different notes with bullet points. Start with uppercase.

genesis/options/solvers.py
```diff
@@ -527,6 +563,16 @@ class FEMOptions(Options):
...
+    They are used in the Rayleigh Damping model, which is a common damping model in FEM simulations.
```
duburcqa, line 578: Add reference for Rayleigh Damping model.

tests/test_fem.py
```diff
@@ -199,23 +199,32 @@ def test_maxvolume(fem_material, show_viewer, box_obj_path):
...
     """Test adding multiple FEM entities to the scene"""
     scene = gs.Scene(
         sim_options=gs.options.SimOptions(
-            dt=1e-2,
+            dt=1 / 60,
+            substeps=5,
         ),
         fem_options=gs.options.FEMOptions(
             use_implicit_solver=True,
+            n_newton_iterations=1,
```
duburcqa: Do not use custom low-level solver options. Unit tests are meant to validate the default options, because it is what matters for most users. If this test cannot validate the behaviour for the default options, then add another test. (of course not `damping_alpha=0.1`, `damping_beta=0.01`)

tests/test_fem.py
```diff
@@ -199,23 +199,32 @@ def test_maxvolume(fem_material, show_viewer, box_obj_path):
...
+            n_pcg_iterations=100,
+            pcg_threshold=1e-6,
+            damping_alpha=0.1,
+            damping_beta=0.01,
         ),
-        show_viewer=show_viewer,
+        coupler_options=gs.options.SAPCouplerOptions(
+            n_sap_iterations=10,
```
duburcqa: same.

tests/test_fem.py
```diff
@@ -199,23 +199,32 @@ def test_maxvolume(fem_material, show_viewer, box_obj_path):
...
+            n_pcg_iterations=100,
+        ),
+        show_viewer=False,
```
duburcqa: `show_viewer=show_viewer`, plus add `show_FPS=False`

tests/test_fem.py
```diff
@@ -199,23 +199,32 @@ def test_maxvolume(fem_material, show_viewer, box_obj_path):
 @pytest.fixture(scope="session")
 def fem_material_linear():
     """Fixture for common FEM linear material properties"""
-    return gs.materials.FEM.Elastic()
+    return gs.materials.FEM.Elastic(friction_mu=0.5)
 
 
 @pytest.mark.parametrize("backend", [gs.cpu])
-def test_multiple_fem_entities_implicit(fem_material_linear, show_viewer):
+def test_sphere_fall_implicit(fem_material_linear, show_viewer):
```
duburcqa: I have the feeling the old unit test is still relevant? I think it would be better to keep them both

genesis/engine/coupler.py
Libero0809: aliases

genesis/engine/coupler.py
Libero0809: aliases

genesis/engine/coupler.py
Libero0809: I want the iteration loop to be serial and the thing inside the iteration to be parallel. If we put the iteration loop inside, there seems to be no simple way to achieve this?

genesis/engine/coupler.py
Libero0809: all the collision pairs are concatenated together

genesis/engine/coupler.py
YilingQiao: yeah, I was curious why pairs and fem_solver don't use the same convention.

genesis/engine/coupler.py
YilingQiao: we may think about other way in the future. I think this one is good for now.

genesis/engine/materials/FEM/elastic.py
Libero0809: Is it OK for the link to exceed 120 limit?

genesis/engine/solvers/fem_solver.py
Libero0809: I will revert it to the original version.

Libero0809: I have put this into this PR description. 
Update the code for modifying the position and velocity. Explicit method should only change the next step position at last, while keep using the previous step position during the computation. The old version set the new step position to be the same as old step position and used in the code, leading to confusion. This change will not affect the output but make the code clearer.

genesis/engine/solvers/fem_solver.py
Libero0809: I will let you guys decide. I was copying from the old code where ti.static was used.

genesis/engine/solvers/fem_solver.py
Libero0809: damping_beta is a user provided parameter so 0.0 means no damping. It's a rare case that user want to use a very small damping factor even smaller than gs.EPS so I guess it would be fine. What's your opinion?

genesis/engine/solvers/fem_solver.py
Libero0809: I think it's the black formatter who did this. Should we change how it works?

genesis/engine/solvers/fem_solver.py
Libero0809: It's a bug. Should I update the PR title as well?

genesis/options/solvers.py
Libero0809: Do you mean the class name or the docstring

tests/test_fem.py
Libero0809: I add back the old unit test and change the test names to be more specific

tests/test_fem.py
```diff
@@ -202,17 +202,19 @@ def fem_material_linear():
     return gs.materials.FEM.Elastic()
 
 
-@pytest.mark.parametrize("backend", [gs.cpu])
-def test_multiple_fem_entities_implicit(fem_material_linear, show_viewer):
+@pytest.mark.parametrize("backend", [gs.gpu])
```
duburcqa: why GPU only ? Just remove this decorator completely to run on the default backend, or use CPU, but forcing GPU is weird.

genesis/engine/entities/fem_entity.py
```diff
@@ -382,6 +387,31 @@ def _add_to_solver(self, in_backward=False):
...
+        For hydroelastic contact: https://drake.mit.edu/doxygen_cxx/group__hydroelastic__user__guide.html
+
+        Notes
+        -----
+        https://github.com/RobotLocomotion/drake/blob/master/geometry/proximity/make_mesh_field.cc
+        TODO: Add margin support
+        Drake's implementation of margin seems buggy.
+        """
+        init_positions = self.init_positions.cpu().numpy()
+        self.pressure_field_np, *_ = signed_distance(init_positions, init_positions, self._surface_tri_np)
```
duburcqa: I would suggest only assigning the attribute `self.pressure_field_np` at the very end of this function.

genesis/engine/solvers/fem_solver.py
duburcqa: Comparing to exact zero is considered a bad practice, probably because it may not go well when the value is parsed from text configuration files, then casted from double / int to float back and forth multiple times in the process, strange things may happen (although exact 0.0 should be preserved in theory).

Beyond that, any damping smaller than EPS should be ignored because it is not numerical significant (within the margin of error), and therefore better sparse irrelevant computations if possible even if non-zero.

genesis/engine/solvers/fem_solver.py
duburcqa: If you write:
```python
self.pcg_state_v[i_b, i_v0].Ap += (
    B[0, 0] * new_p9[0:3] + B[0, 1] * new_p9[3:6] + B[0, 2] * new_p9[6:9]
) * V * damping_beta_factor
```

It gets reformatted? If so, just close this comment it is fine.

genesis/engine/solvers/fem_solver.py
```diff
@@ -931,9 +1013,10 @@ def _kernel_add_elements(
             for j in ti.static(range(3)):
                 self.elements_v[f, i_global, i_b].pos[j] = verts[i_v, j]
             self.elements_v[f, i_global, i_b].vel = ti.Vector.zero(gs.ti_float, 3)
-            self.elements_v_info[i_global].mass_scaled_over_dt2 = 0
+            self.elements_v_info[i_global].mass = 0
```
duburcqa: Try to use float when assigning float variables, same for int. Not a big deal but it helps readability a bit.

genesis/engine/solvers/fem_solver.py
YilingQiao: I think it's okay to use static here because Taichi actually treats this as a static variable during compilation.

genesis/engine/entities/fem_entity.py
YilingQiao: done

genesis/engine/entities/fem_entity.py
YilingQiao: done

genesis/engine/materials/FEM/elastic.py
YilingQiao: done

tests/test_fem.py
YilingQiao: done

tests/test_fem.py
YilingQiao: done

tests/test_fem.py
YilingQiao: done

tests/test_fem.py
YilingQiao: done

genesis/engine/materials/FEM/elastic.py
Libero0809: Done

genesis/engine/solvers/fem_solver.py
Libero0809: Done.

genesis/engine/solvers/fem_solver.py
Libero0809: Done

genesis/engine/solvers/fem_solver.py
Libero0809: Done

genesis/engine/solvers/fem_solver.py
Libero0809: Done

genesis/engine/solvers/fem_solver.py
Libero0809: Done

genesis/engine/solvers/fem_solver.py
Libero0809: Done

genesis/engine/solvers/fem_solver.py
Libero0809: Yes, it will get reformatted.

genesis/engine/solvers/fem_solver.py
Libero0809: Same as above

genesis/engine/solvers/fem_solver.py
Libero0809: Same

genesis/engine/solvers/fem_solver.py
Libero0809: done

genesis/options/solvers.py
Libero0809: Done

genesis/options/solvers.py
Libero0809: Done

genesis/options/solvers.py
Libero0809: Done

tests/test_fem.py
Libero0809: Done. I forgot to change it back cause CPU took a while to run the test.

genesis/engine/entities/fem_entity.py
Libero0809: Done

genesis/engine/solvers/fem_solver.py
Libero0809: Done

genesis/engine/solvers/fem_solver.py
duburcqa: No it is fine.

genesis/engine/materials/FEM/elastic.py
duburcqa: Yes

duburcqa: If they are alone on their line.

genesis/engine/solvers/fem_solver.py
duburcqa: I mean yeah, but it is my point actually. I see this as a problem here, because you compile statically based on sometimes that is entity dependent. Which means that recompiling will still occur even after implementing dynamic array. But we may investigate this later. @YilingQiao

genesis/options/solvers.py
duburcqa: Just docstring

tests/test_fem.py
duburcqa: The remove the parameter entirely so that the user can pick the backend he wants if it works fine in both cases!

duburcqa (approved):

genesis/engine/solvers/fem_solver.py
YilingQiao: Yeah, maybe we can loop over all materials (we don't have many). The number of materials is quite stable unless someone creates a new one. This can avoid most of the recompilation. We can revisit it later.

genesis/options/solvers.py
YilingQiao: done

duburcqa (approved):

PR #1300, opened 2025-06-20
[FEATURE] USD Materials Baking!

ACMLCZH:
Added USD materials baking for USD parsing, now all materials of USD can be parsed into Genesis as `UsdPreviewSurface` with just a little quality loss. But (those Omniverse libraries) only for Python 3.10 and GPU backend now. Remember to install Genesis with `[usd]` option.


duburcqa:
Both python 3.10 and 3.11 are supported for Linux and Windows. Update the CI accordingly.

.github/workflows/generic-cpu.yml
```diff
@@ -90,7 +90,10 @@ jobs:
 
       - name: Install Genesis
         run: |
-          pip install -e '.[dev,render]'
+          pip install -e '.[dev,render,usd]'
+          if [[ "${{ matrix.PYTHON_VERSION }}" == '3.10' ]] ; then
```
duburcqa: Add support of Python 3.11. Exclude MacOS.

genesis/options/textures.py
```diff
@@ -194,3 +200,6 @@ def apply_cutoff(self, cutoff):
         if cutoff is None or self.image_array is None:  # Cutoff does not apply on image file.
             return
         self.image_array = np.where(self.image_array >= 255.0 * cutoff, 255, 0).astype(np.uint8)
+
+    def is_black(self):
+        return np.max(self.image_array) == 0 or all(c == 0.0 for c in self.image_color)
```
duburcqa: Flip the order of these 2 conditions for efficiency.

genesis/utils/usda.py
```diff
@@ -1,11 +1,23 @@
 import genesis as gs
 from . import mesh as mu
+from .usda_bake import BAKE_EXT
+
+try:
+    from pxr import Usd, UsdGeom, UsdShade
+except ImportError as e:
+    gs.raise_exception(
```
duburcqa: Use `gs.raise_exception_from`.

Never "parse" the error traceback yourself as you did, this is almost always a bad idea.

genesis/utils/usda.py
```diff
@@ -1,11 +1,23 @@
...
+        f"Failed to import USD related libs. {e.__class__.__name__}: {e}."
+        "Please make sure that installing Genesis with [usd] option."
+    )
 
-from pxr import Usd, UsdGeom, UsdShade
-import trimesh
-import numpy as np
 from PIL import Image
+import subprocess
```
duburcqa: The order of all the imports in this file are completely messed up. You should first import standards libraries, skip line, then ubiquitous third-party libraries (torch, numpy, tqdm...etc), skip line, then random third-party libraries, skip line, then our own custom libraries, skip line, then relative genesis imports. For each group, make sure it is sorted alphabetically.

genesis/utils/usda.py
```diff
@@ -31,10 +43,12 @@ def get_input_attribute_value(shader, input_name, input_type=None):
...
 
 def get_texture_image(image_path, zipfiles):
-    if zipfiles is None:
-        return np.array(Image.open(image_path.resolvedPath))
-    else:
-        return np.array(Image.open(io.BytesIO(zipfiles.GetFile(image_path.path))))
+    if zipfiles is not None:
+        try:
+            return np.array(Image.open(io.BytesIO(zipfiles.GetFile(image_path.path))))
+        except:
```
duburcqa: Never do `except:`. Please specify which exception you are catching, and add a comment in the except clause to explain why it is fine to pass.

Prefer `np.asarray` in this context.

duburcqa: Do not randomly rely on Try-catch logic to spare checking the validity of your argument. You should rather make sure that `image_path.path` is a valid zip file instead of waiting for it to raise an exception.

genesis/utils/usda.py
```diff
@@ -74,8 +88,11 @@ def parse_component(component_name, component_encode):
 
         # parse emissive
         emissive_texture, emissive_uvname = parse_component("emissiveColor", "srgb")
-        if emissive_uvname is not None and uvname is None:
-            uvname = emissive_uvname
+        if emissive_uvname is not None:
```
duburcqa: add `emissive_texture is not None` or completeness and avoid future type-checker issue.

genesis/utils/usda.py
```diff
@@ -107,7 +124,7 @@ def parse_component(component_name, component_encode):
             "emissive_texture": emissive_texture,
             "normal_texture": normal_texture,
             "ior": ior,
-        }, uvname
+        }, (uvname if uvname else "st")
```
duburcqa: Rather add:
```python
if uvname is None:
    uvname = "st"
```
After line 96. It would be more explicit and more readable.

genesis/utils/usda.py
```diff
@@ -306,19 +191,105 @@ def parse_usd_material(material, surface, zipfiles):
...
+        gs.logger.warning(f"Material require baking:\n{candidates_str}")
+    return material_surface, uv_name, require_bake
 
-def parse_mesh_usd(path, group_by_material, scale, surface):
+
+def parse_mesh_usd(path, group_by_material, scale, surface, bake_cache=True):
     zipfiles = Usd.ZipFile.Open(path) if path.endswith(".usdz") else None
+
+    find_bake_cache = False
+    baked_path = f"{os.path.splitext(path)[0]}_baked.{BAKE_EXT}"
```
duburcqa: Use `pathlib` to do this kind of path processing.

genesis/utils/usda.py
```diff
@@ -306,19 +191,105 @@ def parse_usd_material(material, surface, zipfiles):
...
+    if bake_cache:
+        if os.path.exists(baked_path):
```
duburcqa: Same, rely on `pathlib`.

genesis/utils/usda.py
```diff
@@ -306,19 +191,105 @@ def parse_usd_material(material, surface, zipfiles):
...
+        )
+        gs.logger.warning(f"Material require baking:\n{candidates_str}")
+    return material_surface, uv_name, require_bake
 
-def parse_mesh_usd(path, group_by_material, scale, surface):
+
+def parse_mesh_usd(path, group_by_material, scale, surface, bake_cache=True):
     zipfiles = Usd.ZipFile.Open(path) if path.endswith(".usdz") else None
+
+    find_bake_cache = False
```
duburcqa: `is_baked_file_cached` would be more clear. Try starting boolean flag with `is_` prefix.

genesis/utils/usda.py
```diff
@@ -140,160 +157,28 @@ def parse_component(component_name, component_encode):
         return primvar_name
 
 
-def parse_gltf_surface(shader, source_type, output_name, zipfiles):
```
duburcqa, line 143: I think you should keep `parse_gltf_surface` and `parse_omni_surface` as a fallback when the required dependencies are missing (but still raise a warning). This is important because these extra dependencies are not available on MacOS nor Python 3.12.

genesis/vis/raytracer.py
```diff
@@ -157,19 +157,11 @@ def __init__(self, options, vis_options):
         self.lights = []
         for light in options.lights:
             light_intensity = light.get("intensity", 1.0)
-            self.lights.append(
```
duburcqa, line 160: Remove this change if unrelated. If it is, then keep all arguments on multiple lines for readability.

.github/workflows/generic-cpu.yml
ACMLCZH: Python 3.11 can truly install omniverse-kit, but when using the baking library it will throw an error, so I did not use Python 3.11.

tests/test_mesh.py
```diff
@@ -292,3 +295,33 @@ def test_usd_parse(usd_filename):
         check_gs_textures(
             gs_glb_material.emissive_texture, gs_usd_material.emissive_texture, 0.0, material_name, "emissive"
         )
+
+
+@pytest.mark.skipif(sys.version_info[:2] != (3, 10), reason="USD Baking works only in Python == 3.10")
```
duburcqa: Skip if "not (Python in (3.10, 3.11) and platform in (Linux, Windows)".

tests/test_mesh.py
```diff
@@ -292,3 +295,33 @@ def test_usd_parse(usd_filename):
...
+@pytest.mark.parametrize(
+    "usd_file", ["usd/WoodenCrate/WoodenCrate_D1_1002.usda", "usd/franka_mocap_teleop/table_scene.usd"]
+)
+@pytest.mark.parametrize("backend", [gs.gpu])
+def test_usd_bake(usd_file):
+    asset_path = get_hf_assets(
+        num_retry=1,
+        local_dir="tmp",  # If not using local dir, the symbol link of snapshot download will cause errors.
```
duburcqa: I don't understand your comment: "# If not using local dir, the symbol link of snapshot download will cause errors."

There is no case where specifying "local_dir" should be necessary. You must be doing something wrong and `get_hf_assets` is broken.

tests/utils.py
```diff
@@ -194,11 +202,11 @@ def get_hf_assets(pattern, num_retry: int = 4, retry_delay: float = 30.0, check:
                         raise HTTPError(f"Impossible to parse XML file.") from e
             if not has_files:
                 raise HTTPError("No file downloaded.")
-        except HTTPError:
+        except HTTPError as e:
```
duburcqa: This change is wrong. Remove it. It will raise the original exception after max number of trials anyway.

tests/utils.py
duburcqa, line 1: Remove all changes in this file. They should never be necessary and you are completely messing with the cache mechanism.

genesis/utils/usda.py
ACMLCZH: What is the best practice here, could you please instruct me?

genesis/utils/usda.py
ACMLCZH: However, parse_gltf_surface and parse_omni_surface actually cover few amounts of conditions.

tests/test_mesh.py
ACMLCZH: When using `snapshot_download()` from Hugging Face without specifying the `local_dir` parameter, the assets are stored in a `.blob` directory as binary files **without any file extensions**, and then soft symbolic links are created in the `asset_path`.
However, Omniverse Kit internally resolves the `realpath` when opening a stage (e.g. a `.usd` file). Since the `realpath` points to a file with no extension, the loader fails to recognize it properly, triggering a stage parsing error...
I tried hard to locate this error, and specifying a `local_dir` is the only solution I could come up with. :(

genesis/utils/usda.py
duburcqa: Ok you you think it covers so few cases that it was useless in the first place?

tests/test_mesh.py
duburcqa: Ok I see. I don’t think it is the right way to mitigate this issue though. It should probably be done directly in our utility to parse usd. The idea is to detect whether the path as incorrect extension of resolving it but not before. Then in such a case copy the path with the right extension in a temporary directory and pass the latter to the low-level parser at this point. Because hugging face download is only the tip of the issue. It may fail on many real applications relying on symlinks.

genesis/utils/usda.py
ACMLCZH: True. `parse_gltf_surface` and `parse_omni_surface` are initially used to handle a few special cases in a specific project, especially `parse_omni_surface`.

duburcqa: Ok I see. Let’s remove them in this case.

tests/test_mesh.py
```diff
@@ -292,3 +295,32 @@ def test_usd_parse(usd_filename):
...
+    gs_usd_meshes = [gs_usd_mesh for gs_usd_mesh in gs_usd_meshes if gs_usd_mesh.metadata["baked"]]
+    usd_dir = os.path.dirname(usd_file)
+    texture_folders = [folder for folder in os.listdir(usd_dir) if folder.startswith("baked_textures")]
+    assert len(texture_folders) == len(gs_usd_meshes)
+
+    baked_usd_file = f"{os.path.splitext(usd_file)[0]}_baked.{usda_utils.BAKE_EXT}"
+    os.remove(baked_usd_file)
+    for file in os.listdir(usd_dir):
+        if file.startswith("baked_textures"):
+            shutil.rmtree(os.path.join(usd_dir, file))
```
duburcqa: No need to delete any file. This is not done automatically.

.github/workflows/generic-cpu.yml
duburcqa: This is not a good excuse. You need to find out why it is not working for Python 3.11. Because it should be fine. ignoring Python 3.11 must be very well motivated because it significantly impede adoption of Genesis’ USD parsing by the end-users.

genesis/utils/usda.py
duburcqa:
```python
from pathlib import Path

path = Path(path)
baked_path = path.with_name(path.stem + f"_baked.{BAKE_EXT}")
if bake_cache and baked_path.exists():
    [...]
```

tests/test_mesh.py
ACMLCZH: Will two test functions like `test_usd_bake("usd/WoodenCrate/WoodenCrate_D1_1002.usda")` and `test_usd_bake1("usd/WoodenCrate/WoodenCrate_D1_1002.usda")`(if we have) use the same OS system context? If so, the baked textures generated by the first test will affect the second test. If not, these file deletion codes can be removed.

.github/workflows/generic-cpu.yml
ACMLCZH: When you use USD baking in python 3.11, it will encounter the following errors.
```
2025-07-01 22:50:35 [2,271ms] [Error] [omni.ext.plugin] Failed to resolve extension dependencies. Failure hints:
	Can't find extension with name: omni.replicator.core
 Available versions:
	 (none found)
 Platform incompatible packages:
	 - [omni.replicator.core-1.12.10+107.3.0.la64.d.cp311] (registry)
	 - [omni.replicator.core-1.12.10+107.3.0.wx64.r.cp311] (registry)
	 - [omni.replicator.core-1.12.10+107.3.0.la64.r.cp311] (registry)
	 - [omni.replicator.core-1.12.10+107.3.0.wx64.d.cp311] (registry)
	 - [omni.replicator.core-1.12.10+107.3.0.lx64.d.cp311] (registry)
	 - [omni.replicator.core-1.12.10+107.3.0.lx64.r.cp311] (registry)
 Synced registries:
	 - kit/default         : found 0 packages (couldn't connect or empty)
	 - kit/sdk             : found 0 packages (couldn't connect or empty)
	 - kit/prod/default    : found 485 packages 
	 - kit/prod/sdk        : found 291 packages 

2025-07-01 22:50:35 [2,272ms] [Error] [omni.kit.app.plugin] Exiting app because of dependency solver failure...
```
It is because `omni.replicator.core` in python 3.11 requires `omniverse-kit==107.3`. However, Nvidia does not release version 107.3 for it in PyPI, and only 107.2 for python 3.11 is available. And there is no easy way to install `omniverse-kit` with version 107.3, so I disabled it. What else should I do?

duburcqa: I see. This is a mess and very restrictive overall. Can you add a comment somewhere to summarise what you just said so that we can remember it? Eventually with a link to this discussion.

tests/utils.py
```diff
@@ -181,8 +182,11 @@ def get_hf_assets(pattern, num_retry: int = 4, retry_delay: float = 30.0, check:
             # Make sure that download was successful
             has_files = False
             for path in Path(asset_path).rglob(pattern):
+                cache_path = os.path.join(asset_path, ".cache")
```
duburcqa: Move this path outside the for loop.

tests/utils.py
```diff
@@ -181,8 +182,11 @@ def get_hf_assets(pattern, num_retry: int = 4, retry_delay: float = 30.0, check:
...
                 if not path.is_file():
                     continue
+                if str(path).startswith(cache_path):
```
duburcqa: You should rather do: `cache_path in path`

Bu what is the point in the first place? Can you add a comment to explain the rational behind it?

tests/test_mesh.py
duburcqa: Ok I see. But if this is a concern, there must be a design flaw somewhere. Does it take a long time to generate the backend textures? But if it does not, I would suggest storing the backed textures in some temporary folder unrelated to Genesis source or asset directory. If it does, some kind of caching mechanism is indeed necessary. In this case, I would suggest using the same logic as Taichi and many other packages for the cache directory location: https://github.com/taichi-dev/taichi/blob/a1d99c9c4cef9de5f9b633a515d227677ffab9a0/taichi/common/core.cpp#L33

python-3.12.3-macos11.pkg
duburcqa, line 1: Remove this file.

.github/workflows/linux-gpu.yml
```diff
@@ -53,7 +53,8 @@ jobs:
             --partition=hpc-mid ${NODELIST} --nodes=1 --time="${TIMEOUT_MINUTES}" \
             --job-name=${SLURM_JOB_NAME} \
             bash -c "
-              pip install -e '.[dev,render]' && \
+              pip install --extra-index-url https://pypi.nvidia.com/ omniverse-kit
```
duburcqa: Add trailing ` && \`

.github/workflows/generic-cpu.yml
duburcqa: Just 1 or 2 lines max + reference would be fine. You added too much details here. The point of the comment is to remember that we did not forgot about other python versions, but rather wheels that are either not available for broken in a way.

tests/utils.py
ACMLCZH: Since if you specify `local_dir` in `snapshot_download`, the `.cache` folder will appear in `asset_path` (that is `local_dir`) to store the installed files. Some tmp files will cause `path.stat().st_size` to be `0`. So traversing all files might be unsafe. If you don't specify it, `.cache` folder will be a shared folder outside of `asset_path`.

tests/utils.py
```diff
@@ -169,6 +169,7 @@ def get_hf_assets(pattern, num_retry: int = 4, retry_delay: float = 30.0, check:
 
     for _ in range(num_retry):
         num_trials = 0
+
```
duburcqa: Could you remove this change?

tests/test_mesh.py
```diff
@@ -294,6 +298,35 @@ def test_usd_parse(usd_filename):
         )
 
 
+@pytest.mark.skipif(sys.version_info[:2] != (3, 10), reason="USD Baking works only in Python == 3.10 now")
```
duburcqa: It should be more restrictive. Skip if "not (Python == 3.10 and platform in (Linux, Windows)".

tests/test_mesh.py
```diff
@@ -294,6 +298,35 @@ def test_usd_parse(usd_filename):
...
+@pytest.mark.parametrize(
+    "usd_file", ["usd/WoodenCrate/WoodenCrate_D1_1002.usda", "usd/franka_mocap_teleop/table_scene.usd"]
+)
+@pytest.mark.parametrize("backend", [gs.gpu])
```
duburcqa: Are you sure you need to enforce GPU backend?

ACMLCZH: Yes, Omniverse kit does not support CPU backend

duburcqa: I don't understand. This backend is for Taichi. How is that related to Omniverse kit for USD loading?

genesis/utils/usda_bake.py
```diff
@@ -0,0 +1,127 @@
+import os
+import argparse
+import logging
+import asyncio
+
+os.environ["OMNI_KIT_ACCEPT_EULA"] = "yes"
```
duburcqa: I don't think we should accept EULA in the behalf of the user. You should set this environment variable on the CI, but not directly in Genesis source code.

genesis/utils/usda_bake.py
```diff
@@ -0,0 +1,127 @@
...
+        "--enable",
+        "omni.mdl.distill_and_bake",
+    ]
+    app.startup(app_args)
+    app.update()  # important
+    return app
+
+
+def bake_usd_material(input_file, output_dir, usd_material_paths, device=0, log_level="error"):
+    # sys.stdout = open(os.devnull, "w")
```
duburcqa: Remove comments. If you need to hide some "prints" for other libraries, either use
```python
from contextlib import redirect_stdout

 with open(os.devnull, 'w') as stdout, redirect_stdout(stdout):
     [...]
```
for Python-scope prints.

Or 
```python
from genesis.utils.misc import redirect_libc_stderr

with open(os.devnull, "w") as stderr, redirect_libc_stderr(stderr):
    [...]
```
for C++ scope stderr silencing.

genesis/utils/usda_bake.py
```diff
@@ -0,0 +1,127 @@
...
+    # baked_file = os.path.join(output_dir, f"{output_filename}.{BAKE_EXT}")
+    collector = omni.kit.usd.collect.Collector(stage.GetRootLayer().identifier, output_dir)
+    task = asyncio.ensure_future(collector.collect())
+    while not task.done():
+        app.update()  # Otherwise it will be blocked by omni.kit.app.get_app().next_update_async()
+    success, baked_file = task.result()
+    # stage.GetRootLayer().Export(baked_file)
+
+    logs.append(f"\tExport: {time.time() - start_time}s, {baked_file}.")
+    print("Distill USD material:\n" + "\n".join(logs))
```
duburcqa, line 97: Use `gs.logger.debug` or `gs.logger.info`. No barebone prints.

genesis/utils/usda_bake.py
```diff
@@ -0,0 +1,127 @@
...
+
+    import omni.usd
+    import omni.mdl.distill_and_bake
+    import omni.replicator.core
+    import omni.kit.usd.collect
+
+    # open stage
+    start_time = time.time()
+    omni.usd.get_context().open_stage(input_file)
+    logs.append(f"\tOpen stage: {time.time() - start_time}s.")
```
duburcqa, line 70: Why are you storing prints in a list and waiting for the end of the processing to finally display them all at once?

genesis/utils/usda_bake.py
duburcqa: If you want to store the messages in a buffer and delay printing, you can do this:
```python
import io
from contextlib import redirect_stdout

_outputs = io.StringIO()

with redirect_stdout(_outputs):
    [...]

for line in _outputs.getvalue().splitlines():
    gs.logger.debug(line)
```

genesis/utils/usda_bake.py
```diff
@@ -0,0 +1,127 @@
...
+
+    for usd_material_path in usd_material_paths:
+        material_prim = stage.GetPrimAtPath(usd_material_path)
+        distiller = omni.mdl.distill_and_bake.MdlDistillAndBake(material_prim, ouput_folder=output_dir)
+        distiller.distill()
+    logs.append(f"\tDistill: {time.time() - start_time}s, {material_prim}.")
+
+    # export usd
+    start_time = time.time()
+    # baked_file = os.path.join(output_dir, f"{output_filename}.{BAKE_EXT}")
```
duburcqa: Remove commented code.

tests/test_mesh.py
```diff
@@ -294,6 +298,35 @@ def test_usd_parse(usd_filename):
         )
 
 
+@pytest.mark.skipif(
+    sys.version_info[:2] != (3, 10) or sys.platform not in ("Linux", "Windows"),
+    reason="USD Baking only works in Python == 3.10 and OS == ('Linux', 'Windows') now."
```
duburcqa: You should rather mention what is the technical limitation. The reason you mentioned is already pretty obvious for anybody reading the code.

genesis/utils/usda_bake.py
ACMLCZH: The lib version (like numpy) in Genesis conflict with the internal libs in Omni. That's why I start a new script and use subprocess to run it because they cannot be used together.

tests/test_mesh.py
```diff
@@ -294,6 +298,35 @@ def test_usd_parse(usd_filename):
...
+)
+@pytest.mark.parametrize(
+    "usd_file", ["usd/WoodenCrate/WoodenCrate_D1_1002.usda", "usd/franka_mocap_teleop/table_scene.usd"]
+)
+@pytest.mark.parametrize("backend", [gs.gpu])
+def test_usd_bake(usd_file):
+    asset_path = get_hf_assets(
+        num_retry=1,
```
duburcqa: You should not specify `num_retry=1`

tests/test_mesh.py
```diff
@@ -294,6 +298,35 @@ def test_usd_parse(usd_filename):
...
+        scale=1.0,
+        surface=gs.surfaces.Default(),
+        bake_cache=False
+    )
+    for gs_usd_mesh in gs_usd_meshes:
+        require_bake = gs_usd_mesh.metadata["require_bake"]
+        bake_success = gs_usd_mesh.metadata["bake_success"]
+        assert not require_bake or require_bake and bake_success
+    baked_folder = mu.get_usd_bake_path(gs_usd_meshes[0].metadata["path"])
+    shutil.rmtree(baked_folder)
```
duburcqa: Why is this needed ? I don't get it.

genesis/utils/usda.py
```diff
@@ -1,19 +1,31 @@
...
 from PIL import Image
-import io
+import numpy as np
+import trimesh
+
+try:
+    from pxr import Usd, UsdGeom, UsdShade, Sdf
+except ImportError as e:
+    gs.raise_exception_from(
+        "Failed to import USD related libs. Please make sure that installing Genesis with [usd] option.", e
```
duburcqa: "Genesis with [usd] option." -> "Genesis with 'usd' optional dependencies.

genesis/utils/usda.py
```diff
@@ -1,19 +1,31 @@
...
-import numpy as np
 from PIL import Image
-import io
+import numpy as np
+import trimesh
+
+try:
+    from pxr import Usd, UsdGeom, UsdShade, Sdf
+except ImportError as e:
+    gs.raise_exception_from(
```
duburcqa, line 20: You should raise a different exception depending on whether `[usd]` can be installed or not. Check Linux or Windows + Python 3.10.

genesis/utils/usda.py
```diff
@@ -1,19 +1,31 @@
...
+        "Failed to import USD related libs. Please make sure that installing Genesis with [usd] option.", e
+    )
+
+
+import genesis as gs
+from . import mesh as mu
```
duburcqa: Skip line before relative imports.

genesis/utils/usda.py
```diff
@@ -1,19 +1,31 @@
...
+import trimesh
+
+try:
+    from pxr import Usd, UsdGeom, UsdShade, Sdf
+except ImportError as e:
+    gs.raise_exception_from(
+        "Failed to import USD related libs. Please make sure that installing Genesis with [usd] option.", e
+    )
+
+
```
duburcqa, line 24: Only skip one line, not two.

genesis/utils/usda.py
```diff
@@ -30,14 +42,15 @@ def get_input_attribute_value(shader, input_name, input_type=None):
     return None, None
 
 
-def get_texture_image(image_path, zipfiles):
-    if zipfiles is None:
-        return np.array(Image.open(image_path.resolvedPath))
-    else:
-        return np.array(Image.open(io.BytesIO(zipfiles.GetFile(image_path.path))))
+def get_texture_image(image_path):
+    # if zipfiles is not None:
```
duburcqa: Remove commented code.

genesis/utils/usda.py
```diff
@@ -30,14 +42,15 @@ def get_input_attribute_value(shader, input_name, input_type=None):
     return None, None
 
 
-def get_texture_image(image_path, zipfiles):
-    if zipfiles is None:
-        return np.array(Image.open(image_path.resolvedPath))
-    else:
-        return np.array(Image.open(io.BytesIO(zipfiles.GetFile(image_path.path))))
+def get_texture_image(image_path):
```
duburcqa: Remove this entire function. It is useless for a single line of code that is straightforward.

genesis/utils/usda.py
```diff
@@ -30,14 +42,15 @@ def get_input_attribute_value(shader, input_name, input_type=None):
...
+    # if zipfiles is not None:
+    #     image_data = zipfiles.GetFile(image_path.path)
+    #     if image_data is not None:
+    #         return np.asarray(Image.open(io.BytesIO(image_data)))
+    return np.asarray(Image.open(image_path.resolvedPath))
```
duburcqa: specify `dtype`.

genesis/utils/usda.py
```diff
@@ -306,19 +191,165 @@ def parse_usd_material(material, surface, zipfiles):
...
         )
-    return material_surface, uv_name
+
+    if require_bake:
+        candidates_str = "\n".join(
+            [
+                f"\tShader at {shader_path} with implement {shader_impl} and ID {shader_id}."
+                for shader_path, shader_id, shader_impl in material_candidates
+            ]
+        )
```
duburcqa:
```python
        candidates_str = "\n".join(
            f"\tShader at {shader_path} with implement {shader_impl} and ID {shader_id}."
            for shader_path, shader_id, shader_impl in material_candidates
        )
````

File: genesis/utils/usda.py [hunk 39]
```diff
@@ -306,19 +191,165 @@ def parse_usd_material(material, surface, zipfiles):
...
+        gs.logger.warning(f"Material require baking:\n{candidates_str}")
```
duburcqa: Why a warning ? Is there any issue that the user should be paying attention to?

genesis/utils/usda.py
```diff
@@ -306,19 +191,165 @@ def parse_usd_material(material, surface, zipfiles):
...
+                *baked_material_paths,
+                "--device",
+                str(gs.device.index),
+                "--log_level",
+                logging.getLevelName(gs.logger.level).lower(),
+            ]
+            gs.logger.debug(f"Execute: {' '.join(commands)}")
+
+            # Each material is estimated to bake at most 4s, and boostrap and initialization for 10s.
+            result = subprocess.run(
```
duburcqa, line 319: Only call sub-processes for libraries you don't have control over. In this case, `usda_bake.py` is our own script. Just find a way to expose a function that could be called here.

genesis/utils/usda.py
```diff
@@ -306,19 +191,165 @@ def parse_usd_material(material, surface, zipfiles):
...
+
+            if material_pack is None:
+                material, uv_name, require_bake = parse_usd_material(material_usd, surface)
+                materials[material_id] = (material, uv_name)
+                if not is_bake_cache_found and require_bake:
+                    baked_materials[material_id] = material_usd.GetPath()
+
+    if baked_materials:
+        baked_material_paths = list(map(str, baked_materials.values()))
+        baker_file = os.path.join(os.path.dirname(os.path.abspath(__file__)), "usda_bake.py")
```
duburcqa: This is definitely not the right way to call python script. They should be implemented as proper python "entrypoints" to avoid requiring passing the full path of the file as you are doing.

Anyway, using subprocess should be avoided here, so you can complete forget my comment.

genesis/utils/usda.py
ACMLCZH: I have tried to use a function call but it will initialize Genesis anyway, could you instruct me a way to avoid it?

genesis/utils/usda.py
ACMLCZH: Texture may include some floating type images so it is hard to specify dtype here. And it may be resolved [here](https://github.com/ACMLCZH/Genesis/blob/84017c3acb2b2113046b2ff731819f96541b60df/genesis/options/textures.py#L132)

genesis/utils/usda.py
ACMLCZH: Python version and platform only affects the dependency `omniverse-kit`. `pxr` here (installed by `usd-core`) is not affected.

genesis/utils/usda_bake.py
duburcqa: Could you be a bit more specific? Do you have reproduction script so that I can better understand the issue?

genesis/utils/usda.py
duburcqa: You mean this module (`usda.py`) can still be used in a meaning full way even if `omniverse-kit` cannot be installed?

genesis/utils/usda.py
duburcqa: I see. Then you did the right thing.

genesis/utils/usda.py
duburcqa: Maybe we should have a video call so that you could explain may what is the issue exactly here. Because for me it is not clear why "initialize Genesis anyway" would be an issue.

tests/test_mesh.py
```diff
@@ -294,6 +298,27 @@ def test_usd_parse(usd_filename):
         )
 
 
+@pytest.mark.skipif(
+    sys.version_info[:2] != (3, 10) or sys.platform not in ("Linux", "Windows"),
+    reason="omniverse-kit used by USD Baking cannot be correctly installed other than specified python version and platform now.",
```
duburcqa: `cannot be correctly installed on this platform.` would be fine.

genesis/utils/usda_bake.py
ACMLCZH: Edit `usda_bake.py` to add `import genesis as gs` at the beginning of this script or somewhere. Then run `python /home/zhehuan/Desktop/hz/Genesis/genesis/utils/usda_bake.py --input_file /Path/to/tests/WoodenCrate/WoodenCrate_D1_1002.usda --output_dir /path/to/home/.cache/genesis/usd/bake/f2a276b57e98a72c8070705a3247750b305220fd1fce68cdff861740da572b23 --usd_material_paths /Root/World/Looks/Crate --device 0 --log_level debug`. You will find some conflict errors.

genesis/utils/usda.py
ACMLCZH: Exactly, `pxr` module only depends on `usd-core` installation, but it will take 130MB space and considering not so many users will use USD so we make it an option.

genesis/utils/usda.py
ACMLCZH: I'm OK if you want

genesis/utils/usda_bake.py
ACMLCZH: Initially I was trying to use a function entry point to call USD baking, so I collect the logs for returning values. However, using function failed, but I reserved these codes for potential returning.

ACMLCZH:
Hello, when can we proceed on this PR?

duburcqa:
Yes, I will work on it ASAP

duburcqa (dismissed):

duburcqa (approved):

duburcqa (approved):

PR #1304, opened 2025-06-21
[FEATURE] Add linear_corotated elastic material for fem

Libero0809:
## Description

Add linear_corotated elastic material for fem. This model is more accurate than the linear model. Linear model cannot handle global rotation and will cause artifact.

## Related Issue

## Motivation and Context

A more accurate model for fem simulation.

Below are the videos showing linear elastic model and the new linear corotated model.

https://github.com/user-attachments/assets/35381a4d-cfee-4298-a445-8cbb2328c2a2

https://github.com/user-attachments/assets/c2c22708-8970-4787-bdce-e24b39eaac11

## Checklist:


python-3.12.3-macos11.pkg
SonSang, line 1: let's delete this file

Libero0809: Done

genesis/engine/materials/FEM/elastic.py
```diff
@@ -354,6 +376,147 @@ def compute_energy_stable_neohookean(self, mu, lam, J, F, actu, m_dir):
...
+        -------
+        This implementation assumes small deformations and linear stress-strain relationship.
+        It is adapted from the HOBAKv1 implementation for linear elasticity:
+        https://github.com/theodorekim/HOBAKv1/blob/main/src/Hyperelastic/Volume/LINEAR.cpp
+
+        """
+        I = ti.Matrix.identity(dt=gs.ti_float, n=3)
+        R = self.R[i_b, i_e]
+        F_hat = R.transpose() @ F
+        eps = 0.5 * (F_hat + F_hat.transpose()) - I
```
duburcqa: You should rather just subtract 1.0 from the diagonal terms:
```python
for i in ti.static(range(3)):
    eps[i] = eps[I] - 1.0
```

genesis/engine/materials/FEM/elastic.py
```diff
@@ -354,6 +376,147 @@ def compute_energy_stable_neohookean(self, mu, lam, J, F, actu, m_dir):
...
+        Notes
+        -------
+        This implementation assumes small deformations and linear stress-strain relationship.
+        It is adapted from the HOBAKv1 implementation for linear elasticity:
+        https://github.com/theodorekim/HOBAKv1/blob/main/src/Hyperelastic/Volume/LINEAR.cpp
+
+        """
+        I = ti.Matrix.identity(dt=gs.ti_float, n=3)
+        F_hat = self.R[i_b, i_e].transpose() @ F
+        eps = 0.5 * (F_hat + F_hat.transpose()) - I
```
duburcqa: same.

genesis/engine/materials/FEM/elastic.py
```diff
@@ -354,6 +376,147 @@ def compute_energy_stable_neohookean(self, mu, lam, J, F, actu, m_dir):
...
```
duburcqa: same.

genesis/engine/solvers/fem_solver.py
```diff
@@ -396,6 +399,23 @@ def compute_pos(self, f: ti.i32):
...
+    def precompute_material_data(self, f: ti.i32):
+        for i_b, i_e in ti.ndrange(self._B, self.n_elements):
+            if not self.batch_active[i_b]:
+                continue
+
+            J, F = self._compute_ele_J_F(f, i_e, i_b)  # use last time step's pos to compute
+
+            for mat_idx in ti.static(self._mats_idx):
+                if self.elements_i[i_e].mat_idx == mat_idx:
+                    self._mats[mat_idx].pre_compute(
```
duburcqa: `self._mats[mat_idx].pre_compute(J=J, F=F, i_e=i_e, i_b=i_b)`

genesis/engine/solvers/fem_solver.py
```diff
@@ -483,8 +503,36 @@ def _func_compute_ele_energy(self, f: ti.i32):
...
                         m_dir=self.elements_i[i_e].muscle_direction,
+                        i_e=i_e,
+                        i_b=i_b,
+                    )
+
+            # add linearized damping energy
+            if self._damping_beta > gs.EPS:
+                damping_beta_over_dt = self._damping_beta / self._substep_dt
+                i_v = self.elements_i[i_e].el2v
+                S = ti.Matrix.zero(gs.ti_float, 4, 3)
```
duburcqa: Why `ti.Matrix.zero` instead of `ti.Vector.zero` like everywhere else?

tests/test_fem.py
```diff
@@ -287,3 +287,49 @@ def test_sphere_fall_implicit_fem_sap_coupler(fem_material_linear, show_viewer):
...
         ), f"Entity {entity.uid} minimum Z position {min_pos_z} is not close to 0.0."
+
+
+@pytest.fixture(scope="session")
+def fem_material_linear_corotated():
+    """Fixture for common FEM linear material properties"""
+    return gs.materials.FEM.Elastic(model="linear_corotated")
+
+
+@pytest.mark.parametrize("backend", [gs.gpu])
```
duburcqa: Why forcing GPU? Why not marking this test as required?

tests/test_fem.py
```diff
@@ -287,3 +287,49 @@ def test_sphere_fall_implicit_fem_sap_coupler(fem_material_linear, show_viewer):
...
+def test_linear_corotated_sphere_fall_implicit_fem_sap_coupler(fem_material_linear_corotated, show_viewer):
+    """Test adding multiple FEM entities to the scene"""
+    scene = gs.Scene(
+        sim_options=gs.options.SimOptions(
+            dt=1 / 60,
```
duburcqa: Use floats.

tests/test_fem.py
```diff
@@ -287,3 +287,49 @@ def test_sphere_fall_implicit_fem_sap_coupler(fem_material_linear, show_viewer):
...
+
+    # Run simulation
+    for _ in range(200):
+        scene.step()
+
+    for entity in scene.entities:
+        state = entity.get_state()
+        pos = state.pos.detach().cpu().numpy()
+        min_pos_z = np.min(pos[..., 2])
+        assert_allclose(
```
duburcqa, line 333: This assert is not sufficient. You should detect artifacts and check that the sphere does not get too much distorted.

genesis/engine/materials/FEM/elastic.py
Libero0809: Done

genesis/engine/materials/FEM/elastic.py
Libero0809: Done

genesis/engine/materials/FEM/elastic.py
Libero0809: Done

genesis/engine/solvers/fem_solver.py
Libero0809: Done

genesis/engine/solvers/fem_solver.py
Libero0809: I found that using a matrix would make the code cleaner. Using only the vector would require somethign like this:
```
            s = -B[0, :] - B[1, :] - B[2, :]  # s is the negative sum of B rows
            p9 = ti.Vector([0.0] * 9, dt=gs.ti_float)
            i_v0, i_v1, i_v2, i_v3 = self.elements_i[i_e].el2v

            for i in ti.static(range(3)):
                p9[i * 3 : i * 3 + 3] = (
                    B[0, i] * self.pcg_state_v[i_b, i_v0].p
                    + B[1, i] * self.pcg_state_v[i_b, i_v1].p
                    + B[2, i] * self.pcg_state_v[i_b, i_v2].p
                    + s[i] * self.pcg_state_v[i_b, i_v3].p
                )
```
Maybe I should change where I used vector to matrix.

tests/test_fem.py
Libero0809: Based on our discussion, should I change it to no backend and maybe not mark it as required as it may take some time.

tests/test_fem.py
Libero0809: Done

tests/test_fem.py
Libero0809: Done

genesis/engine/solvers/fem_solver.py
duburcqa: We are using Vector everywhere. I would suggest to rather stick to Vector, even if it makes the code look messy as you pointed out. What do you think @YilingQiao ?

tests/test_fem.py
duburcqa: Yes, no backend and not required for now. Then depending on the CI budget we have, we may revisit this decision very easily.

genesis/engine/solvers/fem_solver.py
Libero0809: Performance wise they should be the same right? I would recommend using both Matrix and Vector since some variables are intrinsically a matrix as the S or B here. In other places, where actually I want to use s static array of 4 vectors but taichi does not support that so that I can only use a 3-by-4 matrix or a vector of 12.

tests/test_fem.py
Libero0809: Done

genesis/engine/materials/FEM/elastic.py
```diff
@@ -354,6 +376,153 @@ def compute_energy_stable_neohookean(self, mu, lam, J, F, actu, m_dir):
...
+        # Zero out the matrix
+        for i in ti.static(ti.grouped(ti.ndrange(3, 3))):
+            hessian_field[i_b, i, i_e].fill(0.0)
+
+        # Identity part
+        for i, k in ti.static(ti.ndrange(3, 3)):
+            hessian_field[i_b, i, i, i_e][k, k] = mu
+
+        for i, j, alpha, beta in ti.static(ti.ndrange(3, 3, 3, 3)):
+            hessian_field[i_b, j, beta, i_e][i, alpha] += mu * R[i, beta] * R[alpha, j] + lam * R[alpha, beta] * R[i, j]
```
KemengHuang, line 435: The linear corotated energy and gradient are accurate, but I'm uncertain about the Hessian, particularly regarding the index. It would be advisable to test the Hessian using the finite difference method or to verify that the energy consistently decreases as expected.

Libero0809: This is Drake's code. I think they are using the same indexing:
```
template <typename T>
void LinearCorotatedModel<T>::CalcFirstPiolaStressDerivativeImpl(
    const Data& data, math::internal::FourthOrderTensor<T>* dPdF) const {
  const Matrix3<T>& R0 = data.R0();
  auto& local_dPdF = (*dPdF);
  /* Add in μ * δₐᵢδⱼᵦ. */
  local_dPdF = math::internal::FourthOrderTensor<T>::MakeMajorIdentity(mu_);
  for (int i = 0; i < 3; ++i) {
    for (int j = 0; j < 3; ++j) {
      for (int alpha = 0; alpha < 3; ++alpha) {
        for (int beta = 0; beta < 3; ++beta) {
          /* Add in  μ *  Rᵢᵦ Rₐⱼ +   λ * Rₐᵦ * Rᵢⱼ. */
          local_dPdF.mutable_data()(3 * j + i, 3 * beta + alpha) +=
              mu_ * R0(i, beta) * R0(alpha, j) +
              lambda_ * R0(alpha, beta) * R0(i, j);
        }
      }
    }
  }
}
```

Libero0809: I can add a test for hessian if you still think it's necessary to do so.

Libero0809: The hessian field is actually a 3 by 3 array of matrix3 so the indexing could be confusing. The full dimension is n_batch * 3 * 3 * n_elements of mat3

KemengHuang: I see, that should be fine then

duburcqa (approved):

PR #1305, opened 2025-06-21
[BUG FIX]  Fix BVH build's radix sort.

Libero0809:
## Description

Maybe continue this PR after the linear corotated is merged into main. I just realized that this branch is based on the corotated one.
- The BVH's radix sort use an unstable sort which will generate a wrong sorting result. Changed it to sequential for each batch so that the result is correct.
- BVH's compute bound is not gpu safe since it does not sync the memory. Changed to multi-kernel launch with forced memory sync.
- Make the tests harder for the bvh. The previous one was too simple.
- Break the large build kernel into several kernels to make it compile faster.

## Checklist:


genesis/engine/bvh.py
```diff
@@ -223,38 +224,44 @@ def expand_bits(self, v):
...
-                self.hist[i_b, j] = 0
+        for i in range(8):
+            self._kernel_radix_sort_morton_codes_one_round(i)
 
-            # Fill histogram
-            for i_b, i_a in ti.ndrange(self.n_batches, self.n_aabbs):
+    @ti.kernel
+    def _kernel_radix_sort_morton_codes_one_round(self, i: int):
+        # Clear histogram
+        for i_b, j in ti.ndrange(self.n_batches, 256):
```
duburcqa: I guess `self.hist.fill(0)` should do the trick?

genesis/engine/coupler.py
```diff
@@ -952,6 +976,102 @@ def fem_floor_detection(self, f: ti.i32):
...
+
+        # Compute data for each contact pair
+        for i_c in range(self.n_fem_floor_contact_pairs[None]):
+            pair = self.fem_floor_contact_pairs[i_c]
+            self.fem_floor_contact_pairs[i_c].active = 1  # mark the contact pair as active
+            i_b = pair.batch_idx
+            i_e = pair.geom_idx
+            intersection_code = pair.intersection_code
+            distance = pair.distance
+            intersected_edges = ti.static(self.kMarchingTetsEdgeTable)[intersection_code]
```
duburcqa: This is a weird syntax. Is static really necessary here?

genesis/engine/coupler.py
```diff
@@ -698,6 +698,30 @@ def init_fem_fields(self):
...
+            Rt=gs.ti_float,  # Regularitaion for tangential
+            vn_hat=gs.ti_float,  # Stablization for normal velocity
+            mu=gs.ti_float,  # friction coefficient
+            mu_hat=gs.ti_float,  # friction coefficient regularized
+            mu_factor=gs.ti_float,  # friction coefficient factor, 1/(1+mu_tilde**2)
+            energy=gs.ti_float,  # energy
+            G=gs.ti_mat3,  # Hessian matrix
+        )
+        self.max_fem_self_contact_pairs = fem_solver.n_surfaces * fem_solver._B * 8
+        self.n_fem_self_contact_pairs = ti.field(gs.ti_int, shape=())
```
duburcqa: This is not used?

genesis/engine/coupler.py
```diff
@@ -952,6 +976,102 @@ def fem_floor_detection(self, f: ti.i32):
...
+                i_v = fem_solver.elements_i[i_e].el2v[i]
+                tet_vertices[:, i] = fem_solver.elements_v[f + 1, i_v, i_b].pos
+                tet_pressures[i] = self.fem_pressure[i_v]
+
+            polygon_vertices = ti.Matrix.zero(gs.ti_float, 3, 4)  # 3 or 4 vertices
+            total_area = gs.EPS  # avoid division by zero
+            total_area_weighted_centroid = ti.Vector([0.0, 0.0, 0.0])
+            for i in range(4):
+                if intersected_edges[i] >= 0:
+                    edge = ti.static(self.kTetEdges)[intersected_edges[i]]
```
duburcqa: same.

genesis/engine/coupler.py
```diff
@@ -952,6 +976,102 @@ def fem_floor_detection(self, f: ti.i32):
...
+            barycentric = tet_barycentric(centroid, tet_vertices)
+            pressure = (
+                barycentric[0] * tet_pressures[0]
+                + barycentric[1] * tet_pressures[1]
+                + barycentric[2] * tet_pressures[2]
+                + barycentric[3] * tet_pressures[3]
+            )
+            self.fem_floor_contact_pairs[i_c].barycentric = barycentric
+
+            C = ti.static(1.0e8)
```
duburcqa: What is this? Create a global constant on top of this file with a proper name.

genesis/engine/coupler.py
```diff
@@ -952,6 +976,102 @@ def fem_floor_detection(self, f: ti.i32):
...
+            deformable_g = C
+            rigid_g = self.fem_pressure_gradient[i_b, i_e].z
+            # TODO A better way to handle corner cases where pressure and pressure gradient are ill defined
+            if total_area < gs.EPS or rigid_g < gs.EPS:
+                self.fem_floor_contact_pairs[i_c].active = 0
+                continue
+            g = 1.0 / (1.0 / deformable_g + 1.0 / rigid_g)  # harmonic average
```
duburcqa: `rigid_g < gs.EPS` would work just fine if you do `g = deformable_g * rigid_g * ti.rsqrt(deformable_g + rigid_g)` Anyway, it would be more robust numerically and faster to compute (even if it is not really a concern here...).

genesis/engine/coupler.py
```diff
@@ -952,6 +976,102 @@ def fem_floor_detection(self, f: ti.i32):
...
+                + barycentric[2] * tet_pressures[2]
+                + barycentric[3] * tet_pressures[3]
+            )
+            self.fem_floor_contact_pairs[i_c].barycentric = barycentric
+
+            C = ti.static(1.0e8)
+            deformable_g = C
+            rigid_g = self.fem_pressure_gradient[i_b, i_e].z
+            # TODO A better way to handle corner cases where pressure and pressure gradient are ill defined
+            if total_area < gs.EPS or rigid_g < gs.EPS:
```
duburcqa, line 941: `rigid_g < gs.EPS` should not be an issue mathematically at least (see following comment).

genesis/engine/materials/FEM/elastic.py
```diff
@@ -63,11 +64,28 @@ def __init__(
...
             gs.raise_exception(f"Unrecognized constitutive model: {model}")
 
         self._model = model
 
+    def build_linear_corotated(self, fem_solver):
+        self.R = ti.field(dtype=gs.ti_mat3, shape=(fem_solver._B, fem_solver.n_elements))
+
+    @ti.func
+    def pre_compute_linear_corotated(self, J, F, i_e, i_b):
+        U, S, V = ti.svd(F)
```
duburcqa, line 86: Add comment for clarity:
```python
# Computing Polar Decomposition instead of calling `R, P = ti.polar_decompose(F)` since `P` is not needed here
```

genesis/engine/bvh.py
Libero0809: Done

genesis/engine/coupler.py
Libero0809: self.kMarchingTetsEdgeTable is a fixed lookup table. I was assuming using static would let the compiler optimize this step of looking up the table. So it wouldn't happen even if I use static?

genesis/engine/coupler.py
Libero0809: Deleted. I accidentally mangled my two branches.

genesis/engine/coupler.py
Libero0809: rigid_g close to zero would lead to g being close to zero and affect following computation's stability. I think the check here is necessary.
Changed to `g = deformable_g * rigid_g / (deformable_g + rigid_g)` for performance.

genesis/engine/coupler.py
Libero0809: I will just use deformable_g then.

genesis/engine/materials/FEM/elastic.py
Libero0809: Done. Actually, I didn't know taichi has a polar_decomp function. You are right, using svd would save some computation.

genesis/engine/coupler.py
duburcqa: I don’t know, but it looks fishy. It is not obviously whether just setting the lookup table as static is enough. You could ask @hughperkins . But marking as static the result of extracting the item first should be fine from what I understand.

genesis/engine/coupler.py
```diff
@@ -934,15 +935,13 @@ def fem_floor_detection(self, f: ti.i32):
             )
             self.fem_floor_contact_pairs[i_c].barycentric = barycentric
 
-            C = ti.static(1.0e8)
-            deformable_g = C
+            deformable_g = ti.static(1.0e8)
```
duburcqa: Removing `C` does not change the fact that `1.0e8` should rather be a global constant defined on top of the file x)

genesis/engine/bvh.py
```diff
@@ -321,31 +330,53 @@ def delta(self, i, j, i_b):
...
+            parent_idx = self.nodes[i_b, i].parent
+            if parent_idx != -1:
+                self.internal_node_ready[i_b, parent_idx] = ti.u8(1)
+            self.internal_node_active[i_b, i] = ti.u8(0)
+            self.updated[None] = True
+
+        for i_b, i in ti.ndrange(self.n_batches, self.n_aabbs - 1):
+            if self.internal_node_ready[i_b, i] == ti.uint8(0):
+                continue
+            self.internal_node_active[i_b, i] = ti.u8(1)
```
duburcqa: `= 1`

genesis/engine/bvh.py
```diff
@@ -321,31 +330,53 @@ def delta(self, i, j, i_b):
...
+            self.nodes[i_b, i].bound.min = ti.min(left_bound.min, right_bound.min)
+            self.nodes[i_b, i].bound.max = ti.max(left_bound.max, right_bound.max)
+            parent_idx = self.nodes[i_b, i].parent
+            if parent_idx != -1:
+                self.internal_node_ready[i_b, parent_idx] = ti.u8(1)
+            self.internal_node_active[i_b, i] = ti.u8(0)
+            self.updated[None] = True
+
+        for i_b, i in ti.ndrange(self.n_batches, self.n_aabbs - 1):
+            if self.internal_node_ready[i_b, i] == ti.uint8(0):
```
duburcqa: `== 0`

genesis/engine/bvh.py
```diff
@@ -321,31 +330,53 @@ def delta(self, i, j, i_b):
...
-                left_bound = self.nodes[i_b, self.nodes[i_b, cur_idx].left].bound
-                right_bound = self.nodes[i_b, self.nodes[i_b, cur_idx].right].bound
-                self.nodes[i_b, cur_idx].bound.min = ti.min(left_bound.min, right_bound.min)
-                self.nodes[i_b, cur_idx].bound.max = ti.max(left_bound.max, right_bound.max)
-                cur_idx = self.nodes[i_b, cur_idx].parent
+    @ti.kernel
+    def _kernel_compute_bounds_one_layer(self):
+        self.updated[None] = False
+        for i_b, i in ti.ndrange(self.n_batches, self.n_aabbs - 1):
+            if self.internal_node_active[i_b, i] == ti.uint8(0):
```
duburcqa: `== 0`

genesis/engine/bvh.py
```diff
@@ -321,31 +330,53 @@ def delta(self, i, j, i_b):
...
+                continue
+            left_bound = self.nodes[i_b, self.nodes[i_b, i].left].bound
+            right_bound = self.nodes[i_b, self.nodes[i_b, i].right].bound
+            self.nodes[i_b, i].bound.min = ti.min(left_bound.min, right_bound.min)
+            self.nodes[i_b, i].bound.max = ti.max(left_bound.max, right_bound.max)
+            parent_idx = self.nodes[i_b, i].parent
+            if parent_idx != -1:
+                self.internal_node_ready[i_b, parent_idx] = ti.u8(1)
```
duburcqa: `= 1`

genesis/engine/bvh.py
```diff
@@ -157,19 +158,27 @@ class Node:
         # Nodes of the BVH, first n_aabbs - 1 are internal nodes, last n_aabbs are leaf nodes
         self.nodes = self.Node.field(shape=(self.n_batches, self.n_aabbs * 2 - 1))
         # Whether an internal node has been visited during traversal
-        self.internal_node_visited = ti.field(ti.u8, shape=(self.n_batches, self.n_aabbs - 1))
+        self.internal_node_active = ti.field(ti.u8, shape=(self.n_batches, self.n_aabbs - 1))
```
duburcqa: `ti.u1`

genesis/engine/bvh.py
```diff
@@ -157,19 +158,27 @@ class Node:
...
+        self.internal_node_ready = ti.field(ti.u8, shape=(self.n_batches, self.n_aabbs - 1))
```
duburcqa: `ti.u1`

genesis/engine/bvh.py
```diff
@@ -321,31 +330,53 @@ def delta(self, i, j, i_b):
...
+        while self.updated[None]:
+            self._kernel_compute_bounds_one_layer()
+
+    @ti.kernel
+    def _kernel_compute_bounds_init(self):
+        self.updated[None] = True
+
         for i_b, i in ti.ndrange(self.n_batches, self.n_aabbs - 1):
-            self.internal_node_visited[i_b, i] = ti.u8(0)
+            self.internal_node_active[i_b, i] = ti.u8(0)
```
duburcqa: `= 0` should be fine, the compiler will do the job, since the variable is already allocated with the right type.

duburcqa: By the way, you could just do:
```python
self.internal_node_active.fill(0)
self.internal_node_ready.fill(0)
```

genesis/engine/bvh.py
```diff
@@ -321,31 +330,53 @@ def delta(self, i, j, i_b):
...
+            self.internal_node_ready[i_b, i] = ti.u8(0)
 
         for i_b, i in ti.ndrange(self.n_batches, self.n_aabbs):
             idx = ti.i32(self.morton_codes[i_b, i])
             self.nodes[i_b, i + self.n_aabbs - 1].bound.min = self.aabbs[i_b, idx].min
             self.nodes[i_b, i + self.n_aabbs - 1].bound.max = self.aabbs[i_b, idx].max
+            parent_idx = self.nodes[i_b, i + self.n_aabbs - 1].parent
+            if parent_idx != -1:
+                self.internal_node_active[i_b, parent_idx] = ti.u8(1)
```
duburcqa: `= 1`

genesis/engine/bvh.py
```diff
@@ -321,31 +330,53 @@ def delta(self, i, j, i_b):
...
+            if self.internal_node_active[i_b, i] == ti.uint8(0):
+                continue
+            left_bound = self.nodes[i_b, self.nodes[i_b, i].left].bound
+            right_bound = self.nodes[i_b, self.nodes[i_b, i].right].bound
+            self.nodes[i_b, i].bound.min = ti.min(left_bound.min, right_bound.min)
+            self.nodes[i_b, i].bound.max = ti.max(left_bound.max, right_bound.max)
+            parent_idx = self.nodes[i_b, i].parent
+            if parent_idx != -1:
+                self.internal_node_ready[i_b, parent_idx] = ti.u8(1)
+            self.internal_node_active[i_b, i] = ti.u8(0)
```
duburcqa: `= 0`

genesis/engine/bvh.py
```diff
@@ -321,31 +330,53 @@ def delta(self, i, j, i_b):
...
+            self.updated[None] = True
+
+        for i_b, i in ti.ndrange(self.n_batches, self.n_aabbs - 1):
+            if self.internal_node_ready[i_b, i] == ti.uint8(0):
+                continue
+            self.internal_node_active[i_b, i] = ti.u8(1)
+            self.internal_node_ready[i_b, i] = ti.u8(0)
```
duburcqa: `= 0`

python-3.12.3-macos11.pkg
duburcqa, line 1: Remove this file

genesis/engine/coupler.py
Libero0809: Done

genesis/engine/bvh.py
Libero0809: Done

genesis/engine/bvh.py
Libero0809: Done

genesis/engine/bvh.py
Libero0809: Done

Libero0809:
👍 = Done. Saves me some typing in the future😊

hughperkins:
> 👍 = Done. Saves me some typing in the future😊

Note that we don't get notifications for emoji. So, once you have addressed everyting you're going to address, you might consider posting something like "addressed all comments". (I'm not sure if that is standard tbh, and I typically often end up DM'ing reviewers directly anyway, but I tend to do this anywya).

Libero0809:
@hughperkins , Noted. 👍

duburcqa (approved):

genesis/engine/coupler.py
```diff
@@ -934,15 +936,13 @@ def fem_floor_detection(self, f: ti.i32):
             )
             self.fem_floor_contact_pairs[i_c].barycentric = barycentric
 
-            C = ti.static(1.0e8)
-            deformable_g = C
+            deformable_g = ti.static(self.default_deformable_g)
```
duburcqa: I do not recommend using `ti.static` here, because changing the value would trigger recompilation, which is not undesirable. Moreover, the benefits for making it static is basically zero here.

genesis/engine/bvh.py
```diff
@@ -186,12 +195,7 @@ def build(self):
             for i in ti.static(range(3)):
                 self.scale[i_b][i] = ti.select(scale[i] > 1e-7, 1.0 / scale[i], 1)
```
duburcqa: I would recommend this instead: `self.scale[i_b][i] = ti.select(scale[i] > gs.EPS, 1.0 / scale[i], 1.0)`

genesis/engine/bvh.py
```diff
@@ -223,38 +227,43 @@ def expand_bits(self, v):
...
+                self.prefix_sum[i_b, j] = self.prefix_sum[i_b, j - 1] + self.hist[i_b, j - 1]
 
-            # Reorder morton codes
-            for i_b, i_a in ti.ndrange(self.n_batches, self.n_aabbs):
-                code = (self.morton_codes[i_b, i_a] >> (i * 8)) & 0xFF
-                idx = ti.i32(self.offset[i_b, i_a] + self.prefix_sum[i_b, ti.i32(code)])
-                self.tmp_morton_codes[i_b, idx] = self.morton_codes[i_b, i_a]
+        # Reorder morton codes
+        for i_b, i_a in ti.ndrange(self.n_batches, self.n_aabbs):
+            code = (self.morton_codes[i_b, i_a] >> (i * 8)) & 0xFF
```
duburcqa, line 258: You are being overzealous here. Just do `idx = self.offset[i_b, i_a] + self.prefix_sum[i_b, ti.i32(code)])`

duburcqa: Actually you are right, because it would be `u32` not `i32`.

genesis/engine/bvh.py
```diff
@@ -223,38 +227,43 @@ def expand_bits(self, v):
...
 
-            # Compute prefix sum
-            for i_b in ti.ndrange(self.n_batches):
-                self.prefix_sum[i_b, 0] = 0
-                for j in range(1, 256):  # sequential prefix sum
-                    self.prefix_sum[i_b, j] = self.prefix_sum[i_b, j - 1] + self.hist[i_b, j - 1]
+        # Compute prefix sum
+        for i_b in ti.ndrange(self.n_batches):
+            self.prefix_sum[i_b, 0] = 0
+            for j in range(1, 256):  # sequential prefix sum
```
erizmr, line 253: Though it has been merged, just to mention there is a built-in parallel prefix sum, which might be useful: https://github.com/taichi-dev/taichi/blob/master/python/taichi/algorithms/_algorithms.py#L42

duburcqa: Oh really! This is interesting!

Libero0809: Thanks for pointing this out. I think if we are doing prefix sum for just 256 elements, sequential would be fine.

PR #1308, opened 2025-06-23
[BUG FIX] Get around MacOS Metal GPU taichi compiler bug.

duburcqa:


genesis/engine/solvers/rigid/collider_decomp.py
```diff
@@ -891,7 +895,7 @@ def _func_narrow_phase_convex_specializations(self):
                 if self._solver.geoms_info[i_ga].type > self._solver.geoms_info[i_gb].type:
                     i_ga, i_gb = i_gb, i_ga
 
-                if ti.static(self._solver._enable_multi_contact):
+                if ti.static(self._solver._enable_multi_contact and sys.platform != "darwin"):
```
YilingQiao: Interesting. We cannot run _func_plane_box_multi_contact on Mac? (not blocking mergeing)

YilingQiao (dismissed):

genesis/engine/solvers/rigid/collider_decomp.py
duburcqa: `_func_plane_box_multi_contact` is still being used on MacOS! But it has to be part of the main convex-convex collision detection path (`_func_mpr`), as it was before improving the compilation time.

Genesis-Embodied-AI (approved):

Issue #1309, opened 2025-06-23
[Feature]: FEM Vertex Constraints


Milotrince:
### What feature or enhancement are you proposing?

The ability to apply a positional constraint on specific vertices on a soft object.

### Motivation

Related issues and discussion:
https://github.com/Genesis-Embodied-AI/Genesis/issues/1091
https://github.com/Genesis-Embodied-AI/Genesis/issues/199
https://github.com/Genesis-Embodied-AI/Genesis/discussions/738
https://github.com/Genesis-Embodied-AI/Genesis/discussions/271

### Potential Benefit

Can attach a soft object to a rigid one by adding vertex constraints to the soft object at the spots it would be attached to the rigid object. Opens up more options for different hybrid robots, like a soft end effector.

### What is the expected outcome of the implementation work?

- [ ] Handle constraints in `FEMSolver` (Dirichlet boundary condition)
  - Explicit FEM -> Can directly set/fix pos?
  - Implicit FEM -> Lagrangian
- [ ] A reasonably straightforward-to-use interface to add/update constraints
- [ ] Tests

PR #1310, opened 2025-06-23
[FEATURE] FEM Vertex Constraints

Milotrince:
## Description
Working on implementing vertex constraints in `FEMSolver`

## Related Issue
https://github.com/Genesis-Embodied-AI/Genesis/issues/1309

## Motivation and Context
See related issue

WIP

## Checklist:


Milotrince:
Hard constraint on explicit FEM:

https://github.com/user-attachments/assets/049875e6-6b07-4513-b3bf-985fcd752fb4

genesis/engine/solvers/fem_solver.py
```diff
@@ -988,6 +1013,18 @@ def get_state_render(self, f):
...
+        """
+        Get forces on all vertices.
+        
+        Returns:
+            np.ndarray:  (B, n_vertices, 3) where B is batch size
+        """
+        if not self.is_active():
+            return None
+
+        return self.elements_v_energy.force.to_numpy()
```
duburcqa: use `ti_field_to_torch`. Numpy arrays are not used for user-facing public API.

genesis/engine/solvers/fem_solver.py
```diff
@@ -1272,3 +1309,120 @@ def n_elements_max(self):
...
+        ----------
+            vertex_indices : array_like
+                List of vertex indices to constrain
+            target_positions : array_like, shape (len(vertex_indices), 3)
+                List of target positions [x, y, z] for each vertex
+            constraint_type: str
+                A "hard" constraint directly set position and zero velocity.
+                A "soft" constraint uses a spring force to pull the vertex towards the target position.
+            stiffness : None | float, optional
+                Fixed position by default. Specify a spring stiffness for a soft constraint. (TODO: what's a reasonable range?)
```
duburcqa: 120 chars line width limit.

genesis/engine/solvers/fem_solver.py
```diff
@@ -1272,3 +1309,120 @@ def n_elements_max(self):
...
+        """
+        if self._use_implicit_solver:
+            gs.logger.warning("Vertex constraints are not supported when using implicit solver. Constraint addition ignored.")
```
duburcqa: same.

genesis/engine/solvers/fem_solver.py
```diff
@@ -1272,3 +1309,120 @@ def n_elements_max(self):
...
+            return
+ 
+        constraint_type_int = 0 if constraint_type == "hard" else 1
+        
+        vertex_indices_np = np.array(vertex_indices, dtype=np.int32)
```
duburcqa: Use torch + device.

genesis/engine/solvers/fem_solver.py
```diff
@@ -1272,3 +1309,120 @@ def n_elements_max(self):
...
+        target_positions_np = np.array(target_positions, dtype=np.float32)
```
duburcqa: same.

genesis/engine/solvers/fem_solver.py
duburcqa: `torch.as_tensor`

genesis/engine/solvers/fem_solver.py
```diff
@@ -1272,3 +1309,120 @@ def n_elements_max(self):
...
+        self._kernel_set_vertex_constraints(vertex_indices_np, target_positions_np, constraint_type_int, stiffness)
+
+    def update_constraint_targets(self, vertex_indices, new_target_positions):
+        """Update target positions for existing constraints."""
+        if not isinstance(vertex_indices, (list, tuple)):
+            vertex_indices = [vertex_indices]
+        if not isinstance(new_target_positions[0], (list, tuple)):
+            new_target_positions = [new_target_positions]
+            
+        vertex_indices_np = np.array(vertex_indices, dtype=np.int32)
```
duburcqa: same.

genesis/engine/solvers/fem_solver.py
```diff
@@ -1272,3 +1309,120 @@ def n_elements_max(self):
...
+        constraint_type_int = 0 if constraint_type == "hard" else 1
+        
+        vertex_indices_np = np.array(vertex_indices, dtype=np.int32)
+        target_positions_np = np.array(target_positions, dtype=np.float32)
+        
+        self._kernel_set_vertex_constraints(vertex_indices_np, target_positions_np, constraint_type_int, stiffness)
+
+    def update_constraint_targets(self, vertex_indices, new_target_positions):
+        """Update target positions for existing constraints."""
+        if not isinstance(vertex_indices, (list, tuple)):
```
duburcqa: Must accept many other types for consistency with Rigid Solver, i.e. slice, range, torch.Tensor, np.ndarray...etc

Take inspiration from `_sanitize_1D_io_variables`

genesis/engine/solvers/fem_solver.py
```diff
@@ -1272,3 +1309,114 @@ def n_elements_max(self):
     @property
     def vol_scale(self):
         return self._vol_scale
+   
+
+    # ------------------------------------------------------------------------------------
+    # -------------------------------- vertex constraints --------------------------------
+    # ------------------------------------------------------------------------------------
+
+    def set_vertex_constraints(self, vertex_indices, target_positions, constraint_type="hard", stiffness=None):
```
duburcqa: `stiffness=0.0` (and update doc)

genesis/engine/solvers/fem_solver.py
```diff
@@ -1272,3 +1309,114 @@ def n_elements_max(self):
...
+        vertex_indices: ti.types.ndarray(),
+        target_positions: ti.types.ndarray(),
+        constraint_type: ti.u1,
+        stiffness: ti.f32,
+    ):
+        one: ti.u1 = 1
+        for i in range(vertex_indices.shape[0]):
+            v_idx = vertex_indices[i]
+            if 0 <= v_idx < self.n_vertices:
+                self.vertex_constraints[v_idx].is_constrained = one
```
duburcqa: Just do `self.vertex_constraints[v_idx].is_constrained = 1`. It should work just fine.

genesis/engine/solvers/fem_solver.py
```diff
@@ -1272,3 +1309,114 @@ def n_elements_max(self):
...
+        self,
+        vertex_indices: ti.types.ndarray(),
+        target_positions: ti.types.ndarray(),
+        constraint_type: ti.u1,
+        stiffness: ti.f32,
+    ):
+        one: ti.u1 = 1
+        for i in range(vertex_indices.shape[0]):
+            v_idx = vertex_indices[i]
+            if 0 <= v_idx < self.n_vertices:
```
duburcqa: Move this check in Python scope and raise an exception if invalid.

genesis/engine/solvers/fem_solver.py
```diff
@@ -1272,3 +1309,114 @@ def n_elements_max(self):
...
+
+    @ti.kernel
+    def _kernel_update_constraint_targets(
+        self,
+        vertex_indices: ti.types.ndarray(),
+        new_target_positions: ti.types.ndarray(),
+    ):
+        for i in range(vertex_indices.shape[0]):
+            v_idx = vertex_indices[i]
+            if 0 <= v_idx < self.n_vertices and self.vertex_constraints[v_idx].is_constrained:
```
duburcqa: Same, move `0 <= v_idx < self.n_vertices` in python scope.

genesis/engine/solvers/fem_solver.py
```diff
@@ -1272,3 +1309,114 @@ def n_elements_max(self):
...
+                for j in ti.static(range(3)):
+                    self.vertex_constraints[v_idx].target_pos[j] = new_target_positions[i, j]
+
+    @ti.kernel
+    def _kernel_remove_specific_constraints(self, vertex_indices: ti.types.ndarray()):
+        for i in range(vertex_indices.shape[0]):
+            v_idx = vertex_indices[i]
+            if 0 <= v_idx < self.n_vertices:
```
duburcqa: same.

genesis/engine/solvers/fem_solver.py
```diff
@@ -1272,3 +1309,114 @@ def n_elements_max(self):
...
+                self.vertex_constraints[v_idx].is_constrained = 0
+
+    @ti.kernel
+    def _kernel_remove_all_constraints(self):
+        for i_v in range(self.n_vertices):
```
duburcqa: Just do : `self.vertex_constraints[i_v].is_constrained.fill(0)`

duburcqa:
Why is it a draft?

Milotrince:
> Why is it a draft?

Hi, thanks for the thorough review! I updated the description, there are several other things I want to add together in this PR before it gets merged in. Since I'm new to contributing, your comments have helped a lot; I'll keep in mind these things as I implement the others

genesis/engine/solvers/fem_solver.py
Milotrince: Yeah, I was a bit annoyed at this warning though when using `= 1`
```
[W 06/25/25 13:58:57.965 2164702] TaichiWarning
_kernel_set_vertex_constraints:
                self.vertex_constraints[v_idx].is_constrained = 1
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Assign may lose precision: u1 <- i32
```
Replaced with `ti.cast(1, ti.u1)`

genesis/engine/solvers/fem_solver.py
```diff
@@ -1405,18 +1408,18 @@ def _kernel_update_constraint_targets(
...
-        for i_v in range(self.n_vertices):
-            self.vertex_constraints[i_v].is_constrained = 0
\ No newline at end of file
+            self.vertex_constraints[v_idx].is_constrained = 0
+    
+    def _assert_vertex_idx_in_bounds(self, vertex_indices):
+        """
+        Assert that all vertex indices are within the valid range.
+        """
+        out_of_bounds = torch.where(vertex_indices >= self.n_vertices or vertex_indices < 0)[0].numel() > 0
```
duburcqa: `(vertex_indices >= self.n_vertices or vertex_indices < 0).any()`

Milotrince:
Updated main comment with examples

genesis/engine/entities/fem_entity.py
```diff
@@ -371,7 +372,7 @@ def _add_to_solver(self, in_backward=False):
         elems_np = self.elems.astype(gs.np_int)
         verts_numpy = self.init_positions.cpu().numpy().astype(gs.np_float)
 
-        self._solver._kernel_add_elements(
+        self.solver._kernel_add_elements(
```
Milotrince: entity uses both `self.solver` and `self._solver`; changed to `self.solver` for consistency

Milotrince:
should I make a doc page for this? it might make more sense to add it once we have a tactile demo

genesis/engine/entities/fem_entity.py
```diff
@@ -750,6 +751,113 @@ def set_muscle_direction(self, muscle_direction):
             muscle_direction=muscle_direction,
         )
 
+    def set_vertex_constraints(
+        self, vertex_indices, target_positions=None, link=None, constraint_type="hard", stiffness=0.0, damping=0.0
```
duburcqa: Would make more sense to use a boolean here, `is_hard_constraint=True`

genesis/engine/entities/fem_entity.py
```diff
@@ -750,6 +751,113 @@ def set_muscle_direction(self, muscle_direction):
...
+        if target_positions is None:
+            target_positions = self.init_positions[vertex_indices]
+        target_positions = torch.as_tensor(target_positions, dtype=gs.tc_float, device=gs.device)
+
+        if link is None:
+            link_idx = -1
+            link_init_pos = np.zeros((3,))
+            link_init_quat = np.zeros((4,))
+        else:
+            assert hasattr(link, "get_pos") and hasattr(
```
duburcqa: This is not a good approach. You should rather use `isinstance(link, [...])`. And since you only support `RigidLink` anyway, I would just do `assert isinstance(link, RigidLink)`.

genesis/engine/entities/fem_entity.py
```diff
@@ -750,6 +751,113 @@ def set_muscle_direction(self, muscle_direction):
...
+                link, "get_quat"
+            ), "Link must have get_pos() and get_quat() methods, e.g. RigidLink"
+            link_init_pos = link.get_pos().detach().cpu().numpy()
```
duburcqa: `link_init_pos = tensor_to_array(link.get_pos())`

Could you extend `tensor_to_array` to also take care of `.detach()` ? I don't think it does this already.

genesis/engine/entities/fem_entity.py
```diff
@@ -750,6 +751,113 @@ def set_muscle_direction(self, muscle_direction):
...
+            link_init_quat = link.get_quat().detach().cpu().numpy()
+
+            def link_func():
+                pos = link.get_pos()
+                quat = link.get_quat()
+                return pos, quat
+
+            link_idx = self.solver._add_link_func(link_func)
```
duburcqa: This looks like some kind of sorcery. This should be avoided because it would mess with the lifetime of  `link`.

I would rather recommend passing `entity_idx` and `link_idx` as input argument, and store these integers (after making sure they are valid of course) instead of object references.

genesis/engine/entities/fem_entity.py
```diff
@@ -750,6 +751,113 @@ def set_muscle_direction(self, muscle_direction):
...
+            constraint_type_int,
+            stiffness,
+            damping,
+            link_idx,
+            link_init_pos,
+            link_init_quat,
+        )
+
+    def _sanitize_input_to_tensor(self, arr, dtype):
+        if np.isscalar(arr):
```
duburcqa: What about removing this and replace `return arr` by `return torch.atleast_1d(arr)` ? It should be slightly more efficient.

genesis/engine/entities/fem_entity.py
```diff
@@ -750,6 +751,113 @@ def set_muscle_direction(self, muscle_direction):
...
+        self.solver._kernel_update_constraint_targets(vertex_indices, target_positions)
+
+    def remove_vertex_constraints(self, vertex_indices=None):
+        """Remove constraints from specified vertices, or all if None."""
+        if not self.solver._constraints_initialized:
+            gs.logger.warning("Ignoring remove_vertex_constraints; constraints have not been initialized.")
+            return
+
+        if vertex_indices is None:
+            vertex_indices = self._v_start + np.arange(self.n_vertices, dtype=gs.np_int)
```
duburcqa: `torch.arange(self.n_vertices, dtype=gs.np_int, device=gs.device)`

genesis/engine/entities/fem_entity.py
duburcqa: Just keep using `self._solver` everywhere if possible.

genesis/engine/solvers/fem_solver.py
```diff
@@ -853,6 +886,7 @@ def linesearch(self, f: ti.i32):
 
     def batch_solve(self, f: ti.i32):
         self.batch_active.fill(True)
+        print("Batch solve for frame", f)
```
duburcqa: No prints except in examples (eventually). You may use `gs.logger.debug` if you think it is worth it but I'm not convinced in this case.

genesis/engine/solvers/fem_solver.py
```diff
@@ -1034,6 +1073,18 @@ def get_state_render(self, f):
...
 
+    def get_forces(self):
+        """
+        Get forces on all vertices.
+
+        Returns:
+            torch.Tensor : shape (B, n_vertices, 3) where B is batch size
+        """
+        if not self.is_active():
+            return None
```
duburcqa, line 1160: Not sure if returning none is a good approach. But let's keep it like this for now.

genesis/engine/solvers/fem_solver.py
```diff
@@ -1317,3 +1368,110 @@ def n_elements_max(self):
...
+        poss: ti.types.ndarray(),
+        quats: ti.types.ndarray(),
+    ):
+        for i_v in range(self.n_vertices):
+            if self.vertex_constraints[i_v].is_constrained and self.vertex_constraints[i_v].link_idx >= 0:
+                i_l = self.vertex_constraints[i_v].link_idx
+                pos = gs.ti_vec3([poss[i_l, 0], poss[i_l, 1], poss[i_l, 2]])
+                quat = gs.ti_vec4([quats[i_l, 0], quats[i_l, 1], quats[i_l, 2], quats[i_l, 3]])
+                offset_pos = self.vertex_constraints[i_v].link_offset_pos
+                offset_quat = ti_quat_mul(quat, ti_inv_quat(self.vertex_constraints[i_v].link_init_quat))
```
duburcqa: Prefer `ti_transform_quat_by_quat`

genesis/engine/solvers/fem_solver.py
```diff
@@ -1317,3 +1368,110 @@ def n_elements_max(self):
...
+                self.vertex_constraints[i_v].target_pos = pos + ti_transform_by_quat(offset_pos, offset_quat)
```
duburcqa: This is equivalent to `ti_transform_by_trans_quat(offset_pos, pos, offset_quat)`, but this looks kinda weird.

genesis/engine/solvers/fem_solver.py
```diff
@@ -1317,3 +1368,110 @@ def n_elements_max(self):
...
+
+    @ti.kernel
+    def apply_hard_constraints(self, f: ti.i32):
+        """Apply hard constraints by directly overriding positions and velocities."""
+        for i_v, i_b in ti.ndrange(self.n_vertices, self._B):
+            if self.vertex_constraints[i_v].is_constrained and self.vertex_constraints[i_v].constraint_type == 0:
+                self.elements_v[f + 1, i_v, i_b].pos = self.vertex_constraints[i_v].target_pos
+                self.elements_v[f + 1, i_v, i_b].vel = ti.Vector.zero(gs.ti_float, 3)
```
duburcqa: `self.elements_v[f + 1, i_v, i_b].vel.fill(0.0)` would not work?

genesis/engine/solvers/fem_solver.py
```diff
@@ -1317,3 +1368,110 @@ def n_elements_max(self):
...
+            if self.vertex_constraints[i_v].is_constrained and self.vertex_constraints[i_v].constraint_type == 1:
+                pos_error = self.elements_v[f, i_v, i_b].pos - self.vertex_constraints[i_v].target_pos
+                vel_error = self.elements_v[f + 1, i_v, i_b].vel - self.elements_v[f, i_v, i_b].vel
+                spring_force = -self.vertex_constraints[i_v].stiffness * pos_error
+                damping_force = -self.vertex_constraints[i_v].damping * vel_error
+
+                dt = self.substep_dt
+                mass = self.elements_v_info[i_v].mass
+                if mass > gs.EPS:
+                    dv = dt * (spring_force + damping_force) / mass
```
duburcqa: What about a "kinematic spring"? i.e. `dv = dt * (spring_force + damping_force)`

genesis/engine/solvers/fem_solver.py
```diff
@@ -1317,3 +1368,110 @@ def n_elements_max(self):
...
+        constraint_type: ti.u1,
+        stiffness: ti.f32,
+        damping: ti.f32,
+        link_idx: ti.i32,
+        link_init_pos: ti.types.vector(3, ti.f32),
+        link_init_quat: ti.types.vector(4, ti.f32),
+    ):
+        for i in range(vertex_indices.shape[0]):
+            v_idx = vertex_indices[i]
+            self.vertex_constraints[v_idx].is_constrained = ti.cast(1, ti.u1)
```
duburcqa: Casting is not useful here. Just do `= 1`

tests/test_fem.py
```diff
@@ -343,3 +344,119 @@ def test_linear_corotated_sphere_fall_implicit_fem_sap_coupler(fem_material_line
...
+
+    # Set initial hard constraints
+    box.set_vertex_constraints(
+        vertex_indices=vertex_indices, target_positions=initial_target_positions, constraint_type="hard"
+    )
+
+    for _ in range(100):
+        scene.step()
+
+    positions = box.solver.get_state(0).pos.cpu()
```
duburcqa: `.cpu()` is useless here. Same for `initial_target_positions.cpu()`

tests/test_fem.py
```diff
@@ -343,3 +344,119 @@ def test_linear_corotated_sphere_fall_implicit_fem_sap_coupler(fem_material_line
...
+
+    new_target_positions = initial_target_positions + gs.tensor(
+        [[0.1, 0.1, 0.1], [0.1, 0.1, 0.1]],
+    )
+    box.update_constraint_targets(vertex_indices=vertex_indices, target_positions=new_target_positions)
+
+    for _ in range(10):
+        scene.step()
+
+    positions_after_update = box.solver.get_state(0).pos.cpu()
```
duburcqa: Same. No need for `.cpu()`

tests/test_fem.py
```diff
@@ -343,3 +344,119 @@ def test_linear_corotated_sphere_fall_implicit_fem_sap_coupler(fem_material_line
...
+    assert_allclose(
+        positions_after_update[0, vertex_indices, :], new_target_positions.cpu(), atol=1e-5, rtol=1e-5
```
duburcqa: Replace `atol=1e-5, rtol=1e-5` by `tol=1e-5`

tests/test_fem.py
```diff
@@ -343,3 +344,119 @@ def test_linear_corotated_sphere_fall_implicit_fem_sap_coupler(fem_material_line
...
+    box.set_vertex_constraints(
+        vertex_indices=vertex_indices, target_positions=initial_target_positions, constraint_type="hard"
+    )
+
+    for _ in range(100):
+        scene.step()
+
+    positions = box.solver.get_state(0).pos.cpu()
+    assert_allclose(
+        positions[0, vertex_indices, :], initial_target_positions.cpu(), atol=1e-5, rtol=1e-5
```
duburcqa: Replace `atol=1e-5, rtol=1e-5` by `tol=1e-5`

duburcqa: Why the accuracy is not better than this? `1e+5` is not great. Maybe you should simulate a bit longer?

tests/test_fem.py
```diff
@@ -343,3 +344,119 @@ def test_linear_corotated_sphere_fall_implicit_fem_sap_coupler(fem_material_line
...
+
+    box.remove_vertex_constraints(vertex_indices)
+
+    for _ in range(10):
+        scene.step()
+
+    positions_after_removal = box.solver.get_state(0).pos.cpu()
+
+    distance_moved = np.linalg.norm(positions_after_removal[0, vertex_indices, :] - new_target_positions.cpu())
+    assert np.any(distance_moved > 1e-3), "Vertices should have moved after removing constraints"
```
duburcqa: I would recommend this pattern:
```python
    with np.testing.assert_raises(AssertionError):
        assert_allclose([...], [...], atol=1e-3)
```

tests/test_fem.py
```diff
@@ -343,3 +344,119 @@ def test_linear_corotated_sphere_fall_implicit_fem_sap_coupler(fem_material_line
...
+        stiffness=1.0e3,
+        damping=1.0e1,
+    )
+
+    for _ in range(100):
+        scene.step()
+
+    positions = box.solver.get_state(0).pos.cpu()
+
+    assert_allclose(
```
duburcqa, line 516: Not possible to get better accuracy? Like increasing `stiffness` / `damping` and reducing simulation `dt` ?

genesis/engine/entities/fem_entity.py
duburcqa: What about forcing critical damping ? So that there is only one parameter to set instead of 2?

genesis/engine/solvers/fem_solver.py
Milotrince: oops yeah left during debug

genesis/engine/solvers/fem_solver.py
Milotrince: hrm yeah, I think the current is more clear with the intention

genesis/engine/solvers/fem_solver.py
Milotrince:
```
W 07/07/25 17:45:12.656 4112073] TaichiWarning
While compiling `_kernel_set_vertex_constraints_c592_0`, File "/home/trinityc/genesis/Genesis/genesis/engine/solvers/fem_solver.py", line 1446, in _kernel_set_vertex_constraints:
            self.vertex_constraints[v_idx].is_constrained = 1
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Assign may lose precision: u1 <- i32
```
It's because of this warning :[ can I keep it?

tests/test_fem.py
Milotrince: Yep the accuracy is exact, updated tol=0

tests/test_fem.py
Milotrince: it's due to gravity force, should it specifically try to counteract gravity?

genesis/engine/entities/fem_entity.py
Milotrince: Yeah.. I've always wanted to be a code wizard but not like this 😅
Is it possible to access `RigidSolver` from `FEMSolver` without storing another reference?

genesis/engine/solvers/fem_solver.py
duburcqa: Just use boolean: `self.vertex_constraints[v_idx].is_constrained = True`

Milotrince: oh. thanks!

genesis/engine/entities/fem_entity.py
Milotrince: ..or should this be handled in the coupler?

duburcqa: It sounds like a responsibility of the coupler indeed. What do you think @YilingQiao ?

tests/test_fem.py
duburcqa: Yes I get that, but if you increase the stiffness you should be able to increase the tolerance ? The other option would be to initialise the simulation with non-zero velocity but zero-gravity and check the final state. May be even better?

genesis/engine/coupler.py
```diff
@@ -345,6 +345,16 @@ def mpm_surface_to_particle(self, f: ti.i32):
                         if sdf_normal.dot(self.mpm_rigid_normal[i_p, i_g, i_b]) >= 0:
                             self.mpm_rigid_normal[i_p, i_g, i_b] = sdf_normal
 
+    def fem_rigid_link_constraints(self):
+        links = self.fem_solver._constraint_links
+        if len(links) > 0:
```
duburcqa: Could you write a kernel to do all this processing ? Because calling `get_links_pos` / `get_links_quat` repeatedly in Python-scope is very inefficient.

tests/test_fem.py
Milotrince: Ok, got it to `tol=1e-5` with smaller dt, more substeps, higher stiffness, velocity instead of gravity

genesis/engine/coupler.py
Milotrince: thanks, fixed! By writing it as a kernel it also cleaned things up a lot too

genesis/engine/solvers/fem_solver.py
```diff
@@ -1377,30 +1374,22 @@ def remove_all_vertex_constraints(self):
...
-        quats: ti.types.ndarray(),
+        links_pos: ti.template(),  # matrix field
+        links_quat: ti.template(), # matrix field
     ):
         for i_v in range(self.n_vertices):
             if self.vertex_constraints[i_v].is_constrained and self.vertex_constraints[i_v].link_idx >= 0:
                 i_l = self.vertex_constraints[i_v].link_idx
-                pos = gs.ti_vec3([poss[i_l, 0], poss[i_l, 1], poss[i_l, 2]])
-                quat = gs.ti_vec4([quats[i_l, 0], quats[i_l, 1], quats[i_l, 2], quats[i_l, 3]])
+                pos = links_pos[0, i_l]
```
duburcqa: You should loop over batch dimension.

tests/test_fem.py
duburcqa: This is great :)

tests/test_fem.py
```diff
@@ -343,3 +344,138 @@ def test_linear_corotated_sphere_fall_implicit_fem_sap_coupler(fem_material_line
...
+            new_target_positions,
+            tol=1e-3
+        ), "Vertices should have moved after removing constraints"
+
+def test_box_soft_vertex_constraint(show_viewer):
+    """Test if a box with strong soft vertex constraints has those vertices near."""
+    scene = gs.Scene(
+        sim_options=gs.options.SimOptions(
+            dt=1e-4,
+            substeps=10,
```
duburcqa: Are you sure this many substeps are really necessary? I would rather use `substeps=1` and `dt=1e-5` if this is really necessary.

tests/test_fem.py
```diff
@@ -343,3 +344,138 @@ def test_linear_corotated_sphere_fall_implicit_fem_sap_coupler(fem_material_line
...
+
+    box.set_vertex_constraints(
+        vertex_indices=vertex_indices,
+        target_positions=target_positions,
+        is_soft_constraint=True,
+        stiffness=1.e9
+    )
+    box.set_velocity(gs.tensor([0.0, 1.0, 0.0]))
+
+    for _ in range(1000):
```
duburcqa: You are simulating for 0.1 second. It sounds very short. What about 0.5s? But then you should probably increase a bit the simulation timestep and reduce a bit the stiffness if necessary, otherwise it may take too much time to simulated. Your time budget is about 300s.

genesis/engine/solvers/fem_solver.py
duburcqa: What do you think @YilingQiao ?

genesis/engine/entities/fem_entity.py
```diff
@@ -744,12 +745,134 @@ def set_muscle_direction(self, muscle_direction):
...
+            self._kernel_get_verts_pos(self._sim.cur_substep_local, target_poss, verts_idx)
+        target_poss = self._sanitize_input_to_batch_tensor(target_poss, gs.tc_float, dims=2)
+
+        assert len(envs_idx) == target_poss.shape[0] == verts_idx.shape[0], \
+            "First dimension should match number of environments."
+        assert target_poss.shape[1] == verts_idx.shape[1], "Target position should be provided for each vertex."
+
+        if link is None:
+            B = self._sim._B
+            link_init_pos = gs.zeros((B, 3))
```
duburcqa: `torch.zeros((self._sim._B, 3), dtype=gs.tc_float, device=gs.device)`

genesis/engine/entities/fem_entity.py
```diff
@@ -744,12 +745,134 @@ def set_muscle_direction(self, muscle_direction):
...
+            link_init_quat = gs.zeros((B, 4))
```
duburcqa: same.

genesis/engine/entities/fem_entity.py
```diff
@@ -744,12 +745,134 @@ def set_muscle_direction(self, muscle_direction):
...
+            link_idx = -1
+        else:
+            assert isinstance(link, RigidLink), "Only RigidLink is supported for vertex constraints."
+            link_init_pos = self._sanitize_input_to_batch_tensor(link.get_pos(), dtype=gs.tc_float)
+            link_init_quat = self._sanitize_input_to_batch_tensor(link.get_quat(), dtype=gs.tc_float)
+            link_idx = link._idx
```
duburcqa: `link_idx = link.idx`

genesis/engine/entities/fem_entity.py
```diff
@@ -744,12 +745,134 @@ def set_muscle_direction(self, muscle_direction):
...
+    def _kernel_get_verts_pos(self, f: ti.i32, pos: ti.types.ndarray(), verts_idx: ti.types.ndarray()):
+        # get current position of vertices
+        n = verts_idx.shape[0]
+        B = verts_idx.shape[1]
+        for i_v, i_b in ti.ndrange(n, B):
+            i_global = verts_idx[i_v, i_b] + self.v_start
+            for j in ti.static(range(3)):
+                pos[i_b, i_v, j] = self._solver.elements_v[f, i_global, i_b].pos[j]
+
+    def _sanitize_input_to_batch_tensor(self, arr, dtype, dims=1):
```
duburcqa: I think the name `_sanitize_input_tensor` would be better.

You should probably move this method right above `def set_vertex_constraints([...]):`

genesis/engine/entities/fem_entity.py
```diff
@@ -10,7 +10,8 @@
 import genesis.utils.mesh as mu
 from genesis.engine.states.cache import QueriedStates
 from genesis.engine.states.entities import FEMEntityState
-from genesis.utils.misc import to_gs_tensor
+from genesis.utils.misc import to_gs_tensor, tensor_to_array
+from genesis.engine.entities.rigid_entity.rigid_link import RigidLink
```
duburcqa: Fix order of imports and simplify RigidLink import:
```python
from genesis.engine.entities.rigid_entity import RigidLink
from genesis.utils.misc import to_gs_tensor, tensor_to_array
```

genesis/engine/entities/fem_entity.py
```diff
@@ -744,12 +745,134 @@ def set_muscle_direction(self, muscle_direction):
...
+        if self._solver._use_implicit_solver:
+            gs.logger.warning("Ignoring vertex constraint; unsupported with FEM implicit solver.")
+            return
+
+        if not self._solver._constraints_initialized:
+            self._solver.init_constraints()
+
+        envs_idx = self._scene._sanitize_envs_idx(envs_idx)
+        verts_idx = self._sanitize_input_to_batch_tensor(verts_idx, dtype=gs.tc_int) + self._v_start
+        self._assert_verts_idx_in_bounds(verts_idx)
```
duburcqa: No need for a dedicated method. It is only one line of code.

genesis/engine/entities/fem_entity.py
```diff
@@ -744,12 +745,134 @@ def set_muscle_direction(self, muscle_direction):
...
+
+        if target_poss is None:
+            target_poss = gs.zeros((verts_idx.shape[0], verts_idx.shape[1], 3), dtype=float, requires_grad=False)
+            self._kernel_get_verts_pos(self._sim.cur_substep_local, target_poss, verts_idx)
+        target_poss = self._sanitize_input_to_batch_tensor(target_poss, gs.tc_float, dims=2)
+
+        assert len(envs_idx) == target_poss.shape[0] == verts_idx.shape[0], \
```
duburcqa: `len(envs_idx) == len(target_poss) == len(verts_idx)`

duburcqa: You should probably take inspiration from `RigidSolver._sanitize_1D_io_variables` and `RigidSolver._sanitize_2D_io_variables`. All the validity checks should be moved alongside casting to torch in your `_sanitize_[...]` method.

Ideally it should be the same for `self._assert_verts_idx_in_bounds`

genesis/engine/entities/fem_entity.py
```diff
@@ -744,12 +745,134 @@ def set_muscle_direction(self, muscle_direction):
...
+        else:
+            verts_idx = self._sanitize_input_to_batch_tensor(verts_idx, dtype=gs.tc_int) + self._v_start
+            self._assert_verts_idx_in_bounds(verts_idx)
+            envs_idx = self._scene._sanitize_envs_idx(envs_idx)
+            self._solver._kernel_remove_specific_constraints(verts_idx, envs_idx)
+
+    @ti.kernel
+    def _kernel_get_verts_pos(self, f: ti.i32, pos: ti.types.ndarray(), verts_idx: ti.types.ndarray()):
+        # get current position of vertices
+        n = verts_idx.shape[0]
```
duburcqa: `n, B = verts_idx.shape`

genesis/engine/entities/fem_entity.py
```diff
@@ -744,12 +745,134 @@ def set_muscle_direction(self, muscle_direction):
...
+        B = verts_idx.shape[1]
+        for i_v, i_b in ti.ndrange(n, B):
```
duburcqa: Just do `ti.ndrange(verts_idx.shape)` directly.

genesis/engine/entities/fem_entity.py
duburcqa: And `n, B` are not good variable names. It should be `n_vertices, n_envs`

genesis/engine/entities/fem_entity.py
```diff
@@ -744,12 +745,134 @@ def set_muscle_direction(self, muscle_direction):
...
+            for j in ti.static(range(3)):
+                pos[i_b, i_v, j] = self._solver.elements_v[f, i_global, i_b].pos[j]
+
+    def _sanitize_input_to_batch_tensor(self, arr, dtype, dims=1):
+        if not isinstance(arr, torch.Tensor):
+            arr = torch.as_tensor(arr, dtype=dtype, device=gs.device)
+        tensor = torch.atleast_1d(arr)
+        if tensor.ndim < dims+1:
+            # add batch dimension if missing
+            tensor = tensor.unsqueeze(0).repeat((self._sim._B, *([1] * dims)))
```
duburcqa: This implementation is weird. What about:
```python
tensor.expand((self._sim._B, *tensor.shape))
```

genesis/engine/entities/fem_entity.py
```diff
@@ -744,12 +745,134 @@ def set_muscle_direction(self, muscle_direction):
...
+        for i_v, i_b in ti.ndrange(n, B):
+            i_global = verts_idx[i_v, i_b] + self.v_start
+            for j in ti.static(range(3)):
+                pos[i_b, i_v, j] = self._solver.elements_v[f, i_global, i_b].pos[j]
+
+    def _sanitize_input_to_batch_tensor(self, arr, dtype, dims=1):
+        if not isinstance(arr, torch.Tensor):
+            arr = torch.as_tensor(arr, dtype=dtype, device=gs.device)
+        tensor = torch.atleast_1d(arr)
+        if tensor.ndim < dims+1:
```
duburcqa: You should rather check before that `tensor.ndim in (dims, dim + 1)`, then do:
```
python
if tensor.ndim == dims:
    [...]
```

genesis/engine/entities/fem_entity.py
duburcqa: `dim` is not clear. What about `dim_unbatched` or something like this?

genesis/engine/entities/fem_entity.py
```diff
@@ -744,12 +745,134 @@ def set_muscle_direction(self, muscle_direction):
...
+            tensor = tensor.unsqueeze(0).repeat((self._sim._B, *([1] * dims)))
+            # tensor = tensor.unsqueeze(0).repeat(self._sim._B, 1)
+        assert tensor.shape[0] == self._sim._B, "Input tensor batch size must match the number of environments."
+        return tensor
+
+    def _assert_verts_idx_in_bounds(self, verts_idx):
+        """
+        Assert that all vertex indices are within the valid range.
+        """
+        out_of_bounds = torch.any(torch.logical_or(verts_idx >= self._solver.n_vertices, verts_idx < 0))
```
duburcqa: `out_of_bounds = not (0 <= verts_idx & verts_idx < self._solver.n_vertices).all()`

genesis/engine/entities/fem_entity.py
```diff
@@ -744,12 +745,134 @@ def set_muscle_direction(self, muscle_direction):
...
+            arr = torch.as_tensor(arr, dtype=dtype, device=gs.device)
+        tensor = torch.atleast_1d(arr)
+        if tensor.ndim < dims+1:
+            # add batch dimension if missing
+            tensor = tensor.unsqueeze(0).repeat((self._sim._B, *([1] * dims)))
+            # tensor = tensor.unsqueeze(0).repeat(self._sim._B, 1)
+        assert tensor.shape[0] == self._sim._B, "Input tensor batch size must match the number of environments."
+        return tensor
+
+    def _assert_verts_idx_in_bounds(self, verts_idx):
```
duburcqa: This function should be removed and moved in `_sanitize_[...]`

tests/test_fem.py
Milotrince: ([d9a0759](https://github.com/Genesis-Embodied-AI/Genesis/pull/1310/commits/d9a07594afe4af72c72e5c6e1a3320d9e7cc95f0)) with `dt=1e-3`, `steps=500` (0.5s), `box.set_velocity(0.01, 0.01, 0.01)`, `stiffness=2.0e5`, `tol=5e-5`, takes about 5 seconds total on my mac with `device=gs.cpu` (and yes the test fails without setting the constraint)

genesis/engine/coupler.py
duburcqa, line 1: `coupler.py` has been renamed and moved in `./couplers/legacy_coupler.py`.

genesis/engine/solvers/fem_solver.py
YilingQiao: I think it's okay for now. Or what do you think would be a better handling? Raise an exception?

genesis/engine/entities/fem_entity.py
YilingQiao: Yeah, this should be better dealt within a top-level coupler and constraint solver. Also, even if it's not implemented in the coupler, I would prefer storing link_idx (the pure data), so that during simulation everything can happen on the CPU instead of relying on magic with link objects or even functions in the Python scope. The API for users can be object-oriented, but our implementation should be more data-centric.

YilingQiao (dismissed):

genesis/engine/entities/fem_entity.py
Milotrince: sorry I left this comment as unresolved, but it is currently implemented at the coupler level! (see `legacy_coupler.py`, not implemented in the SAP coupler)

genesis/engine/entities/fem_entity.py
```diff
@@ -744,12 +744,141 @@ def set_muscle_direction(self, muscle_direction):
...
 
-        self.solver._kernel_set_muscle_direction(
+        self._solver._kernel_set_muscle_direction(
             element_el_start=self._el_start,
             n_elements=self.n_elements,
             muscle_direction=muscle_direction,
         )
 
+    def _sanitize_input_tensor(self, arr, dtype, unbatched_ndim=1, contiguous=False):
+        if not isinstance(arr, torch.Tensor):
```
duburcqa: What is the point of `contiguous=False` instead of unsafe that is more generic?

You could just do this systematically:
```
_tensor = torch.as_tensor(arr, dtype=dtype, device=gs.device).contiguous()
if _tensor is not tensor:
    gs.logger.warning(ALLOCATE_TENSOR_WARNING)
```

if not, you need add least to add `gs.logger.warning(ALLOCATE_TENSOR_WARNING)` after `arr = torch.as_tensor(arr, dtype=dtype, device=gs.device)`

genesis/engine/entities/fem_entity.py
```diff
@@ -744,12 +744,141 @@ def set_muscle_direction(self, muscle_direction):
...
         """
 
-        self.solver._kernel_set_muscle_direction(
+        self._solver._kernel_set_muscle_direction(
             element_el_start=self._el_start,
             n_elements=self.n_elements,
             muscle_direction=muscle_direction,
         )
 
+    def _sanitize_input_tensor(self, arr, dtype, unbatched_ndim=1, contiguous=False):
```
duburcqa: `contiguous=True` should be the default. Try to always be more conservative by default.

genesis/engine/entities/fem_entity.py
```diff
@@ -744,12 +744,141 @@ def set_muscle_direction(self, muscle_direction):
...
+            _tensor = tensor.contiguous()
+            if _tensor is not tensor:
+                gs.logger.warning(ALLOCATE_TENSOR_WARNING)
+            tensor = _tensor
+        return tensor
+
+    def _sanitize_input_verts_idx(self, verts_idx):
+        _verts_idx = self._sanitize_input_tensor(verts_idx, dtype=gs.tc_int, unbatched_ndim=1) + self._v_start
+        verts_idx = _verts_idx.contiguous()
+        if verts_idx is not _verts_idx:
```
duburcqa: No longer relevant since you will always have to allocate memory anyway.

genesis/engine/entities/fem_entity.py
```diff
@@ -744,12 +744,141 @@ def set_muscle_direction(self, muscle_direction):
...
+        if contiguous:
+            _tensor = tensor.contiguous()
+            if _tensor is not tensor:
+                gs.logger.warning(ALLOCATE_TENSOR_WARNING)
+            tensor = _tensor
+        return tensor
+
+    def _sanitize_input_verts_idx(self, verts_idx):
+        _verts_idx = self._sanitize_input_tensor(verts_idx, dtype=gs.tc_int, unbatched_ndim=1) + self._v_start
+        verts_idx = _verts_idx.contiguous()
```
duburcqa: Contiguity of the output of a sum operation is already guaranteed by torch since it is a fresh new allocation anyway.

genesis/engine/entities/fem_entity.py
```diff
@@ -744,12 +744,141 @@ def set_muscle_direction(self, muscle_direction):
...
+        assert tensor.ndim == unbatched_ndim + 1, f"Input tensor ndim is {tensor.ndim}, should be {unbatched_ndim + 1}."
+        assert tensor.shape[0] == self._sim._B, "Input tensor batch size must match the number of environments."
+        if contiguous:
+            _tensor = tensor.contiguous()
+            if _tensor is not tensor:
+                gs.logger.warning(ALLOCATE_TENSOR_WARNING)
+            tensor = _tensor
+        return tensor
+
+    def _sanitize_input_verts_idx(self, verts_idx):
```
duburcqa: This argument should be called `verts_idx_local` apparently?

genesis/engine/entities/fem_entity.py
```diff
@@ -744,12 +744,141 @@ def set_muscle_direction(self, muscle_direction):
...
+            self._kernel_get_verts_pos(self._sim.cur_substep_local, target_poss, verts_idx)
+        target_poss = self._sanitize_input_poss(target_poss)
+
+        assert (
+            len(envs_idx) == len(target_poss) == len(verts_idx)
+        ), "First dimension should match number of environments."
+        assert target_poss.shape[1] == verts_idx.shape[1], "Target position should be provided for each vertex."
+
+        if link is None:
+            B = self._sim._B
```
duburcqa: No need to define `B` if only used twice. Just use it directly.

genesis/engine/entities/fem_entity.py
```diff
@@ -744,12 +744,141 @@ def set_muscle_direction(self, muscle_direction):
...
+            assert isinstance(link, RigidLink), "Only RigidLink is supported for vertex constraints."
+            link_init_pos = self._sanitize_input_tensor(link.get_pos(), dtype=gs.tc_float, contiguous=True)
+            link_init_quat = self._sanitize_input_tensor(link.get_quat(), dtype=gs.tc_float, contiguous=True)
+            link_idx = link.idx
+
+        self._solver._kernel_set_vertex_constraints(
+            self._sim.cur_substep_local,
+            verts_idx,
+            target_poss,
+            1 if is_soft_constraint else 0,
```
duburcqa: pass `is_soft_constraint` directly.

genesis/engine/solvers/fem_solver.py
```diff
@@ -1413,3 +1467,93 @@ def n_surface_vertices(self):
...
+                dv = dt * (spring_force + damping_force)
+                self.elements_v[f + 1, i_v, i_b].vel += dv
+
+    @ti.kernel
+    def _kernel_set_vertex_constraints(
+        self,
+        f: ti.i32,
+        verts_idx: ti.types.ndarray(),  # shape [B, V]
+        target_poss: ti.types.ndarray(),  # shape [B, V, 3]
+        is_soft_constraint: ti.u1,
```
duburcqa: Use `ti.i32` because it turns out  `ti.u1` is not supported on Apple Metal.

genesis/engine/solvers/fem_solver.py
```diff
@@ -1413,3 +1467,93 @@ def n_surface_vertices(self):
...
+        """Apply soft constraints as spring forces for explicit solver."""
+        for i_v, i_b in ti.ndrange(self.n_vertices, self._B):
+            vc = self.vertex_constraints[i_v, i_b]
+            if vc.is_constrained and vc.is_soft_constraint:
+                pos_error = self.elements_v[f, i_v, i_b].pos - vc.target_pos
+                vel_error = self.elements_v[f + 1, i_v, i_b].vel - self.elements_v[f, i_v, i_b].vel
+                spring_force = -vc.stiffness * pos_error
+                damping_force = -2.0 * ti.math.sqrt(vc.stiffness) * vel_error
+
+                dt = self.substep_dt
```
duburcqa: Remove this temporary variable, it is useless.

genesis/engine/solvers/fem_solver.py
```diff
@@ -327,6 +334,29 @@ def compute_surface_elements(self):
     def init_ckpt(self):
         self._ckpt = dict()
 
+    def init_constraints(self):
+        self._constraints_initialized = True
+
+        vertex_constraint_info = ti.types.struct(
+            is_constrained=ti.u1,  # boolean flag indicating if vertex is constrained
```
duburcqa: `gs.ti_bool`

genesis/engine/solvers/fem_solver.py
```diff
@@ -327,6 +334,29 @@ def compute_surface_elements(self):
...
+            target_pos=gs.ti_vec3,  # target position for the constraint
+            is_soft_constraint=ti.u1,  # use spring for soft constraints
```
duburcqa: `gs.ti_bool`

genesis/engine/scene.py
```diff
@@ -1126,6 +1127,40 @@ def load_checkpoint(self, path: str | os.PathLike) -> None:
...
+        if envs_idx is None:
+            return self._envs_idx
+
+        if self.n_envs == 0:
+            gs.raise_exception("`envs_idx` is not supported for non-parallelized scene.")
+
+        if isinstance(envs_idx, slice):
+            return self._envs_idx[envs_idx]
+        if isinstance(envs_idx, int):
+            return self._envs_idx[[envs_idx]]
```
duburcqa: `self._envs_idx[envs_idx: envs_idx+1]` would be WAY more efficient (sadly).

genesis/engine/scene.py
```diff
@@ -1126,6 +1127,40 @@ def load_checkpoint(self, path: str | os.PathLike) -> None:
...
+
+        # Early return if unsafe
+        if unsafe:
+            return envs_idx
+
+        # Perform a bunch of sanity checks
+        _envs_idx = torch.atleast_1d(torch.as_tensor(envs_idx, dtype=gs.tc_int, device=gs.device)).contiguous()
```
duburcqa: Move `torch.atleast_1d` after the check `_envs_idx is not envs_idx`, otherwise it may fail while no memory was actually allocated.

genesis/engine/solvers/fem_solver.py
```diff
@@ -327,6 +334,29 @@ def compute_surface_elements(self):
...
+            link_init_quat=gs.ti_vec4,  # offset rotation of link
+        )
+
+        self.vertex_constraints = vertex_constraint_info.field(
+            shape=self._batch_shape((self.n_vertices)),
+            needs_grad=False,
+            layout=ti.Layout.AOS,
+        )
+
+        self.vertex_constraints.is_constrained.fill(0)
```
duburcqa: `fill(False)`

genesis/engine/solvers/fem_solver.py
```diff
@@ -343,6 +373,10 @@ def build(self):
         # batch fields
         self.init_batch_fields()
 
+        # rendering
+        self.envs_offset = ti.Vector.field(3, dtype=gs.ti_float, shape=self._B)
```
duburcqa: `dtype=ti.f32` if I understand correctly?

genesis/engine/solvers/fem_solver.py
```diff
@@ -343,6 +373,10 @@ def build(self):
...
+        self.envs_offset.from_numpy(self._scene.envs_offset.astype(gs.np_float))
```
duburcqa: `np.float32` if the above comment is correct.

genesis/engine/solvers/fem_solver.py
```diff
@@ -1413,3 +1467,93 @@ def n_surface_vertices(self):
...
+    ):
+        for i_vl, i_bl in ti.ndrange(verts_idx.shape[1], envs_idx.shape[0]):
+            i_b = envs_idx[i_bl]
+            i_v = verts_idx[i_b, i_vl]
+            for j in ti.static(range(3)):
+                self.vertex_constraints[i_v, i_b].target_pos[j] = new_target_poss[i_bl, i_vl, j]
+
+    @ti.kernel
+    def _kernel_remove_specific_constraints(self, verts_idx: ti.types.ndarray(), envs_idx: ti.types.ndarray()):
+        for i_vl, i_bl in ti.ndrange(verts_idx.shape[1], envs_idx.shape[0]):
```
duburcqa: we rather use `i_v_, i_b_` as convention for this.

genesis/engine/solvers/fem_solver.py
Milotrince: can't import gs.ti_bool due to circular import :(

genesis/engine/solvers/fem_solver.py
Milotrince: can we do all the ti.u1 -> gs.bool conversions in a separate PR? i'm not sure if this is correct or what the reasoning behind having it all in `__init__.py` is, but think it would be nice to have all the gs. types in `types.py` and not in `__init__.py` to avoid the circular import issues

duburcqa (dismissed):

duburcqa (approved):

duburcqa (approved):

PR #1314, opened 2025-06-24
[MISC] Various code cleanup and performance improvements.

duburcqa:
* Refactor math helper methods to be more generic and improve performance.
* Speed up primitive mesh generation for viewer.

### DIFF
--- a/genesis/utils/geom.py
+++ b/genesis/utils/geom.py
@@ -1,8 +1,12 @@
+import math
+
 import numpy as np
-import taichi as ti
+import numba as nb
 import torch
 import torch.nn.functional as F
-from scipy.spatial.transform import Rotation
+
+import taichi as ti
+
 import genesis as gs
 
 # ------------------------------------------------------------------------------------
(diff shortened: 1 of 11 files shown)


genesis/utils/geom.py
```diff
@@ -11,197 +15,252 @@
...
 
+    new_v = ti.Vector(
+        [
+            v.x * (q_xx + q_ww - q_yy - q_zz) + v.y * (2.0 * q_xy - 2.0 * q_wz) + v.z * (2.0 * q_xz + 2.0 * q_wy),
+            v.x * (2.0 * q_wz + 2.0 * q_xy) + v.y * (q_ww - q_xx + q_yy - q_zz) + v.z * (-2.0 * q_wx + 2.0 * q_yz),
+            v.x * (-2.0 * q_wy + 2.0 * q_xz) + v.y * (2.0 * q_wx + 2.0 * q_yz) + v.z * (q_ww - q_xx - q_yy + q_zz),
+        ],
+        dt=gs.ti_float,
+    )
+    new_v /= q_ww + q_xx + q_yy + q_zz
```
YilingQiao: should we avoid /= as it's atomic? Or maybe taichi will optmize it and we don't need to worry about it.

YilingQiao:
Looks good to me (after fixing CI).

genesis/utils/geom.py
duburcqa: Yes you are right

Genesis-Embodied-AI (approved):

PR #1315, opened 2025-06-25
[FEATURE] Scene saver

### LINKED ISSUE #716: [Feature]: Function to save/load state of the simulator in/from file or scene.to_dict()

Directly trying to pickle the state fails since the simulation contains Taichi objects (like fields or expressions) that cannot be directly pickled.
``` python
...
state = scene.get_state()

import pickle
# save states
with open("state.pkl", "wb") as f:
    pickle.dump(state, f)
```

throws

``` bash
  File "<string>", line 22, in <module>
TypeError: cannot pickle 'taichi._lib.core.taichi_python.Expr' object
```

### Motivation

Motivation: create dataset

I want to split simulation from rendering. I.e. simulate once and only save the states. And render offline.

### Potential Benefit

This would allow the user to simulate once and render offline.

- provide function save_state()
- provide function load_state()

### Additional information

_No response_

LeonLiu4:
- A simple file-based checkpoint system.
- Helpers gs.states.save_ckpt(sim, path) and gs.states.load_ckpt(sim, path) restore the full simulator state to a compressed NPZ and meta file. 
- Each solver now exposes dump_ckpt_to_numpy and load_ckpt_from_numpy. Scalar fields become ndarrays, vector / matrix fields a 1-elem object array.
- Resolves #716


genesis/engine/solvers/base_solver.py
```diff
@@ -34,6 +34,45 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
     def _add_force_field(self, force_field):
         self._ffs.append(force_field)
 
+    @staticmethod
+    def _is_ti_field(obj) -> bool:
```
duburcqa: This is definitely not the right way to check if something is a taichi field. Just do
`isinstance(obj, taichi.Field)` and remove this method entirely. I checked myself and it works perfectly on both `ti.Vector`, `ti.Matrix` and `ti.types.struct` instances.

genesis/engine/solvers/base_solver.py
```diff
@@ -34,6 +34,45 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
...
+        return callable(getattr(obj, "to_numpy", None)) and callable(getattr(obj, "from_numpy", None))
+
+    def save_ckpt(self, tag: str = "0"):
+        for val in self.__dict__.values():
+            if self._is_ti_field(val):
+                val[tag] = val[None]
```
duburcqa: What are you trying to achieve here? I don't understand and it does not work on my machine if I do:
```python
val = scene.sim.rigid_solver.__dict__['links_T']
val[tag] = val[None]
```

genesis/engine/solvers/base_solver.py
```diff
@@ -34,6 +34,45 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
...
+
+    def load_ckpt(self, tag: str = "0"):
+        for val in self.__dict__.values():
+            if self._is_ti_field(val):
+                val[None] = val[tag]
```
duburcqa: Same. I don't understand.

genesis/engine/solvers/base_solver.py
```diff
@@ -34,6 +34,45 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
...
+        return callable(getattr(obj, "to_numpy", None)) and callable(getattr(obj, "from_numpy", None))
+
+    def save_ckpt(self, tag: str = "0"):
+        for val in self.__dict__.values():
+            if self._is_ti_field(val):
+                val[tag] = val[None]
+
+    def load_ckpt(self, tag: str = "0"):
+        for val in self.__dict__.values():
+            if self._is_ti_field(val):
```
duburcqa: Use `isinstance`

genesis/engine/solvers/base_solver.py
```diff
@@ -34,6 +34,45 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
...
+                val[None] = val[tag]
+
+    def dump_ckpt_to_numpy(self) -> dict[str, np.ndarray]:
+        arrays, prefix = {}, self.__class__.__name__ + "/"
+        for name, val in self.__dict__.items():
+            if not self._is_ti_field(val):
```
duburcqa: Use `isinstance`

genesis/engine/solvers/base_solver.py
```diff
@@ -34,6 +34,45 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
...
+                continue
+            arr = val.to_numpy()
+            if isinstance(arr, np.ndarray):
+                arrays[prefix + name] = arr.copy()
```
duburcqa: * Prefer `".".join((self.__class__.__name__, name))`
* Making a copy is useless here

genesis/engine/solvers/base_solver.py
```diff
@@ -34,6 +34,45 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
...
+            if self._is_ti_field(val):
+                val[None] = val[tag]
+
+    def dump_ckpt_to_numpy(self) -> dict[str, np.ndarray]:
+        arrays, prefix = {}, self.__class__.__name__ + "/"
+        for name, val in self.__dict__.items():
+            if not self._is_ti_field(val):
+                continue
+            arr = val.to_numpy()
+            if isinstance(arr, np.ndarray):
```
duburcqa: Remove this condition. it is useless I think. Just do the same in both cases.

genesis/engine/solvers/base_solver.py
```diff
@@ -34,6 +34,45 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
...
+                arrays[prefix + name] = arr.copy()
+            elif isinstance(arr, dict):
+                arrays[prefix + name] = np.array([arr], dtype=object)
```
duburcqa: Why encapsulating the dict as a numpy array?

genesis/engine/solvers/base_solver.py
```diff
@@ -34,6 +34,45 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
...
+        for val in self.__dict__.values():
+            if self._is_ti_field(val):
+                val[None] = val[tag]
+
+    def dump_ckpt_to_numpy(self) -> dict[str, np.ndarray]:
+        arrays, prefix = {}, self.__class__.__name__ + "/"
+        for name, val in self.__dict__.items():
+            if not self._is_ti_field(val):
+                continue
+            arr = val.to_numpy()
```
duburcqa: You should rather use `ti_field_to_torch` I think. It would be faster and more consistent with the whole codebase.

genesis/engine/solvers/base_solver.py
```diff
@@ -34,6 +34,45 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
...
+    def load_ckpt_from_numpy(self, arr_dict: dict[str, np.ndarray]) -> None:
+        prefix = self.__class__.__name__ + "/"
+        for name, val in self.__dict__.items():
+            if not self._is_ti_field(val):
+                continue
+            key = prefix + name
+            if key not in arr_dict:
+                continue
+            arr = arr_dict[key]
+            if arr.dtype == object and arr.shape == (1,):
```
duburcqa: This edge-case should be removed.

genesis/engine/solvers/base_solver.py
```diff
@@ -34,6 +34,45 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
...
+                arr = arr[0]
+            val.from_numpy(arr)
```
duburcqa: `from_torch`

genesis/engine/states/io.py
```diff
@@ -0,0 +1,44 @@
...
+from pathlib import Path
+import pickle, numpy as np
+import genesis as gs
+
+
+def save_ckpt(sim: gs.Simulator, path: str | Path, tag: str = "0") -> None:
+    path = Path(path)
+    path.parent.mkdir(parents=True, exist_ok=True)
+
+    arrays = {}
```
duburcqa: Just create a big `state` and pickle everything. No need to call `np.savez_compressed` manually and create multiple files.

genesis/engine/states/io.py
```diff
@@ -0,0 +1,44 @@
...
+    for solver in sim.active_solvers:
+        arrays.update(solver.dump_ckpt_to_numpy())
+
+    np.savez_compressed(path.with_suffix(".npz"), **arrays)
+
+    with open(path.with_suffix(".meta"), "wb") as f:
```
duburcqa: I would suggest to rather log all scene-level option attributes.

genesis/engine/states/io.py
```diff
@@ -0,0 +1,44 @@
...
+import pickle, numpy as np
+import genesis as gs
+
+
+def save_ckpt(sim: gs.Simulator, path: str | Path, tag: str = "0") -> None:
+    path = Path(path)
+    path.parent.mkdir(parents=True, exist_ok=True)
+
+    arrays = {}
+    for solver in sim.active_solvers:
```
duburcqa: You should rather call `dump_ckpt_to_numpy()` at scene-level.

genesis/engine/states/io.py
```diff
@@ -0,0 +1,44 @@
...
+
+    for solver in sim.active_solvers:
+        solver.load_ckpt_from_numpy(arrays)
+
+    sim.reset_grad()
+    sim._cur_substep_global = 0
+    gs.logger.info(f"[ckpt] restored from {path}.npz")
+
+
+class _StatesFacade:
```
duburcqa: Remove everything that does after this line and just add `from genesis.engine import state` in `genesis/__init__.py`

tests/test_rigid_physics.py
```diff
@@ -2153,6 +2156,34 @@ def test_urdf_mimic(show_viewer, tol):
     assert_allclose(gs_qpos[-1], gs_qpos[-2], tol=tol)
 
 
+@pytest.mark.required
+@pytest.mark.parametrize("backend", [gs.cpu])
+def test_scene_saver(show_viewer, tol):
```
duburcqa: Just pick the usual franka robot example for this. No need to use a drone and franka is the go-to robot for this kind of generic tests.

genesis/engine/solvers/base_solver.py
LeonLiu4: fixed

genesis/engine/solvers/base_solver.py
LeonLiu4: removed

genesis/engine/solvers/base_solver.py
LeonLiu4: removed

genesis/engine/solvers/base_solver.py
LeonLiu4: fixed

genesis/engine/solvers/base_solver.py
LeonLiu4: fixed

genesis/engine/solvers/base_solver.py
LeonLiu4: resolved

genesis/engine/solvers/base_solver.py
LeonLiu4: I have the check because the val is sometimes not a taichi field and is instead like a UID

genesis/engine/solvers/base_solver.py
LeonLiu4: resolved

genesis/engine/solvers/base_solver.py
LeonLiu4: resolved

genesis/engine/solvers/base_solver.py
LeonLiu4: fixed

genesis/engine/solvers/base_solver.py
LeonLiu4: added

genesis/engine/states/io.py
LeonLiu4: done

genesis/engine/states/io.py
LeonLiu4: done

genesis/engine/states/io.py
LeonLiu4: done

genesis/engine/states/io.py
LeonLiu4: done

tests/test_rigid_physics.py

genesis/engine/scene.py
```diff
@@ -1057,6 +1060,71 @@ def _backward(self):
...
+        -------
+        dict[str, np.ndarray]
+            Mapping ``"Class.attr[.member]" → array`` with raw field data.
+        """
+        arrays: dict[str, np.ndarray] = {}
+
+        scene_prefix = f"{self.__class__.__name__}."
+        for name, field in self.__dict__.items():
+            if isinstance(field, ti.Field):
+                arrays[scene_prefix + name] = field.to_numpy()
```
duburcqa: What about `".".join((self.__class__.__name__, name))`

genesis/engine/scene.py
```diff
@@ -1057,6 +1060,71 @@ def _backward(self):
...
+        """
+        with open(path, "rb") as f:
+            state = pickle.load(f)
+
+        arrays = state["arrays"]
+
+        scene_prefix = f"{self.__class__.__name__}."
+        for name, field in self.__dict__.items():
+            if isinstance(field, ti.Field):
+                key = scene_prefix + name
```
duburcqa: `key = ".".join((self.__class__.__name__, name))`

genesis/engine/solvers/base_solver.py
```diff
@@ -34,6 +36,75 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
...
+
+        for attr_name, field in self.__dict__.items():
+            if not isinstance(field, ti.Field):
+                continue
+
+            key_base = cls_prefix + attr_name
+
+            # ---- fast path: torch → numpy ------------------------------------
+            try:
+                arrays[key_base] = ti_field_to_torch(field).cpu().numpy()
```
duburcqa: Just do `field.to_numpy()` if working with `torch.Tensor` directly is not working out for some reason.

genesis/engine/solvers/base_solver.py
```diff
@@ -34,6 +36,75 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
...
 
+    def dump_ckpt_to_numpy(self) -> dict[str, np.ndarray]:
+        arrays: dict[str, np.ndarray] = {}
+        cls_prefix = f"{self.__class__.__name__}."
+
+        for attr_name, field in self.__dict__.items():
+            if not isinstance(field, ti.Field):
+                continue
+
+            key_base = cls_prefix + attr_name
```
duburcqa: `key_base = ".".join((self.__class__.__name__, attr_name))`

genesis/engine/solvers/base_solver.py
```diff
@@ -34,6 +36,75 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
...
+
+            # ---- fast path: torch → numpy ------------------------------------
+            try:
+                arrays[key_base] = ti_field_to_torch(field).cpu().numpy()
+                continue
+            except (AttributeError, ti.TaichiRuntimeError):
+                pass  # fall back
```
duburcqa: Add a comment to explain in which cases it may fail.

genesis/engine/solvers/base_solver.py
```diff
@@ -34,6 +36,75 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
...
+
+            # ---- generic path ------------------------------------------------
+            data = field.to_numpy()
+
+            # StructField → data is a dict: flatten each member
+            if isinstance(data, dict):
```
duburcqa, line 50: You can remove the fallback if using `to_torch` or `to_numpy` directly.

Ignore my comment about using `ti_field_to_torch`. That was a bad idea. Sorry about that...

genesis/engine/solvers/base_solver.py
```diff
@@ -34,6 +36,75 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
...
+            # ---- Ordinary field ---------------------------------------------
+            if key_base not in arr_dict:
+                continue  # nothing saved for this attribute
+
+            arr = arr_dict[key_base]
+
+            # Try fast torch route if array is numeric
+            if isinstance(arr, np.ndarray) and arr.dtype.kind != "O":
+                try:
+                    field.from_torch(torch.from_numpy(arr))
```
duburcqa: Why not just `field.from_numpy(arr)` ?

tests/test_rigid_physics.py
```diff
@@ -2200,6 +2203,44 @@ def test_urdf_mimic(show_viewer, tol):
     assert_allclose(gs_qpos[-1], gs_qpos[-2], tol=tol)
 
 
+@pytest.mark.required
+@pytest.mark.parametrize("backend", [gs.cpu])
+def test_scene_saver_franka(show_viewer, tol):
+
```
duburcqa: Do not skip space here.

genesis/engine/scene.py
LeonLiu4: fixed

genesis/engine/scene.py
LeonLiu4: fixed

genesis/engine/solvers/base_solver.py
LeonLiu4: done

genesis/engine/solvers/base_solver.py
LeonLiu4: fixed

genesis/engine/solvers/base_solver.py
LeonLiu4: fixed

genesis/engine/solvers/base_solver.py
LeonLiu4: done

genesis/engine/solvers/base_solver.py
LeonLiu4: fixed

tests/test_rigid_physics.py

PR #1316, opened 2025-06-25
[FEATURE] Batched path plan

Kashu7100:
## Description

Add batched (sampling based) path planning algorithms (RRT and RRTConnect) and refactored the path planners.

## Related Issue

## Motivation and Context

We can now perform path planning in batched envs.

Added new example:
`python examples/rigid/grasp_bottle_batch.py -v -a rrt_connect`
`python examples/rigid/grasp_bottle_batch.py -v -a rrt`

video with 100 parallel envs:

https://github.com/user-attachments/assets/e8a38c74-48d9-4759-9f32-baeb92032d69

## Checklist:


examples/rigid/grasp_bottle.py
duburcqa, line 1: Revert these changes.

examples/rigid/grasp_bottle_batch.py
duburcqa, line 1: What about merging this file with the original one, and just add an extra argument `n_envs` that would be equal to either 0 by default (or 49 (7x7), as you want). 100 may too much if the viewer is enabled.

examples/rigid/grasp_bottle_batch.py
```diff
@@ -0,0 +1,143 @@
...
+        camera_fov=30,
+        max_FPS=60,
+    )
+
+    scene = gs.Scene(
+        viewer_options=viewer_options,
+        rigid_options=gs.options.RigidOptions(
+            dt=0.01,
+        ),
+        show_FPS=False,
```
duburcqa: This is deprecated.

examples/rigid/suction_cup.py
duburcqa, line 1: Revert these changes.

examples/tutorials/IK_motion_planning_grasp.py
duburcqa, line 1: Revert these changes.

genesis/engine/solvers/tool_solver.py
```diff
@@ -6,6 +6,7 @@
 from genesis.utils.misc import *
 
 from .base_solver import Solver
+from genesis.engine.entities.tool_entity.tool_entity import ToolEntity
```
duburcqa: Remove this. By the way, `from genesis.engine.entities.tool_entity.tool_entity import ToolEntity` should be moved before relative imports.

tests/test_rigid_physics.py
duburcqa, line 1: Revert these changes.

duburcqa: Plus modify `test_path_planning_avoidance` unit test to leverage batching.

duburcqa: Plus remove `@pytest.mark.skipif(sys.platform == "win32", reason="OMPL is not supported on Windows OS.")` if it is no longer using OMPL under the hood.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+            qpos_goal = qpos_goal.repeat(len(envs_idx), 1)
+        else:
+            assert qpos_goal.shape[0] == len(envs_idx), f"Batch size mismatch: {qpos_goal.shape=} {len(envs_idx)=}"
+
+        if qpos_start.shape[1] != self.n_qs or qpos_goal.shape[1] != self.n_qs:
+            gs.raise_exception("Invalid shape for `qpos_start` or `qpos_goal`.")
+
+        if check_joint_limits:
+            # NOTE: process joint limit
+            if (qpos_start.cpu().numpy() < self.default_q_limit[0]).any() or (
```
duburcqa: `tensor_to_array`

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+                qpos_start.cpu().numpy() > self.default_q_limit[1]
```
duburcqa: `tensor_to_array`

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+        if qpos_goal.shape[0] == 1:
+            qpos_goal = qpos_goal.repeat(len(envs_idx), 1)
+        else:
+            assert qpos_goal.shape[0] == len(envs_idx), f"Batch size mismatch: {qpos_goal.shape=} {len(envs_idx)=}"
+
+        if qpos_start.shape[1] != self.n_qs or qpos_goal.shape[1] != self.n_qs:
+            gs.raise_exception("Invalid shape for `qpos_start` or `qpos_goal`.")
+
+        if check_joint_limits:
+            # NOTE: process joint limit
```
duburcqa: What is this comment?

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+    def n_qs(self):
+        """Number of generalized coordinates."""
+        return self._entity.n_qs
+
+    @property
+    def _q_start(self):
+        """Starting generalized coordinates."""
+        return self._entity._q_start
+
+    def validate_input_qpos_batch(self, qpos_goal, qpos_start, envs_idx, *, check_joint_limits=True):
```
duburcqa: Remove `_batch` suffix.

duburcqa: Why is `envs_idx` mandatory? You could support None.

duburcqa: It should rather be called `_sanitize_[...]` for consistency with the whole codebase.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+            gs.raise_exception("Invalid shape for `qpos_start` or `qpos_goal`.")
+
+        if check_joint_limits:
+            # NOTE: process joint limit
+            if (qpos_start.cpu().numpy() < self.default_q_limit[0]).any() or (
+                qpos_start.cpu().numpy() > self.default_q_limit[1]
+            ).any():
+                gs.logger.warning("`qpos_start` exceeds joint limit.")
+
+            if (qpos_goal.cpu().numpy() < self.default_q_limit[0]).any() or (
```
duburcqa: `tensor_to_array`

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+                qpos_goal.cpu().numpy() > self.default_q_limit[1]
```
duburcqa: `tensor_to_array`

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+            num_samples = sample_ind[i_b, 1] - sample_ind[i_b, 0]
+            for i_q in range(self.n_qs):
+                start = path[i_b, sample_ind[i_b, 0], i_q]
+                end = path[i_b, sample_ind[i_b, 1], i_q]
+                step = (end - start) / num_samples
+                for i_s in range(num_samples):
+                    tensor[i_b, sample_ind[i_b, 0] + i_s, i_q] = start + step * i_s
+
+    def check_collision(self, path, ignore_geom_pairs=None, envs_idx=None):
+        path = path.transpose(1, 0)  # N, B, Dof
```
duburcqa: Just do the same as we do everywhere else:
```python
        for i_keypoint in range(path.shape[1]):
            qpos = path[:, i_keypoint]
            [...]
```

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+                continue
+            num_samples = sample_ind[i_b, 1] - sample_ind[i_b, 0]
+            for i_q in range(self.n_qs):
+                start = path[i_b, sample_ind[i_b, 0], i_q]
+                end = path[i_b, sample_ind[i_b, 1], i_q]
+                step = (end - start) / num_samples
+                for i_s in range(num_samples):
+                    tensor[i_b, sample_ind[i_b, 0] + i_s, i_q] = start + step * i_s
+
+    def check_collision(self, path, ignore_geom_pairs=None, envs_idx=None):
```
duburcqa: I would suggest to implement a single kernel that does the check for all keypoints at once. here you are calling to many kernels, so it will be slow. It may even possible to use parrallelization at keypoint-level but not sure.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+            for i_p in range(ignore_geom_pairs.shape[0]):
+                if (ignore_geom_pairs[i_p, 0] == i_ga and ignore_geom_pairs[i_p, 1] == i_gb) or (
+                    ignore_geom_pairs[i_p, 0] == i_gb and ignore_geom_pairs[i_p, 1] == i_ga
+                ):
+                    ignore = True
+                    break
+            if ignore:
+                continue
+
+            # TODO: handle self-collision (except the case for closed gripper)
```
duburcqa: I would suggest to always take into account all collision pairs that are not ignored. 

Just ignore the starting AND finally active collision pairs.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+        for i_b in envs_idx:
+            collision_detected = self._func_check_collision(ignore_geom_pairs, i_b)
+            tensor[i_b] = tensor[i_b] or collision_detected
+
+    def shortcut_path(self, path_mask, path, iterations=50, ignore_geom_pairs=None, envs_idx=None):
+        """
+        path_mask: torch.Tensor
+            node mask [N,B] for the obtained path
+        path: torch.Tensor
+            the [N,B,Dof] tensor
```
duburcqa: This means nothing.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+        iterations: int
+            the number of refine iterations
+        """
+        path_mask = path_mask.t()  # B, N
```
duburcqa: Could you avoid all these transpose? It is error-prone. Moreover, prefer `.T` in torch.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+        path = path.transpose(1, 0)  # B, N, Dof
+        for i in range(iterations):
+            ind = torch.multinomial(path_mask, 2).sort()[0]  # B, 2
+            ind_mask = (ind[:, 1] - ind[:, 0]) > 1
+            tmp_path = path.clone().contiguous()
```
duburcqa: No need to do `.contiguous()` after cloning.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+            self.interpolate_path(tmp_path, path.contiguous(), ind.contiguous(), ind_mask)
```
duburcqa: Move return arguments at the end, not the beginning.

genesis/utils/path_planing.py
duburcqa: Rename it `tmp_path`

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+        path: ti.types.ndarray(),  # [B, N, Dof]
+        sample_ind: ti.types.ndarray(),  # [B, 2]
+        mask: ti.types.ndarray(),  # [B]
+    ):
+        ti.loop_config(serialize=self._solver._para_level < gs.PARA_LEVEL.ALL)
+        for i_b in range(path.shape[0]):
+            if not mask[i_b]:
+                continue
+            num_samples = sample_ind[i_b, 1] - sample_ind[i_b, 0]
+            for i_q in range(self.n_qs):
```
duburcqa: You cannot interpolate this way. It is just wrong. You need to take into account the Lie algebra space associated with your robot. This is important for spherical joints that are mapping to SO3 space (basically quaternions). This depends on sequence type of joints. I would suggest adding a `lerp` method add `rigid_solver` and `entity` levels (based on a taichi kernel under the hood).

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+        self._rrt_node_info.configuration.fill(0.0)
+        self._rrt_tree_size.fill(0)
+        self._rrt_is_active.fill(0)
+        self._rrt_goal_reached_node_idx.fill(-1)
+
+    @ti.kernel
+    def _kernel_rrt_init(
+        self, qpos_start: ti.types.ndarray(), qpos_goal: ti.types.ndarray(), envs_idx: ti.types.ndarray()
+    ):
+        # NOTE: run IK before this
```
duburcqa: What is this note? It should rather be `# Assuming that IK was executed before calling this method`

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+    ):
+        ti.loop_config(serialize=self._solver._para_level < gs.PARA_LEVEL.ALL)
+        for i_b in envs_idx:
+            if not self._rrt_is_active[i_b]:
+                continue
+
+            random_sample = ti.Vector(
+                [
+                    q_limit_lower[i_q] + ti.random(dtype=gs.ti_float) * (q_limit_upper[i_q] - q_limit_lower[i_q])
+                    for i_q in ti.static(range(self.n_qs))
```
duburcqa: `ti.static(range(self.n_qs))` is not acceptable. Remove static.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+                ]
+            )
+            if ti.random() < self._rrt_goal_bias:
+                random_sample = ti.Vector(
+                    [self._rrt_goal_configuration[i_q, i_b] for i_q in ti.static(range(self.n_qs))]
```
duburcqa: same.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+                dist = (self._rrt_node_info.configuration[i_n, i_b] - random_sample).norm()
+                if dist < nearest_neighbor_dist:
+                    nearest_neighbor_dist = dist
+                    nearest_neighbor_idx = i_n
+
+            # steer from nearest neighbor to random sample
+            nearest_config = self._rrt_node_info.configuration[nearest_neighbor_idx, i_b]
+            direction = random_sample - nearest_config
+            steer_result = ti.Vector.zero(gs.ti_float, self.n_qs)
+            for i_q in ti.static(range(self.n_qs)):
```
duburcqa: same.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+            )
+            if ti.random() < self._rrt_goal_bias:
+                random_sample = ti.Vector(
+                    [self._rrt_goal_configuration[i_q, i_b] for i_q in ti.static(range(self.n_qs))]
+                )
+
+            # find nearest neighbor
+            nearest_neighbor_idx = -1
+            nearest_neighbor_dist = gs.ti_float(1e30)
+            ti.loop_config(serialize=self._solver._para_level < gs.PARA_LEVEL.ALL)
```
duburcqa: Not sure you can do this.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+            for i_n in range(self._rrt_tree_size[i_b]):
+                dist = (self._rrt_node_info.configuration[i_n, i_b] - random_sample).norm()
```
duburcqa: Use `norm_sqr` and only take the root at the end. This would be more efficient.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+                    nearest_neighbor_dist = dist
+                    nearest_neighbor_idx = i_n
+
+            # steer from nearest neighbor to random sample
+            nearest_config = self._rrt_node_info.configuration[nearest_neighbor_idx, i_b]
+            direction = random_sample - nearest_config
+            steer_result = ti.Vector.zero(gs.ti_float, self.n_qs)
+            for i_q in ti.static(range(self.n_qs)):
+                # If the step size exceeds max_step_size, clip it
+                if abs(direction[i_q]) > self._rrt_max_step_size:
```
duburcqa: `ti.abs`

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+                    direction[i_q] = ti.math.sign(direction[i_q]) * self._rrt_max_step_size
```
duburcqa: `(-1.0 if direction[i_q] < 0.0 else 1.0)` is probably what you want.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+            i_ga = self._solver.collider.contact_data[i_c, i_b].geom_a
+            i_gb = self._solver.collider.contact_data[i_c, i_b].geom_b
+
+            ignore = False
+            for i_p in range(ignore_geom_pairs.shape[0]):
+                if (ignore_geom_pairs[i_p, 0] == i_ga and ignore_geom_pairs[i_p, 1] == i_gb) or (
+                    ignore_geom_pairs[i_p, 0] == i_gb and ignore_geom_pairs[i_p, 1] == i_ga
+                ):
+                    ignore = True
+                    break
```
duburcqa: Use `is_` prefix for boolean.

Try avoiding break and continue. Because it will mess up with auto-diff in the future. Just do:
```
for i_c in range(self._solver.collider.n_contacts[i_b]):
    if not is_collision_detected:
        [...]
        for i_p in range(ignore_geom_pairs.shape[0]):
            [...]
            if not is_ignored:
                [...]
```

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+    @ti.kernel
+    def _kernel_rrt_step2(
+        self,
+        ignore_geom_pairs: ti.types.ndarray(),
+        envs_idx: ti.types.ndarray(),
+    ):
+        ti.loop_config(serialize=self._solver._para_level < gs.PARA_LEVEL.ALL)
+        for i_b in envs_idx:
+            if not self._rrt_is_active[i_b]:
+                continue
```
duburcqa: same.

genesis/utils/path_planing.py
Kashu7100: when do we need `ti.static`? I thought `n_qs` is constant after build.

duburcqa: We do not want to force recompilation just because `n_qs` has changed. Because static means it will recompile every time this constant changes. This means that recompilation would be triggered even if we were using taichi dynamic ndarray. Is it correct @YilingQiao ?

examples/rigid/grasp_bottle_batch.py
Kashu7100: which one is deprecated?

genesis/utils/path_planing.py
Kashu7100: Not quite sure how to handle this.

genesis/utils/path_planing.py
Kashu7100: which part? `gs.ti_float(1e30)`?

examples/rigid/grasp_bottle_batch.py
duburcqa:
```python
    profiling_options=gs.options.ProfilingOptions(
        show_FPS=True,
    ),
```

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1165 @@
...
+        path = path.transpose(1, 0)  # N, B, Dof
+        res = torch.zeros(path.shape[1], dtype=gs.tc_int, device=gs.device)
+        for qpos in path:
+            self._entity.set_qpos(qpos, envs_idx=envs_idx)
+            self._solver._kernel_detect_collision()
+            self._kernel_check_collision(res, ignore_geom_pairs, envs_idx)
+        return res
+
+    @ti.func
+    def _func_check_collision(
```
duburcqa: Note that `RigidEntity.get_contacts` and `Collider.get_contacts` have been recently introduced. That may be useful for you.

Kashu7100: I will wait for #1320

tests/test_rigid_physics.py
```diff
@@ -1434,7 +1434,6 @@ def test_path_planning_avoidance(show_viewer):
     hand_quat_ref = torch.tensor([0.3073, 0.5303, 0.7245, -0.2819], device=gs.device)
     qpos = franka.inverse_kinematics(hand, pos=hand_pos_ref, quat=hand_quat_ref)
     qpos[-2:] = 0.04
-
```
duburcqa, line 1393: Remove this change, it is unnecessary.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1588,184 +1589,34 @@ def plan_path(
         ignore_joint_limit : bool, optional
             This option has been deprecated and is not longer doing anything.
         planner : str, optional
-            The name of the motion planning algorithm to use. Supported planners: 'PRM', 'RRT', 'RRTConnect', 'RRTstar', 'EST', 'FMT', 'BITstar', 'ABITstar'. Defaults to 'RRTConnect'.
+            The name of the motion planning algorithm to use. Supported planners: 'RRT', 'RRTConnect'. Defaults to 'RRTConnect'.
```
duburcqa: 120 chars line width limit.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1588,184 +1589,34 @@ def plan_path(
...
+        planner = RRTConnect(self)
+        path = planner.plan(
+            qpos_goal, 
+            qpos_start,
+            resolution=resolution,
+            timeout=timeout,
+            max_retry=max_retry,
+            smooth_path=smooth_path,
+            num_waypoints=num_waypoints,
+            # ignore_collision=ignore_collision,
```
duburcqa: Why is this commented out?

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -2745,10 +2596,7 @@ def get_contacts(self, with_entity=None, exclude_self_contact=False):
                 ),
             )
 
-        if self._solver.n_envs == 0:
-            contacts_info = {key: value[valid_mask] for key, value in contacts_info.items()}
-        else:
-            contacts_info = {key: value[:, valid_mask] for key, value in contacts_info.items()}
+        contacts_info = {key: value[valid_mask] for key, value in contacts_info.items()}
```
duburcqa: Remove this change. Fixed in https://github.com/Genesis-Embodied-AI/Genesis/pull/1320

genesis/utils/path_planing.py
duburcqa: Merged!

genesis/utils/path_planing.py
Kashu7100: I would create a separate PR for handling spherical joints since this seem to take longer.
Meanwhile I will raise an error with entity with gs.JOINT_TYPE.SPHERICAL.
@duburcqa is this ok with you?

duburcqa: Yes, same for `FREE` joints. But do not resolve this comment.

genesis/utils/path_planing.py
duburcqa: Why not just make `self.default_q_limit` as torch tensor ? That would avoid moving data from GPU to CPU.

genesis/utils/path_planing.py
duburcqa: After thinking twice, that would be a pain. Let's ignore this for now. But do not resolve this comment for now.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1127 @@
...
+                continue
+            num_samples = sample_ind[i_b, 1] - sample_ind[i_b, 0]
+            for i_q in range(self.n_qs):
+                start = path[i_b, sample_ind[i_b, 0], i_q]
+                end = path[i_b, sample_ind[i_b, 1], i_q]
+                step = (end - start) / num_samples
+                for i_s in range(num_samples):
+                    tensor[i_b, sample_ind[i_b, 0] + i_s, i_q] = start + step * i_s
+
+    def check_collision(self, path, ignore_geom_pairs=None, envs_idx=None):
```
duburcqa: Neither `ignore_geom_pairs` or `envs_idx=None` is supported. Remove the default values.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1127 @@
...
+        valid_mask = torch.logical_and(valid_mask, contact_indices < n_contacts)
+
+        max_env_collisions = int(torch.max(n_contacts).item())
+        valid_mask = valid_mask[:max_env_collisions]
+        geom_a = scene_contact_info["geom_a"][:max_env_collisions][valid_mask]  # N
+        geom_b = scene_contact_info["geom_b"][:max_env_collisions][valid_mask]  # N
+        assert len(geom_a) == len(geom_b)
+
+        # NOTE: we will reduce the contacts in batch dim assuming internal geom collisions are the same for a robot
+        stacked_tensors = torch.stack((geom_a, geom_b), dim=1)
```
duburcqa: This is not a proper name. Just remove this temporary.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1127 @@
...
+        qpos_cur = self._entity.get_qpos()
+        qpos_goal, qpos_start = self.validate_input_qpos_batch(qpos_goal, qpos_start, envs_idx)
+
+        self._init_rrt_fields()
+        self._reset_rrt_fields()
+        self._kernel_rrt_init(qpos_start.contiguous(), qpos_goal.contiguous(), envs_idx)
+
+        self._entity.set_qpos(qpos_start)
+        self._solver._kernel_detect_collision()
+        scene_contact_info = self._solver.collider.contact_data.to_torch(gs.device)
```
duburcqa: Use the new API instead.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1127 @@
...
+        n_contacts = self._solver.collider.n_contacts.to_torch(gs.device)
```
duburcqa: Do not use `self._solver.collider.n_contacts.to_torch(gs.device)` but rather the new API.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1127 @@
...
+
+        valid_mask = torch.logical_or(
```
duburcqa: It seems that you could just call `self._entity.get_contacts()` directly.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,1127 @@
...
+    IS_OMPL_AVAILABLE = False
+
+    __all__ = [
+        "RRTConnect",
+        "RRT",
+    ]
+
+
+@ti.data_oriented
+class PathPlanner(ABC):
```
duburcqa, line 12: I don't see the point of doing this. I would recommend explicitly calling `self._entity` and `self._solver`. This is clearer and more consistent with what we are doing elsewhere. Because I'm not convinced these attributes should be part of the public API. But if you think they should and also @YilingQiao, then you can keep the current design.

Kashu7100:
[WIP] still some problem(s) and local test not passing.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1548,35 +1546,39 @@ def _kernel_forward_kinematics(
...
+            The goal state. [B, Nq] or [1, Nq]
         qpos_start : None | array_like, optional
-            The start state. If None, the current state of the rigid entity will be used. Defaults to None.
+            The start state. If None, the current state of the rigid entity will be used. Defaults to None. [B, Nq] or [1, Nq]
         resolution : float, optiona
             Joint-space resolution in pourcentage. It corresponds to the maximum distance between states to be checked
             for validity along a path segment. Default to 1%.
         timeout : float, optional
-            The maximum time (in seconds) allowed for the motion planning algorithm to find a solution. Defaults to 5.0.
+            This option has been deprecated and is not longer doing anything.
```
duburcqa: Why? I think it was useful.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1588,184 +1590,84 @@ def plan_path(
...
+        if "ignore_joint_limit" in kwargs:
+            gs.logger.warning("`ignore_joint_limit` is deprecated")
+
+        ee_link_idx = None
+        if ee_link_name is not None:
+            assert with_entity is not None, "`with_entity` must be specified."
+            ee_link_idx = self.get_link(ee_link_name).idx
+        if with_entity is not None:
+            assert ee_link_name is not None, "reference link of the robot must be specified."
+            assert len(with_entity.links) == 1, "only non-articulated object is supported for now."
```
duburcqa, line 1711: What are you trying to check here?

tests/test_rigid_physics.py
```diff
@@ -1569,27 +1561,26 @@ def test_path_planning_avoidance(show_viewer):
...
-        resolution=0.002,
+        resolution=0.05,
+        max_nodes=4000,
+        max_retry=40,
     )
-    assert free_path
-    assert_allclose(free_path[0], 0.0, tol=gs.EPS)
-    assert_allclose(free_path[-1], qpos, tol=gs.EPS)
+    assert_allclose(avoidance_path[0].cpu(), 0, tol=gs.EPS)
+    assert_allclose(avoidance_path[-1].cpu(), qpos.cpu(), tol=gs.EPS)
```
duburcqa: You can remove `.cpu()` while you are here. No longer needed.

tests/test_rigid_physics.py
duburcqa: As a reminder:
> Modify test_path_planning_avoidance unit test to leverage batching

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,981 @@
...
+def rrt_valid_mask(tensor: torch.Tensor) -> torch.Tensor:
+    """
+    Returns valid mask of the rrt connect result node indicies
+
+    Parameters
+    ----------
+    tensor: torch.Tensor
+        path tensor in [N, B]
+    """
+    mask = tensor > 0
```
duburcqa: Just do `(tensor > 0.0).to(gs.tc_float)` directly.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,981 @@
...
+    mask_float = mask.float().T.unsqueeze(1)
```
duburcqa: This implementation looks very inefficient and convoluted. Could you explain what I does? It should be possible to write it in a better way.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,981 @@
...
+    Returns valid mask of the rrt connect result node indicies
+
+    Parameters
+    ----------
+    tensor: torch.Tensor
+        path tensor in [N, B]
+    """
+    mask = tensor > 1
+    mask_float = mask.float().T.unsqueeze(1)
+    kernel = torch.ones(1, 1, 3, device=tensor.device, dtype=gs.tc_float)
```
duburcqa, line 1075: same.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,981 @@
...
+    mask: torch.Tensor
+        the masking of path, indicating active waypoints
+    num_points: int
+        the number of the desired waypoints
+
+    Returns
+    -------
+        A new 2D PyTorch tensor [num_points, B, Dof]
+    """
+    res = torch.zeros(path.shape[1], num_points, path.shape[-1], device=gs.device)
```
duburcqa: All these transpose everywhere... You should be able to avoid the last one at least.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,981 @@
...
+    for i_b in range(path.shape[1]):
+        if mask[:, i_b].sum() == 0:
+            continue
+        res[i_b] = torch.nn.functional.interpolate(
```
duburcqa: First compute the return arguments, and do any manipulation before storing the result in `res` separately.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,981 @@
...
+        if self._solver.n_envs > 0:
+            self._entity.set_qpos(qpos_cur, envs_idx=envs_idx)
+        else:
+            self._entity.set_qpos(qpos_cur)
+
+        if is_plan_with_obj:
+            self.update_object(ee_link_idx, obj_link_idx, T_robot_obj, envs_idx)
+
+        if not is_valid.all():
+            gs.logger.warning(f"rrt connect planning failed in {int((~is_valid).sum().cpu())} environments")
```
duburcqa: Create a variable for the total number of failure.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,981 @@
...
+                    T_robot_object=T_robot_obj,
+                )
+            else:
+                sol = self.shortcut_path(
+                    torch.ones_like(sol[..., 0]), sol, iterations=10, ignore_geom_pairs=unique_pairs, envs_idx=envs_idx
+                )
+        sol = align_weypoints_length(sol, torch.ones_like(sol[..., 0]).bool(), num_waypoints)
+
+        if ignore_collision:
+            is_valid = ~is_active
```
duburcqa: Remove `is_valid` and rather define `is_invalid`

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,981 @@
...
+        else:
+            if is_plan_with_obj:
+                is_valid = ~torch.logical_or(
```
duburcqa: `is_invalid = torch.`

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,981 @@
...
+                        is_plan_with_obj=is_plan_with_obj,
+                        obj_geom_start=obj_geom_start,
+                        obj_geom_end=obj_geom_end,
+                        ee_link_idx=ee_link_idx,
+                        obj_link_idx=obj_link_idx,
+                        T_robot_object=T_robot_obj,
+                    ),
+                )
+            else:
+                is_valid = ~torch.logical_or(
```
duburcqa: `is_valid = torch.`

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,981 @@
...
+                )
+            else:
+                sol = self.shortcut_path(
+                    torch.ones_like(sol[..., 0]), sol, iterations=10, ignore_geom_pairs=unique_pairs, envs_idx=envs_idx
+                )
+
+        sol = align_weypoints_length(sol, torch.ones_like(sol[..., 0]).bool(), num_waypoints)
+
+        if ignore_collision:
+            is_valid = ~is_active
```
duburcqa: `is_invalid`

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1588,184 +1590,84 @@ def plan_path(
...
-                        gs.logger.warning("Path planning did not converge. Returning approximation path.")
-                    else:
-                        gs.logger.warning("Path planning did not converge. Returning empty path.")
-                    break
-            gs.logger.info("Path solution found successfully.")
-            break
+        for i in range(max_retry):
+            if not is_valid.all():
+                gs.logger.info(f"planning failed. retrying for {(~is_valid).sum()} environments")
+                retry_path, retry_is_valid = p.plan(
```
duburcqa: Just return `is_invalid` directly. Makes more sense.

genesis/engine/entities/rigid_entity/rigid_entity.py
Kashu7100: what is the best way to implement the timeont with python? I thought it might require threading or something and over complicate the code

duburcqa: I should have a closer look to your implementation. But yes if threading is needed, but not support it. But since IK is iterative, it should be easy to check the elapsed time after each iteration and abort if taking too long. That would not be exact timeout but who cares?

genesis/engine/entities/rigid_entity/rigid_entity.py
Kashu7100: `with_entity` has to be single body non-articulated object for now. And to track the robot, we need a reference link in the robot when we want to use `with_entity`

genesis/utils/path_planing.py
```diff
@@ -931,23 +941,24 @@ def align_weypoints_length(
...
         the number of the desired waypoints
 
     Returns
     -------
         A new 2D PyTorch tensor [num_points, B, Dof]
     """
-    res = torch.zeros(path.shape[1], num_points, path.shape[-1], device=gs.device)
-    for i_b in range(path.shape[1]):
+    t_path = path.permute(1, 2, 0) # [B, Dof, N]
+    res = torch.zeros(num_points, t_path.shape[0], t_path.shape[1], device=gs.device) # [num_points, B, Dof]
```
duburcqa: `torch.zeros((num_points, t_path.shape[0], t_path.shape[1]), dtype=gs.tc_float, device=gs.device)`

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
...
+    mask: torch.Tensor
+        the masking of path, indicating active waypoints [N, B]
+    num_points: int
+        the number of the desired waypoints
+
+    Returns
+    -------
+        A new 2D PyTorch tensor [num_points, B, Dof]
+    """
+    t_path = path.permute(1, 2, 0) # [B, Dof, N]
```
duburcqa: What about doing `t_path = path.permute(1, 0, 2)`? This way you should be able to remove transpose in
`res[:, i_b] = interpolated_path`? Not a big deal though. It is already readable enough.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
...
+    res = torch.zeros(num_points, t_path.shape[0], t_path.shape[1], device=gs.device) # [num_points, B, Dof]
+    for i_b in range(t_path.shape[0]):
+        if mask[:, i_b].sum() == 0:
```
duburcqa: `if not mask[:, i_b].any():`

genesis/utils/path_planing.py
Kashu7100: In this way, we need to put transpose inside the interporate() so maybe we can keep as is if it is ok!

duburcqa: Ah yes I see what you mean. Let's do this.

genesis/engine/entities/rigid_entity/rigid_entity.py
duburcqa: I see. Makes sense.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
...
+        if ee_link_idx is not None and obj_entity is not None:
+            is_plan_with_obj = True
+            obj_geom_start = obj_entity.geom_start
+            obj_geom_end = obj_entity.geom_end
+            obj_link_idx = obj_entity._links[0].idx
+            T_robot_obj = self.get_link_transformation(ee_link_idx, obj_link_idx)[envs_idx]
+
+        self._init_rrt_fields(max_nodes=max_nodes, max_step_size=resolution)
+        self._reset_rrt_fields()
+        self._kernel_rrt_init(qpos_start.contiguous(), qpos_goal.contiguous(), envs_idx)
```
duburcqa: Just make sure that `format_input` returns contiguous tensors in the first place.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
...
+        if qpos_start is None:
+            qpos_start = qpos_cur.clone()
+            if self._solver.n_envs == 0:
+                qpos_start = qpos_start.unsqueeze(0)
+        else:
+            if qpos_start.ndim == 1:
+                qpos_start = qpos_start.unsqueeze(0)
+            assert qpos_start.ndim == 2
+            if qpos_start.shape[0] == 1:
+                qpos_start = qpos_start.repeat(len(envs_idx), 1)
```
duburcqa: rather use `torch.expand`

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
...
+
+        if qpos_goal.ndim == 1:
+            qpos_goal = qpos_goal.unsqueeze(0)
+        assert qpos_goal.ndim == 2
+        if qpos_goal.shape[0] == 1:
+            qpos_goal = qpos_goal.repeat(len(envs_idx), 1)
```
duburcqa: same.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
...
+        return qpos_cur, qpos_goal, qpos_start, envs_idx
```
duburcqa, line 89: You should probably check that the shape of all the tensors are valid before returning.

duburcqa: At some point, you had `gs.raise_exception("Invalid shape for 'qpos_start' or 'qpos_goal'.")` but it has been removed. I'm afraid the exception may not be clear enough for the user without checking for size explicitly.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
...
+            self._solver.set_base_links_pos(trans, obj_link_idx, envs_idx=envs_idx)
+            self._solver.set_base_links_quat(quat, obj_link_idx, envs_idx=envs_idx)
+        else:
+            self._solver.set_base_links_pos(trans, obj_link_idx)
+            self._solver.set_base_links_quat(quat, obj_link_idx)
+
+    # ------------------------------------------------------------------------------------
+    # ------------------------------ util funcs ------------------------------------------
+    # ------------------------------------------------------------------------------------
+    def format_input(self, qpos_cur, qpos_goal, qpos_start, envs_idx):
```
duburcqa, line 77: As I said, it should rather be called `_sanitize_[...]` for consistency with the whole codebase.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
...
+        path: torch.Tensor
+            the [N,B,Dof] tensor containing batched waypoints
+        iterations: int
+            the number of refine iterations
+        """
+        for i in range(iterations):
+            ind = torch.multinomial(path_mask.T, 2).sort()[0]  # B, 2
+            ind_mask = (ind[:, 1] - ind[:, 0]) > 1
+            result_path = path.clone()
+            self.interpolate_path(result_path.contiguous(), path.contiguous(), ind.contiguous(), ind_mask)
```
duburcqa: `self.interpolate_path(result_path, path.contiguous(), ind, ind_mask)`

duburcqa: As I already said, move return arguments at the end, not the beginning.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
...
+            valid waypoint mask [N,B] for the obtained path
+        path: torch.Tensor
+            the [N,B,Dof] tensor containing batched waypoints
+        iterations: int
+            the number of refine iterations
+        """
+        for i in range(iterations):
+            ind = torch.multinomial(path_mask.T, 2).sort()[0]  # B, 2
+            ind_mask = (ind[:, 1] - ind[:, 0]) > 1
+            result_path = path.clone()
```
duburcqa: I would prefer `torch.empty_like(path, dtype=gs.tc_float, device=gs.device)`

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
...
+                    random_sample = ti.Vector(
+                        [self._rrt_goal_configuration[i_q, i_b] for i_q in range(self._entity.n_qs)]
+                    )
+
+                # find nearest neighbor
+                nearest_neighbor_idx = -1
+                nearest_neighbor_dist = gs.ti_float(1e30)
+                ti.loop_config(serialize=self._solver._para_level < gs.PARA_LEVEL.ALL)
+                for i_n in range(self._rrt_tree_size[i_b]):
+                    dist = (self._rrt_node_info.configuration[i_n, i_b] - random_sample).norm()
```
duburcqa: Use `norm_sqr` and only take the root at the end. This would be more efficient.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1548,35 +1546,40 @@ def _kernel_forward_kinematics(
...
         Plan a path from `qpos_start` to `qpos_goal`.
 
         Parameters
         ----------
         qpos_goal : array_like
-            The goal state.
+            The goal state. [B, Nq] or [1, Nq]
         qpos_start : None | array_like, optional
-            The start state. If None, the current state of the rigid entity will be used. Defaults to None.
+            The start state. If None, the current state of the rigid entity will be used. Defaults to None. [B, Nq] or [1, Nq]
```
duburcqa: 120 chars line width limit.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1588,184 +1591,84 @@ def plan_path(
...
+        if ee_link_name is not None:
+            assert with_entity is not None, "`with_entity` must be specified."
+            ee_link_idx = self.get_link(ee_link_name).idx
+        if with_entity is not None:
+            assert ee_link_name is not None, "reference link of the robot must be specified."
+            assert len(with_entity.links) == 1, "only non-articulated object is supported for now."
+
+        match planner:
+            case "RRT":
+                p = RRT(self)
```
duburcqa: `p` is not a good name. What about `planner_obj` or something like this?

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1588,184 +1591,84 @@ def plan_path(
...
+            case "RRTConnect":
+                p = RRTConnect(self)
+            case _:
+                gs.raise_exception(f"invalid planner {planner} specified.")
+
+        path, is_invalid = p.plan(
```
duburcqa: Remove this to only rely on the for-loop right below.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1588,184 +1591,84 @@ def plan_path(
...
-                    continue
-                else:
-                    if waypoints:
-                        gs.logger.warning("Path planning did not converge. Returning approximation path.")
-                    else:
-                        gs.logger.warning("Path planning did not converge. Returning empty path.")
-                    break
-            gs.logger.info("Path solution found successfully.")
-            break
+        for i in range(max_retry):
```
duburcqa: Init `is_invalid` right before the for-loop.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1588,184 +1591,84 @@ def plan_path(
...
+                    max_nodes=max_nodes,
+                    smooth_path=smooth_path,
+                    num_waypoints=num_waypoints,
+                    ignore_collision=ignore_collision,
+                    envs_idx=envs_idx,
+                    ee_link_idx=ee_link_idx,
+                    obj_entity=with_entity,
+                )
+                path[:, is_invalid] = retry_path[:, is_invalid]
+                is_invalid = torch.logical_and(is_invalid, retry_is_invalid)
```
duburcqa: `is_invalid &= retry_is_invalid`

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1588,184 +1591,84 @@ def plan_path(
...
+                    timeout=timeout,
+                    max_nodes=max_nodes,
+                    smooth_path=smooth_path,
+                    num_waypoints=num_waypoints,
+                    ignore_collision=ignore_collision,
+                    envs_idx=envs_idx,
+                    ee_link_idx=ee_link_idx,
+                    obj_entity=with_entity,
+                )
+                path[:, is_invalid] = retry_path[:, is_invalid]
```
duburcqa, line 1741: Add a comment to explain what this does. It took me too long to understand by myself.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
...
+                gs.raise_exception("planning for the gs.JOINT_TYPE.FREE is not supported (yet)")
+            elif joint.type == gs.JOINT_TYPE.SPHERICAL:
+                gs.raise_exception("planning for the gs.JOINT_TYPE.SPHERICAL is not supported (yet)")
+
+    @abstractmethod
+    def plan(
+        self,
+        qpos_goal,
+        qpos_start=None,
+    ): ...
```
duburcqa, line 28: Skip line.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
...
+
+    def get_link_transformation(self, robot_g_link_idx, obj_g_link_idx):
```
duburcqa: `get_link_pose`

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
...
+        """
+        get transformation from robot link to the object link
```
duburcqa: This formatting is messed up. Skip line after short description, start sentences with uppercase and finish with dot:
```python
Get transformation from robot link to the object link.

robot_g_link_idx: int
    Global link idx of the link of the robot.
[...]
```

duburcqa: All, the description is not clear:
`Get the relative pose of a given robot link wrt some object link as an homogeneous transformation matrix.`

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
...
+            global link idx of the link of the robot
+        obj_g_link_idx: int
+            global link idx of the base link of the object
+        """
+        robot_T = gg.trans_quat_to_T(
+            self._solver.get_links_pos(links_idx=robot_g_link_idx),
+            self._solver.get_links_quat(links_idx=robot_g_link_idx),
+        )
+        obj_T = gg.trans_quat_to_T(
+            self._solver.get_links_pos(links_idx=obj_g_link_idx), self._solver.get_links_quat(links_idx=obj_g_link_idx)
```
duburcqa: Skip line for style consistency.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
+from abc import ABC, abstractmethod
```
duburcqa, line 1: Can you fix all import ordering? Here is the rule: You should first import standards libraries, skip line, then ubiquitous third-party libraries (torch, numpy, tqdm...etc), skip line, then random third-party libraries, skip line, then our own custom libraries, skip line, then relative genesis imports. For each group, make sure it is sorted alphabetically.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
...
+import taichi as ti
+import torch
+import torch.nn.functional as F
+import time
+import genesis as gs
+import genesis.utils.geom as gg
```
duburcqa: `import genesis.utils.geom as gu` for consistency.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
...
+            global link idx of the base link of the object
+        """
+        robot_T = gg.trans_quat_to_T(
+            self._solver.get_links_pos(links_idx=robot_g_link_idx),
+            self._solver.get_links_quat(links_idx=robot_g_link_idx),
+        )
+        obj_T = gg.trans_quat_to_T(
+            self._solver.get_links_pos(links_idx=obj_g_link_idx), self._solver.get_links_quat(links_idx=obj_g_link_idx)
+        )
+        T_robot_object = torch.linalg.inv(robot_T) @ obj_T
```
duburcqa: Don't do this. This is inefficient. We should add some optimised `inv_T` geom helper:
```python
def inv_T(T):
    if isinstance(T, torch.Tensor):
        T_inv = torch.empty_like(T)
    elif isinstance(T, np.ndarray):
        T_inv = np.empty_like(T)
    else:
        gs.raise_exception(
            f"the input must be torch.Tensor or np.ndarray. got: {type(T)=}"
        )
    
    trans, R = T[..., :3, 3], T[..., :3, :3]
    T_inv[..., 3, 3] = 1.0
    T_inv[..., :3, 3] = - R.T @ trans
    T_inv[..., :3, :3] = R.T

    return T_inv
```

I will make a PR for it.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
...
+            robot_T = gg.trans_quat_to_T(
+                self._solver.get_links_pos(ee_link_idx, envs_idx=envs_idx),
+                self._solver.get_links_quat(ee_link_idx, envs_idx=envs_idx),
+            )
+        else:
+            robot_T = gg.trans_quat_to_T(
+                self._solver.get_links_pos(ee_link_idx), self._solver.get_links_quat(ee_link_idx)
+            )
+        obj_T = robot_T @ T_robot_object
+        trans, quat = gg.T_to_trans_quat(obj_T)
```
duburcqa: Just manipulate `trans, quat` from the start and use `transform_pos_quat_by_trans_quat`

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
...
+            self._solver.get_links_pos(links_idx=robot_g_link_idx),
+            self._solver.get_links_quat(links_idx=robot_g_link_idx),
+        )
+        obj_T = gg.trans_quat_to_T(
+            self._solver.get_links_pos(links_idx=obj_g_link_idx), self._solver.get_links_quat(links_idx=obj_g_link_idx)
+        )
+        T_robot_object = torch.linalg.inv(robot_T) @ obj_T
+        return T_robot_object
+
+    def update_object(self, ee_link_idx, obj_link_idx, T_robot_object, envs_idx):
```
duburcqa: I don't think storing pose in `T_robot_object` is a good idea. Just keep them as `_pos, _quat` pair. It would be more efficient and consistent with the codebase.

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
...
+                qpos_start = qpos_start.repeat(len(envs_idx), 1)
+
+        if qpos_goal.ndim == 1:
+            qpos_goal = qpos_goal.unsqueeze(0)
+        assert qpos_goal.ndim == 2
+        if qpos_goal.shape[0] == 1:
+            qpos_goal = qpos_goal.repeat(len(envs_idx), 1)
+        return qpos_cur, qpos_goal, qpos_start, envs_idx
+
+    def get_exclude_geom_pairs(self, qpos_goal, qpos_start, envs_idx):
```
duburcqa: I would prefer to avoid duplicating everything for `start` and `goal`. What about a function that generates the list of excluded pairs for some `qpos`? Then you call it twice and merge the result? For even better, you accept passing a list of configurations as input arguments: `def get_exclude_geom_pairs(self, qposs, envs_idx):`

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
...
+        else:
+            self._entity.set_qpos(qpos_goal[0])
+        self._solver._kernel_detect_collision()
+        scene_contact_info = self._entity.get_contacts()
+        geom_a_goal = scene_contact_info["geom_a"]
+        geom_b_goal = scene_contact_info["geom_b"]
+        if self._solver.n_envs > 0:
+            valid_mask = scene_contact_info["valid_mask"]
+            geom_a_goal = geom_a_goal[valid_mask]
+            geom_b_goal = geom_b_goal[valid_mask]
```
duburcqa: This seems broken, you were doing nothing with `geom_a_goal`

genesis/utils/path_planing.py
```diff
@@ -0,0 +1,995 @@
...
+        tensor: ti.types.ndarray(),  # [N, B, Dof]
+        path: ti.types.ndarray(),  # [N, B, Dof]
+        sample_ind: ti.types.ndarray(),  # [B, 2]
+        mask: ti.types.ndarray(),  # [B]
+    ):
+        ti.loop_config(serialize=self._solver._para_level < gs.PARA_LEVEL.ALL)
+        for i_b in range(path.shape[1]):
+            if not mask[i_b]:
+                continue
+            num_samples = sample_ind[i_b, 1] - sample_ind[i_b, 0]
```
duburcqa: Compute `num_samples_inv = 1.0 / (sample_ind[i_b, 1] - sample_ind[i_b, 0])` directly for efficiency.

Kashu7100: Not sure if this is a good idea because I use num_samples in the range() below.

genesis/utils/path_planing.py
Kashu7100: Do you mean replace `<CODE>.norm()` by `ti.sqrt(<CODE>.norm_sqr())`?

genesis/utils/path_planing.py
Kashu7100: I would like to keep this since qpos_start can be None.

genesis/utils/path_planing.py
Kashu7100: fixed!

genesis/utils/path_planing.py
Kashu7100: changed to trans, quat operations

genesis/utils/path_planing.py
duburcqa: I see. You are right.

genesis/utils/path_planing.py
duburcqa: I don't get your point. Then do not add `qpos_start` to the list if it is None.

genesis/utils/path_planing.py
duburcqa: No. In pseudo-code:
```python
dist2 = inf
idx = -1
for i in range(N):
    dist2_i = vecs[i].norm_sqr()
    if dist2_i < dist2:
        dist2 = dist2_i
        idx = i
dist = ti.sqrt(dist2)
```
This way, you only take the square root once.

genesis/utils/path_planing.py
Kashu7100: My mistake, never mind. I can put `qposs = [qpos_goal, qpos_start]` but it might be dangerous because there is no assurance of which goes first (`[qpos_goal, qpos_start]` vs `[qpos_start, qpos_goal]`)

duburcqa: Why the order should matter here?

genesis/utils/path_planing.py
Kashu7100: I think `path.clone()` is necessary since it uses the `path` value directly as result when the sampled inds are next to each other.

duburcqa: This is not a good API. Result should only be used to store the result (i.e. write only and not read/write). If the sampled inds are next to each other, then just copy the data from path to result_path.

The other option would be to use `path` as input and output argument, and completely get rid of `result_path`, then you should keep your clone and only pass `result_path` as argument in practice.

genesis/utils/path_planing.py
Kashu7100: Simplified the function based on @YilingQiao suggestion. It needs to be goal -> start as this function set_qpos inside and it is desirable to be in the start config when exiting this function.

duburcqa: Setting the initial configuration should not be the responsibility of this method. And I’m not convinced this tiny optimization is worth it.

duburcqa: (Even if we agree this method should call set_qpos internally of course)

Kashu7100: So what you suggest is change to `qposs` and run set_qpos for the initial pose outside again?

Kashu7100:
Seems `test_inverse_kinematics` is failing, which is not the part of this PR.

genesis/utils/path_planing.py
duburcqa: Yes!

duburcqa:
> Seems test_inverse_kinematics is failing, which is not the part of this PR.

Actually our IK test is very unit, and also relies on path planning. So your changes are having any impact on this test!

YilingQiao (approved):

Kashu7100:
updated the unit test for batching, addressed all comments.

YilingQiao:
Thank you @duburcqa for the reviews! As all comments are addressed, I think it's good to merge once CI passes.

PR #1321, opened 2025-06-26
[MISC] Add stub code for mouse interaction

gasnica:
## Description
Add ViewerInteraction class that hooks into pyrender/viewer with callbacks that handle chosen mouse events, and an on_draw() callback.

Currently, when activated, this only displays an arrow indicating a mouse ray hit on the ground plane.

To activate, set ViewerOptions.use_viewer_interaction = True

## Motivation and Context
This is work leading to enabling mouse-click selection of objects, and dragging them.

Only manual interactive tests.

image

## Checklist:

### DIFF
--- a/genesis/ext/pyrender/interaction/ray.py
+++ b/genesis/ext/pyrender/interaction/ray.py
@@ -1,3 +1,4 @@
+from dataclasses import dataclass
 from genesis.ext.pyrender.interaction.vec3 import Vec3
 
 class Ray:
@@ -6,6 +7,42 @@ class Ray:
 
     def __init__(self, origin: Vec3, direction: Vec3):
         self.origin = origin
-        self.direction = direction
+        self.direction = direction.normalized()
 
     def __repr__(self) -> str: return f"Ray(origin={self.origin}, direction={self.direction})"
+
+
+@dataclass
+class RayHit:
+    is_hit: bool
+    distance: float    
+    normal: Vec3
+    position: Vec3
+    object_idx: int
+
+
+class Plane:
+    normal: Vec3
+    distance: float  # distance from plane to origin along normal
+
+    def __init__(self, normal: Vec3, point: Vec3):
+        self.normal = normal
+        self.distance = -normal.dot(point)
+
+    def raycast(self, ray: Ray) -> RayHit:
+        dot = ray.direction.dot(self.normal)
+        dist = ray.origin.dot(self.normal) + self.distance
+
+        if -1e-6 < dot or dist < 1e-6:
+            return RayHit(is_hit=False, distance=0, normal=Vec3.zero(), position=Vec3.zero(), object_idx=-1)
+
+        dist_along_ray = dist / -dot
+
+        return RayHit(
+            is_hit=True,
+            distance=dist_along_ray,
+            normal=self.normal,
+            position=ray.origin + ray.direction * dist_along_ray,
+            object_idx=0
+        )
+
--- a/genesis/ext/pyrender/interaction/ray.py
+++ b/genesis/ext/pyrender/interaction/ray.py
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
-from genesis.ext.pyrender.interaction.vec3 import Vec3
+
+from .vec3 import Vec3
 
 
 EPSILON = 1e-6
--- a/genesis/ext/pyrender/interaction/ray.py
+++ b/genesis/ext/pyrender/interaction/ray.py
@@ -1,10 +1,9 @@
 from dataclasses import dataclass
-
-import numpy as np
-
 from genesis.ext.pyrender.interaction.vec3 import Vec3
 
-EPSILON32: np.float32 = np.float32(1e-6)
+
+EPSILON = 1e-6
+
 
 class Ray:
     origin: Vec3
@@ -21,7 +20,7 @@ class Ray:
 @dataclass
 class RayHit:
     is_hit: bool
-    distance: np.float32
+    distance: float
     normal: Vec3
     position: Vec3
     object_idx: int
@@ -29,7 +28,7 @@ class RayHit:
 
 class Plane:
     normal: Vec3
-    distance: np.float32  # distance from plane to origin along normal
+    distance: float  # distance from plane to origin along normal
 
     def __init__(self, normal: Vec3, point: Vec3):
         self.normal = normal
@@ -39,8 +38,8 @@ class Plane:
         dot = ray.direction.dot(self.normal)
         dist = ray.origin.dot(self.normal) + self.distance
 
-        if -EPSILON32 < dot or dist < EPSILON32:
-            return RayHit(is_hit=False, distance=np.float32(0), normal=Vec3.zero(), position=Vec3.zero(), object_idx=-1)
+        if -EPSILON < dot or dist < EPSILON:
+            return RayHit(is_hit=False, distance=0, normal=Vec3.zero(), position=Vec3.zero(), object_idx=-1)
 
         dist_along_ray = dist / -dot
 
--- a/genesis/ext/pyrender/interaction/ray.py
+++ b/genesis/ext/pyrender/interaction/ray.py
@@ -1,5 +1,10 @@
 from dataclasses import dataclass
+
+import numpy as np
+
 from genesis.ext.pyrender.interaction.vec3 import Vec3
+
+EPSILON32: np.float32 = np.float32(1e-6)
 
 class Ray:
     origin: Vec3
@@ -16,7 +21,7 @@ class Ray:
 @dataclass
 class RayHit:
     is_hit: bool
-    distance: float    
+    distance: np.float32
     normal: Vec3
     position: Vec3
     object_idx: int
@@ -24,7 +29,7 @@ class RayHit:
 
 class Plane:
     normal: Vec3
-    distance: float  # distance from plane to origin along normal
+    distance: np.float32  # distance from plane to origin along normal
 
     def __init__(self, normal: Vec3, point: Vec3):
         self.normal = normal
@@ -34,8 +39,8 @@ class Plane:
         dot = ray.direction.dot(self.normal)
         dist = ray.origin.dot(self.normal) + self.distance
 
-        if -1e-6 < dot or dist < 1e-6:
-            return RayHit(is_hit=False, distance=0, normal=Vec3.zero(), position=Vec3.zero(), object_idx=-1)
+        if -EPSILON32 < dot or dist < EPSILON32:
+            return RayHit(is_hit=False, distance=np.float32(0), normal=Vec3.zero(), position=Vec3.zero(), object_idx=-1)
 
         dist_along_ray = dist / -dot
 
--- a/genesis/ext/pyrender/interaction/ray.py
+++ b/genesis/ext/pyrender/interaction/ray.py
@@ -1,0 +1,11 @@
+from genesis.ext.pyrender.interaction.vec3 import Vec3
+
+class Ray:
+    origin: Vec3
+    direction: Vec3
+
+    def __init__(self, origin: Vec3, direction: Vec3):
+        self.origin = origin
+        self.direction = direction
+
+    def __repr__(self) -> str: return f"Ray(origin={self.origin}, direction={self.direction})"
--- a/genesis/ext/pyrender/interaction/ray.py
+++ b/genesis/ext/pyrender/interaction/ray.py
@@ -9,7 +9,8 @@ class Ray:
         self.origin = origin
         self.direction = direction.normalized()
 
-    def __repr__(self) -> str: return f"Ray(origin={self.origin}, direction={self.direction})"
+    def __repr__(self) -> str:
+        return f"Ray(origin={self.origin}, direction={self.direction})"
 
 
 @dataclass
--- a/genesis/ext/pyrender/interaction/vec3.py
+++ b/genesis/ext/pyrender/interaction/vec3.py
@@ -11,8 +11,8 @@ class Vec3:
 
     def __add__(self, other: 'Vec3') -> 'Vec3': return Vec3(self.v + other.v)
     def __sub__(self, other: 'Vec3') -> 'Vec3': return Vec3(self.v - other.v)
-    def __mul__(self, other: float) -> 'Vec3': return Vec3(self.v * np.float32(other))
-    def __rmul__(self, other: float) -> 'Vec3': return Vec3(self.v * np.float32(other))
+    def __mul__(self, other: np.float32) -> 'Vec3': return Vec3(self.v * np.float32(other))
+    def __rmul__(self, other: np.float32) -> 'Vec3': return Vec3(self.v * np.float32(other))
     def dot(self, other: 'Vec3') -> np.float32: return np.dot(self.v, other.v)
     def cross(self, other: 'Vec3') -> 'Vec3': return Vec3(np.cross(self.v, other.v))
 
--- a/genesis/ext/pyrender/interaction/vec3.py
+++ b/genesis/ext/pyrender/interaction/vec3.py
@@ -11,9 +11,9 @@ class Vec3:
 
     def __add__(self, other: 'Vec3') -> 'Vec3': return Vec3(self.v + other.v)
     def __sub__(self, other: 'Vec3') -> 'Vec3': return Vec3(self.v - other.v)
-    def __mul__(self, other: np.float32) -> 'Vec3': return Vec3(self.v * np.float32(other))
-    def __rmul__(self, other: np.float32) -> 'Vec3': return Vec3(self.v * np.float32(other))
-    def dot(self, other: 'Vec3') -> np.float32: return np.dot(self.v, other.v)
+    def __mul__(self, other: float) -> 'Vec3': return Vec3(self.v * np.float32(other))
+    def __rmul__(self, other: float) -> 'Vec3': return Vec3(self.v * np.float32(other))
+    def dot(self, other: 'Vec3') -> float: return np.dot(self.v, other.v).item()
     def cross(self, other: 'Vec3') -> 'Vec3': return Vec3(np.cross(self.v, other.v))
 
     def normalized(self) -> 'Vec3': return Vec3(self.v / (np.linalg.norm(self.v) + 1e-24))
--- a/genesis/ext/pyrender/interaction/vec3.py
+++ b/genesis/ext/pyrender/interaction/vec3.py
@@ -6,13 +6,14 @@ class Vec3:
 
     def __init__(self, v: NDArray[np.float32]):
         assert v.shape == (3,), f"Vec3 must be initialized with a 3-element array, got {v.shape}"
+        assert v.dtype == np.float32, f"Vec3 must be initialized with a float32 array, got {v.dtype}"
         self.v = v
 
     def __add__(self, other: 'Vec3') -> 'Vec3': return Vec3(self.v + other.v)
     def __sub__(self, other: 'Vec3') -> 'Vec3': return Vec3(self.v - other.v)
-    def __mul__(self, other: float) -> 'Vec3': return Vec3(self.v * other)
-    def __rmul__(self, other: float) -> 'Vec3': return Vec3(self.v * other)
-    def dot(self, other: 'Vec3') -> float: return np.dot(self.v, other.v)
+    def __mul__(self, other: float) -> 'Vec3': return Vec3(self.v * np.float32(other))
+    def __rmul__(self, other: float) -> 'Vec3': return Vec3(self.v * np.float32(other))
+    def dot(self, other: 'Vec3') -> np.float32: return np.dot(self.v, other.v)
     def cross(self, other: 'Vec3') -> 'Vec3': return Vec3(np.cross(self.v, other.v))
 
     def normalized(self) -> 'Vec3': return Vec3(self.v / (np.linalg.norm(self.v) + 1e-24))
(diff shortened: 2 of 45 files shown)


genesis/utils/tools.py
```diff
@@ -65,6 +65,8 @@ def reset(self):
         self.just_reset = True
         if self.level == 0 and not self.skip:
             print("─" * os.get_terminal_size()[0])
+        if self.ti_sync:
```
duburcqa: Please rebase on main.

genesis/options/vis.py
```diff
@@ -41,6 +41,7 @@ class ViewerOptions(Options):
     camera_lookat: tuple = (0.0, 0.0, 0.5)
     camera_up: tuple = (0.0, 0.0, 1.0)
     camera_fov: float = 40
+    use_viewer_interaction: bool = False
```
duburcqa: Why not just `enable_interaction`? It is actually the same viewer no?

genesis/ext/pyrender/viewer.py
```diff
@@ -376,6 +381,20 @@ def __init__(
         self.scene.main_camera_node = self._camera_node
         self._reset_view()
 
+        # Setup mouse interaction
+
+        # Note: can't import ViewerIntaction at file scope while genesis.engine.scene is still being imported
+        from genesis.ext.pyrender.interaction.viewer_interaction import ViewerInteraction
```
duburcqa: Just use relative import and move them back on top (after all non-relative imports). It should work:
```python
from .interaction.viewer_interaction import ViewerInteraction
from .interaction.viewer_interaction_base import ViewerInteractionBase
```

genesis/ext/pyrender/viewer.py
```diff
@@ -64,6 +64,10 @@
 
 pyglet.options["shadow_window"] = False
 
+# Used for type checking:
+from pyglet.event import EVENT_HANDLE_STATE
```
duburcqa: Move this import after `from pyglet import clock`

genesis/ext/pyrender/viewer.py
```diff
@@ -64,6 +64,10 @@
 
 pyglet.options["shadow_window"] = False
 
+# Used for type checking:
```
duburcqa: You can remove this comment

genesis/ext/pyrender/viewer.py
```diff
@@ -64,6 +64,10 @@
...
+from pyglet.event import EVENT_HANDLE_STATE
+from genesis.vis.rasterizer_context import RasterizerContext
```
duburcqa: Move this import before relative imports.

Here is the rule: You should first import standards libraries, skip line, then ubiquitous third-party libraries (torch, numpy, tqdm...etc), skip line, then random third-party libraries, skip line, then our own custom libraries, skip line, then relative genesis imports. For each group, make sure it is sorted alphabetically.

genesis/ext/pyrender/interaction/viewer_interaction_base.py
```diff
@@ -0,0 +1,34 @@
...
+    """Base class for handling pyglet.window.Window events.
+    """
+
+    log_events: bool
+
+    def __init__(self, log_events: bool = False):
+        self.log_events = log_events
+
+    def on_mouse_motion(self, x: int, y: int, dx: int, dy: int) -> EVENT_HANDLE_STATE:
+        if self.log_events: print(f"Mouse moved to {x}, {y}")
```
duburcqa: Use `gs.logger`. No in-line `if` statements.

genesis/ext/pyrender/interaction/viewer_interaction_base.py
```diff
@@ -0,0 +1,34 @@
...
+# Notes:
+# Viewer window is based on pyglet.window.Window, mouse events are defined in pyglet.window.BaseWindow
+
+class ViewerInteractionBase():
+    """Base class for handling pyglet.window.Window events.
+    """
+
+    log_events: bool
+
+    def __init__(self, log_events: bool = False):
```
duburcqa, line 13: This flag is unnecessary. Remove it and always log at "DEBUG" level. (See next comment)

genesis/ext/pyrender/interaction/viewer_interaction_base.py
duburcqa: i.e. `gs.logger.debug(f"Mouse moved to {x}, {y}")`

genesis/ext/pyrender/interaction/viewer_interaction.py
```diff
@@ -0,0 +1,115 @@
...
+        return Ray(position, forward)
+
+    def _raycast_against_ground(self, ray: Ray) -> RayHit:
+        ground_plane = Plane(Vec3.from_xyz(0, 0, 1), Vec3.zero())
+        return ground_plane.raycast(ray)
+
+    def _draw_arrow(self, pos: Vec3, dir: Vec3, color: tuple[float, float, float, float] = (1, 1, 1, 1)) -> None:
+        # Only draws arrowhead:
+        self.scene.draw_debug_arrow(
+            pos=pos.v,
```
duburcqa: one line would do fine. No need to split on multiple line in this case it is already very readable.

genesis/ext/pyrender/interaction/viewer_interaction.py
```diff
@@ -0,0 +1,115 @@
...
+        mtx = self.camera.matrix
+        position = Vec3(mtx[:3, 3])
+        forward = Vec3(-mtx[:3, 2])
+        return Ray(position, forward)
+
+    def _raycast_against_ground(self, ray: Ray) -> RayHit:
+        ground_plane = Plane(Vec3.from_xyz(0, 0, 1), Vec3.zero())
+        return ground_plane.raycast(ray)
+
+    def _draw_arrow(self, pos: Vec3, dir: Vec3, color: tuple[float, float, float, float] = (1, 1, 1, 1)) -> None:
```
duburcqa: `tuple[float, float, float, float] = (1, 1, 1, 1)`. Typing is wrong since you specify ints x)

genesis/ext/pyrender/interaction/vec3.py
```diff
@@ -0,0 +1,31 @@
+import numpy as np
+from numpy.typing import NDArray
+
+class Vec3:
```
duburcqa, line 5: I'm not convinced by the relevant of this class. It just add one extra layer of indirection that is hindering performance will not really making code much clearer. This kind of approach is rarely used in Python.

genesis/ext/pyrender/interaction/ray.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def __init__(self, normal: Vec3, point: Vec3):
+        self.normal = normal
+        self.distance = -normal.dot(point)
+
+    def raycast(self, ray: Ray) -> RayHit:
+        dot = ray.direction.dot(self.normal)
+        dist = ray.origin.dot(self.normal) + self.distance
+
+        if -1e-6 < dot or dist < 1e-6:
```
duburcqa: what about using `gs.EPS` ?

genesis/ext/pyrender/interaction/vec3.py
```diff
@@ -0,0 +1,31 @@
...
+        self.v = v
+
+    def __add__(self, other: 'Vec3') -> 'Vec3': return Vec3(self.v + other.v)
+    def __sub__(self, other: 'Vec3') -> 'Vec3': return Vec3(self.v - other.v)
+    def __mul__(self, other: float) -> 'Vec3': return Vec3(self.v * other)
+    def __rmul__(self, other: float) -> 'Vec3': return Vec3(self.v * other)
+    def dot(self, other: 'Vec3') -> float: return np.dot(self.v, other.v)
+    def cross(self, other: 'Vec3') -> 'Vec3': return Vec3(np.cross(self.v, other.v))
+
+    def normalized(self) -> 'Vec3': return Vec3(self.v / (np.linalg.norm(self.v) + 1e-24))
```
duburcqa: Personally I prefer `self.v / max(np.linalg.norm(self.v), gs.EPS)`

genesis/ext/pyrender/interaction/ray.py
```diff
@@ -0,0 +1,48 @@
...
+
+class Ray:
+    origin: Vec3
+    direction: Vec3
+
+    def __init__(self, origin: Vec3, direction: Vec3):
+        self.origin = origin
+        self.direction = direction.normalized()
+
+    def __repr__(self) -> str: return f"Ray(origin={self.origin}, direction={self.direction})"
```
duburcqa: no inlined statements.

gasnica: gotcha -- no one-line functions and no one-line "if xx: then" blocks

genesis/ext/pyrender/interaction/ray.py
gasnica: Seems like gs.EPS is a user-defined param (albeit with a default value). But for that reason I'd pass.

I can add a local Epsilon definition for use in any further raycasting code. I'm also used to 1e-6 ;-) gs.EPS is 1e-7.

genesis/ext/pyrender/interaction/vec3.py
gasnica: Yes, I totally agree that it looks out of place. Possibly we can come up with some good alternative.

My thinking:
- I'd really like to strong type or lint all my code (I already found a function I needed to use that dropped linting in an overridden function, renamed a param name, and hence made it misleading -- needed to spend a moment looking up documentation)
- for typing vectors, I'd like to have a short name Vec3 is shorter than np.array[np.float32], so that fits nicely into function definitions
- in our setup/demo code we create vectors with code such as np.array[0, 1, 2] or np.array[1.0, 2.0, 3.0], which creates arrays of either ints or float64, which are then reallocated, converted to float32 and deallocated. If such vectors are reused then we're paying more pref for converting them to float32 on every use. 
- also there's no checking of size of vectors. I'd like to be able to work clearly with 2d & 3d vectors, and not mix them up. In taichi we have types for 2d & 3d vectors of floats and ints. I'd love to have something that matches that for python scope. 
- though I agree this is akward, so let's come up with something that we like. --- Here I was already trying to be minimal, and this basic implementation reflects what you'd create in c++ or c# (apart from the extra dereference :-) )

genesis/ext/pyrender/interaction/vec3.py
gasnica: Yeah, this is a trick I started using at some point, and it hasn't backfired yet. It is definitely faster. For any vector length greater than epsilon it has zero effect due to numerical accuracy. And for zero vectors it takes care of the problem. You endup with some denormalized length vectors for vectors shorter than 1e-12. 

Using the max(xx, gs.EPS) already produces bad results for vectors shorter than gs.EPS.

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: Thanks for spotting this :-)

Am I thinking correctly -- typing seems fine, it's just the default value that will get converted to tuple of floats when used?

I did that, because this is the format we often use elsewhere, just using ints. 

Let me know again, please, if we'd like to change it to (1.0, 1.0, 1.0, 1.0) ??

genesis/ext/pyrender/interaction/vec3.py
YilingQiao: Hi @duburcqa, we have some type definitions here.
Do you think we could also define something similar, like np_vec3 or np_vec4, that can be used for type annotations?

https://github.com/Genesis-Embodied-AI/Genesis/blob/fd06778cae5790ae2fb2478b19a2746974636e53/genesis/__init__.py#L118

genesis/ext/pyrender/interaction/viewer_interaction_base.py
gasnica: So I just had this originally, to see if and how the events work.

But when those are enabled they will just flood the output console, spitting out a line for every frame a mouse is moved. That's why I think this should have an explicit "enable debug logs" flag independent of system-wide debug settings. Or I can just comment out the code altogether if we don't like it ?

Let me know, please.

genesis/ext/pyrender/viewer.py
gasnica: Moving to after `import genesis as gs`

Thank you for this! I must write down the order -- it makes sense, but there's a lot of categorization :-) Do we have it noted down in some coding style guidelines?

What is our take on relative imports ?? From what I read up on python, some good-practiced discourage use of relative paths altogether. --- And I'm not using it in my new code. Do we enforce its use in our style guidelines though?

genesis/ext/pyrender/viewer.py
gasnica: It does not work ---

ImportError: cannot import name 'Scene' from partially initialized module 'genesis.engine.scene' (most likely due to a circular import) (/home/adrian/Genesis/genesis/engine/scene.py)

genesis/options/vis.py
gasnica: Much better name XD

genesis/utils/tools.py
gasnica: We can't do rebase after creating a PR, can we?? 

Will merge from main and push.

genesis/ext/pyrender/interaction/viewer_interaction.py
```diff
@@ -0,0 +1,115 @@
...
+            projection_matrix = self.camera.camera.get_projection_matrix(*self.viewport_size)
+            x = x - 0.5 * self.viewport_size[0]
+            y = y - 0.5 * self.viewport_size[1]
+            x = 2.0 * x / self.viewport_size[0] / projection_matrix[0, 0]
+            y = 2.0 * y / self.viewport_size[1] / projection_matrix[1, 1]
+
+        # Note: ignoring pixel aspect ratio
+
+        mtx = self.camera.matrix
+        position = Vec3(mtx[:3, 3])
```
gasnica: @YilingQiao @duburcqa On the topic of strong typing.

I just realized, that here, in code, we still use an incorrect dtype, cos camera.matrix returns an array of float64's -- And while the expected type is linted in Vec3 constructor, it was not asserted at runtime.

So, just having an alias for types is not sufficient for guarding data type correctness at this moment. Possibly we could run some strong linter on our code as a required commit check. But currently, I can only rely on runtime checks, afaiu. 

I'll add another classmethod to Vec3 to convert explicitly from an array[np.float64]

genesis/ext/pyrender/interaction/ray.py
duburcqa: > Seems like gs.EPS is a user-defined param

Yeah, It is probably a bad idea actually.

> I can add a local Epsilon definition for use in any further raycasting code. 

Yes that would be nice.

genesis/ext/pyrender/interaction/vec3.py
duburcqa: As you wish, just mathematical purism probably.

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: btw. I just made a simple tests of adding np.arrays with float32 & float64, and the floats seem nearly x2 faster.

Not sure how much that can influence our e.g. debug-drawing code, but a bit maybe.

genesis/ext/pyrender/interaction/vec3.py
gasnica: @duburcqa Alexis, are we good with making exceptions here and using one-line function definitions?

It is actually more readable in this format, imho, where you can just see the slight implementation differences line-to-line. So I think it could qualify as a good exception :-)

genesis/options/vis.py
gasnica: Is there a python tool that helps with name-changing/refactoring? 

In VisualStudio/c#/c++ I can change all occurrences and uses of a variable reliably in entire codebase in one go. Here in Cursor my rename did not touch any of the other uses, and requires manual code edit everywhere.

Would you have any tools or VSCode/Cursor extensions to recommend maybe? I heard good things of pycharm, do you have experience with it? @duburcqa @YilingQiao

gasnica: btw. I noticed that linting types for our functions' parameters allows us to do more reliable and automatic renames & refactors, instead of manual editing of all occurences. And it also right away enables intellisense and lists fields of my types, after I press the dot in my editor. Amazing :-)

In my perspective, such things hugely increase coding efficiency @YilingQiao 

Do your editors support intellisense without linting (and I'm not talking about copilots/cursor's suggestions, which are often-enough incorrect)?

YilingQiao:
```
such things hugely increase coding efficiency
```
yeah, that's an advantage

genesis/ext/pyrender/interaction/viewer_interaction.py
YilingQiao: yeah, (1.0, 1.0, 1.0, 1.0) is better

genesis/ext/pyrender/viewer.py
duburcqa: There must be something wrong. Less discuss this later.

genesis/ext/pyrender/viewer.py
duburcqa: > some good-practiced discourage use of relative paths altogether.

I don't get the point of disabling relative important. That may have been causing issues 10 years ago I don't know, but never faced any issue since then. Moreover, then clarify what are internal lower-level modules from external dependences, which I think is great. Finally, I would expect to help sorting out some circular dependency issues but apparently it does not for some reason, I need to check.

genesis/ext/pyrender/interaction/viewer_interaction_base.py
duburcqa: Is it flooding this much? Anyway debug is only used for...debug :p 

If so, then do both, logger (any level you want, likely info) + flag to disable.

genesis/ext/pyrender/viewer.py
gasnica: From my understanding 
 - when we lint types we need to include the linted type in a module. And we haven't been doing that much.
 - the Scene module imports ViewerInteraction which imports Scene again, before its processing is done, and that is not allowed
 - so I'm postponing import of the problematic ViewerInteraction till after the Scene module is imported, to the moment when its actually needed 
 - one proper solution I can think of is to loosen the coupling between the classes and add some interfaces, but that's more code and a bit of an overengineering maybe at this moment
 - yes, let's talk to see if we can do other things
 - possibly linting types in 'quotes' (like you do withing a class that's being defined) will work and can solve the problem? (haven't tried)
 - but after asking chatGpt, i read its a common practice to delay imports -- and it's also a common practice do to this kind of 'lazy improt' to speed up start up time

genesis/ext/pyrender/viewer.py
gasnica: PEP8 recommends absolute paths: "Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured"

And allows relative: "However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose"

genesis/ext/pyrender/viewer.py
duburcqa: > the Scene module imports ViewerInteraction which imports Scene again, before its processing is done, and that is not allowed

OK so the problem is here ! Just delay import of Scene inside `ViewerInteraction`, and use relative import inside `ViewerInteraction`.

> possibly linting types in 'quotes' (like you do withing a class that's being defined) will work and can solve the problem? (haven't tried)

This is rather the way to go:
```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from genesis.engine.simulator import Simulator
```

> but after asking chatGpt, i read its a common practice to delay imports -- and it's also a common practice do to this kind of 'lazy improt' to speed up start up time

It is, but it must be well-motivated. If it is necessary then it should be avoided.

genesis/ext/pyrender/interaction/vec3.py
duburcqa: > Do you think we could also define something similar, like np_vec3 or np_vec4, that can be used for type annotations?

I guess so!

genesis/ext/pyrender/interaction/vec3.py
duburcqa: > Alexis, are we good with making exceptions here and using one-line function definitions?

I don't think we make any exception. This would be the way to go in C++ definitely, but not in Python I think. This kind of in-lined function is very rarely seen in "modern" Python packages in practice (post 2010).

genesis/ext/pyrender/interaction/viewer_interaction.py
duburcqa: > Not sure how much that can influence our e.g. debug-drawing code, but a bit maybe.

The complete rendering pipeline could (should?) probably be enforced in 32 bits.

genesis/ext/pyrender/viewer.py
gasnica: Oh wonderful !! I needed that :-)

Should I also do the same with the other imports that I added in genesis files purely for typechecking?

gasnica: Oh no -- it does not work. My class does not process when I make Scene/Node import conditional on TYPE_CHECKING:

class ViewerInteraction(ViewerInteractionBase):
    camera: Node
    scene: Scene
    [...]
    def __init__(self, 
        camera: Node, 
        scene: Scene, 
        [...]
        
Both places where Scene or Node is used don't compile.

genesis/ext/pyrender/interaction/vec3.py
gasnica: Expanded all lines and resolving.

genesis/ext/pyrender/viewer.py
duburcqa: It has to be combined with quote typing :)

duburcqa: > Should I also do the same with the other imports that I added in genesis files purely for typechecking?

if it avoids circular import issues, then yes. Otherwise it is up to you :)

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: This may be a future discussion -- resolving now.

genesis/ext/pyrender/viewer.py
gasnica: Fixed after some communication on Slack. Resolving.

genesis/ext/pyrender/interaction/viewer_interaction.py
```diff
@@ -0,0 +1,114 @@
+from typing import override, TYPE_CHECKING
+
+import numpy as np
+
+from pyglet.event import EVENT_HANDLE_STATE
+
+from genesis.ext.pyrender.interaction.ray import Plane, Ray, RayHit
```
duburcqa: Could you switch to relative import?

genesis/ext/pyrender/viewer.py
```diff
@@ -14,6 +14,9 @@
 from OpenGL.GL import *
 
 import genesis as gs
+from genesis.ext.pyrender.interaction.viewer_interaction import ViewerInteraction
```
duburcqa: Same. Use relative import here.

genesis/ext/pyrender/viewer.py
```diff
@@ -14,6 +14,9 @@
...
+from genesis.ext.pyrender.interaction.viewer_interaction_base import ViewerInteractionBase
+from genesis.vis.rasterizer_context import RasterizerContext
```
duburcqa, line 17: But not for this one of course

genesis/ext/pyrender/interaction/ray.py
```diff
@@ -0,0 +1,53 @@
+from dataclasses import dataclass
+from genesis.ext.pyrender.interaction.vec3 import Vec3
```
duburcqa: Skip one line and relative import:
```python
from dataclasses import dataclass

from .vec3 import Vec3
```

duburcqa (approved):

Issue #1113, opened 2025-05-06
[Feature]: Env-wise gravity vector


nootnootpengupengu:
### What feature or enhancement are you proposing?

The ability to set a unique gravity vector for each of my simulation environments, not a global vector for all envs. 

### Motivation

Part of my domain randomization scheme requires varying the gravity vector between all envs during intraining. It is currently not possible to do so as it's a global parameter in `SimOptions`. 

### Potential Benefit

Improved policy training and robustness due to increased resilience from different gravity force. 

### What is the expected outcome of the implementation work?

The ability to set a `SimOptions.gravity = gravity_tensor`, where `gravity_tensor.shape = (num_envs, 3)`

bb7332193:
This would be very useful, would a [PerlinNoiseField](https://genesis-world.readthedocs.io/en/latest/_modules/genesis/engine/force_fields.html#PerlinNoiseField) work for this purpose?

Milotrince:
PR #1324 added this, but there is a bug. I will fix this.

PR #1324, opened 2025-06-26
[FEATURE] Environment-wise gravity

LeonLiu4:
This PR introduces support for environment-specific gravity settings. Different environments in the simulation can now have distinct gravity configurations, enabling more flexible and realistic physics behavior tailored to each context.

Example use case with two spheres: 

```
import torch
import genesis as gs

gs.init(backend=gs.cpu)

scene = gs.Scene(
    show_viewer=True,
        dt=0.01,
        substeps=1,
    ),
)

scene.build(n_envs=2)

scene.sim.set_gravity(torch.tensor([0.0, 0.0, -9.8]), envs_idx=0)
scene.sim.set_gravity(torch.tensor([0.0, 0.0, 9.8]), envs_idx=1)

for _ in range(200):
    scene.step()
```

Addresses: #1113 

Below is a video demonstrating two parallel environments one with more negative gravity making the drone fall while the other hovers: 

https://github.com/user-attachments/assets/ac3c148f-ab3e-4d26-a47f-c56d67d48601

### DIFF
--- a/genesis/engine/simulator.py
+++ b/genesis/engine/simulator.py
@@ -106,7 +106,8 @@ class Simulator(RBC):
         self._steps_local = options._steps_local
 
         self._cur_substep_global = 0
-        self._gravity = np.array(options.gravity)
+        self._g_np = np.asarray(options.gravity, dtype=np.float32)
+        self._g_ti = None
 
         # solvers
         self.tool_solver = ToolSolver(self.scene, self, self.tool_options)
@@ -188,9 +189,15 @@ class Simulator(RBC):
         self._B = self.scene._B
         self._para_level = self.scene._para_level
 
+        if self._g_np.ndim == 1:
+            self._g_np = np.repeat(self._g_np[None], self._B, axis=0)
+        self._g_ti = ti.Vector.field(3, gs.ti_float, shape=self._B)
+        self._g_ti.from_numpy(self._g_np)
+
         # solvers
         self._rigid_only = self.rigid_solver.is_active()
         for solver in self._solvers:
+            solver._finalize_batch(self._B)
             solver.build()
             if solver.is_active():
                 self._active_solvers.append(solver)
@@ -418,7 +425,7 @@ class Simulator(RBC):
     @property
     def gravity(self):
         """The gravity vector."""
-        return self._gravity
+        return self._g_ti
 
     @property
     def requires_grad(self):
--- a/genesis/engine/simulator.py
+++ b/genesis/engine/simulator.py
@@ -196,10 +196,13 @@ class Simulator(RBC):
         self._g_ti = ti.Vector.field(3, gs.ti_float, shape=self._B)
         self._g_ti.from_numpy(g_np)
 
+        for solver in self._solvers:
+            solver._gravity = self._g_ti
+
         # solvers
         self._rigid_only = self.rigid_solver.is_active()
         for solver in self._solvers:
-            solver._finalize_batch(self._B)
+            # solver._finalize_batch(self._B)
             solver.build()
             if solver.is_active():
                 self._active_solvers.append(solver)
--- a/genesis/engine/simulator.py
+++ b/genesis/engine/simulator.py
@@ -202,7 +202,6 @@ class Simulator(RBC):
         # solvers
         self._rigid_only = self.rigid_solver.is_active()
         for solver in self._solvers:
-            # solver._finalize_batch(self._B)
             solver.build()
             if solver.is_active():
                 self._active_solvers.append(solver)
--- a/genesis/engine/simulator.py
+++ b/genesis/engine/simulator.py
@@ -1,6 +1,7 @@
 from typing import TYPE_CHECKING
 import numpy as np
 import taichi as ti
+import torch
 
 import genesis as gs
 from genesis.engine.entities.base_entity import Entity
@@ -106,8 +107,8 @@ class Simulator(RBC):
         self._steps_local = options._steps_local
 
         self._cur_substep_global = 0
-        self._g_np = np.asarray(options.gravity, dtype=np.float32)
-        self._g_ti = None
+        self._g = torch.as_tensor(options.gravity, dtype=torch.float32).clone()
+        self._g_ti: ti.Field | None = None
 
         # solvers
         self.tool_solver = ToolSolver(self.scene, self, self.tool_options)
@@ -189,10 +190,11 @@ class Simulator(RBC):
         self._B = self.scene._B
         self._para_level = self.scene._para_level
 
-        if self._g_np.ndim == 1:
-            self._g_np = np.repeat(self._g_np[None], self._B, axis=0)
+        g_np = self._g.numpy()
+        if g_np.ndim == 1:
+            g_np = np.repeat(g_np[None], self._B, axis=0)
         self._g_ti = ti.Vector.field(3, gs.ti_float, shape=self._B)
-        self._g_ti.from_numpy(self._g_np)
+        self._g_ti.from_numpy(g_np)
 
         # solvers
         self._rigid_only = self.rigid_solver.is_active()
@@ -425,7 +427,19 @@ class Simulator(RBC):
     @property
     def gravity(self):
         """The gravity vector."""
-        return self._g_ti
+        return self._g_ti if self._g_ti is not None else self._g
+
+    @gravity.setter
+    def gravity(self, new_g):
+        """Set the gravity vector for the simulator."""
+        # store as torch tensor
+        self._g = torch.as_tensor(new_g, dtype=gs.tc_float)
+        # if we've already built, update the Taichi field in-place
+        if self._g_ti is not None:
+            g_np = self._g.numpy()
+            if g_np.ndim == 1:
+                g_np = np.repeat(g_np[None], self._B, axis=0)
+            self._g_ti.from_numpy(g_np)
 
     @property
     def requires_grad(self):
--- a/genesis/engine/simulator.py
+++ b/genesis/engine/simulator.py
@@ -1,7 +1,6 @@
 from typing import TYPE_CHECKING
 import numpy as np
 import taichi as ti
-import torch
 
 import genesis as gs
 from genesis.engine.entities.base_entity import Entity
@@ -184,6 +183,7 @@ class Simulator(RBC):
             solver._add_force_field(force_field)
 
     def build(self):
+
         self.n_envs = self.scene.n_envs
         self._B = self.scene._B
         self._para_level = self.scene._para_level
--- a/genesis/engine/simulator.py
+++ b/genesis/engine/simulator.py
@@ -107,8 +107,7 @@ class Simulator(RBC):
         self._steps_local = options._steps_local
 
         self._cur_substep_global = 0
-        self._g = torch.as_tensor(options.gravity, dtype=torch.float32).clone()
-        self._g_ti: ti.Field | None = None
+        self._gravity = np.array(options.gravity)
 
         # solvers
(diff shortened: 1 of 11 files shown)


tests/test_rigid_physics.py
```diff
@@ -2200,6 +2200,36 @@ def test_urdf_mimic(show_viewer, tol):
...
+@pytest.mark.required
+@pytest.mark.parametrize("backend", [gs.cpu])
+def test_gravity(show_viewer):
+    base_rpm = 14468.429183500699
+    scene = gs.Scene(
+        show_viewer=show_viewer,
+        sim_options=gs.options.SimOptions(
+            dt=0.01,
+            substeps=1,
+            gravity=[(0.0, 0.0, -9.8), (0.0, 0.0, -10.00)],
```
duburcqa: Just try `(0.0, 0.0, +9.81)`

tests/test_rigid_physics.py
```diff
@@ -2200,6 +2200,36 @@ def test_urdf_mimic(show_viewer, tol):
...
+        ),
+    )
+
+    drone = scene.add_entity(
```
duburcqa: Just simulate a dropping sphere. A drone is way too complex for this.

tests/test_rigid_physics.py
```diff
@@ -2200,6 +2200,36 @@ def test_urdf_mimic(show_viewer, tol):
...
+        gs.morphs.Drone(file="urdf/drones/cf2x.urdf", pos=(0, 0, 1.0)),
+    )
+
+    scene.build(n_envs=2)
+
+    for _ in range(500):
```
duburcqa: This is too long. 200 steps should be fine.

tests/test_rigid_physics.py
```diff
@@ -2200,6 +2200,36 @@ def test_urdf_mimic(show_viewer, tol):
...
+        drone.set_propellels_rpm([[base_rpm, base_rpm, base_rpm, base_rpm], [base_rpm, base_rpm, base_rpm, base_rpm]])
+        scene.step()
+
+    first_pos = drone.get_dofs_position()[0, 2]
+    second_pos = drone.get_dofs_position()[1, 2]
+    assert_allclose(
```
duburcqa: For the dropping sphere, you should be able to check analytically that the positions match the theory.

YilingQiao:
Could you resolve the conflicts? Thanks!

genesis/engine/simulator.py
```diff
@@ -106,7 +106,8 @@ def __init__(
         self._steps_local = options._steps_local
 
         self._cur_substep_global = 0
-        self._gravity = np.array(options.gravity)
+        self._g_np = np.asarray(options.gravity, dtype=np.float32)
```
duburcqa: All of this seems over complicated. Only store gravity as a torch tensor at this point. Then check in the attribute getter if the simulation has been built already. If so, then fetch the value from the solver, otherwise return the local value.

genesis/engine/solvers/base_solver.py
```diff
@@ -34,6 +31,15 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
     def _add_force_field(self, force_field):
         self._ffs.append(force_field)
 
+    def _finalize_batch(self, B: int):
```
duburcqa: Remove this method and move its content line 23.

duburcqa:
I don't understand why do you have to change the logic this much just for changing the way the gravity is initialised.

Moreover, it would be nice to add a setter for the gravity (both pre- and post-build).

tests/test_rigid_physics.py

tests/test_rigid_physics.py

tests/test_rigid_physics.py

tests/test_rigid_physics.py

genesis/engine/simulator.py
LeonLiu4: fixed

genesis/engine/solvers/base_solver.py
LeonLiu4: done

duburcqa (changes requested):
Currently, different solvers may have different gravity. This is kind of hacky and should be removed at some point, but for now that is how it is. This means that what you are doing is not going to work. 

However, it is not possible to just update the code block at line 25 to take into account the batch size because it is unknown at this point and will only be specified during build. Anyway, all options should be batch-size agnostic by design. I do not think it is acceptable to have to options that only work for a specific batch size as you are currently doing.

As a result, I would suggest to force `option.gravity` being a vector, and a base `build` method to `Solver` class that would be responsible for both setting the batch size and initialising the gravity field. Then add a setter with the following signature to class`Solver`:
```python
def set_gravity(self, gravity, envs_idx=None):
```
if `envs_idx=None`, set the gravity for all environments as you are doing, otherwise set the gravity for one specific environment. Just pass without doing anything if `self._gravity = None`.

Then, remove all your changes at scene-level for now, and add a `set_gravity` setter that would call `set_gravity` for all the underlying solvers sequentially.

genesis/engine/solvers/base_solver.py
```diff
@@ -36,6 +30,22 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
     def _add_force_field(self, force_field):
         self._ffs.append(force_field)
 
+
+    def build(self, B: int):
```
duburcqa: What about removing `B: int` from the input arguments and just do `self._B = self._sim._B` inside this method? Then call `super.build()` in all derived implementations of `Solver`.

genesis/engine/solvers/base_solver.py
```diff
@@ -21,13 +21,7 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
...
         self._dt: float = options.dt
         self._substep_dt: float = options.dt / sim.substeps
-
-        if hasattr(options, "gravity"):
-            self._gravity = ti.field(dtype=gs.ti_vec3, shape=())
-            self._gravity.from_numpy(np.array(options.gravity, dtype=gs.np_float))
-        else:
-            self._gravity = None
-
+        self._gravity = None
```
duburcqa: You need to keep track of the solver-specific gravity:
```python
self._init_gravity = getattr(options, "gravity")
```

genesis/engine/solvers/base_solver.py
```diff
@@ -36,6 +30,22 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
     def _add_force_field(self, force_field):
         self._ffs.append(force_field)
 
+
+    def build(self, B: int):
+        self._B = B
+        g_np = np.asarray(self._sim._gravity)
```
duburcqa:
```python
if self._init_gravity is not None:
    gravity = np.asarray(self._init_gravity, dtype=gs.np_float)
    [...]
```

tests/test_rigid_physics.py
```diff
@@ -2210,6 +2210,33 @@ def test_urdf_mimic(show_viewer, tol):
 
 @pytest.mark.required
 @pytest.mark.parametrize("backend", [gs.cpu])
+def test_gravity(show_viewer, tol):
+    scene = gs.Scene(
```
duburcqa, line 2305: Just do:
```python
scene = gs.Scene(
       show_viewer=show_viewer,
    )
```

LeonLiu4: fixed

genesis/engine/solvers/base_solver.py
LeonLiu4: added

genesis/engine/solvers/base_solver.py
LeonLiu4: added

genesis/engine/solvers/base_solver.py
LeonLiu4: removed

genesis/engine/solvers/base_solver.py
```diff
@@ -36,6 +31,29 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
     def _add_force_field(self, force_field):
         self._ffs.append(force_field)
 
+    def build(self):
+        self._B = self._sim._B
+
+        g_np = np.asarray(
```
duburcqa: If think you should rather completely ignore gravity if `self._init_gravity` is None:

```python
if self._init_gravity is not None:
    gravity = np.asarray(self._init_gravity, dtype=np.float32)
    gravity = np.repeat(g_np[None], self._B, axis=0)

    self._gravity = ti.Vector.field(3, dtype=gs.ti_float, shape=self._B)
    self._gravity.from_numpy(gravity)
```

genesis/engine/solvers/base_solver.py
```diff
@@ -36,6 +31,29 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
...
+            self._init_gravity if self._init_gravity is not None else self._sim._gravity,
+            dtype=np.float32,
+        )
+        g_np = np.repeat(g_np[None], self._B, axis=0)
+
+        self._gravity = ti.Vector.field(3, dtype=gs.ti_float, shape=self._B)
+        self._gravity.from_numpy(g_np)
+
+        if self._B == 1:
+            self._gravity.to_numpy = lambda _orig=self._gravity.to_numpy: _orig().squeeze(0)
```
duburcqa: You cannot do that. Overwriting instance methods is very hacky and should only be used in last resort. Why would this be required in the first place? It is required by `set_gravity`, then just update `set_gravity` accordingly to avoid relying on this mechanism.

genesis/engine/solvers/base_solver.py
```diff
@@ -36,6 +31,29 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
...
+
+    def set_gravity(self, gravity, envs_idx=None):
```
duburcqa, line 43: Add `@gs.assert_built` decorator.

genesis/engine/solvers/base_solver.py
```diff
@@ -36,6 +31,28 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
...
 
+    def build(self):
+        self._B = self._sim._B
+        if self._init_gravity is not None:
+            g_np = np.asarray(self._init_gravity, dtype=np.float32)
+            g_np = np.repeat(g_np[None], self._B, axis=0)
+            self._gravity = ti.Vector.field(3, dtype=gs.ti_float, shape=self._B)
+            self._gravity.from_numpy(g_np)
+        else:
+            self._gravity = None
```
duburcqa: It is already None. No need to set it again.

duburcqa (approved):

duburcqa (approved):

genesis/engine/solvers/base_solver.py
```diff
@@ -36,6 +31,26 @@ def __init__(self, scene: "Scene", sim: "Simulator", options):
...
+        self._B = self._sim._B
+        if self._init_gravity is not None:
+            g_np = np.asarray(self._init_gravity, dtype=np.float32)
+            g_np = np.repeat(g_np[None], self._B, axis=0)
+            self._gravity = ti.Vector.field(3, dtype=gs.ti_float, shape=self._B)
+            self._gravity.from_numpy(g_np)
+
+    @gs.assert_built
+    def set_gravity(self, gravity, envs_idx=None):
+        if self._gravity is None:
```
duburcqa, line 44: `if self._init_gravity is None:`

duburcqa (dismissed):

YilingQiao:
@LeonLiu4 could you please fix the unit test? Next time you can run them on you local branch first.

duburcqa (approved):

Issue #1329, opened 2025-06-28
[Feature]: Make Genesis compile faster


hughperkins:
### What feature or enhancement are you proposing?

Make Genesis compile faster

### Motivation

As a simulation user
I want to be able to compile Genesis quickly
Because I don't want to wait for it to compile

### Potential Benefit

- more efficient use of simulation user time

### What is the expected outcome of the implementation work?

- Genesis compiles faster
- There will likely be a trade-off though: faster compilation vs faster runtime
     - so should probably be optinoal

MRiabov:
I would also want to add that it would be useful to cache merging of convex hulls:
``` 
[2025-06-30 17:08:45.460] [CoACD] [info] threshold               0.1
[2025-06-30 17:08:45.460] [CoACD] [info] max # convex hull       -1
[2025-06-30 17:08:45.460] [CoACD] [info] preprocess mode         auto
[2025-06-30 17:08:45.460] [CoACD] [info] preprocess resolution   30
[2025-06-30 17:08:45.460] [CoACD] [info] pca                     false
[2025-06-30 17:08:45.460] [CoACD] [info] mcts max depth          3
[2025-06-30 17:08:45.460] [CoACD] [info] mcts nodes              20
[2025-06-30 17:08:45.460] [CoACD] [info] mcts iterations         100
[2025-06-30 17:08:45.460] [CoACD] [info] merge                   true
[2025-06-30 17:08:45.460] [CoACD] [info] decimate                false
[2025-06-30 17:08:45.460] [CoACD] [info] max_ch_vertex           256
[2025-06-30 17:08:45.460] [CoACD] [info] extrude                 false
[2025-06-30 17:08:45.460] [CoACD] [info] extrude margin          0.1
[2025-06-30 17:08:45.460] [CoACD] [info] approximate mode        ch
[2025-06-30 17:08:45.460] [CoACD] [info] seed                    0
[2025-06-30 17:08:45.468] [CoACD] [info]  - Manifold Check
[2025-06-30 17:08:45.537] [CoACD] [info]        Unclosed mesh
[2025-06-30 17:08:45.537] [CoACD] [info] Manifold Check Time: 0.070015s
[2025-06-30 17:08:45.588] [CoACD] [info] Mesh Manifoldness: false
[2025-06-30 17:08:45.589] [CoACD] [info]  - Preprocess
[2025-06-30 17:08:45.589] [CoACD] [info] Preprocess resolution: 30
[...]

PR #1338, opened 2025-06-30
[BUG FIX] Fix terrain collision detection.

duburcqa:
## Description

* Exposing low-level MPR interface allowing to explicitly passing the desired geometries' center as input argument of MPR

## Motivation and Context

The collision detection algorithm from ANY vs TERRAIN is based on MPR under the hood. This involves dynamically computing the "center" of terrain by is modifying `geoms_info[i_gb].center` on-the-fly. This should be avoided because it creates collisions when dealing with batched environments.

This bug was uncovered by passing `cfg_optimization=False` to `taichi.init` method. Taichi's optimisations are apparently hiding the bug, possibly because of some temporary variable elision of something. But it should have failed from the start.

## Related Issue

Related to https://github.com/Genesis-Embodied-AI/Genesis/issues/1329

### DIFF
--- a/genesis/engine/solvers/rigid/collider_decomp.py
+++ b/genesis/engine/solvers/rigid/collider_decomp.py
@@ -572,16 +572,14 @@ class Collider:
                 )
             )
 
-            for i in range(6):
-                i_axis = i % 3
-                i_m = i // 3
-
-                sign = gs.ti_float(1 - i_m * 2)
-                direction = ti.Vector([i_axis == 0, i_axis == 1, i_axis == 2], dt=gs.ti_float)
-                direction = direction * sign
-
+            for i_axis, i_m in ti.ndrange(3, 2):
+                direction = ti.Vector.zero(gs.ti_float, 3)
+                if i_m == 0:
+                    direction[i_axis] = 1.0
+                else:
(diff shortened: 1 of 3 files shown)


genesis/engine/solvers/rigid/collider_decomp.py
```diff
@@ -622,24 +620,27 @@ def _func_contact_mpr_terrain(self, i_ga, i_gb, i_b):
...
+                                    center_a = gu.ti_transform_by_trans_quat(
+                                        self._solver.geoms_info[i_ga].center, ga_pos, ga_quat
+                                    )
+                                    center_b = ti.Vector.zero(gs.ti_float, 3)
                                     for i_p in ti.static(range(6)):
-                                        pos = pos + self.prism[i_p, i_b]
+                                        center_b = center_b + self.prism[i_p, i_b]
+                                    center_b = center_b / 6.0
 
-                                    self._solver.geoms_info[i_gb].center = pos / 6
```
YilingQiao, line 629: ahh... surprised it did not cause error before

duburcqa: Yes!

PR #1361, opened 2025-07-05
[Migration] Initial template for migration

YilingQiao:
## Description

This is a template for Stage 1 of the migration:
- After this migration, the kernels will no longer rely on self, and each kernel will be stateless.
- This migration can be done kernel by kernel and does not change the logic of the code.
- In the next stages, when struct-of-ndarray is supported, we will add ndarray support (to avoid recompilation) and may move some kernels to new files for better separation of logic.

When migrating a kernel, you can pay attention to those things
- Change `self` argument to `self_unused`
- For well defined link-related (or joint, geom, dof related) arguments like `self.links_state`, `self.dofs_info`, you can put them into the argument list.
- For global arguments like `self.mass_parent_mask`, you can put them into `RigidGlobalInfo` and pass `_rigid_global_info` as an argument (similar to what we did to `n_awake_dofs`)
- For static variables (like `self._para_level`) that will be used during compile time, you can put them into `RigidGlobalInfo` and pass `_rigid_global_info` as an argument.

## Related Issue

## Checklist:


gasnica (dismissed):
This looks great and super clean!

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -490,38 +518,46 @@ def _kernel_init_dof_fields(
...
+                dofs_info[I].force_range[j] = dofs_force_range[i, j]
+
+            dofs_info[I].armature = dofs_armature[i]
+            dofs_info[I].invweight = dofs_invweight[i]
+            dofs_info[I].stiffness = dofs_stiffness[i]
+            dofs_info[I].damping = dofs_damping[i]
+            dofs_info[I].kp = dofs_kp[i]
+            dofs_info[I].kv = dofs_kv[i]
+
+        ti.loop_config(serialize=ti.static(static_rigid_sim_config.para_level) < gs.PARA_LEVEL.PARTIAL)
```
duburcqa: This is inconsistent. Sometimes flagged as ti.static and sometimes not.

YilingQiao: Nice catch. Fixed. I deleted the ti.static since Python objects parsed as ti.template are already implicitly static (as far as I remember).

duburcqa: You are probably right.

PR #1370, opened 2025-07-07
[FEATURE] Add 'get_weld_constraints' API.

### LINKED ISSUE #1365: [Feature]: Create an API to get weld constraints

Hello, 
There are two functions to add rigid (weld) constraints as per #961. However, there seems to be no "get_weld_constraint", in my case (where there are a number of such constraints).  This makes it more complex to add e.g. sanity checks when removing two constraints between such bodies.

### Motivation

It would be useful to find out which bodies have the constraints for tests and assertions.

As a RL researcher working with fasteners,
I want to assert that my code behaves as expected (and debug)
This enables me to write the code that works.

### Potential Benefit

Improved tests of user code. Additionally, it currently is challenging to remove all such weld constraints in case of custom scene reset - we don't really know which lines constrain what unless we store it explicitly (or there is a method, but it is undocumented)

- A function/property in scene.rigid_solver outputting a tensor (?) of connected links. e.g - a tensor of shape (5,2) outputting two links for 5 bodies.

### Additional information

_No response_

LeonLiu4:
Adds RigidSolver.get_weld_constraints() – returns an (N, 3) NumPy array (env_id, link1_idx, link2_idx) of all active weld constraints, enabling quick inspection and test assertions without poking internal Taichi fields.

Resolves #1365


duburcqa (changes requested):
Could you take inspiration of `get_contacts`? i.e. return a dict instead of a raw tendit, and extract all taichi field values in a single kernel call.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -5002,6 +5002,41 @@ def _kernel_delete_weld_constraint(
...
+        envs_idx = np.atleast_1d(envs_idx)
+
+        env_col, obj_a_col, obj_b_col = [], [], []
+        WELD = gs.EQUALITY_TYPE.WELD
+
+        for env in envs_idx:
+            n_eq = int(self.constraint_solver.ti_n_equalities[env])
+            for i in range(n_eq):
+                rec = self.equalities_info[i, env]
+                if rec.eq_type == WELD:
```
duburcqa: Just use `gs.EQUALITY_TYPE.WELD` directly.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -5002,6 +5002,41 @@ def _kernel_delete_weld_constraint(
...
+                    env_col.append(env)
+                    obj_a_col.append(int(rec.eq_obj1id))
+                    obj_b_col.append(int(rec.eq_obj2id))
+
+        obj_a = np.asarray(obj_a_col, dtype=np.int32)
+        obj_b = np.asarray(obj_b_col, dtype=np.int32)
```
duburcqa: Move all of this in a taichi kernel.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -5002,6 +5002,41 @@ def _kernel_delete_weld_constraint(
...
+
+        if self.n_envs == 0:
+            return {"obj_a": obj_a, "obj_b": obj_b}
+
+        env_arr = np.asarray(env_col, dtype=np.int32)
+        valid_mask = np.zeros((self.n_envs, obj_a.shape[0]), dtype=bool)
```
duburcqa: You took inspiration from the wrong function. It was supposed to be `Collider.get_contacts`, not `RigidEntity.get_contacts`, sorry about that. Which mean that the API should be `def get_weld_constraints(self, as_tensor: bool = True, to_torch: bool = True):`

Then you could also add its counterpart in `RigidEntity` if you think it is worth it.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
LeonLiu4: fixed

genesis/engine/solvers/rigid/rigid_solver_decomp.py
LeonLiu4: added

genesis/engine/solvers/rigid/rigid_solver_decomp.py
LeonLiu4: fixed the above comment I will work on implementing it into RigidEntity as well

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -4998,6 +4998,69 @@ def _kernel_delete_weld_constraint(
...
+            "env": env_arr,
+            "obj_a": obj_a,
+            "obj_b": obj_b,
+            "valid_mask": valid,
+        }
+
+    @ti.kernel
+    def _kernel_collect_welds(self, buf: ti.types.ndarray()):
+        WELD = gs.EQUALITY_TYPE.WELD
+        n_envs = ti.static(max(1, self.n_envs))
```
duburcqa: Just use `self._n_envs`

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -4998,6 +4998,69 @@ def _kernel_delete_weld_constraint(
...
+        stride = buf.shape[0] // n_envs
```
duburcqa: Use hard-code stride it is fine. This method is private and only called once, and it should crash anyway if the buffer shape is not correct.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
LeonLiu4: fixed

genesis/engine/solvers/rigid/rigid_solver_decomp.py
LeonLiu4: fixed

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -5715,6 +5715,68 @@ def _kernel_delete_weld_constraint(
                         ]
                     self.constraint_solver.ti_n_equalities[i_b] = self.constraint_solver.ti_n_equalities[i_b] - 1
 
+    @gs.assert_built
+    def get_weld_constraints(self, as_tensor: bool = True, to_torch: bool = True):
```
duburcqa: This is still wrong. If you look more closely, there is no such thing as `valid_mask` in `RigidSolver.get_contacts`, only in `RigidEntity.get_contacts`. You should do the same here.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -5715,6 +5715,68 @@ def _kernel_delete_weld_constraint(
...
+        return {
+            "env": env_arr,
+            "obj_a": obj_a,
+            "obj_b": obj_b,
+            "valid_mask": valid,
+        }
+
+    @ti.kernel
+    def _kernel_collect_welds(self, buf: ti.types.ndarray()):
+        WELD = gs.EQUALITY_TYPE.WELD
```
duburcqa: Remove all these variables and use their value directly in the implementation.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -5715,6 +5715,68 @@ def _kernel_delete_weld_constraint(
...
+            obj_b = empty((0,), dtype=np.int32)
+            if n_envs == 1:
+                return {"obj_a": obj_a, "obj_b": obj_b}
+            return {
+                "env": obj_a,
+                "obj_a": obj_a,
+                "obj_b": obj_b,
+                "valid_mask": zeros((n_envs, 0), dtype=bool),
+            }
+        shape3 = (n_envs, stride, 3)
```
duburcqa: Could you follow more strictly the pattern in `RigidSolver.get_contacts`? You should not pass a 3D tensor because it would allocate more memory than necessary but most importantly, it will break "C"-contiguity, which would be problematic if forwarded as input argument of another taichi kernel.

tests/test_rigid_physics.py
```diff
@@ -2342,6 +2342,75 @@ def test_drone_advanced(show_viewer):
...
+    cube1.set_dofs_velocity(
+        velocity=np.array([1.0, 0.0, 0.0], dtype=np.float32),
+        dofs_idx_local=np.array([0, 1, 2], dtype=np.int32),
+        envs_idx=None,
+    )
+
+    for _ in range(200):
+        scene.step()
+
+    vel1 = np.asarray(cube1.get_vel())
```
duburcqa: No need to call `np.asarray` here

tests/test_rigid_physics.py
```diff
@@ -2342,6 +2342,75 @@ def test_drone_advanced(show_viewer):
...
+    vel2 = np.asarray(cube2.get_vel())
+
+    assert_allclose(vel1, vel2, tol=1e-7)
```
duburcqa: Just do `assert_allclose(cube1.get_vel(), cube2.get_vel(), tol=1e-7)`

tests/test_rigid_physics.py
```diff
@@ -2342,6 +2342,75 @@ def test_drone_advanced(show_viewer):
...
+    link_a = torch.tensor([cube1.base_link.idx], dtype=gs.tc_int, device=gs.device)
+    link_b = torch.tensor([cube2.base_link.idx], dtype=gs.tc_int, device=gs.device)
+
+    rigid.add_weld_constraint(link_a, link_b)
+    scene.step()
+
+    welds_single = cube1.get_weld_constraints(as_tensor=True, to_torch=False)
+
+    assert_allclose(
+        np.array([int(welds_single["obj_a"][0]), int(welds_single["obj_b"][0])]),
```
duburcqa: Remove `np.array` and `int`

tests/test_rigid_physics.py
```diff
@@ -2342,6 +2342,75 @@ def test_drone_advanced(show_viewer):
...
+        np.array([link_a.item(), link_b.item()], dtype=np.int32),
```
duburcqa: Same.

tests/test_rigid_physics.py
```diff
@@ -2342,6 +2342,75 @@ def test_drone_advanced(show_viewer):
...
+@pytest.mark.parametrize("backend", [gs.cpu])
+def test_get_weld_constraints_entity(show_viewer, tol):
+    scene = gs.Scene(show_viewer=show_viewer)
+
+    cube1 = scene.add_entity(gs.morphs.Box(size=(0.05,) * 3, pos=(0.0, 0.0, 0.05)))
+    cube2 = scene.add_entity(gs.morphs.Box(size=(0.05,) * 3, pos=(0.2, 0.0, 0.05)))
+
+    scene.build(n_envs=1)
+
+    rigid = scene.sim.rigid_solver
```
duburcqa: Do not define this alias. This is useless here.

tests/test_rigid_physics.py
```diff
@@ -2342,6 +2342,75 @@ def test_drone_advanced(show_viewer):
...
+    vel2 = np.asarray(cube2.get_vel())
+
+    assert_allclose(vel1, vel2, tol=1e-7)
+
+
+@pytest.mark.required
+@pytest.mark.parametrize("backend", [gs.cpu])
+def test_get_weld_constraints_entity(show_viewer, tol):
+    scene = gs.Scene(show_viewer=show_viewer)
+
```
duburcqa: No need to skip line here.

tests/test_rigid_physics.py
```diff
@@ -2342,6 +2342,75 @@ def test_drone_advanced(show_viewer):
...
+    cube1 = scene.add_entity(gs.morphs.Box(size=(0.05,) * 3, pos=(0.0, 0.0, 0.05)))
+    cube2 = scene.add_entity(gs.morphs.Box(size=(0.05,) * 3, pos=(0.2, 0.0, 0.05)))
+
```
duburcqa: and here.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
LeonLiu4: removed valid_mask

genesis/engine/solvers/rigid/rigid_solver_decomp.py
LeonLiu4: fixed

genesis/engine/solvers/rigid/rigid_solver_decomp.py
LeonLiu4: modified

tests/test_rigid_physics.py

tests/test_rigid_physics.py

tests/test_rigid_physics.py

tests/test_rigid_physics.py

tests/test_rigid_physics.py

tests/test_rigid_physics.py

tests/test_rigid_physics.py

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -6005,6 +6005,61 @@ def _kernel_delete_weld_constraint(
...
+        if max_welds == 0:
+            empty = torch.empty if to_torch else np.empty
+            obj_a = empty((0,), dtype=np.int32)
+            obj_b = empty((0,), dtype=np.int32)
+            if n_envs == 1:
+                return {"obj_a": obj_a, "obj_b": obj_b}
+            return {
+                "obj_a": [obj_a] * n_envs,
+                "obj_b": [obj_b] * n_envs,
+            }
```
duburcqa: This branch should be useless. Just skip `self._kernel_collect_welds(buf)` if `max_welds == 0`

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -6005,6 +6005,61 @@ def _kernel_delete_weld_constraint(
...
+
+        total = n_envs * max_welds
```
duburcqa: This is wrong. See my previous comment:

> Could you follow more strictly the pattern in RigidSolver.get_contacts? You should not pass a 3D tensor because it would allocate more memory than necessary but most importantly, it will break "C"-contiguity, which would be problematic if forwarded as input argument of another taichi kernel.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
LeonLiu4: fixed

genesis/engine/solvers/rigid/rigid_solver_decomp.py
LeonLiu4: fixed

davidhommie:
Too much additions leads to error

YilingQiao:
@LeonLiu4 can you merge the latest main? You might need to modify the kernel a little bit to adapt to the new struct of array access pattern

YilingQiao:
@davidhommie can you provide a script? Or a link to the issue you mentioned?

YilingQiao (dismissed):

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -2296,6 +2296,49 @@ def update_verts_for_geom(self, i_g):
...
+                self._static_rigid_sim_config,
+            )
+
+        if as_tensor:
+            if n_envs > 0:
+                buf = buf.view(n_envs, n_max, 3) if to_torch else buf.reshape(n_envs, n_max, 3)
+            env_idx, obj_a, obj_b = buf[..., 0], buf[..., 1], buf[..., 2]
+            return {"env": env_idx, "obj_a": obj_a, "obj_b": obj_b}
+
+        if n_envs == 0:
```
duburcqa: Move this return on the top, right after `n_eqs = ...`

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -2296,6 +2296,49 @@ def update_verts_for_geom(self, i_g):
...
         )
 
+    def get_weld_constraints(self, as_tensor: bool = True, to_torch: bool = True):
+        n_eqs = tuple(self.constraint_solver.constraint_state.ti_n_equalities.to_numpy())
+        n_envs = len(n_eqs)
+        n_max = max(n_eqs) if n_eqs else 0
+
+        if as_tensor:
+            out_size = n_envs * n_max
+            splits = None
```
duburcqa: Do not define `splits` in this case, it is useless.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
duburcqa: Then you should remove the branches `if n_eqs: ...` everywhere.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -2296,6 +2296,49 @@ def update_verts_for_geom(self, i_g):
...
+        if as_tensor:
+            if n_envs > 0:
+                buf = buf.view(n_envs, n_max, 3) if to_torch else buf.reshape(n_envs, n_max, 3)
+            env_idx, obj_a, obj_b = buf[..., 0], buf[..., 1], buf[..., 2]
+            return {"env": env_idx, "obj_a": obj_a, "obj_b": obj_b}
+
+        if n_envs == 0:
+            return {"obj_a": [], "obj_b": []}
+
+        parts = torch.split(buf, n_eqs) if to_torch else np.split(buf, splits)
```
duburcqa: `data_chunks` is more common name.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -2296,6 +2296,49 @@ def update_verts_for_geom(self, i_g):
...
+        a = [p[:, 1] for p in parts]
```
duburcqa: `a` and `b` are not suitable variable names. Just use `obj_a, obj_b`.

Create tuple to keep the data immutable. And generate both output at once:
```python
obj_a, obj_b = tuple(zip(*((data[:, 1], data[:, 2]) for data in data_chunks)))
```

duburcqa: If you find it more readable, you can still generate `obj_a, obj_b` separately:
```
obj_a = tuple(data[:, 1] for for data in data_chunks)
obj_b = tuple(data[:, 2] for for data in data_chunks)
```

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -2296,6 +2296,49 @@ def update_verts_for_geom(self, i_g):
...
+            )
+
+        if as_tensor:
+            if n_envs > 0:
+                buf = buf.view(n_envs, n_max, 3) if to_torch else buf.reshape(n_envs, n_max, 3)
+            env_idx, obj_a, obj_b = buf[..., 0], buf[..., 1], buf[..., 2]
+            return {"env": env_idx, "obj_a": obj_a, "obj_b": obj_b}
+
+        if n_envs == 0:
+            return {"obj_a": [], "obj_b": []}
```
duburcqa, line 2333: You should add `"env"` key, even if it is empty.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -2296,6 +2296,49 @@ def update_verts_for_geom(self, i_g):
...
+                self.constraint_solver.constraint_state,
+                self.equalities_info,
+                self._static_rigid_sim_config,
+            )
+
+        if as_tensor:
+            if n_envs > 0:
+                buf = buf.view(n_envs, n_max, 3) if to_torch else buf.reshape(n_envs, n_max, 3)
+            env_idx, obj_a, obj_b = buf[..., 0], buf[..., 1], buf[..., 2]
+            return {"env": env_idx, "obj_a": obj_a, "obj_b": obj_b}
```
duburcqa: Adding key `"env"` should not be necessary. It was not there in `Collider.get_contacts`.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1834,6 +1834,32 @@ def set_pos(self, pos, envs_idx=None, *, relative=False, zero_velocity=True, uns
         if zero_velocity:
             self.zero_all_dofs_velocity(envs_idx, unsafe=unsafe)
 
+    @gs.assert_built
+    def get_weld_constraints(self, as_tensor: bool = True, to_torch: bool = True):
```
duburcqa: Remove the options `as_tensor: bool = True, to_torch: bool = True`. It should be enforced in this public method. No need for this level of flexibility here. This will simplify a lot the implementation.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1834,6 +1834,32 @@ def set_pos(self, pos, envs_idx=None, *, relative=False, zero_velocity=True, uns
...
+        welds = self.scene.sim.rigid_solver.get_weld_constraints(as_tensor=as_tensor, to_torch=to_torch)
+        obj_a = welds["obj_a"]
+        obj_b = welds["obj_b"]
+        if as_tensor:
+            mask = (obj_a == self.idx) | (obj_b == self.idx)
+            if self._solver.n_envs > 0:
+                welds["valid_mask"] = mask
+            for k in ("obj_a", "obj_b"):
+                welds[k] = welds[k][mask]
+            if "env" in welds:
```
duburcqa: Optional keys should be avoided unless strictly necessary. Why do you need this?

tests/test_rigid_physics.py
```diff
@@ -2352,6 +2352,62 @@ def test_drone_advanced(show_viewer):
...
+    cube2 = scene.add_entity(gs.morphs.Box(size=(0.05,) * 3, pos=(0.2, 0.0, 0.05)))
+    scene.build(n_envs=1)
+
+    link_a = torch.tensor([cube1.base_link.idx], dtype=gs.tc_int, device=gs.device)
+    link_b = torch.tensor([cube2.base_link.idx], dtype=gs.tc_int, device=gs.device)
+
+    scene.sim.rigid_solver.add_weld_constraint(link_a, link_b)
+    scene.step()
+
+    welds = scene.sim.rigid_solver.get_weld_constraints(as_tensor=True, to_torch=False)
```
duburcqa: You should check the 4 possible combinations.

tests/test_rigid_physics.py
```diff
@@ -2352,6 +2352,62 @@ def test_drone_advanced(show_viewer):
...
+    env_id = 0 if "env" not in welds else int(welds["env"][0])
+
+    row = np.array([env_id, int(welds["obj_a"][0]), int(welds["obj_b"][0])], dtype=np.int32)
```
duburcqa: Why converting to int is necessary ? It should be int already, and anyway, you are enforcing dtype.

tests/test_rigid_physics.py
```diff
@@ -2352,6 +2352,62 @@ def test_drone_advanced(show_viewer):
...
+    scene.build(n_envs=1)
+
+    link_a = torch.tensor([cube1.base_link.idx], dtype=gs.tc_int, device=gs.device)
+    link_b = torch.tensor([cube2.base_link.idx], dtype=gs.tc_int, device=gs.device)
+
+    scene.sim.rigid_solver.add_weld_constraint(link_a, link_b)
+    scene.step()
+
+    welds = scene.sim.rigid_solver.get_weld_constraints(as_tensor=True, to_torch=False)
+    env_id = 0 if "env" not in welds else int(welds["env"][0])
```
duburcqa: Same. Casting to int should not be necessary.

tests/test_rigid_physics.py
```diff
@@ -2352,6 +2352,62 @@ def test_drone_advanced(show_viewer):
...
+    assert_allclose(row, ref, tol=tol)
+
+    cube1.set_dofs_velocity(
+        velocity=np.array([1.0, 0.0, 0.0], dtype=np.float32),
+        dofs_idx_local=np.array([0, 1, 2], dtype=np.int32),
+        envs_idx=None,
+    )
+    for _ in range(200):
+        scene.step()
+    assert_allclose(cube1.get_vel(), cube2.get_vel(), tol=tol)
```
duburcqa, line 2386: What is the point of this ? You are only validating API, not physics. This is done in another test? Just remove it.

duburcqa: What is the point of this ? You are only validating API, not physics. This is done in another test? Just remove it.

tests/test_rigid_physics.py
```diff
@@ -2352,6 +2352,62 @@ def test_drone_advanced(show_viewer):
...
+
+
+@pytest.mark.required
+@pytest.mark.parametrize("backend", [gs.cpu])
+def test_get_weld_constraints_entity(show_viewer, tol):
```
duburcqa: Just keep them both in the same test. An call the test `test_get_weld_constraints_api`

genesis/engine/solvers/rigid/rigid_solver_decomp.py
LeonLiu4: fixed

genesis/engine/solvers/rigid/rigid_solver_decomp.py
LeonLiu4: removed

genesis/engine/solvers/rigid/rigid_solver_decomp.py
LeonLiu4: fixed

genesis/engine/solvers/rigid/rigid_solver_decomp.py
LeonLiu4: fixed

genesis/engine/solvers/rigid/rigid_solver_decomp.py
LeonLiu4: removed

genesis/engine/entities/rigid_entity/rigid_entity.py
LeonLiu4: removed

genesis/engine/entities/rigid_entity/rigid_entity.py
LeonLiu4: removed

tests/test_rigid_physics.py
LeonLiu4: added

tests/test_rigid_physics.py
LeonLiu4: removed

tests/test_rigid_physics.py
LeonLiu4: removed

tests/test_rigid_physics.py
LeonLiu4: removed

tests/test_rigid_physics.py

MRiabov:
Folks, maybe you would merge a basic version of this? It's been a month since https://github.com/Genesis-Embodied-AI/Genesis/issues/1365. I currently have to write strange logic to assert weld constraints existence in my tests, e.g. moving a part by an arbitrary distance, and asserting that welded parts have to be equally moved.

duburcqa (approved):

duburcqa:
@MRiabov I understand your frustration but we only merge features when it is ready, and we were busy with other features during the last few weeks.

Issue #909, opened 2025-03-22
[Bug]: `self.drone.get_quat()` return `nan` in hover env.


Nam-dada:
### Bug Description

I am currently trying to use the algorithms from [EvoX](https://github.com/EMI-Group/evox) to train environments provided in Genesis. While attempting to train the `hover_env`, I found that after a certain number of iterations, the returned reward becomes `nan`. 

To investigate this, I conducted local debugging and discovered that in the definition of `hover_env.py`, the return value of `self.drone.get_quat()` may become `nan` for some individuals. This subsequently causes an error in the computation of `self.base_euler`, which in turn affects the calculation of `crash_condition`, `reset_buf`, and `_reward_yaw`, eventually leading to incorrect behavior.

### Steps to Reproduce

Since the code is still under development, I am currently unable to share all the scripts. However, I will list some of the PyTorch-related settings I am using below.

```python
import torch
import genesis as gs
from hover_env import HoverEnv

torch.set_float32_matmul_precision("high")
torch.set_default_device("cuda" if torch.cuda.is_available() else "cpu")
seed = 1234
torch.manual_seed(seed)
torch.cuda.manual_seed_all(seed)

def get_cfgs():
    env_cfg = {
        "num_actions": 4,
        # termination
        "termination_if_roll_greater_than": 180,  # degree
        "termination_if_pitch_greater_than": 180,
        "termination_if_close_to_ground": 0.1,
        "termination_if_x_greater_than": 3.0,
        "termination_if_y_greater_than": 3.0,
        "termination_if_z_greater_than": 2.0,
        # base pose
        "base_init_pos": [0.0, 0.0, 1.0],
        "base_init_quat": [1.0, 0.0, 0.0, 0.0],
        "episode_length_s": 15.0,
        "at_target_threshold": 0.1,
        "resampling_time_s": 3.0,
        "simulate_action_latency": True,
        "clip_actions": 1.0,
        # visualization
        "visualize_target": False,
        "visualize_camera": False,
        "max_visualize_FPS": 60,
    }
    obs_cfg = {
        "num_obs": 17,
        "obs_scales": {
            "rel_pos": 1 / 3.0,
            "lin_vel": 1 / 3.0,
            "ang_vel": 1 / 3.14159,
        },
    }
    reward_cfg = {
        "yaw_lambda": -10.0,
        "reward_scales": {
            "target": 10.0,
            "smooth": -1e-4,
            "yaw": 0.01,
[...]

PR #1374, opened 2025-07-08
[FIX]: RPM conversion to radians/s in hover_env.py

LeonLiu4:
Convert input from RPM → rad/s (rpm * 2 π / 60) inside set_propellers_rpm before passing it to the Taichi kernel. The kernel expects angular velocity (rad/s); feeding raw RPM caused unrealistically large thrust/torque and simulation. 

Addresses: #909 the buggy repro script does not return nan now


YilingQiao:
This is not a proper fix. That kernel does not expect angular velocity.

duburcqa:
Let's keep this PR

PR #1376, opened 2025-07-09
[MIGRATION] Collider migration

SonSang:
## Description

First stage of migration for Collider functions based on #1361 .

## Related Issue

## Checklist:


genesis/engine/solvers/rigid/collider_decomp.py
```diff
@@ -322,7 +354,21 @@ def detection(self) -> None:
...
         # timer.stamp("func_update_aabbs")
-        self._func_broad_phase()
+        self._func_broad_phase(
+            self._solver.links_state,
+            self._solver.links_info,
+            self._solver.geoms_state,
+            self._solver.geoms_info,
+            self._solver._rigid_global_info,
+            self._solver._static_rigid_sim_config,
+            self.collision_pair_validity,
```
YilingQiao: can you put `collision_pair_validity ` into `broad_buffer`. We can wrap those variables into that struct to make the argument list shorter and cleaner.

genesis/engine/solvers/rigid/collider_decomp.py
```diff
@@ -322,7 +354,21 @@ def detection(self) -> None:
...
+            self.broad_buffer,
+            self.broad_first_time,
```
YilingQiao: same, can we put it into `broad_buffer `

genesis/engine/solvers/rigid/collider_decomp.py
```diff
@@ -322,7 +354,21 @@ def detection(self) -> None:
...
+            self.n_broad_pairs,
+            self.broad_collision_pairs,
+            self._max_collision_pairs,
```
YilingQiao: Can we store this in a ti.field? If it's a constant, the kernel will need to be recompiled when its value changes.

genesis/engine/solvers/rigid/collider_decomp.py
```diff
@@ -50,6 +51,42 @@ def rotmatx(matin, i0, i1, i2, f0, f1, f2):
     return matres
 
 
+@ti.func
+def check_collision_valid(
```
YilingQiao: Can you keep it as a class method, in the same place with the same name, instead of turning it into a standalone function, to minimize code changes?

genesis/engine/solvers/rigid/collider_decomp.py
```diff
@@ -50,6 +51,42 @@ def rotmatx(matin, i0, i1, i2, f0, f1, f2):
...
+        if (links_state[i_la, i_b].hibernated and links_info[I_lb].is_fixed) or (
+            links_state[i_lb, i_b].hibernated and links_info[I_la].is_fixed
+        ):
+            is_valid = False
+
+    return is_valid
+
+
+@ti.func
+def is_geom_aabbs_overlap(geoms_state: array_class.GeomsState, i_ga: int, i_gb: int, i_b: int):
```
YilingQiao: can you keep it as a class method instead of a standalone function to minimize code change?

genesis/engine/solvers/rigid/collider_decomp.py
```diff
@@ -171,25 +208,20 @@ def _init_collision_fields(self) -> None:
         )
 
         ############## broad phase SAP ##############
-        # This buffer stores the AABBs along the search axis of all geoms
-        struct_sort_buffer = ti.types.struct(value=gs.ti_float, i_g=gs.ti_int, is_max=gs.ti_int)
-        self.sort_buffer = struct_sort_buffer.field(
-            shape=self._solver._batch_shape(2 * self._solver.n_geoms_),
-            layout=ti.Layout.SOA,
+        self.broad_buffer = array_class.ColliderBroadPhaseBuffer(
```
YilingQiao: can you call it something like 
```
self.collision_global_info = CollisionGlobalInfo
```
and also put narrow-phase-related data into this wrapper? That way, all collision-related data are defined within this class.

genesis/engine/solvers/rigid/collider_decomp.py
```diff
@@ -268,7 +300,7 @@ def _kernel_reset(
         ti.loop_config(serialize=self._solver._para_level < gs.PARA_LEVEL.ALL)
         for i_b_ in range(envs_idx.shape[0]):
             i_b = envs_idx[i_b_]
-            self.first_time[i_b] = 1
+            self.broad_first_time[i_b] = 1
```
gasnica: I think this is fine -- but in general, we'd like to avoid changing variable names during this migration

genesis/engine/solvers/rigid/array_class.py
```diff
@@ -8,10 +8,33 @@
...
     def __init__(self, n_dofs: int, n_entities: int, n_geoms: int, f_batch: Callable):
         self.n_awake_dofs = ti.field(dtype=gs.ti_int, shape=f_batch())
         self.awake_dofs = ti.field(dtype=gs.ti_int, shape=f_batch(n_dofs))
+
+
+# =========================================== Collider ===========================================
+
+
+@ti.data_oriented
+class ColliderBroadPhaseBuffer:
```
gasnica: it's multiple buffers, so maybe use plural in name.?

duburcqa: Personally I think the terminology `[...]State` or `[...]Data` with trailing `s`. `Buffer` is confusing because it is usually associated with temporary data that are in undefined state after processing but pre-allocated for efficiency only.

genesis/engine/solvers/rigid/collider_decomp.py
SonSang: I pulled them out because if we keep it as a class method, I should use `self_unused.check_collision_valid(...)`. Would it be okay to do in this way? I assume it would be fine, but just for confirmation.

genesis/engine/solvers/rigid/array_class.py
gasnica: Yeah, we spoke with Yiling yesterday, and it'd be good to split between static/const and changing data. Probably using State for chaning, and some good name for static -- for now we can use -Info, like DofsState & DofsInfo.

duburcqa: We already have this I think. It is `[...]_info` vs `[...]_state` no?

In pinocchio they are using the terminology : `[...]_model` VS `[...]_data`. Not sure if it is better.

EDIT: Ok just finished reading your comment x)

SonSang: Thanks for your comments! But when we see [here](https://github.com/Genesis-Embodied-AI/Genesis/blob/b531c32fff0abda587566896c4219a7c3a0ee785/genesis/engine/solvers/rigid/array_class.py#L14), we use `GlobalInfo` also for the changing data. Also, when we see [here](https://github.com/Genesis-Embodied-AI/Genesis/blob/b531c32fff0abda587566896c4219a7c3a0ee785/genesis/engine/solvers/rigid/rigid_solver_decomp.py#L71), we use `Static...` for the static data. 

Therefore I was using the similar naming -- `ColliderGlobalInfo` for storing every changing data, and `StaticColliderInfo` for static data (even though I did not push it yet), just following these naming conventions. Does it sound good to you guys? I could change them to `ColliderState` and `ColliderInfo`, but I think we need an agreement on this naming convention.

YilingQiao: Yeah, let's use `ColliderState` and `ColliderInfo`. I will also update to `RigidGlobalState` and `RigidGlobalInfo` accordingly

genesis/engine/solvers/rigid/collider_decomp.py
YilingQiao: oh, I see, maybe we can keep your current implementation. Keep them standalone.

genesis/engine/solvers/rigid/array_class.py
```diff
@@ -4,14 +4,248 @@
...
 
 @ti.data_oriented
 class RigidGlobalInfo:
-    def __init__(self, n_dofs: int, n_entities: int, n_geoms: int, f_batch: Callable):
+    def __init__(self, n_dofs: int, n_entities: int, n_geoms: int, _B: int, f_batch: Callable):
         self.n_awake_dofs = ti.field(dtype=gs.ti_int, shape=f_batch())
         self.awake_dofs = ti.field(dtype=gs.ti_int, shape=f_batch(n_dofs))
+
+        self.n_geoms = ti.field(dtype=gs.ti_int, shape=())
+        self.n_geoms[None] = n_geoms
```
YilingQiao: we don't need to store n_geoms, _B, etc in a variable. They could be accessed by (e.g. geoms_state.shape[0], geoms_state.shape[1])

genesis/engine/solvers/rigid/array_class.py
```diff
@@ -4,14 +4,248 @@
...
+
+        # This buffer stores indexes of active geoms during SAP search
+        if use_hibernation:
+            self.active_buffer_awake = ti.field(dtype=gs.ti_int, shape=f_batch(n_geoms))
+            self.active_buffer_hib = ti.field(dtype=gs.ti_int, shape=f_batch(n_geoms))
+        self.active_buffer = ti.field(dtype=gs.ti_int, shape=f_batch(n_geoms))
+
+        # Stores the validity of the collision pairs
+        self.collision_pair_validity = ti.field(dtype=gs.ti_int, shape=(n_geoms, n_geoms))
+        n_possible_pairs = self._init_collision_pair_validity(solver)
```
YilingQiao: Ideally, we want to separate the data from the logic. Ultimately, classes like ColliderState should be a pure dataclasses that only store data. The initialization logic should go into the Collider or RigidSolver class. Maybe I’ll push to your PR and show you what I mean.

genesis/engine/solvers/rigid/collider_decomp.py
```diff
@@ -52,80 +54,122 @@ def rotmatx(matin, i0, i1, i2, f0, f1, f2):
 
 @ti.data_oriented
 class Collider:
+    @dataclass(frozen=True)
+    class ColliderInfo:
```
YilingQiao, line 58: Those should be static config. Let's address them in the next commit

genesis/engine/solvers/rigid/array_class.py
```diff
@@ -4,14 +4,140 @@
...
+        f_batch = solver._batch_shape
+        n_geoms = solver.n_geoms_
+        n_verts = solver.n_verts_
+        max_collision_pairs = min(solver._max_collision_pairs, n_possible_pairs)
+        max_contact_pairs = max_collision_pairs * collider_info.n_contacts_per_pair
+        use_hibernation = solver._static_rigid_sim_config.use_hibernation
+        box_box_detection = solver._static_rigid_sim_config.box_box_detection
+
+        ############## vertex connectivity ##############
+        self.vert_neighbors = ti.field(dtype=gs.ti_int, shape=max(1, n_vert_neighbors))
```
YilingQiao, line 57: Those should be in ColliderInfo. Let's address in a next PR.

PR #1378, opened 2025-07-09
[MISC] Add viewer code to handle moving objects

gasnica:
## Description
Objects can be held with LMB. 
When objects are held, camera rotation (via mouse-drag) is suspended.

Temporary implementation of object movement is added, where only position is updated, not velocities, so that causes interpenetrations of objects.

* Add simple box-moving interaction.
* Lazy load tensor in vec3.py, add Vec3/Quat.from_tensor/from_array.
* Rename oobb to obb.
* Vec3/Quat.as_tensor() is now a method.

## Motivation and Context

Allow moving of objects and kinematic chains by using external user force.

Manual testing only

image

## Checklist:

### DIFF
--- a/genesis/ext/pyrender/interaction/vec3.py
+++ b/genesis/ext/pyrender/interaction/vec3.py
@@ -4,9 +4,22 @@ from typing import Union
 import numpy as np
 from numpy.typing import NDArray
 
+from genesis.utils.misc import tensor_to_array
+
 # If not needing runtime checks, we can just use annotated types:
 # Vec3 = Annotated[npt.NDArray[np.float32], (3,)]
 # Aabb = Annotated[npt.NDArray[np.float32], (2, 3)]
+
+_is_torch_imported: bool = False
+
+def _ensure_torch_imported() -> None:
+    global _is_torch_imported
+    if not _is_torch_imported:
+        _is_torch_imported = True
+        global gs
+        import genesis as gs
+        global torch
+        import torch
 
 
 class Vec3:
@@ -68,32 +81,27 @@ class Vec3:
     def z(self) -> float:
         return self.v[2]
 
+    @property
+    def as_tensor(self) -> 'torch.Tensor':
+        _ensure_torch_imported()
+        return torch.tensor(self.v, dtype=gs.tc_float) 
+
     @classmethod
     def from_xyz(cls, x: float, y: float, z: float) -> 'Vec3':
         return cls(np.array([x, y, z], dtype=np.float32))
 
     @classmethod
-    def from_int32(cls, v: NDArray[np.int32]) -> 'Vec3':
+    def from_array(cls, v: np.ndarray) -> 'Vec3':
         assert v.shape == (3,), f"Vec3 must be initialized with a 3-element array, got {v.shape}"
-        assert v.dtype == np.int32, f"from_int32 must be initialized with a int32 array, got {v.dtype}"
+        assert v.dtype == np.int32 or v.dtype == np.int64 or v.dtype == np.float32 or v.dtype == np.float64, \
+            f"from_array must be initialized with a array of ints/floats 32/64-bit, got {v.dtype}"
         return cls.from_xyz(*v)
 
     @classmethod
-    def from_int64(cls, v: NDArray[np.int64]) -> 'Vec3':
-        assert v.shape == (3,), f"Vec3 must be initialized with a 3-element array, got {v.shape}"
-        assert v.dtype == np.int64, f"from_int64 must be initialized with a int64 array, got {v.dtype}"
-        return cls.from_xyz(*v)
-
-    @classmethod
-    def from_float64(cls, v: NDArray[np.float64]) -> 'Vec3':
-        assert v.shape == (3,), f"Vec3 must be initialized with a 3-element array, got {v.shape}"
-        assert v.dtype == np.float64, f"from_float64 must be initialized with a float64 array, got {v.dtype}"
-        return cls.from_xyz(*v)
-
-    @classmethod
-    def from_any_array(cls, v: np.ndarray) -> 'Vec3':
-        assert v.shape == (3,), f"Vec3 must be initialized with a 3-element array, got {v.shape}"
-        return cls.from_xyz(*v)
+    def from_tensor(cls, v: 'torch.Tensor') -> 'Vec3':
+        _ensure_torch_imported()
+        array: np.ndarray = tensor_to_array(v)
+        return cls.from_array(array)
 
 
     @classmethod
@@ -158,15 +166,26 @@ class Quat:
     def z(self) -> float:
         return self.v[3]
 
+    @property
+    def as_tensor(self) -> 'torch.Tensor':
+        _ensure_torch_imported()
+        return torch.tensor(self.v, dtype=gs.tc_float) 
 
     @classmethod
     def from_wxyz(cls, w: float, x: float, y: float, z: float) -> 'Quat':
         return cls(np.array([w, x, y, z], dtype=np.float32))
 
     @classmethod
-    def from_any_array(cls, v: np.ndarray) -> 'Quat':
+    def from_array(cls, v: np.ndarray) -> 'Quat':
         assert v.shape == (4,), f"Quat must be initialized with a 4-element array, got {v.shape}"
         return cls.from_wxyz(*v)
+
+    @classmethod
+    def from_tensor(cls, v: 'torch.Tensor') -> 'Quat':
+        _ensure_torch_imported()
+        array: np.ndarray = tensor_to_array(v)
+        return cls.from_array(array)
+
 
 
 @dataclass
--- a/genesis/ext/pyrender/interaction/vec3.py
+++ b/genesis/ext/pyrender/interaction/vec3.py
@@ -35,6 +35,9 @@ class Vec3:
     def __rmul__(self, other: float) -> 'Vec3':
         return Vec3(self.v * np.float32(other))
 
+    def __neg__(self) -> 'Vec3':
+        return Vec3(-self.v)
+
     def dot(self, other: 'Vec3') -> float:
         return np.dot(self.v, other.v).item()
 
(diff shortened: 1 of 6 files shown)


genesis/ext/pyrender/interaction/viewer_interaction.py
```diff
@@ -60,17 +56,55 @@ def on_mouse_motion(self, x: int, y: int, dx: int, dy: int) -> EVENT_HANDLE_STAT
...
+                delta_3d_pos: Vec3 = new_mouse_3d_pos - self.prev_mouse_3d_pos
+                self.prev_mouse_3d_pos = new_mouse_3d_pos
+
+                use_force: bool = False
+                if use_force:
+                    # apply force
+                    pass
+                else:
+                    #apply displacement
+                    pos = Vec3(self.picked_entity.get_pos().cpu().numpy())
```
duburcqa: You should rather use `tensor_to_array`.

genesis/ext/pyrender/interaction/viewer_interaction.py
```diff
@@ -3,16 +3,15 @@
 
 import numpy as np
 from numpy.typing import NDArray
-
-from pyglet.event import EVENT_HANDLE_STATE
+import torch
```
duburcqa: Could you delay torch import inside the function using it? Because it will initialise CUDA runtime, which depends on environment variables, so the later it is imported the more leeway it gives to the end-user for modifying these environment variables at runtime.

genesis/ext/pyrender/interaction/viewer_interaction.py
```diff
@@ -43,13 +34,18 @@ def __init__(self,
...
-        self.scene = scene
-        self.viewport_size = viewport_size
-        self.camera_yfov = camera_yfov
+        self.camera: 'Node' = camera
+        self.scene: 'Scene' = scene
+        self.viewport_size: tuple[int, int] = viewport_size
+        self.camera_yfov: float = camera_yfov
+
+        self.tan_half_fov: float = np.tan(0.5 * self.camera_yfov)
+        self.prev_mouse_pos: tuple[int, int] = tuple(np.array(viewport_size) / 2)
```
duburcqa: I would recommend `(0.5 * viewport_size[0], 0.5 * viewport_size[1])`

genesis/ext/pyrender/interaction/viewer_interaction.py
```diff
@@ -91,6 +125,20 @@ def on_draw(self) -> None:
...
                 self._draw_entity_unrotated_oobb(hit_entity)
 
+            if self.picked_entity:
+                assert self.mouse_drag_plane is not None
+                assert self.picked_point_in_local is not None
+
+                # draw held point
+                pose: Pose = self.get_pose_of_first_geom(self.picked_entity)
+                held_point: Vec3 = pose.transform_point(self.picked_point_in_local)
+                self.scene.draw_debug_sphere(held_point.v, 0.02, Color.red().tuple())
```
duburcqa, line 133: Why not just store pre-defined colours in a frozen dict of tuples whose keys would be some `IntEnum` (e.g. `COLOR.RED`) ? That would be more pythonic.

genesis/ext/pyrender/interaction/viewer_interaction.py
```diff
@@ -117,6 +165,10 @@ def screen_position_to_ray(self, x: float, y: float) -> Ray:
         direction = forward + right * x + up * y
         return Ray(position, direction)
 
+    def get_camera_forward(self) -> Vec3:
+        mtx = self.camera.matrix
+        return Vec3.from_float64(-mtx[:3, 2])
```
duburcqa: `Vec3.from_float64` is confusing, this is an array. Why not `from_array` or `from_tensor`? Why `float64` matters here ? You would do the same in both cases?

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: Oh, cool -- that sounds neat :-)

One thing we possibly could not do with a dict of tuples is modify the color inline, like this: `Color.red().with_alpha(0.5).tuple()` or is there a good way to do that?

Also if needed, a class can do color blending, or modify the color in HSV space easily.

Let me know whether we should keep it or change. I'll be happy with what you suggest :-)

duburcqa: > Color.red().with_alpha(0.5).tuple() or is there a good way to do that?

No you are right. If we want this feature, then your approach is probably the way to go.

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: Yeah, this was just for being verbose about underlying floating type change.

Oh, there's from_any_array() there already.

But I was thinking of adding .from_any() too :-) I'll change this.

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: I moved conversion to withing Vec3. And I'll do delayed import of genesis & torch there.

genesis/ext/pyrender/interaction/viewer_interaction.py
duburcqa: > Yeah, this was just for being verbose about underlying floating type change.

What about adding `dtype` optional argument, that would be `np.float64` by default? It would be more pythonic.

duburcqa: `def from_any(data: tp.Any, *, dtype: tp.Type[np.generic] = np.float64) -> Vec3:`

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: done (?) using Vec3.from_tensor

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: done

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: resolving as-is

genesis/ext/pyrender/interaction/vec3.py
```diff
@@ -4,10 +4,23 @@
...
 
+from genesis.utils.misc import tensor_to_array
+
 # If not needing runtime checks, we can just use annotated types:
 # Vec3 = Annotated[npt.NDArray[np.float32], (3,)]
 # Aabb = Annotated[npt.NDArray[np.float32], (2, 3)]
 
+_is_torch_imported: bool = False
+
+def _ensure_torch_imported() -> None:
```
duburcqa, line 14: You don't need to do this I think, you may just call `import torch` or `import genesis as gs`, it will be crazy fast if already imported before.

```python
In [19]: %timeit import torch
48.2 ns ± 0.452 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

I doubt you are chasing 50ns for now :p

genesis/ext/pyrender/interaction/vec3.py
```diff
@@ -158,16 +166,27 @@ def y(self) -> float:
     def z(self) -> float:
         return self.v[3]
 
+    @property
+    def as_tensor(self) -> 'torch.Tensor':
```
duburcqa: I don't like the idea of making this a proper instead of a classic method, because the result is a copy and not a reference, which is confusing for a property I think.

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: oh, nice -- and also new & confusing to me :-)

for now -- i just created from_array & from_tensor, and do runtime checks for array size/underlying type

gasnica: > What about adding dtype optional argument, that would be np.float64 by default?

... adding a dtype just to do checks on the input type would be a bit counter intuitive too, wouldn't it ?

genesis/ext/pyrender/interaction/vec3.py
gasnica: cool, thanks -- simplifying -- just keeping the function name

gasnica: and resolving

genesis/ext/pyrender/interaction/vec3.py
gasnica: oh, good point

and -- oh, is there a general assumption in python that a property is a reference ? 

removing @property

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: resolving, thread continued elsewhere

genesis/ext/pyrender/interaction/viewer_interaction.py
duburcqa: In my case it was to define the internal type being stored by `Vec3`, independently from the input type.

genesis/ext/pyrender/interaction/vec3.py
duburcqa: > oh, is there a general assumption in python that a property is a reference ?

Not necessarily but fairly common since `property` kinda imply "no processing" contract. For something like a vector abstraction, It would be reasonable to assume that `.as_tensor` returns the underlying storage. This is partly mitigated by `as_` prefix, but still I think that the benefits of removing the property decorator outweighs to drawbacks here.

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: Gotcha. 

My intention would be to always use consistent type for all Vec3's across the build. So if at all -- this could internally set array-element type gs.ti_float

Thank you for all suggestions and hints, Alexis :-)

I'm assuming we're good to leave it as is (with .from_array() and .from_tensor() added now)

Resolving

duburcqa:
Ready to merge after rebase.

genesis/ext/pyrender/interaction/vec3.py
gasnica: That makes very good sense. 

And true to the `as_` prefix. Leaving the prefix as-is for now, unless you prefer more verbose but longer `get_as_tensor`. 

Changed the property to a method.

Resolving

gasnica:
and rebased.
tests run ok locally a bit earlier.

Issue #1362, opened 2025-07-05
[Feature]: Sensor API & Tactile Force Sensing


Milotrince:
### What feature or enhancement are you proposing?

Add support for a kind of generic `SensorEntity` that extract data from its environment, and visualize the collected data live.
Then, implement a TactileSensor which collects 6-axis force/torques from the solver.

### Motivation

Sensor support has been on the roadmap and is a popularly requested feature: https://github.com/Genesis-Embodied-AI/Genesis/issues/55

For tactile: there is not yet an accurate, straightforward-to-use simulator for tactile sensors available. 

Genesis could be the go-to platform for roboticists who use tactile sensors. 
There's a lot of different types of tactile sensors, like:
- Vision-based (aka visuo-tactile or optical)
  - GelSight, Meta's DIGIT
- Magnetic
  - Melexis's Tactaxis
- Capacitative
  - XELA's uSkin
- Mechanical
- Piezoresistive

Each has their pros/cons but in the end, all of these methods are trying to approximate force.
The ability to measure/extract forces in Genesis would hopefully open up possibilities for sim-to-real with tactile sensors.

### Potential Benefit

Genesis could be the de-facto simulation platform for tactile sensing 

### What is the expected outcome of the implementation work?

- [ ] Sensor API
  - [ ] Data collection
  - [ ] Visualization
- [ ] Example sensors
  - [ ] ContactSensor
  - [ ] ForceSensor

### Additional information

I will be working on this!

PR #1381, opened 2025-07-09
[FEATURE] Sensors

Milotrince:
## Description
Add support for sensors.
Sensors observe the scene, which means they mostly just extract existing properties from solvers. But since we want this data often, implemented sensors should make this as efficient as possible. The base sensor class also provides methods to stream and collect the data.
- [x] Added base `Sensor` class
- [x] Added example sensors
    - [x] Rigid tactile
    - [ ] Distance (LiDAR) -> Future PR (@gasnica ?)
    - [x] Migrate `Camera` as a `Sensor`
        - Temporary implementation applied. I suggest refactoring `Camera` into several classes (`RasterizerCamera`, `RaytracerCamera`, ...) in a future PR.
- [x] Added utility `SensorDataRecorder` class
- [x] Example sensor usage script
  - [ ] RL in-hand rotation with tactile sensors demo
      - could be in a separate PR?
- [x] Docs PR: https://github.com/Genesis-Embodied-AI/genesis-doc/pull/61

## Related Issue
https://github.com/Genesis-Embodied-AI/Genesis/issues/1362

## Motivation and Context
Sensors are an important part of robotics simulation.

## How Has This Been Tested?
Full example with a RigidContactForceGridSensor: 

https://github.com/user-attachments/assets/67a38b15-c40f-4560-b42c-3f9347f131a3

with allegro hand: [script](https://gist.github.com/Milotrince/62e9a53f4ec5b46410ca4a5083db819f)

https://github.com/user-attachments/assets/cace529e-cf6b-425c-b719-3b7ab25ef428

## Checklist:

- [x] I updated the [documentation](https://github.com/Genesis-Embodied-AI/genesis-doc) accordingly: https://github.com/Genesis-Embodied-AI/genesis-doc/pull/61
- [x] I have added ~~tests to cover my changes~~ example scripts.


Kashu7100:
Thanks for initiating this! I have a suggestion regarding the performance of the Sensor.read method, which might be beneficial to discuss.

At present, `read()` computes the sensory readings every time it is invoked. In a robot learning context, where this method might be called multiple times per simulation step, this could introduce non-trivial overhead.

My suggestion is to implement a caching mechanism or a buffer for the Sensor class. This would store the computed sensory reading for the current `step()`, and subsequent calls to `read()` within the same step would return the cached value. This would prevent redundant calculations.

Also, I am thinking there might be potential issue with non-negligible overhead if there are multiple sensors attached to an entity that performs the similar computation (that could be shared) but recomputes because of the Sensor abstraction. If there is a way to exchange some intermediate states (buffers?) among the sensors, that might be good! 
This is not very straight forward how to do it (something like sensor manager can be implemented?). For my personal repo, I have Robot abstraction with buffers and each sensors are just returning the values from the buffers (with minimal compute).

Happy to discuss!

Milotrince:
@Kashu7100 Agreed, I've been thinking about this. Would implementing caching/batching on the class level be acceptable? e.g. `@classmethod`s for `build()`, `update_buffer()`. I want to avoid having a separate `SensorManager` for every sensor type

Kashu7100:
We need to think if we can share the class level buffers among sensors.

Milotrince:
@Kashu7100 
> We need to think if we can share the class level buffers among sensors.

Yeah they can. Each sensor can access its own class properties. What do you think about [this implementation](https://github.com/Genesis-Embodied-AI/Genesis/blob/3a2558e87cfd0eae0aa688d1fd07b30463308a12/genesis/sensors/tactile.py)?

Kashu7100:
Yeah this looks nice! I think this is promising

yun-long:
High-Level Comments on `sensors`

- **Separation from Entity Abstraction**:
Ideally, `sensors` should be independent of Entity. According to the current definition:

> An Entity object is the abstraction of everything that requires physics simulation (except for rendering) in a scene. This includes rigid or deformable objects, liquids, etc., simulated with various physics solvers such as MPM, PBD, etc.

Coupling sensors to Entity is conceptually inconsistent.

- **`Sensor` Placement and API Layering**:
Genesis is a simulator at its core. The composition of a robot or object that includes sensors should happen at a higher-level API (e.g., within an RL environment). From a software architecture perspective, this keeps the simulator core clean and modular.

- **Data Handling Responsibility**:
`Sensors` typically do not function as data recorders. They provide time-stamped observations of the current state, but do not persist historical data. Logging or data storage should be handled at a higher layer (e.g., environment wrappers, logging utilities), not within the sensor implementation itself.

Milotrince:
re: @yun-long 
> Separation from Entity Abstraction

I actually think under the definition of entity given, it makes sense that sensors are associated with an entity. Some sensors (like tactile) are highly dependent on the kind of entity, and in fact, the physical object itself *is* the sensor. Cameras and distance (LiDAR) sensors can also still take physical form (they are very much solid objects in the real world!) and even if the collision/mesh doesn't matter, they have position/rotation.

But I can also make sensor "own" an entity instead, which would also address:

> Sensor Placement and API Layering

And as for

> Data Handling Responsibility

I agree with this, but I was actually following the pattern of the existing `Camera` class which handles its own data recording. I can also make `DataCollector` own `Sensor` instead of the other way around.

@duburcqa @YilingQiao thoughts on which makes more sense?
1. `Entity` has `Entity.sensors` and `Sensor` has `DataCollector`
```python
sensor = entity.add_sensor(...)  # can also do scene.add_sensor(...) which adds Static entity with sensor
sensor.start_recording(...)

for i in range(steps):
    scene.step()
    
    value = sensor.read()
```
2. `Sensor` has `Sensor.entity` and `DataCollector` has `Sensor`(s)
```python
sensor = Sensor(entity, ...)
data_collector = DataCollector(sensors, ...)

for i in range(steps):
    scene.step()
    data_collector.update()

    value = sensor.read()
```

I think both are fine

Milotrince:
Ok we talked it over in a meeting, going with option 2 (`Sensor` has `Sensor.entity`).
The main/first comment of this issue has been updated

Milotrince:
I wanted to add a RL in-hand rotation with tactile sensors demo, should that be in a separate PR or with this one? can I add files to `assets` (allegro hand urdf)? didnt add it yet because it adds a lot of bloat

Kashu7100:
I would say the RL example will go to a separate PR (each PR is preferred to be atomic). I think new assets are now need to be hosted on HF not here to keep Genesis repo light. You can ask @duburcqa to add you to the HF or ask him to upload your assets.

YilingQiao (dismissed):
looks good to me

genesis/sensors/data_collector.py
```diff
@@ -0,0 +1,191 @@
+import genesis as gs
+import threading
+import queue
+from dataclasses import dataclass
+from .data_handlers import DataHandler
+from .base_sensor import Sensor
+
+
+@dataclass
+class RecordingOptions:
```
Kashu7100: this should go to under gs.options?

genesis/sensors/data_collector.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+class SensorDataRecorder:
+    """
+    Utility class to automatically collect data from sensors and process it using specified handlers.
+
+    Parameters
+    ----------
+    sensor: gs.sensors.Sensor
+
```
Kashu7100: doc string for other args missing

genesis/sensors/tactile.py
```diff
@@ -0,0 +1,214 @@
+import torch
+import taichi as ti
+import genesis as gs
+from genesis.engine.entities import RigidEntity
+from genesis.utils.misc import tensor_to_array
+from genesis.utils.geom import ti_inv_transform_by_quat, inv_transform_by_quat, transform_by_trans_quat
+from typing import Dict, List, Any, Optional
+from .base_sensor import Sensor
+import numpy as np
```
Kashu7100: The import ordering.

genesis/sensors/data_handlers.py
```diff
@@ -0,0 +1,252 @@
+import genesis as gs
+import cv2
+import csv
+import os
+import numpy as np
+from typing import Optional
+from collections.abc import Iterable
+from genesis.utils.tools import animate
```
Kashu7100, line 11: The import order.

genesis/sensors/data_collector.py
```diff
@@ -0,0 +1,191 @@
+import genesis as gs
+import threading
+import queue
+from dataclasses import dataclass
+from .data_handlers import DataHandler
+from .base_sensor import Sensor
```
Kashu7100: What I learned from Alexis: You should first import standards libraries, skip line, then ubiquitous third-party libraries (torch, numpy, tqdm...etc), skip line, then random third-party libraries, skip line, then our own custom libraries, skip line, then relative genesis imports. For each group, make sure it is sorted alphabetically.

genesis/sensors/data_collector.py
Milotrince: sure

genesis/sensors/data_collector.py
Milotrince: oops thanks, added!

genesis/sensors/data_handlers.py
Milotrince: i wish black formatter did auto organize imports too.. do you happen to have auto format set up for imports?

genesis/sensors/data_collector.py
Milotrince: no way you're doing this manually o_o

Milotrince: isort vs code extension does Organize Imports; I haven't set it up to do it automatically though

genesis/sensors/data_handlers.py
Kashu7100: no... I wish too

genesis/sensors/data_collector.py
Kashu7100: let me know if you do that!

YilingQiao (dismissed):

YilingQiao:
I'm okay with the current PR. Do you want to have a quick look? @duburcqa

examples/sensors/tactile/tactile_fingertips.py
```diff
@@ -0,0 +1,188 @@
...
+
+    ########################## scene setup ##########################
+    scene = gs.Scene(
+        sim_options=gs.options.SimOptions(
+            dt=args.dt,
+        ),
+        profiling_options=gs.options.ProfilingOptions(
+            show_FPS=False,
+        ),
+        rigid_options=gs.options.RigidOptions(use_gjk_collision=True),
```
YilingQiao: can you add in rigid_options

```
    constraint_timeconst=max(0.01, 2 * args.dt / args.substeps),
```

YilingQiao:
I'll run the CI and merge it. Please add some unit tests in your next PR with the kernel version of the sensors.

YilingQiao (approved):

PR #1385, opened 2025-07-10
[FEATURE] expose joint position query

Kashu7100:
## Description

Add two new API `get_joints_anchor_pos` and `get_joints_anchor_axis` that returns the joint position in world coordinate and the axis direction in unit vector. 

## Related Issue

## Checklist:

- [] I updated the [documentation](https://github.com/Genesis-Embodied-AI/genesis-doc) accordingly or no change is needed.

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -2444,6 +2444,46 @@ class RigidEntity(Entity):
         return self._solver.get_qpos(qs_idx, envs_idx, unsafe=unsafe)
 
     @gs.assert_built
+    def get_joints_anchor_pos(self, joints_idx_local=None, envs_idx=None, *, unsafe=False):
+        """
+        Returns anchor position of the entity's joints. This is the position of the joint in the world frame.
+
+        Parameters
+        ----------
+        joints_idx_local : None | array_like, optional
+            The indices of the dofs to get. If None, all dofs will be returned. Note that here this uses the local `q_idx`, not the scene-level one. Defaults to None.
+        envs_idx : None | array_like, optional
+            The indices of the environments. If None, all environments will be considered. Defaults to None.
+
+        Returns
+        -------
+        anchor_pos : torch.Tensor, shape (n_joints, 3) or (n_envs, n_joints, 3)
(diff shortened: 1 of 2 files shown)


duburcqa:
Can you explain what is the use-case for this?

As already discussed several times in the past, I think exposing the “position” of joints is a bad idea because it is error prone. Most of the user will not understand what this is and will use it incorrectly. In my view, only link and geom poses should be exposed to the user.

Kashu7100:
I was thinking of using this when I convert some code using different simulator but I think I can use the links to do the same!

Kashu7100:
(Maybe) `get_joints_anchor_pos` is still worth keeping it. Feel free to reopen if needed.

duburcqa:
Now that I remember, there is already `RigidJoint.get_anchor_pos` and `RigidJoint.get_anchor_axis`.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -2443,6 +2443,46 @@ def get_qpos(self, qs_idx_local=None, envs_idx=None, *, unsafe=False):
         qs_idx = self._get_idx(qs_idx_local, self.n_qs, self._q_start, unsafe=True)
         return self._solver.get_qpos(qs_idx, envs_idx, unsafe=unsafe)
 
+    @gs.assert_built
+    def get_joints_anchor_pos(self, joints_idx_local=None, envs_idx=None, *, unsafe=False):
+        """
+        Returns anchor position of the entity's joints. This is the position of the joint in the world frame.
```
duburcqa, line 2449: This means nothing: "This is the position of the joint in the world frame."

duburcqa: Here is the actual definition: https://github.com/Genesis-Embodied-AI/Genesis/blob/32904b3c1bc3a2840dae268d70509501e677193a/genesis/engine/entities/rigid_entity/rigid_joint.py#L104

duburcqa:
After think twice, maybe we could factorize the implementations at joint-level with what you did? If you do so, I'm open to merge it!

duburcqa:
Closing since there has been no activity since quite a while.

Issue #1383, opened 2025-07-09
[Bug]: Weld constraint does not work for multi-environment


abhijitmajumdar:
### Bug Description

When using weld constraint to simulate suction cup on robot, it does not work for all robots when running multi-environment. The [source code](https://github.com/Genesis-Embodied-AI/Genesis/blob/39a2951007a8acbbe0fc73e45a66144e157218ac/genesis/engine/solvers/rigid/rigid_solver_decomp.py#L4916) indicates that there is a provision to specify `env_idx`, implying weld constraint _should_ work across multi environment setups TMU. I believe there might be a bug where it is not taking effect, or I may be using the weld constraint for multi environment incorrectly. Please advise

### Steps to Reproduce

```py
# Demo: suction not working in multi-env setup
import genesis as gs
import numpy as np

def n_conv(vec, num):
    return np.tile(np.array(vec), [num, 1])

def main():
    gs.init()

    scene = gs.Scene(
        show_viewer=True,
        viewer_options= gs.options.ViewerOptions(
            max_FPS       = 30,
        ),
    )
    plane = scene.add_entity(
        gs.morphs.Plane(),
    )
    cube = scene.add_entity(
        gs.morphs.Box(
            size=(0.04, 0.04, 0.04),
            pos=(0.65, 0.0, 0.02),
        ),
        surface=gs.surfaces.Plastic(color=(1, 0, 0)),
    )
    ur = scene.add_entity(
        gs.morphs.MJCF(file="xml/universal_robots_ur5e/ur5e.xml"),
    )
    NENV = 4
    scene.build(NENV, (3., 3.))

    end_effector = ur.get_link("ee_virtual_link")

    # move to pre-grasp pose
    qpos = ur.inverse_kinematics(
        link=end_effector,
        pos=n_conv(np.array([0.65, 0.0, 0.5]), NENV),
        quat=n_conv(np.array([0, 1, 0, 0]), NENV),
    )
    ur.set_qpos(qpos)
    ur.control_dofs_position(qpos)
    for i in range(200):
        scene.step()

    # reach
    qpos = ur.inverse_kinematics(
        link=end_effector,
        pos=n_conv(np.array([0.65, 0.0, 0.140]), NENV),
        quat=n_conv(np.array([0, 1, 0, 0]), NENV),
    )
    ur.control_dofs_position(qpos)
    print('reaching')
    for i in range(100):
        scene.step()

    # add suction / weld constraint
    print('adding suction')
[...]

PR #1386, opened 2025-07-10
[FIX BUG]: Weld constraint in multi-env 

LeonLiu4:
If envs_idx contains multiple environments but link1_idx/link2_idx are length-1, we now np.repeat those indices so their length matches envs_idx. This prevents silent out-of-bounds reads that dropped welds in all but the first environment. 

Addresses: #1383


genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -4920,6 +4920,19 @@ def add_weld_constraint(self, link1_idx, link2_idx, envs_idx=None, *, unsafe=Fal
         _, link2_idx, envs_idx = self._sanitize_1D_io_variables(
             None, link2_idx, self.n_links, envs_idx, idx_name="links_idx", skip_allocation=True, unsafe=unsafe
         )
+
```
duburcqa, line 4923: Do not cast from torch to numpy, this is irrelevant.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -4920,6 +4920,19 @@ def add_weld_constraint(self, link1_idx, link2_idx, envs_idx=None, *, unsafe=Fal
...
+        if torch.is_tensor(link1_idx):
+            link1_idx = link1_idx.cpu().numpy()
+        if torch.is_tensor(link2_idx):
+            link2_idx = link2_idx.cpu().numpy()
+        if envs_idx is not None and torch.is_tensor(envs_idx):
+            envs_idx = envs_idx.cpu().numpy()
+
+        if envs_idx is not None and envs_idx.shape[0] > 1:
+            if link1_idx.shape[0] == 1:
+                link1_idx = np.repeat(link1_idx, envs_idx.shape[0])
```
duburcqa, line 4933: Use torch.repeat

duburcqa: By the way, no need to check if it is a tensor. It has been sanitized as torch tensor already.

duburcqa:
Why the sanitize helper was not able to catch this dimensionality issue ? This is basically the whole point of this method.

duburcqa:
Superseded by https://github.com/Genesis-Embodied-AI/Genesis/pull/1390

PR #1394, opened 2025-07-10
[MISC] Add Generic GPU CI.

duburcqa:


duburcqa:
Let's stick to what we have for now. Adding cross-platform GPU runner will significantly increase the cost of the CI for limited advantage.

Issue #1112, opened 2025-05-06
[Feature]: One-sided plane


nootnootpengupengu:
### What feature or enhancement are you proposing?

Is it possible for the default plane to render as one-sided?

### Motivation

One sided plane makes it easier to visualize what's going on in my application. 

### Potential Benefit

Improved visualisation

### What is the expected outcome of the implementation work?

A boolean field option to add to `gs.morphs.Plane` that lets me set `render_as_one_sided`.

PR #1396, opened 2025-07-10
Feature/one sided plane

LeonLiu4:
create_plane() now accepts one_sided=True, returning a single-sided quad with the same footprint/UVs.

Addresses #1112

### DIFF
--- a/genesis/options/morphs.py
+++ b/genesis/options/morphs.py
@@ -331,10 +331,14 @@ class Plane(Primitive):
         Whether the entity needs to be visualized. Set it to False if you need a invisible object only for collision purposes. Defaults to True. `visualization` and `collision` cannot both be False. **This is only used for RigidEntity.**
     collision : bool, optional
         Whether the entity needs to be considered for collision checking. Defaults to True. `visualization` and `collision` cannot both be False. **This is only used for RigidEntity.**
+    one_sided : bool, optional
+        Whether the entity needs to be rendered as a one-sided or two-sided plane.
     """
 
     fixed: bool = True
     normal: tuple = (0, 0, 1)
+    one_sided: bool = False
+    double_sided: bool | None = None
 
     def __init__(self, **data):
         super().__init__(**data)
@@ -347,6 +351,8 @@ class Plane(Primitive):
 
         if self.requires_jac_and_IK:
             gs.raise_exception("`requires_jac_and_IK` must be False for `Plane`.")
+
+        self.double_sided = not self.one_sided
 
         self.normal = tuple(np.array(self.normal) / np.linalg.norm(self.normal))
 
--- a/genesis/utils/mesh.py
+++ b/genesis/utils/mesh.py
@@ -792,32 +792,60 @@ def create_cylinder(radius, height, sections=None, color=(1.0, 1.0, 1.0, 1.0)):
     return mesh
 
 
-def create_plane(size=1e3, color=None, normal=(0, 0, 1)):
+def create_plane(size=1e3, color=None, normal=(0, 0, 1), one_sided: bool = False):
     thickness = 1e-2  # for safety
-    mesh = trimesh.creation.box(extents=[size, size, thickness])
+    if one_sided:
+        half = size * 0.5
+        verts = np.array(
+            [
+                [-half, -half, 0.0],
+                [half, -half, 0.0],
+                [half, half, 0.0],
+                [-half, -half, 0.0],
+                [half, half, 0.0],
+                [-half, half, 0.0],
+            ],
+            dtype=np.float32,
+        )
+        faces = np.arange(6, dtype=np.int32).reshape(-1, 3)
+        uv = np.array(
+            [
+                [0, 0],
+                [size, 0],
+                [size, size],
+                [0, 0],
+                [size, size],
+                [0, size],
+            ],
+            dtype=np.float32,
+        )
+        mesh = trimesh.Trimesh(verts, faces, process=False)
+    else:
+        mesh = trimesh.creation.box(extents=[size, size, thickness])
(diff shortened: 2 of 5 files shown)


genesis/options/morphs.py
```diff
@@ -331,10 +331,14 @@ class Plane(Primitive):
         Whether the entity needs to be visualized. Set it to False if you need a invisible object only for collision purposes. Defaults to True. `visualization` and `collision` cannot both be False. **This is only used for RigidEntity.**
     collision : bool, optional
         Whether the entity needs to be considered for collision checking. Defaults to True. `visualization` and `collision` cannot both be False. **This is only used for RigidEntity.**
+    one_sided : bool, optional
+        Whether the entity needs to be rendered as a one-sided or two-sided plane.
     """
 
     fixed: bool = True
     normal: tuple = (0, 0, 1)
```
duburcqa: Why 2 flags ? Only define one or the other. Moreover, I think it should be one-sided by default.

LeonLiu4: fixed. Though the one sided plane renders a much darker top than two sided. Not sure what is causing this

duburcqa: Hum... There must be something wrong. It should be fixed before merging.

LeonLiu4: I changed the logic a bit and now I think it works at least the one sided plane is not too dark.

YilingQiao: can you post a image comparison between one-sided vs two-sided? are they the same?

LeonLiu4: yes

genesis/options/morphs.py
```diff
@@ -348,6 +351,8 @@ def __init__(self, **data):
         if self.requires_jac_and_IK:
             gs.raise_exception("`requires_jac_and_IK` must be False for `Plane`.")
 
+        self.double_sided = self.double_sided
```
duburcqa: This is useless. It will be done automatically.

genesis/options/morphs.py
```diff
@@ -331,10 +331,13 @@ class Plane(Primitive):
         Whether the entity needs to be visualized. Set it to False if you need a invisible object only for collision purposes. Defaults to True. `visualization` and `collision` cannot both be False. **This is only used for RigidEntity.**
     collision : bool, optional
         Whether the entity needs to be considered for collision checking. Defaults to True. `visualization` and `collision` cannot both be False. **This is only used for RigidEntity.**
+    double_sided : bool, optional
```
duburcqa: Terrain morph should also have this option.

genesis/vis/rasterizer_context.py
```diff
@@ -296,9 +296,7 @@ def on_rigid(self):
                             mesh=mesh,
                             poses=geom_T,
                             smooth=geom.surface.smooth if "collision" not in rigid_entity.surface.vis_mode else False,
-                            double_sided=(
-                                geom.surface.double_sided if "collision" not in rigid_entity.surface.vis_mode else False
-                            ),
+                            double_sided=rigid_entity.morph.double_sided,
```
duburcqa: I don't think this change is a good idea.

genesis/options/morphs.py
duburcqa: By the way, adding this option at morph-level should be avoided. You should rather change the default here:
https://github.com/Genesis-Embodied-AI/Genesis/blob/40eac852e464d1fe25ccdef9041273f2f04c4b4f/genesis/engine/scene.py#L310

By I think it is already properly set. So nothing to change.

genesis/options/morphs.py
duburcqa: Looking good I think.

genesis/utils/mesh.py
```diff
@@ -860,24 +860,36 @@ def create_box(extents=None, color=(1.0, 1.0, 1.0, 1.0), bounds=None, wireframe=
...
+            [half, -half, 0.0],
+            [half, half, 0.0],
+            [-half, -half, 0.0],
+            [half, half, 0.0],
+            [-half, half, 0.0],
+        ],
+        dtype=np.float32,
+    )
+    faces = np.arange(6, dtype=np.int32).reshape(-1, 3)
+    uv = np.array(
```
duburcqa: Move back `uv` where it was initially defined.

LeonLiu4: fixed

genesis/vis/rasterizer_context.py
LeonLiu4: reverted

genesis/options/morphs.py
LeonLiu4: reverted

genesis/options/morphs.py
LeonLiu4: fixed

duburcqa:
Is this patch all that was needed to fix one-sided plane?! That would be awesome :)

genesis/options/morphs.py
LeonLiu4: will work on terrain asw

LeonLiu4:
> Is this patch all that was needed to fix one-sided plane?! That would be awesome :)

Unfortunately that patch created a mesh that was not a solid volume causing there to be errors with stuff like collisions. However I have rectified that by differentiating between a visual and collision mesh

duburcqa:
@LeonLiu4 Can you confirm this PR is ready to merge and the videos in description are still valid?

duburcqa (approved):

LeonLiu4:
> @LeonLiu4 Can you confirm this PR is ready to merge and the videos in description are still valid?

Yes the videos are correct and its ready to merge

PR #1413, opened 2025-07-14
[BUG FIX] Fix draw_debug tutorial types

RaiaN:
## Description
Simple fix for `draw_debug` tutorial types as it currently seems to be broken due to the following error:

```
assert trans.shape[:-1] == R.shape[:-2]
AttributeError: 'tuple' object has no attribute 'shape'
```

## Motivation and Context
Looking to contribute to Genesis development. First step is to ensure all tutorials are working fine.

Run `draw_debug` script without errors.


YilingQiao:
Thanks, I will merge it after a few other PRs are merged. Then I can update it to the main branch.

duburcqa (changes requested):
Could you rather modify `draw_debug_line` to systematically cast input to numpy using `np.asarray(start, dtype=np.float32)` (same for `end` and probably color?).

PR #1416, opened 2025-07-15
[FEATURE] Integrating Madrona Batch Renderer

yuhongyi:
This PR migrates all Madrona batch rendering changes from https://github.com/genesis-company/Genesis, which is temporary to the official repo.  All the code has already been reviewed before except for a few changes I commented in this PR to resolve merge conflicts. See [this link](https://github.com/genesis-company/Genesis/pulls?q=is%3Apr+is%3Aclosed) for previous PRs.

## Description

This PR add support of batch rendering for Genesis.

## Batch Renderer

- **[New API]** Enable batch renderer
    
    ```
        scene = gs.Scene(
    				...
            renderer = gs.options.renderers.BatchRenderer(
                use_rasterizer=True,
            )
        )
    ```
    
    - use_rasterizer - If true, use rasterizer for batch renderer, otherwise raytracer will be used. 
    Default to false
- **[New API]** Batch render - `Scene.render_all_cameras()`. A tuple of 4-d arrays will be returned by this function, with len() of the tuple being the number of cameras, first dimension of each tensor being environment/world.  Right now only rgb and depth is support.  Normal and segmentation output will be supported later
- **[New Behavior]** Single-camera render - `Camera.render()` still works, and it will return a tensor with output only from the camera from which render() is called.  In the case of n_envs >0, the first dimension is the environment. When `Camera.render()` is called, actually `Scene.batch_render()` will be invoked internally, with outputs from all cameras cached, which means if `Camera.render()` is later called from another camera, the output will be returned immediately without any additional rendering.  This behavior may change later to only render for the requesting camera.

## Scene

- When n_envs == 0 or is not specified, treat it as n_envs == 1 for batch renderer. The first dimension of outputs from camera.render() or scene.batch_render() is the environment.

## Lights

- **[New API]** Add a second `Scene.add_light()` with a new signature to add non-mesh light.  This `add_light()` can be used to add point lights, spot lights and directional lights, which are supported by the lighting model of the batch renderer.
- The signature of the new add_light() function

```
    def add_light(
        self,
        pos,
        dir,
        intensity,
        directional,
        castshadow,
        cutoff,
    ):
```

- Light attenuation and light color will be supported in future updates.
- There are 2 `add_light()` functions now, with different signatures.  If BatchRenderer is used, only the new add_light() can be used, otherwise, there will be a warning and lights won’t be added.
- If BatchRenderer is not used, only the old `add_light()` can be used, otherwise, there will be a warning and lights won’t be added.
- The old `add_light()` might be renamed to `add_mesh_light()` in future and be supported in the ray tracing pipeline in the batch renderer, but right now only directional lights, point lights and spot lights are supported.
- Non-batch rasterizer uses the lights defined in VisOptions.  Now we have 3 different places adding lights for different pipelines, which is confusing and hard to maintain.  The APIs to add lights will be unified in future.

## Shadow
- Shadow will be casted from lights added with `castshadow=true`.  The raytracer pipeline of batch rendering can cast shadow from all lights, while the rasterizer pipeline can only cast shadow from the first light added with shadow casting enabled.

## Camera

- Cameras can be added by the existing `Scene.add_camera()` function, but aperture and focus_dis is not supported right now.
- **[New Benavior]** The resolution of batch camera now uses the resolution of the first camera.  Batch render with cameras with different resolutions will be supported in future.
- **[New Feature]** Attach cameras to links - Attaching camera to links is now supported with batch rendering.  Each camera can only be attached to one link, but the camera positions will be updated for each environment.
- **[New Feature]** Camera follows entity - Following entity is now supported with batch rendering.  Each camera an only follow one entity, but camera positions will be updated for each environment.
- **[New API]** Support multi-environment pose update for cameras by adding a new parameter `env_idx` to `Camera.set_pose()` to specify the indices of environments to update camera transforms.
- Transform updates from attaching cameras to links supersedes updates from following entities.

## Related Issue

Resolves Genesis-Embodied-AI/Genesis/issues/1232
Resolves Genesis-Embodied-AI/Genesis/issues/21

Tests have been done with example script single_franka.py and single_franka_batch_render.py, to cover combinations of
- batch rasterizer / batch raytracer / non-batch rasterizer / non-batch raytracer
- n_envs > 0 / n_envs == 0
- camera attached to link / camera not attached to link
- rgb / depth
- [Performance benchmark] (https://www.notion.so/Madrona-benchmark-results-2070c06194e181b58753d2362c174bce)

rgb_cam1_env0_000 (Copy)

## Checklist:


genesis/vis/rasterizer_context.py
```diff
@@ -875,7 +875,7 @@ def create_node_seg(self, seg_key, seg_node):
             self.seg_node_map[seg_node] = self.seg_idxc_to_idxc_rgb(seg_idxc)
 
     def remove_node_seg(self, seg_node):
-        self.seg_node_map.pop(seg_node)
+        self.seg_node_map.pop(seg_node, None)
```
yuhongyi: Fixed a crash here in case seg_node is not a valid key.
This is more a workaround to prevent fix.  We need to find out why seg_node could be an invalid key.

tests/test_render.py
```diff
@@ -279,8 +279,8 @@ def test_batched_mounted_camera_rendering(show_viewer, tol):
     T = np.eye(4)
     T[:3, :3] = np.array([[1, 0, 0], [0, -1, 0], [0, 0, -1]])
     T[:3, 3] = np.array([0.1, 0.0, 0.1])
-    for nenv, cam in enumerate(cams):
-        cam.attach(robot.get_link("hand"), T, nenv)
+    for cam in cams:
+        cam.attach(robot.get_link("hand"), T)
```
yuhongyi: Passing camera index as env index is wrong.
The new camera attach implementation in this PR allows attaching a camera to the same link and update the camera location to follow the position of the link in each environment independently, so env index is not needed any more.

genesis/vis/camera.py
```diff
@@ -138,26 +145,9 @@ def attach(self, rigid_link, offset_T, env_idx: int | None = None):
...
-
-        Raises
-        ------
-        Exception
-            If running parallel simulations but env_idx is not specified.
-        Exception
-            If invalid env_idx is specified (env_idx >= n_envs)
         """
         self._attached_link = rigid_link
-        self._attached_offset_T = offset_T
```
yuhongyi, line 155: Someone implemented a attaching camera 2 weeks ago, but the implementation is limited to only attach to the location of one environment, so this implementation is removed with the implementation in this PR, which allows attaching cameras to the same rigid link and update their location independently in each environment.

genesis/vis/visualizer.py
```diff
@@ -162,8 +182,9 @@ def build(self):
                 self._viewer.update(auto_refresh=True)
             else:
                 # viewer creation will compile rendering kernels if viewer is not created, render here once to compile
-                self._rasterizer.update_scene()
-                self._rasterizer.render_camera(self._cameras[0])
+                if not self._use_batch_renderer:
+                    self._rasterizer.update_scene()
```
yuhongyi: @AnisB I moved the update_scene() you added in a previous PR under the if-statement, since when batch renderer is enabled, _rasterizer, which is the legacy pyrender renderer, will not be used

genesis/utils/geom.py
```diff
@@ -1136,8 +1136,61 @@ def scale_to_T(scale):
...
 
 
+def z_up_to_R(z, up=np.array([0, 0, 1]), out=None):
+    if isinstance(z, torch.Tensor):
+        return _tc_z_up_to_R(z, up, out)
+    else:
+        return _np_z_up_to_R(z, up, out)
+
+
+def _tc_z_up_to_R(z, up=None, out=None):
```
yuhongyi: Added back the torch implementation.
Will do a thorough benchmark to find out which is faster.

- Numba + CPU->GPU memory copy
- Torch tensor + all GPU memory

genesis/vis/rasterizer_context.py
yuhongyi: Without the workaround here, it still crashes even after merging the latest main, which includes a fix in https://github.com/Genesis-Embodied-AI/Genesis/pull/1426, so I'll keep it for now.

genesis/vis/visualizer.py
duburcqa: Yes, this is the way to go.

yuhongyi: Skipped `_rasterizer.render_camera()` when batch renderer is enabled, as we discussed.

README.md
```diff
@@ -172,6 +172,8 @@ Genesis's development has been made possible thanks to these open-source project
 - [libccd](https://github.com/danfis/libccd): Reference for collision detection.
 - [PyRender](https://github.com/mmatl/pyrender): Rasterization-based renderer.
 - [LuisaCompute](https://github.com/LuisaGroup/LuisaCompute) and [LuisaRender](https://github.com/LuisaGroup/LuisaRender): Ray-tracing DSL.
+- [Madrona](https://github.com/shacklettbp/madrona): Madrona batch renderer
+- [Madrona-mjx](https://github.com/shacklettbp/madrona_mjx): Madrona MJX integration
```
YilingQiao: can we put the in the same line?

[Madrona](https://github.com/shacklettbp/madrona) and [Madrona-mjx](https://github.com/shacklettbp/madrona_mjx): Batch renderer backend

examples/rigid/single_franka_batch_render.py
```diff
@@ -0,0 +1,98 @@
...
+
+import genesis as gs
+from genesis.options.renderers import BatchRenderer
+from genesis.utils.geom import trans_to_T
+from genesis.utils.image_exporter import FrameImageExporter
+
+N_ENVS = 3
+N_STEPS = 2
+IS_RENDER_ALL_CAMERAS = True
+OUTPUT_DIR = "img_output/test"
```
YilingQiao: can we put them into
```
parser.add_argument
```

examples/rigid/single_franka_batch_render.py
```diff
@@ -0,0 +1,98 @@
...
+    gs.init(backend=gs.cpu if args.cpu else gs.gpu)
+
+    ########################## create a scene ##########################
+    scene = gs.Scene(
+        viewer_options=gs.options.ViewerOptions(
+            camera_pos=(3.5, 0.0, 2.5),
+            camera_lookat=(0.0, 0.0, 0.5),
+            camera_fov=40,
+        ),
+        show_viewer=args.vis,
```
YilingQiao: let's put 
```
show_viewer=args.vis
```
after 
```
renderer=gs.options.renderers.BatchRenderer(
            use_rasterizer=True,
        ),
```

genesis/__init__.py
```diff
@@ -356,6 +356,11 @@ def _custom_excepthook(exctype, value, tb):
     except ImportError:
         pass
 
+    try:
```
YilingQiao: what's this for?

README.md
yuhongyi: Yes, I could

tests/test_render.py
YilingQiao: Did you add new tests for madrona batched renderer?

genesis/utils/image_exporter.py
```diff
@@ -0,0 +1,128 @@
...
+import numpy as np
+import torch
+from functools import partial
+from concurrent.futures import ThreadPoolExecutor
+
+import genesis as gs
+from genesis.utils.misc import tensor_to_array
+
+
+class FrameImageExporter:
```
YilingQiao: when you add batched rendering test, can you also test this function?

tests/test_render.py
YilingQiao: maybe we can start with some some tests for API, not necessarily the correctness test of rendering results

yuhongyi: I tried, but it's more work than I thought.  The test failed probably because the container needs to be updated due to extra dependencies. I think it's better to do it in a separate PR.

genesis/__init__.py
yuhongyi: I copy-pasted from the code for Luisa above.  I think the point is to import the library earlier.
BTW, this code is not new.  It has been reviewed before.

.gitmodules
```diff
@@ -7,3 +7,6 @@
 [submodule "genesis/ext/ParticleMesher"]
 	path = genesis/ext/ParticleMesher
 	url = https://github.com/ACMLCZH/ParticleMesher
+[submodule "genesis/ext/gs-madrona"]
```
duburcqa: I don't think `gs-madrona` should be a submodules. Python wheels will be compiled and distributed on PyPI completely independently from Genesis. `gs-madrona` will then be added to the Python requirements of Genesis simulator.

examples/rigid/single_franka_batch_render.py
```diff
@@ -0,0 +1,98 @@
...
+        castshadow=1,
+        cutoff=45.0,
+        intensity=0.5,
+    )
+
+    ########################## build ##########################
+    scene.build(n_envs=N_ENVS)
+
+    # Create an image exporter
+    exporter = FrameImageExporter(OUTPUT_DIR)
```
duburcqa: Could you add a comment to explain what this frame image exporter does and why you are using it instead of the usual `start|stop_recording` public API? This would be helpful for practitioners.

yuhongyi:
> `@pytest.mark.xfail(reason="gs-madrona must be built and installed manually for now.")`.

Good idea.  I'll add one

genesis/__init__.py
duburcqa: The case of LuisaRender is a bit specific and steam from a bad infra in the first place. LuisaRender should not be a submodule that is build along with Genesis. There should be some pre-compiled binaries distributed on PyPI instead.

This is what we are planning to do for `gs-madrona`. In the meantime, I think you can just remove this early import and it will be fine. I will make sure wheels are available before the next public release of Genesis simulator.

genesis/engine/scene.py
```diff
@@ -1030,6 +1086,30 @@ def draw_debug_path(self, qposs, entity, link_idx=-1, density=0.3, frame_scaling
...
+            Whether to render the depth image.
+        normal : bool, optional
+            Whether to render the normal image.
+        segmentation : bool, optional
+            Whether to render the segmentation image.
+        force_render : bool, optional
+            Whether to force render the scene.
+
+        Returns:
+            A tuple of tensors of shape (n_envs, H, W, 3) if rgb is not None, otherwise a list of tensors of shape (n_envs, H, W, 1) if depth is not None.
```
duburcqa: 120 chars linewidth limit.

genesis/ext/gs-madrona
duburcqa, line 1: Remove this submodule. It will be managed differently. On the long run we should get rid of all these submodules, it is not a good approach.

genesis/utils/geom.py
duburcqa: Did you check whether the 2 implementation were matching?

Did you do the benchmark that you suggested?

examples/rigid/single_franka_batch_render.py
yuhongyi: I could.  Basically this does batch exporting

.gitmodules
yuhongyi: @YilingQiao Do you agree to remove gs-madrona as a submodule?

genesis/utils/image_exporter.py
duburcqa: Yes, and a fairly detailed documentation (I would expect at least 4-5 lines).

README.md
yuhongyi: Updated

examples/rigid/single_franka_batch_render.py
yuhongyi: Updated

examples/rigid/single_franka_batch_render.py
yuhongyi: Updated

genesis/utils/image_exporter.py
yuhongyi: Added

examples/rigid/single_franka_batch_render.py
duburcqa: Yes, but why it works differently for batch rendering?

yuhongyi: Added comments to explain the reason to use FrameImageExporter for batch exporting

genesis/engine/scene.py
yuhongyi: Split into 2 lines

genesis/ext/gs-madrona
yuhongyi: Removed.  Yes, we should remove LuisaRender as well.

yuhongyi:
> > `@pytest.mark.xfail(reason="gs-madrona must be built and installed manually for now.")`.
> 
> Good idea. I'll add one

Added a test case using batch renderer and marked xfail.  Tested locally

tests/test_render.py
yuhongyi: Added one test case for batch renderer

.gitmodules
yuhongyi: Removed

tests/test_render.py
duburcqa: > I think it's better to do it in a separate PR.

Note that we don't merge any new feature without accompanying unit tests. No exceptions.

But it seems you added one. Even if it only runs locally on your machine, this is fine for now.

.gitmodules
```diff
@@ -6,4 +6,4 @@
 	url = https://github.com/Alif-01/LuisaRender
 [submodule "genesis/ext/ParticleMesher"]
 	path = genesis/ext/ParticleMesher
-	url = https://github.com/ACMLCZH/ParticleMesher
+	url = https://github.com/ACMLCZH/ParticleMesher
```
duburcqa: Remove this change.

.gitignore
```diff
@@ -148,3 +148,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# Batch renderer output
```
duburcqa: Remove this change. I don't think it should be the default. It is the responsibility of the user to choose a folder outside the repository if desired.

genesis/vis/rasterizer_context.py
duburcqa: That makes no sense, the exact same workaround is part of https://github.com/Genesis-Embodied-AI/Genesis/pull/1426. Anyway.

tests/test_render.py
duburcqa: > The new camera attach implementation in this PR allows attaching a camera to the same link and update the camera location to follow the position of the link in each environment independently.

This was not the case before? From what I understand, it does not make sense anymore to create multiple cameras or is this still necessary?

tests/test_render.py
```diff
@@ -324,3 +328,92 @@ def test_batched_mounted_camera_rendering(show_viewer, tol):
...
+        visualize_contact=True,
+    )
+    cam_0 = scene.add_camera(
+        res=(512, 512),
+        pos=(1.5, 0.5, 1.5),
+        lookat=(0.0, 0.0, 0.5),
+        fov=45,
+        GUI=True,
+    )
+    cam_0.attach(franka.links[6], trans_to_T(np.array([0.0, 0.5, 0.0])))
```
duburcqa: Remove this. It is out of scope of this unit test. You should rather parameterise `test_batched_mounted_camera_rendering` to run with both `pyrender` and `gs-madrona` (with the one corresponding to `gs-madrona` being marked as `xfail`).

If you want to add multiple camera to the scene, just call `scene.add_camera` twice.

.gitmodules
yuhongyi: Restored

tests/test_render.py
```diff
@@ -324,3 +328,92 @@ def test_batched_mounted_camera_rendering(show_viewer, tol):
...
                 assert np.count_nonzero(diff) > diff_tol * np.prod(diff.shape)
+
+
+@pytest.mark.parametrize("use_rasterizer", [True, False])
+@pytest.mark.parametrize("render_all_cameras", [True, False])
+@pytest.mark.parametrize("n_envs", [1, 3])
+@pytest.mark.parametrize("n_steps", [1, 1000])
+@pytest.mark.required
+@pytest.mark.xfail(reason="gs-madrona must be built and installed manually for now.")
+def test_madrona_batch_rendering(use_rasterizer, render_all_cameras, n_envs, n_steps):
```
duburcqa: What about just update `test_batched_offscreen_rendering` to run for both `pyrender` and `gs-madrona`, and add 2 fixed cameras in the scene to make the test more comprehensive?

tests/test_render.py
```diff
@@ -324,3 +328,92 @@ def test_batched_mounted_camera_rendering(show_viewer, tol):
             for i in range(n_envs):
                 diff = frames_t[i] - frames_t_minus_1[i]
                 assert np.count_nonzero(diff) > diff_tol * np.prod(diff.shape)
+
+
+@pytest.mark.parametrize("use_rasterizer", [True, False])
+@pytest.mark.parametrize("render_all_cameras", [True, False])
+@pytest.mark.parametrize("n_envs", [1, 3])
+@pytest.mark.parametrize("n_steps", [1, 1000])
```
duburcqa: I don't see the point.

tests/test_render.py
```diff
@@ -324,3 +328,92 @@ def test_batched_mounted_camera_rendering(show_viewer, tol):
             for i in range(n_envs):
                 diff = frames_t[i] - frames_t_minus_1[i]
                 assert np.count_nonzero(diff) > diff_tol * np.prod(diff.shape)
+
+
+@pytest.mark.parametrize("use_rasterizer", [True, False])
+@pytest.mark.parametrize("render_all_cameras", [True, False])
+@pytest.mark.parametrize("n_envs", [1, 3])
```
duburcqa: I think `@pytest.mark.parametrize("n_envs", [0, 3])` would be better.

tests/test_render.py
```diff
@@ -324,3 +328,92 @@ def test_batched_mounted_camera_rendering(show_viewer, tol):
...
+    scene.add_light(
+        pos=[0.0, 0.0, 1.5],
+        dir=[1.0, 1.0, -2.0],
+        directional=1,
+        castshadow=1,
+        cutoff=45.0,
+        intensity=0.5,
+    )
+    scene.add_light(
+        pos=[4, -4, 4],
```
duburcqa: Use floats for floats. And prefer tuple over list.

tests/test_render.py
```diff
@@ -324,3 +328,92 @@ def test_batched_mounted_camera_rendering(show_viewer, tol):
...
+        res=(512, 512),
+        pos=(1.5, -0.5, 1.5),
+        lookat=(0.0, 0.0, 0.5),
+        fov=45,
+        GUI=True,
+    )
+    scene.add_light(
+        pos=[0.0, 0.0, 1.5],
+        dir=[1.0, 1.0, -2.0],
+        directional=1,
```
duburcqa: Use bools for bools.

tests/test_render.py
```diff
@@ -324,3 +328,92 @@ def test_batched_mounted_camera_rendering(show_viewer, tol):
...
+    scene.build(n_envs=n_envs)
+
+    # Create an image exporter
+    # FrameImageExporter exports images from all cameras and all environments in batch and parallelly,
+    # while Camera.start|stop_recording only exports images from a single camera and environment.
+    export_dir = "img_output"
+    exporter = FrameImageExporter(export_dir)
+
+    # Only verify functionality works without crashes and output dimension matches for now
+    # To verify whether the output is correct pixel-wise, we need to use a more sophisticated test
```
duburcqa: This is not enough. See unit test `test_batched_offscreen_rendering`. You should check that the standard deviation for each frame is large enough, that 2 frames that are supposed to match are truly matching, and that 2 frames that are supposed to be different are truly different (easy if you have 2 cameras at different locations).

Having good extensive unit tests is more important than adding new features. We do not merge PRs if their unit tests are not comprehensive enough.

tests/test_render.py
```diff
@@ -324,3 +328,92 @@ def test_batched_mounted_camera_rendering(show_viewer, tol):
...
+        castshadow=1,
+        cutoff=45.0,
+        intensity=0.5,
+    )
+    scene.build(n_envs=n_envs)
+
+    # Create an image exporter
+    # FrameImageExporter exports images from all cameras and all environments in batch and parallelly,
+    # while Camera.start|stop_recording only exports images from a single camera and environment.
+    export_dir = "img_output"
```
duburcqa: You should use a proper temporary directly. This is issue, just add `tmp_path` as input argument of this unit test. pytest will automatically create a temporary directory that is unique to this test.

tests/test_render.py
```diff
@@ -324,3 +328,92 @@ def test_batched_mounted_camera_rendering(show_viewer, tol):
...
+            assert depth[1].shape == torch.Size([n_envs, 512, 512, 1])
+            exporter.export_frame_all_cameras(i, rgb=rgba, depth=depth)
+        else:
+            rgba, depth, _, _ = cam_0.render(rgb=True, depth=True)
+            assert rgba.shape == torch.Size([512, 512, 4])
+            assert depth.shape == torch.Size([512, 512, 1])
+            exporter.export_frame_single_camera(i, cam_0.idx, rgb=rgba, depth=depth)
+
+    # Clean up export_dir
+    if os.path.exists(export_dir):
```
duburcqa: Remove this. It is useless and a bad practice. See above comment.

genesis/utils/geom.py
yuhongyi: If you are asking the whether the implementation of tc version matches the implementation of your new np version, the answer is no, but the tc version is based on the previous implementation and expanded it into the 2-dimension case.  I didn't rewrite the algorithm.
I think we discussed this earlier and agreed that I can use the previously implemented torch version for this PR.  That implementation was already tested and approved in the PR under genesis-company/Genesis.

The benchmarking will be done later on.

tests/test_render.py
yuhongyi: This is not about multiple cameras.
My implementation could, for each camera, update the camera location to follow the link it is attaching to in all environments, but the current implementation added by some one 2 weeks ago only support one environment, by specifying an env_idx.
For example, the position of a link could be at different location in different environment.  My implementation can move camera to different locations in different environments.  The camera transforms are updated in batch.
I didn't know how the current implementation got approved.  It is not a good implementation.
I added this feature per request from @YilingQiao

tests/test_render.py
yuhongyi: Removed

.gitignore
yuhongyi: Removed

tests/test_render.py
yuhongyi: Good point

tests/test_render.py
yuhongyi: Removed

tests/test_render.py
yuhongyi: Updated

tests/test_render.py
yuhongyi: Updated

tests/test_render.py
yuhongyi: Oh, I didn't know.  Updated

tests/test_render.py
yuhongyi: Removed

tests/test_render.py
yuhongyi: Updated

tests/test_render.py
yuhongyi: I was hesitant of adding something to compare mean and std for an image because I don't think this is the right way to compare rendering results.  It might work as some point, but it is far from ideal, because rendering result may change very frequently for a system under development.

A rendering test system should not be part of CI but not part of the unit test framework.  It should compare images pixel by pixel and show the percentage of differences, and allow users to update golden images when necessary.

But anyway, since we don't have such a system yet, I'm OK to just compare mean and std here.

tests/test_render.py
yuhongyi: Good point.  Let me see how I could do it.
I also found that test_batched_offscreen_rendering should be updated. I just went through the test case.  It was not written using cameras and environments properly.

yuhongyi: If I update test_batched_offscreen_rendering()  , I'll have to make it temporarily xfail.  How about we have 2 separate test functions for now.  When the wheels are ready, and the test case for batched camera can pass, we merge them into one.

.gitignore
YilingQiao: it's not removed?

tests/test_render.py
YilingQiao: yeah, I merged that one as a temporay solution for pyrender. You can replace it with a better solution.

.gitignore
yuhongyi: I'll push the change

yuhongyi: Actually removed now

tests/test_render.py
yuhongyi: Updated with comparing mean and std

YilingQiao (approved):

genesis/engine/scene.py
```diff
@@ -466,6 +468,13 @@ def add_light(
         beam_angle : float
             The beam angle of the light.
         """
+        if isinstance(self.renderer_options, gs.renderers.BatchRenderer):
+            gs.logger.warning(
+                "This add_light() function is only supported when NOT using BatchRenderer."
+                "Please use add_light(self, pos, dir, intensity, directional, castshadow, cutoff) instead."
+            )
+            return
```
duburcqa: In this we should rather raise an exception. No need to tolerate such mistake.

genesis/engine/scene.py
```diff
@@ -478,6 +487,43 @@ def add_light(
...
+            The intensity of the light.
+        directional : bool
+            Whether the light is directional.
+        castshadow : bool
+            Whether the light casts shadows.
+        cutoff : float
+            The cutoff angle of the light in degrees.
+        """
+        if not isinstance(self.renderer_options, gs.renderers.BatchRenderer):
+            gs.logger.warning(
```
duburcqa: same.

genesis/utils/image_exporter.py
duburcqa: Still no documentation about what this class is doing.

genesis/utils/image_exporter.py
```diff
@@ -0,0 +1,128 @@
...
+                executor.map(rgb_job, np.arange(len(rgb)))
+
+        if depth is not None:
+            depth = torch.as_tensor(depth, dtype=torch.float32, device=gs.device)
+
+            # Unsqueeze depth to (n_envs, H, W, 1)
+            if depth.ndim == 3:
+                depth = depth.unsqueeze(0)
+            elif depth.ndim == 2:
+                depth = depth.reshape(1, depth.shape[0], depth.shape[1], 1)
```
duburcqa: `.reshape((1, *depth.shape, 1))`

genesis/vis/batch_renderer.py
```diff
@@ -0,0 +1,223 @@
...
+import torch
+import taichi as ti
+
+import genesis as gs
+from genesis.repr_base import RBC
+
+try:
+    from gs_madrona.renderer_gs import MadronaBatchRendererAdapter
+except ImportError as e:
+    gs.raise_exception(f"Failed to import Madrona batch renderer. {e.__class__.__name__}: {e}")
```
duburcqa: It would be better to sort out whether gs-madrona can be installed on this platform (i.e. Linux x86 with NVIDIA GPU), and if some, mention that it should be installed via pip. Otherwise mention that native batch rendering is not supported on this platform.

genesis/vis/batch_renderer.py
```diff
@@ -0,0 +1,223 @@
...
+
+        Returns
+        -------
+        rgb_torch : tuple of tensors
+            The rgb image.
+        depth_torch : tuple of tensors
+            The depth image.
+        """
+        if normal:
+            raise NotImplementedError("Normal rendering is not implemented")
```
duburcqa: `not supported from now.`

genesis/vis/batch_renderer.py
```diff
@@ -0,0 +1,223 @@
...
+        if segmentation:
+            raise NotImplementedError("Segmentation rendering is not implemented")
```
duburcqa: same.

genesis/vis/batch_renderer.py
```diff
@@ -0,0 +1,223 @@
...
+        # Update last_t to current time to avoid re-rendering if the scene is not updated
+        self._last_t = self._visualizer.scene.t
+        self._last_render_options = render_options
+        self.update_scene()
+
+        rigid = self._visualizer.scene.rigid_solver
+        camera_pos = self._visualizer.camera_pos
+        camera_quat = self._visualizer.camera_quat
+
+        render_options_array = np.array([rgb, depth, normal, segmentation, aliasing], dtype=np.uint32)
```
duburcqa: `render_options_np = np.array(render_options, dtype=np.uint32)`

genesis/vis/batch_renderer.py
```diff
@@ -0,0 +1,223 @@
...
+
+        rgb_torch, depth_torch = self._renderer.render(rigid, camera_pos, camera_quat, render_options_array)
+
+        if rgb_torch is not None:
```
duburcqa: You are doing the same processing twice. Code duplication can be avoided easily.

genesis/vis/batch_renderer.py
```diff
@@ -0,0 +1,223 @@
...
+        self._depth_torch = None
+
+    @property
+    def lights(self):
+        return self._lights
+
+    def has_lights(self):
+        return len(self._lights) > 0
+
+    @property
```
duburcqa: I'm not a huge fan of all these properties. They are inefficient, arguably useless for the end-user, and no following the usual pattern for public getters (i.e. torch tensor on GPU device)

genesis/vis/camera.py
```diff
@@ -85,10 +87,12 @@ def __init__(
...
         self._near = near
         self._far = far
-        self._pos = pos
-        self._lookat = lookat
-        self._up = up
-        self._transform = transform
+        self._initial_pos = torch.as_tensor(pos, dtype=gs.tc_float, device=gs.device)
+        self._initial_lookat = torch.as_tensor(lookat, dtype=gs.tc_float, device=gs.device)
+        self._initial_up = torch.as_tensor(up, dtype=gs.tc_float, device=gs.device)
+        self._initial_transform = (
```
duburcqa: I would recommend this syntax, that is more clearer I think:
```python
self._initial_transform = None
if transform is not None:
     self._initial_transform = torch.as_tensor(transform, dtype=gs.tc_float, device=gs.device)
```

genesis/vis/camera.py
```diff
@@ -183,29 +168,63 @@ def move_to_attach(self):
...
         """
-        Obtain the RGB camera view.
-        This is a temporary implementation to make Camera a Sensor.
+        Render the camera view with batch renderer.
         """
-        rgb, _, _, _ = self.render()
-        return rgb
+        assert self._visualizer._use_batch_renderer, "Batch renderer is not enabled."
+
+        rgb_arr, depth_arr, seg_arr, normal_arr = self._batch_renderer.render(
```
duburcqa, line 306: `self._batch_renderer.render` should cache the frames internally to avoid computing the same frames multiple times when updating different cameras.

genesis/vis/camera.py
```diff
@@ -183,29 +168,63 @@ def move_to_attach(self):
...
+        rgb=True,
+        depth=False,
+        segmentation=False,
+        colorize_seg=False,
+        normal=False,
+        force_render=False,
+        antialiasing=False,
+    ):
         """
         Render the camera view. Note that the segmentation mask can be colorized, and if not colorized, it will store an object index in each pixel based on the segmentation level specified in `VisOptions.segmentation_level`. For example, if `segmentation_level='link'`, the segmentation mask will store `link_idx`, which can then be used to retrieve the actual link objects using `scene.rigid_solver.links[link_idx]`.
```
duburcqa: It would be great to update this messy formatting.

genesis/vis/camera.py
```diff
@@ -290,7 +316,7 @@ def render(self, rgb=True, depth=False, segmentation=False, colorize_seg=False,
             if depth:
                 depth_min = depth_arr.min()
                 depth_max = depth_arr.max()
-                depth_normalized = (depth_arr - depth_min) / (depth_max - depth_min)
+                depth_normalized = (depth_arr - depth_min) / max(depth_max - depth_min, gs.EPS)
                 depth_normalized = 1 - depth_normalized  # closer objects appear brighter
```
duburcqa: Just write directly: 
```python
if depth_max - depth_min > gs.EPS:
    depth_normalized = (depth_max - depth_arr) / max(depth_max - depth_min, gs.EPS)
    depth_img = (depth_normalized * 255).astype(np.uint8)
else:
    depth_img = np.zeros_like(depth_arr, dtype=np.uint8)
```

genesis/vis/camera.py
```diff
@@ -432,11 +458,48 @@ def backproject_depth_to_pointcloud(K: np.ndarray, depth: np.ndarray, pose, worl
         else:
             gs.raise_exception("We need a rasterizer to render depth and then convert it to pount cloud.")
 
+    # quat for Madrona needs to be transformed to y-forward
+    def _T_to_quat_for_madrona(self, T):
```
duburcqa: This method should be standalone or at least static.

genesis/vis/camera.py
```diff
@@ -448,24 +511,76 @@ def set_pose(self, transform=None, pos=None, lookat=None, up=None):
...
+        if pos is not None:
+            pos = torch.as_tensor(pos, dtype=gs.tc_float, device=gs.device)
+        if lookat is not None:
+            lookat = torch.as_tensor(lookat, dtype=gs.tc_float, device=gs.device)
+        if up is not None:
+            up = torch.as_tensor(up, dtype=gs.tc_float, device=gs.device)
+
+        # Expand to n_envs
+        if env_idx is None:
+            env_idx = torch.arange(self.n_envs)
```
duburcqa: a slice would be more efficient

genesis/vis/camera.py
```diff
@@ -717,22 +858,24 @@ def aspect_ratio(self):
     @property
     def pos(self):
         """The current position of the camera."""
-        return np.array(self._pos)
+        return tensor_to_array(self._multi_env_pos_tensor[0])
```
duburcqa: The documentation should be updated to mention that it only extract the position of the camera associated with the first environment, from what I understand. I'm not sure if this is desirable though. Why not returning them all?

genesis/vis/camera.py
```diff
@@ -750,3 +893,7 @@ def intrinsics(self):
         cx = 0.5 * self._res[0]
         cy = 0.5 * self._res[1]
         return np.array([[f, 0, cx], [0, f, cy], [0, 0, 1]])
+
+    @property
```
duburcqa: I this property really needed? This is weird to have it here.

genesis/vis/visualizer.py
```diff
@@ -244,3 +272,15 @@ def connected_to_display(self):
     @property
     def cameras(self):
         return self._cameras
+
+    @property
+    def camera_pos(self):
```
duburcqa: Where are used these getters? Are they relevant? Not all of them are properly enforcing dtype and device.

tests/test_render.py
duburcqa: > A rendering test system should not be part of CI but not part of the unit test framework.

Can you elaborate? This is not very clear to be. We should indeed have ground truth to compare against. That can be manually updated whenever it is necessary and easy to inspect. I will implement something based on https://github.com/syrupy-project/syrupy which is designed precisely for this.

tests/test_render.py
```diff
@@ -240,6 +244,7 @@ def test_batched_offscreen_rendering(show_viewer, tol):
 
 
 @pytest.mark.required
+@pytest.mark.flaky(reruns=3, condition=(sys.platform == "darwin"))
```
duburcqa: If this really necessary? this is concerning.

tests/test_render.py
```diff
@@ -381,3 +386,109 @@ def test_debug_draw(show_viewer):
     scene.step()
     rgb_array, *_ = cam.render(rgb=True, depth=False, segmentation=False, colorize_seg=False, normal=False)
     assert_allclose(np.std(rgb_array.reshape((-1, 3)), axis=0), 0.0, tol=gs.EPS)
+
+
+@pytest.mark.parametrize("use_rasterizer", [True, False])
+@pytest.mark.parametrize("render_all_cameras", [True, False])
+@pytest.mark.parametrize("n_envs", [0, 3])
+@pytest.mark.required
+@pytest.mark.xfail(reason="gs-madrona must be built and installed manually for now.")
```
duburcqa: It should be now!

duburcqa:
@yuhongyi I will take care of updating this PR myself

genesis/engine/scene.py
```diff
@@ -478,6 +487,43 @@ def add_light(
             mesh, color, intensity, morph.pos, morph.quat, revert_dir, double_sided, beam_angle
         )
 
+    @gs.assert_unbuilt
+    def add_light(
```
duburcqa: Python does not support function overloading. You cannot rely on this approach to discriminate your two implementations for 'add_light'.

tests/test_render.py
yuhongyi: This is not my test case.  I didn't touch it  It's just flaky so I marked it as flaky. This test doesn't even use batch renderer.

tests/test_render.py
yuhongyi: I can update it

genesis/engine/scene.py
yuhongyi: Yes, but they have different signatures. Since this is a public API, what do you recommend? Pick a different name?

tests/test_render.py
yuhongyi: When I was working on other game engines, we usually have a independent rendering test system that is not part of the unit test system, but it's part of the CI.  Unit test can only show whether the image outputs are closer to the golden images or not, but it can't show how it is different from the golden image.  An ideal rendering test system needs to be able to compare images visually and manage the golden images easily if case they do need to be updated.
I think we could build something on top of syrupy, but ideally we also need some UIs to manage the tests and visualize testing results.

genesis/vis/camera.py
yuhongyi: Yes, but env_idx is expected to be a tensor. All the code below uses env_idx with the assumption that it is a tensor.  If we use slice here we could gain some performance for indexing but will pay more cost to handle potentially different types of env_idx, which will make the code more complex and may be overall more expensive.
But thank you for showing me another trick in python.

genesis/vis/camera.py
yuhongyi: I did it this way because the camera.pos is being used in a few places in Genesis, where they all assume that the camera position of all environments are the same.  However, with per-environment camera attaching, this is no longer true.  In order to minimize the change and not break anything, here it just returns the first camera.  Ideally we should return all camera positions, but this may require some design change and refactoring in a larger scope.  We should do that, but we need some design proposal for how it works and API changes first, if we want to return pos for each environment.

yuhongyi:
> @yuhongyi I will take care of updating this PR myself

Thank you

genesis/utils/geom.py
yuhongyi: Added unit tests for new geom util functions

duburcqa (dismissed):

duburcqa (approved):

tests/test_render.py
duburcqa: > Unit test can only show whether the image outputs are closer to the golden images or not, but it can't show how it is different from the golden image. 

Building this on top of `syrupy` is very straightforward. At most 50 lines of code.

> I think we could build something on top of syrupy, but ideally we also need some UIs to manage the tests and visualize testing results.

Same, it should be very easy to implement such feature based on syrupy.

PR #1426, opened 2025-07-16
[BUG FIX] Fix broken interactive viewer.

duburcqa:
## Description

* Fix segmentation node removal.
* Cleanup rendering context state update.

## Motivation and Context

Having all rendering backend managing their own context update state induces additional boilerplate without any advantage. It should be centralised in the context itself. Moreover, the viewer initialisation logic is overly complicated and should be simplified. 

`python examples/rigid/single_franka.py --vis`

## Checklist:

### DIFF
--- a/genesis/vis/visualizer.py
+++ b/genesis/vis/visualizer.py
@@ -130,14 +130,15 @@ class Visualizer(RBC):
 
         self._context.reset()
 
-        # Need to update viewer once here, because otherwise camera will update scene if render is called right after
-        # build, which will lead to segfault.
-        if self._viewer is not None:
-            if self._viewer.is_alive():
-                self._viewer.update(auto_refresh=True)
-
         if self._raytracer is not None:
             self._raytracer.reset()
+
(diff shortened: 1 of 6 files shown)


genesis/vis/visualizer.py
```diff
@@ -130,15 +130,16 @@ def reset(self):
...
-        # Need to update viewer once here, because otherwise camera will update scene if render is called right after
-        # build, which will lead to segfault.
-        if self._viewer is not None:
-            if self._viewer.is_alive():
-                self._viewer.update(auto_refresh=True)
-
         if self._raytracer is not None:
             self._raytracer.reset()
 
+        if self.viewer_lock is not None:
```
yuhongyi, line 136: Is this correct?
viewer_lock is either a real viewer lock or a DummyViewerLock.  It will never be None unless it is not built.

duburcqa: Yes I think so!

PR #1436, opened 2025-07-19
[MISC] Use 0.02 as the default timeconst

YilingQiao:
## Description

I found that using 2 * dt as the default parameter is not a safe choice for a general audience. Many users try using a smaller timestep to make the simulation more stable, but this actually causes the simulation to blow up—which is unintuitive. @Milotrince also ran into this issue today. Shall we revert it to 0.02 to follow MuJoCo’s convention?

## Related Issue

## Checklist:

### DIFF
--- a/genesis/utils/geom.py
+++ b/genesis/utils/geom.py
@@ -1546,7 +1546,7 @@ def default_solver_params():
 
     Reference: https://mujoco.readthedocs.io/en/latest/modeling.html#solver-parameters
     """
-    return np.array([0.0, 1.0e00, 9.0e-01, 9.5e-01, 1.0e-03, 5.0e-01, 2.0e00])
+    return np.array([0.02, 1.0e00, 9.0e-01, 9.5e-01, 1.0e-03, 5.0e-01, 2.0e00])
 
 
 def default_friction():
--- a/genesis/vis/rasterizer_context.py
+++ b/genesis/vis/rasterizer_context.py
@@ -723,7 +723,9 @@ class RasterizerContext:
             self.add_light(light)
 
     def draw_debug_line(self, start, end, radius=0.002, color=(1.0, 0.0, 0.0, 1.0)):
-        mesh = mu.create_line(start, end, radius, color)
+        mesh = mu.create_line(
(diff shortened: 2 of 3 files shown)


duburcqa:
> Many users try using a smaller timestep to make the simulation more stable

I think the statement raises from a misunderstanding of how rigid body simulation and collision detection works. Decreasing the simulation timestep does not make the simulation more stable in general, especially when using 32bits precision.

As already discussed, I'm not in favour of merging this. Using a small timestep is not a good idea for many reasons: 
* It becomes impractical for data collection
* It is not necessary in case of poly-articulated robots that are inherently compliant

It is useful for simulating stiff objects bouncing around, but I disagree that the statement that this is the main application of Genesis. I think Genesis is mainly used for large scale data collection and we should use parameters that are suitable for this application by default. As a result, making the simulation work "better" (which is questionable) for smaller timestep is not a good approach.

There is already a warning that clearly state that using a small timestep is discouraged because it makes the simulation numerically unstable. Users are expected to read the warnings carefully, there is a reason why there are here: https://github.com/Genesis-Embodied-AI/Genesis/blob/5cd71cb959053428ffedbebcad29ed0766f3124d/genesis/engine/solvers/rigid/rigid_solver_decomp.py#L100

Maybe it is more stable with GJK, but still, decreasing the simulation timestep without decreasing the time constant of the constraints make little sense for rigid body dynamics alone.

duburcqa:
After thinking twice. Here is the current status:
* GJK is way more robust than MPR to small simulation timestep because contact information (depth and normal) is more reliable.
* When coupling multiple solvers, it may be necessary to shrink the simulation timestep to satisfy the requirement of the most demanding solver among them, even if it would not be necessary for the rigid solver itself

Based on this, I suggest this compromise:
* Switching to GJK by default
* Replacing `RigidOption.constraint_timeconst = None` by `RigidOption.constraint_timeconst_min = 0.005`. This way, the default constraint time constant will be bounded as expected, without changing the current behaviour for larger timestep.

duburcqa:
Superseded by https://github.com/Genesis-Embodied-AI/Genesis/pull/1439

PR #1438, opened 2025-07-19
[FEATURE] Add support of 'contype'/'conaffinity' for primitive and mesh entities.

Kashu7100:
## Description

Extend `contype` and `conaffinity` to entity level. 

This can be used for RL setup where you only want the plane to collide with object not a hand for grasping task.

https://github.com/user-attachments/assets/f77b60d1-d74f-4c24-bb63-6ba919c0b813

Note: contype and conaffinity are 32-bit integer bitmasks used for contact filtering of contact pairs.
When the contype of one geom and the conaffinity of the other geom share a common bit set to 1, two geoms can collide.
- Plane:            contype=0xFFFF, conaffinity=0xFFFF (1111 1111 1111 1111) default value
- Red Cube:     contype=1, conaffinity=1 (0001) -> collide with Plane and Blue Cube
- Green Cube: contype=2, conaffinity=2 (0010) -> collide with Plane and Blue Cube
- Blue Cube:    contype=3, conaffinity=3 (0011) -> collide with Plane, Red Cube, and Green Cube
- Dragon:         contype=4, conaffinity=4 (0100) -> collide with Plane only

## Checklist:

### DIFF
--- a/genesis/_main.py
+++ b/genesis/_main.py
@@ -91,7 +91,7 @@ def get_motors_info(robot):
             if len(dofs_idx_local) == 1:
                 dofs_name = [joint.name]
             else:
-                dofs_name = [f"{joint.name}_{i_d}" for i_d in range(dofs_idx_local)]
+                dofs_name = [f"{joint.name}_{i_d}" for i_d in dofs_idx_local]
             motors_dof_idx += dofs_idx_local
             motors_dof_name += dofs_name
     return motors_dof_idx, motors_dof_name


YilingQiao:
Can you add a unit test for each feature added?

examples/collision/contype.py
```diff
@@ -0,0 +1,90 @@
...
+            camera_lookat=(0.0, 0.0, 0.5),
+            camera_fov=40,
+            max_FPS=200,
+        ),
+        show_viewer=args.vis,
+    )
+
+    scene.add_entity(gs.morphs.Plane())
+
+    # NOTE: contype and conaffinity are 32-bit integer bitmasks used for contact filtering of contact pairs.
```
duburcqa: Move this comment on the very top of this file.

examples/collision/contype.py
```diff
@@ -0,0 +1,90 @@
...
+            quat=(0, 0, 0, 1),
+            size=(0.1, 0.1, 0.1),
+            contype=1,
+            conaffinity=1,
+        ),
+        surface=gs.surfaces.Default(
+            color=(1.0, 0.0, 0.0, 1.0),
+        ),
+    )
+
```
duburcqa: Do not skip line between each `add_entity`

genesis/engine/solvers/rigid/collider_decomp.py
```diff
@@ -197,7 +197,7 @@ def _compute_collision_pair_validity(self):
                         continue
 
                 # contype and conaffinity
-                if links_entity_idx[i_la] == links_entity_idx[i_lb] and not (
```
duburcqa, line 200: I'm not OK with this change. I see what you are trying to achieve and it makes sense, but it has unacceptable side-effects I think.

Typically, if you import twice the same robot, then the same mask will apply to each robot, which means that some parts of different robots will not be able to collide with each other anymore. To be more specific, if contype/conaffinity is defined on some humanoid robot in a way to prevent the hip from colliding with the tight, and if you add 2 robots, the hip of the first robot will not be able to collide with the tight of the other robot either, which is probably undesirable.

I would suggest to add `is_collision_mask_local: bool = True` to RigidEntity morph options. This boolean would control whether contype and conaffinity masks are entity-local and global for the entire scene. If there are global, there is nothing to do, but if there are global, you must find a way to update the masks accordingly, which seems to be tricky.

What do you think @YilingQiao ?

genesis/options/morphs.py
```diff
@@ -12,7 +12,8 @@
 
 """
 We define all types of morphologies here: shape primitives, meshes, URDF, MJCF, and soft robot description files.
-These are independent of backend solver type and are shared by different solvers. E.g. a mesh can be either loaded as a rigid object / MPM object / FEM object.
+These are independent of backend solver type and are shared by different solvers.
```
duburcqa, line 15: Please refrain yourself from reformatting documentation that you did not update (I mean adding or removing actual content). This is making the diff much harder to read.

Kashu7100: Ok, just wanted to enforce the 120 char rule for the file. I won't do this for the future commits.

genesis/engine/solvers/rigid/collider_decomp.py
Kashu7100: Thanks for pointing out. Yes I think this can cause a problem.

YilingQiao: I prefer the revised version for several reasons:

1. This matches MuJoCo’s behavior (i.e., the entity is not considered when filtering based on coaffinity). We don’t need to introduce additional concepts.
https://github.com/google-deepmind/mujoco/blob/4f43a7d08e4f1289f4d76a4c48b8f42aca5bb38b/src/engine/engine_collision_driver.c#L1479

2. Importing the same robot twice is not a common use case. We haven’t done this in our demos or applications. We might not have to increase the system complexity / divergence from mujoco to support a rare condition.

3. If users really want to use two robots, they can modify the coaffinity bitmask of the two robots—perhaps not through a well-exposed API, but by manually editing the coaffinity in the XML file—to ensure that the pair across entities is not filtered out.

4. Or we can have `is_collision_mask_local`, but maybe default to `False` to match mujoco, unless some advanced users want?

Kashu7100: @duburcqa What do you prefer?

YilingQiao: I will probably go with the current implementation:
1. The expressiveness of "only considering contype within one entity" is a subset of this implementation.
2. It follows MuJoCo's behavior.
3. We can add a localized contype flag if there are actual users who require it, or if we ourselves encounter such a scenario.

duburcqa:
> just wanted to enforce the 120 char rule for the file.

This change should have been reverted before merging @YilingQiao @Kashu7100 . Anyway now that it is merged...

duburcqa:
> This matches MuJoCo’s behavior (i.e., the entity is not considered when filtering based on coaffinity). We don’t need to introduce additional concepts.

As far I as know, this is not true, because you cannot import multiple MJCF files on the same scene.

Because on the current status, I think we should create a PR that explicitly forbid doing so in genesis, because I think it is currently not supported based on the current status and what you just said.

YilingQiao:
> This change should have been reverted before merging @YilingQiao @Kashu7100 . Anyway now that it is merged...

Okay, my mistake, I will follow this.

PR #1439, opened 2025-07-19
[MISC] Enable GJK collision detection algorithm by default.

duburcqa:
## Description

* Cleanup motion planning algorithm.
* Replace option 'constraint_timeconst' by 'constraint_timeconst_min'.
* Enable GJK by default.

### DIFF
--- a/genesis/options/solvers.py
+++ b/genesis/options/solvers.py
@@ -261,8 +261,7 @@ class RigidOptions(Options):
     max_dynamic_constraints : int, optional
         Maximum number of dynamic constraints (like suction cup). Defaults to 8.
     use_gjk_collision: bool, optional
-        Whether to use GJK for collision detection. Defaults to False, because it requires more compilation time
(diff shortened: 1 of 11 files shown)


genesis/options/solvers.py
```diff
@@ -292,7 +291,7 @@ class RigidOptions(Options):
     ls_tolerance: float = 1e-2
     sparse_solve: bool = False
     contact_resolve_time: Optional[float] = None
-    constraint_timeconst: Optional[float] = None
+    constraint_timeconst_min: float = 0.004
```
YilingQiao: can we change it to 0.01 to be safer?

YilingQiao (dismissed):

genesis/options/solvers.py
duburcqa: done!

YilingQiao (approved):

PR #1443, opened 2025-07-21
[MISC] Picking boxes and Panda links with a MouseSpring

gasnica:
## Description
Added:
- maths for object picking, with spherical inertia only
   - implemented just applying external force & torque, outside of solver step
   - will add non-spherical inertia handling, and motion damping in later pass
- cleaned up & refactored a bit:
   - MouseSpring & ViewerInterction refrence held_link, instead of geom or entity
   - raycasting functions always return a RayHit, and it holds hit Geometry
- thread locking was needed when running viewer in threaded mode. On a single debugging attempt, it seemed window-event-callbacks were called on a different thread than simulaton-update callback (which makes sense)
- simple placeholder OBBs are used to raycast against and pick up geometries/links of Panda robot
- split out OBB from AABB for maybe slightly cleaner code
- some linting added
- also there's the thought of adding `pose` property to RigidLinks, but that's just an idea

## Motivation and Context
Allow interaction with the scene.

Manual testing only ST/MT

https://github.com/user-attachments/assets/1bb6f095-9d8a-45fc-ac40-8eca98a2dce0

## Checklist:


genesis/ext/pyrender/interaction/aabb.py
```diff
@@ -79,3 +72,33 @@ def from_center_and_size(cls, center: Vec3, size: Vec3) -> 'AABB':
         max = center + 0.5 * size
         bounds = np.stack((min.v, max.v))
         return cls(bounds)
+
+
+class OBB(AABB):
```
duburcqa: Does not make sense. You should probably have some `BaseBoundingBox` abstraction.

genesis/ext/pyrender/interaction/aabb.py
```diff
@@ -79,3 +72,33 @@ def from_center_and_size(cls, center: Vec3, size: Vec3) -> 'AABB':
...
+    pose: Pose
+
+    def __init__(self, v: NDArray[np.float32], pose: Pose):
+        super().__init__(v)
+        self.pose = pose
+
+    def raycast(self, ray: Ray) -> RayHit:
+        inv_pose = self.pose.get_inverse()
```
duburcqa: Do not compute inverse explicit. This is slow. You should rather implement `inv_transform_point`, `inv_transform_direction`

genesis/ext/pyrender/interaction/aabb.py
```diff
@@ -79,3 +72,33 @@ def from_center_and_size(cls, center: Vec3, size: Vec3) -> 'AABB':
...
+
+    def __repr__(self) -> str:
+        return f"OBB: Min({self.min.x}, {self.min.y}, {self.min.z}) Max({self.max.x}, {self.max.y}, {self.max.z}) " \
+            + f"Pose: {self.pose}"
+
+    @classmethod
+    def from_center_and_size(cls, size: Vec3, pose: Pose) -> 'OBB':
+        min = - 0.5 * size
+        max = + 0.5 * size
+        bounds = np.stack((min.v, max.v))
```
duburcqa: Always specify on which access you stack for readability.

genesis/ext/pyrender/interaction/mouse_spring.py
```diff
@@ -16,65 +16,84 @@ def _ensure_torch_imported() -> None:
...
         self.prev_control_point = control_point
 
     def detach(self):
-        self.held_geom = None
+        self.held_link = None
 
     def apply_force(self, control_point: Vec3, delta_time: float):
         _ensure_torch_imported()
+
+        # note when threaded: apply_force is called before attach!
```
duburcqa, line 36: Could you confirm if it is fixed and clean these notes (which does NOT mean removing them entirely!)

genesis/ext/pyrender/interaction/mouse_spring.py
```diff
@@ -16,65 +16,84 @@ def _ensure_torch_imported() -> None:
...
-        # pos = Vec3.from_tensor(self.held_geom.entity.get_pos())
-        # pos = pos + delta
-        # self.held_geom.entity.set_pos(pos.as_tensor())
         self.prev_control_point = control_point
 
         # do simple force on COM only:
-        link: RigidLink = self.held_geom.link
-        link_pos: Vec3 = Vec3.from_tensor(link.get_pos())
+        link: RigidLink = self.held_link
+        assert link
```
duburcqa: This makes no sense if the type of link is `RigidLink`. Either the assert should be removed or typing should be replaced by `RigidLink | None`

genesis/ext/pyrender/interaction/mouse_spring.py
```diff
@@ -16,65 +16,84 @@ def _ensure_torch_imported() -> None:
...
         lin_vel: Vec3 = Vec3.from_tensor(link.get_vel())
         ang_vel: Vec3 = Vec3.from_tensor(link.get_ang())
+        link_pose: Pose = Pose.from_link(link)
+        held_point_in_world: Vec3 = link_pose.transform_point(self.held_point_in_local)
+
+        # note: we should assert earlier that link inertial_pos/quat are not None
```
duburcqa, line 50: If the simulation is built, `inertial_pos` and `inertial_quat` should not be none. We should update these property getters to read values from taichi field if the simulation is already built. See `sol_params` implementation: https://github.com/Genesis-Embodied-AI/Genesis/blob/767d536b8689cba5f382fe184df8bd270f889ea2/genesis/engine/entities/rigid_entity/rigid_joint.py#L145

genesis/ext/pyrender/interaction/mouse_spring.py
```diff
@@ -16,65 +16,84 @@ def _ensure_torch_imported() -> None:
...
+        # todo: verify inertial_pos/quat are stored in local frame
+        link_T_principal: Pose = Pose(Vec3.from_arraylike(link.inertial_pos), Quat.from_arraylike(link.inertial_quat))
+        world_T_principal: Pose = link_pose * link_T_principal
+
+        arm_in_principal: Vec3 = link_T_principal.get_inverse() * self.held_point_in_local   # for non-spherical inertia
```
duburcqa: Same. No `get_inverse`, this is slow due to memory allocation.

What about overloading operation "/" to multiply by inverse, à-la-Matlab.

genesis/ext/pyrender/interaction/mouse_spring.py
```diff
@@ -16,65 +16,84 @@ def _ensure_torch_imported() -> None:
...
+        arm_in_world: Vec3 = world_T_principal.rot * arm_in_principal  # for spherical inertia
 
-        pos_err_v: Vec3 = control_point - link_pos
-        vel_err_v: Vec3 = Vec3.zero() - lin_vel
+        pos_err_v: Vec3 = control_point - held_point_in_world
         inv_mass: float = float(1.0 / link.get_mass() if link.get_mass() > 0.0 else 0.0)
+        inv_spherical_inertia: float = float(1.0 / link.inertial_i[0, 0] if link.inertial_i[0, 0] > 0.0 else 0.0)
```
duburcqa, line 60: `inv_spherical_inertia = 0.0` make sense ?

genesis/ext/pyrender/interaction/mouse_spring.py
```diff
@@ -16,65 +16,84 @@ def _ensure_torch_imported() -> None:
...
 
         inv_dt: float = 1.0 / delta_time
-        # these are temporary values, till we fix an issue with apply_links_external_force.
-        # after fixing it, use tau = damp = 1.0:
         tau: float = MOUSE_SPRING_POSITION_CORRECTION_FACTOR
         damp: float = MOUSE_SPRING_VELOCITY_CORRECTION_FACTOR
 
         total_impulse: Vec3 = Vec3.zero()
+        total_torque_impulse: Vec3 = Vec3.zero()
```
duburcqa, line 67: I suggest to remove typing when it is irrelevant. We already know that `Vec3.zero()` has type `Vec3`.

genesis/ext/pyrender/interaction/ray.py
```diff
@@ -24,14 +29,16 @@ class RayHit:
...
     @property
     def is_hit(self) -> bool:
-        return 0 <= self.distance
+        assert 0.0 <= self.distance
+        return self.distance < _MAX_RAY_DISTANCE
 
     @classmethod
     def no_hit(cls) -> 'RayHit':
-        return RayHit(-1.0, Vec3.zero(), Vec3.zero())
+        return RayHit(_MAX_RAY_DISTANCE, Vec3.zero(), Vec3.zero(), None)
```
duburcqa, line 41: What about add default values to `RayHit` constructor? This would make the code more readable, but also faster because `Vec3.zero()` requires memory allocation (useless you don't do any copy in RayHit).

genesis/ext/pyrender/interaction/vec3.py
```diff
@@ -62,6 +63,9 @@ def normalized(self) -> 'Vec3':
     def magnitude(self) -> float:
         return np.linalg.norm(self.v)
 
+    def sqr_magnitude(self) -> float:
+        return np.inner(self.v, self.v)
```
duburcqa: `np.dot` is much more common.

genesis/ext/pyrender/interaction/viewer_interaction.py
```diff
@@ -96,64 +106,58 @@ def on_resize(self, width: int, height: int) -> EVENT_HANDLE_STATE:
...
+                    self.prev_mouse_3d_pos = new_mouse_3d_pos
+
+                    use_force: bool = True
+                    if use_force:
+                        # apply force
+                        self.mouse_spring.apply_force(new_mouse_3d_pos, self.scene.sim.dt)
+                    else:
+                        #apply displacement
+                        pos = Vec3.from_tensor(self.picked_link.entity.get_pos())
+                        pos = pos + delta_3d_pos
```
duburcqa: `pos += delta_3d_pos` is more efficient.

genesis/ext/pyrender/interaction/viewer_interaction.py
```diff
@@ -96,64 +106,58 @@ def on_resize(self, width: int, height: int) -> EVENT_HANDLE_STATE:
...
-                    pos = Vec3.from_tensor(self.picked_entity.get_pos())
-                    pos = pos + delta_3d_pos
-                    self.picked_entity.set_pos(pos.as_tensor())
+        with self.lock:
+            if self.picked_link:
+                mouse_ray: Ray = self.screen_position_to_ray(*self.prev_mouse_pos)
+                ray_hit: RayHit = self.mouse_drag_plane.raycast(mouse_ray)
+                assert ray_hit.is_hit
+                if ray_hit.is_hit:
+                    new_mouse_3d_pos: Vec3 = ray_hit.position
```
duburcqa, line 115: useless typing

genesis/ext/pyrender/interaction/viewer_interaction.py
```diff
@@ -96,64 +106,58 @@ def on_resize(self, width: int, height: int) -> EVENT_HANDLE_STATE:
...
+                    delta_3d_pos: Vec3 = new_mouse_3d_pos - self.prev_mouse_3d_pos
+                    self.prev_mouse_3d_pos = new_mouse_3d_pos
+
+                    use_force: bool = True
```
duburcqa, line 119: same. But you can keep them if you like them x)

genesis/ext/pyrender/interaction/viewer_interaction.py
```diff
@@ -197,53 +201,62 @@ def _raycast_against_ground_plane(self, ray: Ray) -> RayHit:
...
-        closest_hit = None
-        hit_entity: RigidEntity | None = None
-        for entity in self.get_entities():
-            ray_hit = self.raycast_against_entity_obb(entity, ray)
-            if ray_hit.is_hit and (closest_hit is None or ray_hit.distance < closest_hit.distance):
+        else:
+            closest_hit = RayHit.no_hit()
+            for link in entity.links:
+                # how to check if the link is dynamic ??
+                if 0.0 < link.get_mass():
```
duburcqa: `if link.get_mass() > 0.0:` is more natural for me.

genesis/ext/pyrender/interaction/viewer_interaction.py
```diff
@@ -197,53 +201,62 @@ def _raycast_against_ground_plane(self, ray: Ray) -> RayHit:
...
-    def raycast_against_entities(self, ray: Ray) -> tuple[RayHit | None, RigidEntity | None]:
-        closest_hit = None
-        hit_entity: RigidEntity | None = None
-        for entity in self.get_entities():
-            ray_hit = self.raycast_against_entity_obb(entity, ray)
-            if ray_hit.is_hit and (closest_hit is None or ray_hit.distance < closest_hit.distance):
+        else:
+            closest_hit = RayHit.no_hit()
+            for link in entity.links:
+                # how to check if the link is dynamic ??
```
duburcqa: What does this comment means? Whether it can move ? If so, you need to change for its parent index. If it is "-1", then it is fixed wrt world (95% sure). Another option is to check the value of `is_fixed`, which is arguably much clearer.

genesis/ext/pyrender/interaction/viewer_interaction.py
```diff
@@ -197,53 +201,62 @@ def _raycast_against_ground_plane(self, ray: Ray) -> RayHit:
...
+        box: gs.morphs.Box = box_entity.morph
+        pose = Pose.from_link(box_entity.links[0])
+        size = Vec3.from_xyz(*box.size)
+        return OBB.from_center_and_size(size, pose)
 
-    def get_pose_of_first_geom(self, entity: RigidEntity) -> 'Pose':
-        return Pose.from_geom(entity.geoms[0])
+    def _get_geom_placeholder_obb(self, geom: 'RigidGeom') -> OBB:
+        pose = Pose.from_geom(geom)
+        size = 0.125 * Vec3.one()
```
duburcqa: Could you add `def full(fill_value):` ? That would be more convenient and efficient.

genesis/ext/pyrender/interaction/viewer_interaction.py
```diff
@@ -197,53 +201,62 @@ def _raycast_against_ground_plane(self, ray: Ray) -> RayHit:
...
-            size = Vec3.from_xyz(*box.size)
-            geom: RigidGeom = entity.geoms[0]
-            assert geom._solver.n_envs == 0, "ViewerInteraction only supports single-env for now"
-            # geom.get_pos() and .get_quat() are squeezed if n_envs == 0
-            pos = Vec3.from_tensor(geom.get_pos())
-            quat = Quat.from_tensor(geom.get_quat())
-            aabb = AABB.from_center_and_size(pos, size)
+        if obb:
+            aabb: AABB = AABB.from_center_and_size(obb.pose.pos, obb.extents)
             aabb.expand(0.01)
```
duburcqa: Use keyword arguments when passing constants to methods. This makes the code more self-explanatory.

genesis/ext/pyrender/interaction/aabb.py
gasnica: actually, I think, pose.get_inverse has the same cost as transform_point -- one rotation of vector by quaternion being the main cost. 

But yes, with the extra inv_pose, it's 3 such function calls instead of 2. Fixing

genesis/ext/pyrender/interaction/mouse_spring.py
gasnica: - I don't think I can confirm with threaded nondeterministic execution (?)
- the early return here was the first work around, and it fixed the crash
- later I added locks, which might also fix the cause of the earlier crash
- I have not tested removal of the original work around, but it's nondeterministic, so that wouldn't confirm anything, so I'm just leaving it here, with comments. 

Would you like me to add the above notes to the source, or do you want me to write something else specific instead?

genesis/ext/pyrender/interaction/mouse_spring.py
gasnica: roger that, yes. 

yeah, we'd need to have the \ operator here, which we don't .. hmm

genesis/ext/pyrender/interaction/mouse_spring.py
gasnica: that would be a body that's fixed rotationally, or is infinitely inertive and torques don't affect its ang_vel

genesis/ext/pyrender/interaction/mouse_spring.py
gasnica: true. 

Though, I started using it as if in a strongly-typed language, to also indicate that he variable is being declared here, and not reused/modified.

I'll leave it as is for now, or tell me again, please, to remove it.

gasnica: ... and code seems to me more consistent this way, when all new var have their types explicitly stated.

genesis/ext/pyrender/interaction/ray.py
gasnica: 1. I'd leave no_hit() there, cos, imho, it actually makes calling code more readable -- stating explicitly that a code path results in no hit. Returning just Raycast() is vague.

2. As for the Vec3 allocations -- true. I can change it, just from my short experience it feels that using attributes linted as `SomeType | None` makes all the code later more dirty, cos you have to assert the value is there before using the var, otherwise the linter and/or in-editor linter complains. 

So I feel having the values there always, and being able to treat them almost like value types, allows for much cleaner linted code later.

Let me know what you think..

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: maybe let's hop on a call to figure this out

my intuition is that if we're adding typing, then should be consistently in doing so. And here it's informative, as it tells you the type of ray_hit.position, which otherwise could be an np.array, a tensor, or None, a tuple, a list and what not

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: are we good keeping this ??

I bought it previously in an inline if-else statement, but otherwise prefer to use < sign only, which shows value relation positionally too

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: Thanks for picking that up. -- yes, whether it can move, and more precisely, whether it can accelerate when acted on with a force.

- yes, RigidLink.is_fixed seems to have the value I need. Do we want to correct it's value though ?? Cos RigidLink._build() sets it to an np.int32 instead of bool -- don't know why.

Also -- for my fixed root link, get_mass() still returns non-zero value. Do we want or need to correct that?

Another question -- what does RigidEntity.is_free mean ?? Here, the Panda entity is_free, while the root link is_fixed (?)

gasnica: Also does checking for parent_idx != -1 work for kinematic chains that are not attached at the root?

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: Thanks :-) Keeping & resolving XD

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: Oh, btw. how's += faster ?? we don't define the += operators, so isn't it just syntactic sugar?

Or is it just faster, cos there's less script tokens to process?

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: I'm assuming I can keep this, just like the other case below :-) resolving

YilingQiao:
Do you have a scripts that I can test and play with?

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: even if there's one possible and required argument only; and it's somehow self explanatory like here?

gasnica:
Sent the demo file to you via Slack, Yiling :-)

genesis/ext/pyrender/interaction/aabb.py
gasnica: Made OBB an unrelated class. --- Just using a temp aabb instance internally in raycast().

genesis/ext/pyrender/interaction/aabb.py
gasnica: done. resolving.

genesis/ext/pyrender/interaction/aabb.py
gasnica: done. resolving.

genesis/ext/pyrender/interaction/mouse_spring.py
gasnica: removed the assert. resolving.

genesis/ext/pyrender/interaction/mouse_spring.py
gasnica: done.

btw. currently, I'm doing 3 allocations anyways, when rotating a vector in Quat.__mul__. --- Haven't been concerned with optimizing that, since it's slow anyways.

genesis/ext/pyrender/interaction/vec3.py
gasnica: changed. resolving.

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: done. resolving.

YilingQiao:
Just tried and It's fun to play with! I found that it doesn't work well if I use gpu as the backend. Is it the same on your device? Is it expected? Should we force the backend to be cpu if `enable_interaction=True`?

genesis/ext/pyrender/interaction/viewer_interaction.py
```diff
@@ -197,53 +201,60 @@ def _raycast_against_ground_plane(self, ray: Ray) -> RayHit:
...
+        box: gs.morphs.Box = box_entity.morph
+        pose = Pose.from_link(box_entity.links[0])
+        half_extents = 0.5 * Vec3.from_xyz(*box.size)
+        return OBB(pose, half_extents)
 
-    def get_pose_of_first_geom(self, entity: RigidEntity) -> 'Pose':
-        return Pose.from_geom(entity.geoms[0])
+    def _get_geom_placeholder_obb(self, geom: 'RigidGeom') -> OBB:
+        pose = Pose.from_geom(geom)
+        half_extents = Vec3.full(0.5 * 0.125)
```
YilingQiao, line 241: I see, right now we are using the same box for all general geoms. After raycasting, we then don't need OBB?

YilingQiao (approved):
I've played with it. It is fun. Although not very smooth and seems a little bit problematic on GPU backend.

genesis/ext/pyrender/interaction/mouse_spring.py
duburcqa: That would be great to confirm whether it is still necessary or not. If you don't know and do not want to sort this out, you can leave the comments as-is.

genesis/ext/pyrender/interaction/mouse_spring.py
duburcqa: ok I see. But is it really what we expect from a body having zero inertia in the first place?

genesis/ext/pyrender/interaction/ray.py
duburcqa: Let's keep it like this for now.

genesis/ext/pyrender/interaction/viewer_interaction.py
duburcqa: You can keep this if you want, but I don't think you should "prefer to use < sign only", because of order of the argument is important. The first argument is the comparison should be what matters most. In this case, what you are doing is "checking that the mass is strictly positive i.e. larger than zero" not that "zero is strictly smaller than the mass", which is extremely weird and confusing to me.

genesis/ext/pyrender/interaction/viewer_interaction.py
duburcqa: > yes, RigidLink.is_fixed seems to have the value I need. Do we want to correct its value though ?? Cos RigidLink._build() sets it to an np.int32 instead of bool -- don't know why.

Good spot, I will fix it. It should be `ti_bool`, which is `ti.u1` aka "bool" on any platform except MacOS where this type is not supported so it fallbacks as `ti.int32`.

> Also -- for my fixed root link, get_mass() still returns non-zero value. Do we want or need to correct that?

No, because it is not because an object is fixed wrt world that it should have no mass. I agree that the mass is indeed pointless in this context and arguably ill-defined, but keeping the mass non-zero aligns with what Mujoco is doing.

> Another question -- what does RigidEntity.is_free mean ?? Here, the Panda entity is_free, while the root link is_fixed (?)

Honestly I don't know. This flag is weird and poorly documented. We should improve this at some point.

duburcqa: > Also does checking for parent_idx != -1 work for kinematic chains that are not attached at the root?

Yes I think so. But it may be good to double check.

genesis/ext/pyrender/interaction/viewer_interaction.py
duburcqa: Not really, there is only one argument but what is  this argument is not self-explanatory I think.

genesis/ext/pyrender/interaction/mouse_spring.py
gasnica: Yep, leaving the comment as is for now -- better to have the code and comment there, than have things crash later unexpectedly. 

Resolving

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: I'm using link.is_fixed for now. Thanks for all the info, Alexis. Resolving.

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: yes. I'll need to add some sensible OBB to the geoms though for a 'broadphase' check. -- do we have something I can use ?? We have that geoms_init_AABB ... do we have anything else that might be suitable?

YilingQiao: There is geoms_state.aabb_max and geoms_state.aabb_min

gasnica:
It seems running demos with or without interaction on gpu vs cpu, is much slower, and brings down fps significantly.
Interaction just adds to it. 

I started timing parts of my code, and indeed interaction is quite prohibitative adding some (!) 40ms per frame. I haven't looked at it closer yet, but I'm guessing it's caused by the code running in python and reading Poses of RigidLinks from taichi fields, individually for each link.

Will address that in later PRs.

genesis/ext/pyrender/interaction/mouse_spring.py
gasnica: resolving this. -- assuming we're not doing anything about the allocations for now.

genesis/ext/pyrender/interaction/mouse_spring.py
gasnica: oh, zero-inertia is invalid. So I'm just interpreting it as encoding for inf-inertia, and then math for computing effective/virtual mass of a constraint and impulse applications works.

Possibly this code is just unneeded here, and I can add asserts instead -- do we ever have zero masses/inertias in the system -- I do not know what our convention is. With what I'm used to I'd store inv mass/inertia in body definition, and that encodes mass for both dynamic and fixed bodies in a unified way.

genesis/ext/pyrender/interaction/viewer_interaction.py
gasnica: resolving

genesis/ext/pyrender/interaction/mouse_spring.py
gasnica: Ookay,  yeah -- not sure what's better <-- reading from taichi fields means kernel launch from python code, and that's very slow. So maybe just make sure it's always correctly set at build?

With the scenes I'm using inertial_pos/quat are there. 

Leaving as-is for now, or I can add asserts when we attach the mouse to the link.

genesis/ext/pyrender/interaction/mouse_spring.py
duburcqa: > Possibly this code is just unneeded here, and I can add asserts instead -- do we ever have zero masses/inertias in the system -- I do not know what our convention is.

That would be great to enforced strictly positive mass for non-fixed bodies at entity-level, then assume it will always be positive in the viewer interaction codepath.

duburcqa:
We need to find a way to speed up data accessors in taichi. I think there should be improved at Taichi-level. It seems like a priority, now that we are heavily relying on this for both interactive viewer and sensors. @YilingQiao

YilingQiao (approved):

PR #1451, opened 2025-07-23
[FEATURE] Rigid Tactile Sensors

Milotrince:
## Description
Added Normal+Tangential force type sensors, and rewrote tactile sensors with kernels.

## Related Issue
Continuation of https://github.com/Genesis-Embodied-AI/Genesis/pull/1381

## Motivation and Context
While starting on making the RL demo example, I thought about more formats that the tactile data could be in, and added that. So now there is:
1. `RigidContactSensor` returns boolean for contact/no contact
2. `RigidContactForceSensor` returns [Fx, Fy, Fz] force
3. `RigidNormalTangentialForceSensor` returns [|Fn|, |Ft|, tx, ty] normal and tangential magnitude and tangential direction (unit vector)
4. `RigidContactGridSensor` is grid version of `RigidContactSensor`
5. `RigidContactForceGridSensor` is grid version of `RigidContactForceSensor`
6. `RigidNormalTangentialForceGridSensor` is grid version of `RigidNormalTangentialForceSensor`

and I rewrote them as kernels while I was at it!

## How Has This Been Tested?

Test for tactile sensors is in `tests/test_sensors.py`

## Checklist:

- [x] I updated the [documentation](https://github.com/Genesis-Embodied-AI/genesis-doc) accordingly or **no change is needed**.

### DIFF
--- a/genesis/sensors/tactile.py
+++ b/genesis/sensors/tactile.py
@@ -1,12 +1,12 @@
-from typing import Any, Dict, List, Optional
+from typing import List, Optional
 
 import numpy as np
-import gstaichi as ti
+import taichi as ti
 import torch
 
 import genesis as gs
 from genesis.engine.entities import RigidEntity
-from genesis.utils.geom import inv_transform_by_quat, ti_inv_transform_by_quat, transform_by_trans_quat
+from genesis.utils.geom import ti_inv_transform_by_quat, transform_by_trans_quat
 from genesis.utils.misc import tensor_to_array
 
 from .base_sensor import Sensor
@@ -25,37 +25,47 @@ class RigidContactSensor(Sensor):
         The index of the link to which this sensor is attached. If None, defaults to the base link of the entity.
     """
 
-    _all_contacts: Dict[str, Any] | None = None
-    _last_contacts_update_step = -1
-
     def __init__(self, entity: RigidEntity, link_idx=None):
-        self._cls = self.__class__
-        self._cached = None
-        self._entity = entity
+        assert isinstance(entity, RigidEntity), "entity must be a RigidEntity"
+        self._cached: torch.Tensor | None = None
+        self._cached_step: int = -1
+        self._entity: RigidEntity = entity
         self._sim = entity._sim
+        self._scene = entity._scene
+        self._solver = self._sim.rigid_solver
         self.link_idx = link_idx if link_idx is not None else entity.base_link_idx
 
-    def read(self, envs_idx: Optional[List[int]] = None):
-        if self._cls._last_contacts_update_step == self._sim.cur_step_global:
+    def read(self, envs_idx: Optional[List[int]] = None) -> np.ndarray:
+        """Returns np.ndarray of shape (batch_size,) of bool values."""
+        if self._cached_step == self._sim.cur_step_global:
             return self._cached
+        self._cached_step = self._sim.cur_step_global
 
-        self._cls._last_contacts_update_step = self._sim.cur_step_global
-        self._cls._all_contacts = self._sim.rigid_solver.collider.get_contacts(
-            as_tensor=True, to_torch=True, keep_batch_dim=True
+        all_contacts = self._solver.collider.get_contacts(as_tensor=True, to_torch=True)
+        contact_links = torch.cat(
+            [
+                self._preprocess_kernel_input(all_contacts["link_a"]),
+                self._preprocess_kernel_input(all_contacts["link_b"]),
+            ],
+            dim=1,
         )
-
-        contact_links = torch.cat([self._cls._all_contacts["link_a"], self._cls._all_contacts["link_b"]], dim=1)
         is_contact = (contact_links == self.link_idx).any(dim=1)
 
-        self._cached = is_contact
+        self._cached = tensor_to_array(is_contact)
+        return self._cached[envs_idx] if envs_idx is not None else self._cached
 
-        return is_contact[envs_idx] if envs_idx is not None else is_contact
+    def _preprocess_kernel_input(self, tensor: torch.Tensor) -> torch.Tensor:
+        # adds missing batch dim for n_envs=0 and ensures tensor is continuous
+        if self._scene.n_envs > 0:
+            return tensor.contiguous()
+        else:
+            return tensor.unsqueeze(0).contiguous()
 
 
 @ti.data_oriented
 class RigidContactForceSensor(RigidContactSensor):
     """
-    Sensor that returns contact force and position based on its associated RigidLink's collision info.
+    Sensor that returns contact force (Fx, Fy, Fz) in the associated RigidLink's local frame.
 
     Parameters
     ----------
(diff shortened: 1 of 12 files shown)


genesis/engine/solvers/rigid/collider_decomp.py
```diff
@@ -2866,7 +2860,7 @@ def _func_box_box_contact(
                             i_b,
                         )
 
-    def get_contacts(self, as_tensor: bool = True, to_torch: bool = True, keep_batch_dim: bool = False):
```
Milotrince: i added this change in the last sensor PR but i want to revert it back because it was unnecessary

genesis/engine/solvers/rigid/collider_decomp.py
```diff
@@ -1,25 +1,22 @@
...
-from genesis.utils.misc import ti_field_to_torch
-import genesis.engine.solvers.rigid.array_class as array_class
 
-from .mpr_decomp import MPR
 from .gjk_decomp import GJK
+from .mpr_decomp import MPR
 from .support_field_decomp import SupportField
 
-from enum import IntEnum
-
```
Milotrince: these are all auto formatting changes because I saved the file

Milotrince:
I'm not really sure if the way I wrote the kernels is good practice, but here's my line of thinking for the current implementation:
1. big kernel for each sensor class to compute the class level cache (processing data for contacts on all links) -- the idea is to do all the computation that doesn't require link-specific logic here
2. then read from the class level cache, extracting for the relevant link the sensor is attached to
3. use inherited class properties when possible
4. some code is repeated in each sensor, but i thought it was more readable/clear this way
5. there are several `if not None` checks because the Sensor object may be created before `scene.build()`-- we could require sensors to be instantiated after build, but since `scene.add_camera` happens before, i thought it would be confusing

YilingQiao:
I ran the benchmark code. 

RigidContactSensor, no sensor 0.5348, old sensor 0.9645, new sensor 0.8373
RigidContactForceGridSensor, no sensor 0.6095, old sensor 1.8706, new sensor 1.6334

I'm not sure if the performance gain is worth the code duplication.
Regarding the code structure, could we implement a caching mechanism in the base tactile sensor class while still using the existing get_contact function? So different sensor could inherit from it.
Are users expected to call add_sensor after build? I thought it would be more natural to do it before build.
I'm not very clear on those parts—maybe we can discuss them tomorrow.

Milotrince:
> RigidContactSensor, no sensor 0.5348, old sensor 0.9645, new sensor 0.8373
> RigidContactForceGridSensor, no sensor 0.6095, old sensor 1.8706, new sensor 1.6334

with the bool sensor and default batch size, n_envs, grid size-- yes it's probably negligible. What do you think about the numbers I posted above?

> Regarding the code structure, could we implement a caching mechanism in the base tactile sensor class while still using the existing get_contact function? So different sensor could inherit from it.

Yes that's pretty much how it is in the version in main right now

> Are users expected to call add_sensor after build? I thought it would be more natural to do it before build.

Agreed, it is before build right now (hence the if checks for first time read)

Milotrince:
A benchmark comparsion: [script](https://github.com/Milotrince/genesis-test/blob/main/src/sensor_demos/benchmark.py)
Total sim loop time = includes `scene.step()` and  `sensor.read()` for all sensors
Results on 4090:
```
batch size: 2048 , num sensors: 10 , grid size: (10, 10, 1) , sim steps: 1000

Sensor Type                      Total sim loop time (s)
No Sensors                       11.5345 

RigidContactSensor               12.2464             
OldRigidContactSensor            12.3388 
-
RigidContactForceSensor          12.8124
OldRigidContactForceSensor       19.7384
-          
RigidContactForceGridSensor      17.6419             
OldRigidContactForceGridSensor   21.9638 
```
So there is a negligible difference for `RigidContactSensor` (bool) and speed up of 35% and 20% for others.
The grid sensor almost doubles the time of the sim loop just by reading... Is that acceptable?

Results on M1 Macbook Air   (note the fewer sim steps)
```
batch size: 2048 , num sensors: 10 , grid size: (10, 10, 1) , sim steps: 100

Sensor Type                      Total sim loop time (s)
No Sensors                       4.8083              

RigidContactSensor               5.7372              
OldRigidContactSensor            6.3240          
-    
RigidContactForceSensor          5.9683              
OldRigidContactForceSensor       13.0323      
-       
RigidContactForceGridSensor      25.1927             
OldRigidContactForceGridSensor   13.3129
```
There is a large slow down with the more kernel-ized version when running on my macbook, likely due to Apple metal limitations with taichi.

YilingQiao:
Thanks very much for your benchmark! The acceleration from kernel-ized version is ok-ish, kind of boarderline to me.

`doubles the time of the sim loop just by reading... Is that acceptable?`
We observe something similar and @duburcqa has a method to mitigate the slowdown from data reading.

genesis/engine/solvers/rigid/collider_decomp.py
duburcqa: Good idea!

genesis/engine/solvers/rigid/collider_decomp.py
duburcqa: They are valid. Let's keep them. But usually it is better to avoid unnecessary formatting changes.

genesis/sensors/tactile.py
```diff
@@ -17,6 +18,8 @@ class RigidContactSensor(Sensor):
     """
     Sensor that returns bool based on whether associated RigidLink is in collision.
 
+    read() -> np.ndarray of shape (batch_size,) of bool values
```
duburcqa: What is this comment? Formatting is weird and this should not be part of high-level documentation. Maybe something like `One can retrieve the current value of a sensor by calling the method 'read'. See corresponding documentation for details.`

genesis/sensors/tactile.py
```diff
@@ -25,83 +28,224 @@ class RigidContactSensor(Sensor):
...
+    _cached_link_is_colliding: torch.Tensor | None = None
+    _last_shared_cache_step = -1
 
-    def __init__(self, entity: RigidEntity, link_idx=None):
-        self._cls = self.__class__
-        self._cached = None
+    def __init__(
+        self,
+        entity: RigidEntity,
+        link_idx=None,
```
duburcqa: Do not split arguments on multiple lines unless it is necessary.

genesis/sensors/tactile.py
```diff
@@ -25,83 +28,224 @@ class RigidContactSensor(Sensor):
...
+    ):
         self._entity = entity
```
duburcqa: You should check if it is a rigid entity, and raise an exception if not.

genesis/sensors/tactile.py
```diff
@@ -25,83 +28,224 @@ class RigidContactSensor(Sensor):
...
+        self._solver = self._sim.rigid_solver
         self.link_idx = link_idx if link_idx is not None else entity.base_link_idx
+        self._cached = None
+        self._last_cache_step = -1
 
-    def read(self, envs_idx: Optional[List[int]] = None):
-        if self._cls._last_contacts_update_step == self._sim.cur_step_global:
-            return self._cached
+    def read(self, envs_idx: Optional[List[int]] = None) -> np.ndarray:
+        if type(self)._cached_link_is_colliding is None:
```
duburcqa: First, you should never call `type(self)` yourself. Second, relying on class attribute is never a good idea because it is not scoped well enough and many unexpected behaviour may happen. Typically, we support simulating two scenes "in parrallel". Your mechanism will break in such a case. You have no other choice than moving you shared buffer either at scene-level or solver-level.

duburcqa (changes requested):

genesis/sensors/tactile.py
```diff
@@ -25,83 +28,224 @@ class RigidContactSensor(Sensor):
...
+
+
+@ti.data_oriented
+class RigidNormalTangentialForceSensor(RigidContactSensor):
+    """
+    Sensor that returns (|fn|, |ft|, tx, ty) for contact force normal, tangential shear force, and tangential
+    direction as a unit vector in the associated RigidLink's local frame.
+    The tangential direction will not be computed if the tangential force < FORCE_EPSILON, since it will be noise.
+
+    read() -> np.ndarray of shape (batch_size, 4) representing the normal and tangential forces.
```
duburcqa: his should be part of `read` documentation. If necessary, you may add a comment stating that one should have a look to the documentation of the base implementation for details.

genesis/sensors/tactile.py
Milotrince: I see, I wanted to try to avoid modifying the scene/solver for this. What would be a clean way to add custom attribute/fields to the scene?

duburcqa:
You should take inspiration from `Collider.get_contacts` to optimise the way to retrieve data from taichi fields. The take away is:
* Allocate at build time a single 2D continuous shared caching buffer to store all the return arguments.
* Expose a set of views of appropriate shape in python, pre-computed at build time, and owned by each individual sensor.

Calling `read` for a single sensor while update the caching buffer for all the sensors at once, everything else will be giving access to torch views.

duburcqa:
If you want I can help you on this!

Milotrince:
The "old" implementation does use `Collider.get_contacts` while this PR currently tries to skip calling `get_contacts` and gets only the necessary information from `collider_info`. There is a slight speedup, but Yiling and I discussed it's probably not worth it because of code duplication (if anything were to be changed with the contacts info in the future, then you would have to update the code of 6 different kernels here 😅 instead of only `get_contacts`).
So my plan for this PR is:
1. Use `Collider.get_contacts`
2. Adds `*NormalTangentialForceSensor`
3. Use buffer on scene-level instead of the dangerous `type(self)`/class

Milotrince:
```batch size: 2048 , num sensors: 10 , grid size: (10, 10, 1) , sim steps: 1000
Sensor Type                      Total sim loop time (s)
No Sensors                       11.2603             
RigidContactSensor               12.5755             
OldRigidContactSensor            12.2114             
RigidContactForceSensor          15.9091             
RigidNormalTangentialForceSensor 16.2792             
OldRigidContactForceSensor       20.1169             
RigidContactForceGridSensor      19.7755             
RigidNormalTangentialForceGridSensor 20.9395             
OldRigidContactForceGridSensor   21.8198

**Comment by Milotrince** [comment]
oops, did not mean to close issue with comment.

here's the numbers from the implementation in https://github.com/Genesis-Embodied-AI/Genesis/pull/1451/commits/c5d1fd714a124c2b1d6f7b58ca6b9211a68ec211
```
batch size: 2048 , num sensors: 10 , grid size: (10, 10, 1) , sim steps: 1000

Sensor Type                      Total sim loop time (s)
No Sensors                       11.2603             

RigidContactSensor               12.5755             
OldRigidContactSensor            12.2114             

RigidContactForceSensor          15.9091             
OldRigidContactForceSensor       20.1169         
    
RigidContactForceGridSensor      19.7755                   
OldRigidContactForceGridSensor   21.8198

RigidNormalTangentialForceSensor     16.2792          
RigidNormalTangentialForceGridSensor 20.9395          
```

genesis/sensors/tactile.py
Milotrince: I removed the class caching (since collider.get_contacts already caches) in the latest commit!

YilingQiao (approved):

Milotrince:
review after IMU PR https://github.com/Genesis-Embodied-AI/Genesis/pull/1594

PR #1482, opened 2025-07-31
[MISC] Update pre-commit to use ruff formatting and auto-organize import instead of black

Milotrince:
## Description
Updates pre-commit hook and pyproject.toml. I also added torch to the pyproject because it was missing.

## Motivation and Context
We still have unorganized imports merged into main; we should update the pre-commit hook to mitigate this.

I tested with trying to commit an ugly file on my repo.

## Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,8 +3,6 @@ requires = ["setuptools", "wheel", "cython>=3.0.0", "numpy>=1.26.4"]
 build-backend = "setuptools.build_meta"
 
 [project]
-# name = "genesis-world-nightly"
-# version = "0.0.3"
 name = "genesis-world"
 version = "0.2.1"
 description = "A universal and generative physics engine"
@@ -18,6 +16,8 @@ dependencies = [
     "trimesh",
     "scipy >= 1.14",
     "libigl",
+    "ompl>=1.7.0; platform_system != 'Windows'",
+    "torch>=2.7.1",
     # Cross-platform library to get various kind of CPU information such as model name
     "py-cpuinfo",
     # * Mujoco 3.3 made Native CCD an opt-out option instead of opt-in,
@@ -60,7 +60,6 @@ dependencies = [
     # Motion Planning library
     # * 1.7.0: First version distributed on PyPI (no pre-compiled binaries for Windows OS).
     #          Fix OMPL DLL search directory.
-    "ompl>=1.7.0; platform_system != 'Windows'"
 ]
 
 [tool.setuptools.packages.find]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -18,6 +16,8 @@ dependencies = [
     "trimesh",
     "scipy >= 1.14",
     "libigl",
+    "ompl>=1.7.0; platform_system != 'Windows'",
+    "torch>=2.7.1",
```
YilingQiao: why?

YilingQiao:
The CI is still using black. Maybe we need to change together @duburcqa 
```
      - name: Black Format Check
        if: matrix.OS == 'ubuntu-22.04' && matrix.PYTHON_VERSION == '3.12'
        run: |
          pip install black
          black --line-length 120 --check .
```

pyproject.toml
Milotrince: torch is not included in the pyproject right now so if you `uv pip install Genesis` with a fresh venv you would be missing torch 😅 ompl was added too when i did `uv add torch`, im not sure about that though

YilingQiao: We deliberately not put torch here because pip install can not resolve the version reliably. It's safer to go to official website and manually download the corresponding version (based on cuda and system combination).

Milotrince: ah ok!

Milotrince:
> ```
>         if: matrix.OS == 'ubuntu-22.04' && matrix.PYTHON_VERSION == '3.12'
> ```
why is this check necessary?

YilingQiao:
We only need to do the check on one system configuration within the matrix. Even if the format check fails on one configuration, other configurations could still run the test to check the correctness

Milotrince:
im a bit afraid that updating the CI may cause it to fail / we would have to do a mass reformatting?
Screenshot 2025-07-31 at 7 34 39 PM

YilingQiao:
Yeah, that’s what would happen if you apply a new formatter. Maybe we should wait until all ongoing changes are merged before updating the CI. 

Or we could choose not to change the CI for now. In that case, I'm not sure whether the pre-commit hook's formatter is consistent with the CI's formatter.

Milotrince:
Yeah there's no urgency on this, we can apply this later

Milotrince:
ruff tries to be very similar to black, but they still have their differences. If we wanted to have the pre-commit hook change now, I would probably remove the formatting check on the CI and add it after everything is migrated. Or we just wait to apply everything at once later.

YilingQiao:
We could apply it later. 

> I would probably remove the formatting check on the CI and add it after everything is migrated. 

Newly contributed code could become messy without a formatter enforced in CI, since not everyone enables the pre-commit hook.

duburcqa (changes requested):
Remove torch requirement. Add back OMPL.

duburcqa:
No mass reformatting changes for now. It is irrelevant. Black is doing the job so far. No need to unify formatting with other repositories at the moment. Let’s focus on what brings value to the users and developers.

PR #1493, opened 2025-08-03
Yunlong/batchrender example mani

yun-long:
## Description

This PR adds an end‑to‑end training script for a two‑stage manipulation pipeline:

- **Stage 1** – Privileged RL (teacher): Train a PPO policy with state privileges using rsl-rl-lib’s OnPolicyRunner.

- **Stage 2** – Vision Behavior Cloning (student): Distill the teacher into a vision-based policy that takes RGB + low‑dim state using a lightweight CNN encoder.

**Motivation**

- To demonstrate using madrona render for vision-based manipulation 
- Provide a simple, reproducible path to train a performant teacher and distill it into a student policy suitable for vision-only or vision+state deployment. To simulate RGB images, it uses `gs-madrona` batch render. 

The policy controls the robot from a stereo camera (left and right) fixed in the real world. 

https://github.com/user-attachments/assets/bcb12057-761c-4fc5-97fe-f285b207d2b5

## Related Issue

## Checklist:


YilingQiao:
can you provide some instructions about the commands that I need to run it on my side?

yun-long:
> can you provide some instructions about the commands that I need to run it on my side?

I have updated the descriptions. You can to test the example via the provided commands.

examples/manipulation/behavior_cloning.py
```diff
@@ -0,0 +1,425 @@
...
+
+        # Position loss (MSE)
+        pos_loss = F.mse_loss(pred_pos, target_pos)
+
+        # Orientation loss (quaternion distance)
+        # Normalize quaternions
+        pred_quat = F.normalize(pred_quat, p=2, dim=1)
+        target_quat = F.normalize(target_quat, p=2, dim=1)
+
+        # Quaternion distance: 1 - |dot(q1, q2)|
```
duburcqa, line 155: Note that the "natural" quaternion distance is given by `2 * arccos(dot(q1, q2))`, i.e. the rotation angle between these two rotations. Your formula is yet another valid norm since this quantity is monotonically increasing the angle and equal to zero when matching. However, even though all norm are equivalent by definition, the non-linear transformation between these two quantities when affect not only the gradient by also the global optimum. 

Not sure yields better performance in practice.

examples/manipulation/behavior_cloning.py
```diff
@@ -0,0 +1,425 @@
...
+    def add(
+        self,
+        rgb_obs: torch.Tensor,
+        robot_pose: torch.Tensor,
+        object_poses: torch.Tensor,
+        actions: torch.Tensor,
+    ) -> None:
+        """Add experience to buffer."""
+        ptr = self._ptr % self._max_size
+        self._rgb_obs[ptr].copy_(rgb_obs)
```
duburcqa: Why not just `self._rgb_obs[ptr] = rgb_obs` ?

examples/manipulation/behavior_cloning.py
```diff
@@ -0,0 +1,425 @@
...
+        self._actions[ptr].copy_(actions)
+        self._ptr = self._ptr + 1
+        self._size = min(self._size + 1, self._max_size)
+
+    def get_batches(self, mini_batches_size: int, num_epochs: int):
+        """Generate batches for training."""
+        buffer_size = self._size * self._num_envs
+        indices = torch.randperm(buffer_size, device=self._device)
+        # calculate the size of each mini-batch
+        num_batches = min(buffer_size // mini_batches_size, 10)
```
duburcqa: why there should be a minimum number of batches ? And 10 seems kind of arbitrary here, and redundant with num_epochs.

examples/manipulation/behavior_cloning.py
```diff
@@ -0,0 +1,425 @@
...
+        for _ in range(num_epochs):
+            for batch_idx in range(num_batches):
```
duburcqa: `indices = torch.randperm(buffer_size, device=self._device)` should be randomised for each epoch no ?

examples/manipulation/behavior_cloning.py
```diff
@@ -0,0 +1,425 @@
...
+                    "rgb_obs": self._rgb_obs.view(-1, *self._img_shape)[mb_indices],
+                    "robot_pose": self._robot_pose.view(-1, self._state_dim)[mb_indices],
+                    "object_poses": self._object_poses.view(-1, 7)[mb_indices],
+                    "actions": self._actions.view(-1, self._action_dim)[mb_indices],
+                }
+                yield batch
+
+    def clear(self) -> None:
+        """Clear the buffer."""
+        self._rgb_obs.zero_()
```
duburcqa, line 298: Actually zero-ing data is irrelevant but I can understand the motivation. It looks nicer ;)

examples/manipulation/behavior_cloning.py
```diff
@@ -0,0 +1,425 @@
...
+        self.mlp = self._build_mlp(mlp_cfg)
+
+        # MLP for pose prediction
+        pose_mlp_cfg = config["pose_head"]
+        pose_mlp_cfg["input_dim"] = vision_encoder_output_dim
+        pose_mlp_cfg["output_dim"] = 7
+        self.pose_mlp = self._build_mlp(pose_mlp_cfg)
+
+        # Force float32 for better performance
+        self.float()
```
duburcqa: Why is this not the responsibility of the practitioner ?

examples/manipulation/behavior_cloning.py
```diff
@@ -0,0 +1,425 @@
...
+
+    def _build_cnn(self, config: dict) -> nn.Sequential:
```
duburcqa: This method is static, why not to mark it as such ? And why to make it a member method in the first place ?

examples/manipulation/behavior_cloning.py
```diff
@@ -0,0 +1,425 @@
...
+                ]
+            )
+
+        # Add adaptive pooling if specified
+        if config.get("pooling") == "adaptive_avg":
+            layers.append(nn.AdaptiveAvgPool2d((4, 4)))
+
+        return nn.Sequential(*layers)
+
+    def _build_mlp(self, config: dict) -> nn.Sequential:
```
duburcqa: same.

examples/manipulation/grasp_env.py
```diff
@@ -2,22 +2,28 @@
 import math
 from typing import Literal
 import genesis as gs
```
duburcqa, line 5: Since you are updating import formatting, skip line between standard libraries and other import while you are here.

examples/manipulation/grasp_env.py
```diff
@@ -74,18 +83,24 @@ def __init__(
...
-                fov=50,
-                GUI=True,
-            )
+        # == add stero camera ==
+        self.left_cam = self.scene.add_camera(
+            res=(self.image_width, self.image_height),
+            pos=(1.25, 0.3, 0.3),
+            lookat=(0.0, 0.0, 0.0),
+            fov=60,
+            GUI=False,
```
duburcqa: `GUI=self.env_cfg["visualize_camera"]`

examples/manipulation/grasp_env.py
```diff
@@ -122,10 +138,9 @@ def reset_idx(self, envs_idx):
         random_z = torch.ones(num_reset, device=self.device) * 0.025  # 0.15 ~ 0.15
         random_pos = torch.stack([random_x, random_y, random_z], dim=-1)
 
-        # randomly yaw the object
+        # downward facing quaternion to align with the hand
         q_downward = torch.tensor([0.0, 1.0, 0.0, 0.0], device=self.device).repeat(num_reset, 1)
-        PI = 3.1415926
-        random_yaw = (torch.rand(num_reset, device=self.device) * 2 * PI - PI) * 0.25
+        random_yaw = (torch.rand(num_reset, device=self.device) * 2 * math.pi - math.pi) * 0.25
         q_yaw = torch.stack(
```
duburcqa, line 155: Why not just to define goal_yaw directly?
```python
goal_yaw = torch.stack(
    [
        torch.zeros(num_reset, device=self.device),
        torch.cos(random_yaw / 2),
        -torch.sin(random_yaw / 2),
        torch.zeros(num_reset, device=self.device),
    ],
    dim=-1,
)
```

examples/manipulation/grasp_env.py
```diff
@@ -122,10 +138,9 @@ def reset_idx(self, envs_idx):
         random_z = torch.ones(num_reset, device=self.device) * 0.025  # 0.15 ~ 0.15
         random_pos = torch.stack([random_x, random_y, random_z], dim=-1)
 
-        # randomly yaw the object
```
duburcqa, line 125: Why to remove this comment ? It is still valid.

examples/manipulation/grasp_env.py
```diff
@@ -149,14 +164,14 @@ def reset_idx(self, envs_idx):
...
+    def reset(self) -> tuple[torch.Tensor, dict]:
         self.reset_buf[:] = True
         self.reset_idx(torch.arange(self.num_envs, device=gs.device))
 
         obs, self.extras = self.get_observations()
-        return obs, None
+        return obs, self.extras
 
-    def step(self, actions):
+    def step(self, actions: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor, dict]:
```
duburcqa, line 185: That would be great to have some type alias to make the various outputs explicit.

examples/manipulation/grasp_env.py
```diff
@@ -198,18 +213,19 @@ def is_episode_complete(self):
         self.extras["time_outs"][time_out_idx] = 1.0
         return self.reset_buf.nonzero(as_tuple=True)[0]
 
-    def get_observations(self):
+    def get_observations(self) -> tuple[torch.Tensor, dict]:
```
duburcqa, line 227: Same.

examples/manipulation/grasp_env.py
```diff
@@ -198,18 +213,19 @@ def is_episode_complete(self):
...
         )
         obj_pos, obj_quat = self.object.get_pos(), self.object.get_quat()
+        #
         obs_components = [
             finger_pos - obj_pos,  # 3D position difference
             finger_quat,  # current orientation (4D quaternion)
-            obj_pos,  # goal pose (7D: pos + quat)
-            obj_quat,  # goal pose (7D: pos + quat)
+            obj_pos,  # goal position
+            obj_quat,  # goal orientation (4D quaternion)
```
duburcqa: Specify it is quat (x,y,z,w) or (w,x,y,z).

examples/manipulation/grasp_env.py
```diff
@@ -219,8 +235,37 @@ def rescale_action(self, action: torch.Tensor) -> torch.Tensor:
         rescaled_action = action * self.action_scales
         return rescaled_action
 
+    def get_depth_image(self, normalize: bool = True) -> torch.Tensor:
+        # Render depth image from the camera
+        _, depth, _, _ = self.batch_cam.render(rgb=False, depth=True)
+        depth = depth.permute(0, 3, 1, 2)  # shape (B, 1, H, W)
+        if normalize:
+            depth = torch.clamp(depth, min=0.0, max=10)
```
duburcqa: Use floats for floats: `max=10.0`

examples/manipulation/grasp_env.py
```diff
@@ -219,8 +235,37 @@ def rescale_action(self, action: torch.Tensor) -> torch.Tensor:
...
+            depth = (depth - 0.0) / (10.0 - 0.0)  # normalize to [0, 1]
```
duburcqa: Define on top of this module right after imports a global constant in upper case for maximum depth instead of some hard-coded value inside this method. You could also make it an option for the user if relevant.

examples/manipulation/grasp_env.py
```diff
@@ -219,8 +235,37 @@ def rescale_action(self, action: torch.Tensor) -> torch.Tensor:
...
+        return depth
+
+    def get_stereo_rgb_images(self, normalize: bool = True) -> torch.Tensor:
+        rgb_left, _, _, _ = self.left_cam.render(rgb=True, depth=False)
+        rgb_right, _, _, _ = self.right_cam.render(rgb=True, depth=False)
+
+        # Convert to proper format
+        rgb_left = rgb_left.permute(0, 3, 1, 2)[:, :3]  # shape (B, 3, H, W)
```
duburcqa, line 254: It is already enforced to be RGB (3 channels) and unit-tested as such.

examples/manipulation/grasp_env.py
duburcqa: Why not just `depth /= 10.0` ?

examples/manipulation/grasp_env.py
```diff
@@ -219,8 +235,37 @@ def rescale_action(self, action: torch.Tensor) -> torch.Tensor:
...
+        rgb_right = rgb_right.permute(0, 3, 1, 2)[:, :3]  # shape (B, 3, H, W)
+
+        # Normalize if requested
+        if normalize:
+            rgb_left = torch.clamp(rgb_left, min=0.0, max=255.0)
+            rgb_left = (rgb_left - 0.0) / (255.0 - 0.0)
```
duburcqa: Same. What about `rgb_left /= 255.0` ? Or even `rgb_left = torch.clamp(rgb_left, min=0.0, max=255.0) / 255.0`

examples/manipulation/grasp_env.py
```diff
@@ -219,8 +235,37 @@ def rescale_action(self, action: torch.Tensor) -> torch.Tensor:
...
+        # Render depth image from the camera
+        _, depth, _, _ = self.batch_cam.render(rgb=False, depth=True)
+        depth = depth.permute(0, 3, 1, 2)  # shape (B, 1, H, W)
+        if normalize:
+            depth = torch.clamp(depth, min=0.0, max=10)
+            depth = (depth - 0.0) / (10.0 - 0.0)  # normalize to [0, 1]
+        return depth
+
+    def get_stereo_rgb_images(self, normalize: bool = True) -> torch.Tensor:
+        rgb_left, _, _, _ = self.left_cam.render(rgb=True, depth=False)
```
duburcqa: You could just call `scene.render_all` directly to get a unified buffer out of the box. The current processing is more flexible but less efficient since internally a single `render_all` is done, then data are split for all camera. Here you are just going back to the initial memory layout but without one extra copy.

May be it is not a big deal in this case.

examples/manipulation/grasp_eval.py
```diff
@@ -19,73 +22,234 @@
...
+def load_rl_policy(env, train_cfg, log_dir):
+    """Load reinforcement learning policy."""
+    runner = OnPolicyRunner(env, train_cfg, log_dir, device=gs.device)
+
+    # Find the latest checkpoint
+    checkpoint_files = [f for f in os.listdir(log_dir) if re.match(r"model_\d+\.pt", f)]
+    if not checkpoint_files:
+        raise FileNotFoundError(f"No checkpoint files found in {log_dir}")
+
+    last_ckpt = sorted(checkpoint_files)[-1]
```
duburcqa: I usually prefer this syntax: `*_, last_ckpt = sorted(checkpoint_files)`

examples/manipulation/grasp_eval.py
```diff
@@ -19,73 +22,234 @@
...
 from grasp_env import GraspEnv
+from behavior_cloning import BehaviorCloning
+
+
+def load_rl_policy(env, train_cfg, log_dir):
+    """Load reinforcement learning policy."""
+    runner = OnPolicyRunner(env, train_cfg, log_dir, device=gs.device)
+
+    # Find the latest checkpoint
+    checkpoint_files = [f for f in os.listdir(log_dir) if re.match(r"model_\d+\.pt", f)]
```
duburcqa: Could be nice to finally move to `pathlib`

examples/manipulation/grasp_eval.py
duburcqa: By the way, you could just do:
```python
try:
    *_, last_ckpt = sorted(f for f in os.listdir(log_dir) if re.match(r"model_\d+\.pt", f))
except ValueError as e:
    raise FileNotFoundError(f"No checkpoint files found in {log_dir}") from e
```

examples/manipulation/grasp_env.py
```diff
@@ -219,8 +235,37 @@ def rescale_action(self, action: torch.Tensor) -> torch.Tensor:
         rescaled_action = action * self.action_scales
         return rescaled_action
 
+    def get_depth_image(self, normalize: bool = True) -> torch.Tensor:
+        # Render depth image from the camera
+        _, depth, _, _ = self.batch_cam.render(rgb=False, depth=True)
```
duburcqa: I would suggest adding `segmentation=False, normal=False` explicitly even if unsupported by Madrona so far. It is more explicit and this will soon not be the case anymore.

examples/manipulation/grasp_eval.py
```diff
@@ -19,73 +22,234 @@
...
+    print(f"Loaded BC checkpoint from {last_ckpt_file}")
+    bc_runner.load(last_ckpt_file)
+
+    return bc_runner._policy
+
+
+def get_stereo_frame(env, step_count):
+    """Get stereo frame as numpy array."""
+    # Get individual camera images
+    rgb_left, _, _, _ = env.left_cam.render(rgb=True, depth=False)
```
duburcqa: You keep applying the same processing to right and left for this entire method and finally concatenate both image. Maybe you could store them in a list to avoid duplication?

examples/manipulation/grasp_eval.py
```diff
@@ -19,73 +22,234 @@
...
+    rgb_right, _, _, _ = env.right_cam.render(rgb=True, depth=False)
+
+    # Convert to numpy arrays and format for OpenCV
+    left_img = rgb_left[0].cpu().numpy()  # Take first environment
+    right_img = rgb_right[0].cpu().numpy()  # Take first environment
+
+    # Convert from RGB to BGR for OpenCV
+    left_img = cv2.cvtColor(left_img, cv2.COLOR_RGB2BGR)
```
duburcqa: What about `cv2.cvtColor(left_img, left_img, cv2.COLOR_RGB2BGR)` ?

examples/manipulation/grasp_eval.py
```diff
@@ -19,73 +22,234 @@
...
+    # Resize images for better display
+    scale = 2.0
+    left_img = cv2.resize(left_img, (int(left_img.shape[1] * scale), int(left_img.shape[0] * scale)))
+    right_img = cv2.resize(right_img, (int(right_img.shape[1] * scale), int(right_img.shape[0] * scale)))
+    # Concatenate images horizontally (side by side)
+    stereo_img = np.hstack([left_img, right_img])
+    # Add a vertical line separator between the two images
+    separator_x = left_img.shape[1]
+    cv2.line(
+        stereo_img,
```
duburcqa: Could you specify all the keywords ?

examples/manipulation/grasp_eval.py
```diff
@@ -19,73 +22,234 @@
...
+    """Save a list of frames as a video."""
+    if not frames:
+        print("No frames to save!")
+        return
+
+    # Get frame dimensions from the first frame
+    height, width = frames[0].shape[:2]
+
+    # Create video writer
+    fourcc = cv2.VideoWriter_fourcc(*"mp4v")
```
duburcqa: cv2 is not the best option to record video in python. I would recommend [PyAV](https://github.com/PyAV-Org/PyAV). But it is probably acceptable in this context.

examples/manipulation/grasp_eval.py
```diff
@@ -19,73 +22,234 @@
...
+        # For RL, load the standard configs
+        env_cfg, reward_cfg, robot_cfg, rl_train_cfg, bc_train_cfg = pickle.load(open(f"{log_dir}/cfgs.pkl", "rb"))
+    else:
+        # For BC, we need to load the configs and create BC config
+        env_cfg, reward_cfg, robot_cfg, rl_train_cfg, bc_train_cfg = pickle.load(open(f"{log_dir}/cfgs.pkl", "rb"))
 
-    # visualize the target
-    env_cfg["visualize_target"] = True
-    # for video recording
-    env_cfg["visualize_camera"] = args.record
```
duburcqa, line 39: Why removing these options ?

examples/manipulation/grasp_train.py
```diff
@@ -96,43 +147,62 @@ def get_cfgs():
     return env_cfg, reward_scales, robot_cfg
 
 
+def load_teacher_policy(env, rl_train_cfg, exp_name):
+    # load teacher policy
+    log_dir = f"logs/{exp_name + '_' + 'rl'}"
```
duburcqa: Would be better to use  `pathlib`

examples/manipulation/behavior_cloning.py
yun-long: Yes, this is a cheap proxy of the optimal loss for pose/orientation estimation. However, given that the main task here is to minimize the action loss, the impact of using a proxy orientation loss should be minor.

examples/manipulation/behavior_cloning.py
yun-long: yes. it should sit inside the first epoch loop. will address

examples/manipulation/behavior_cloning.py
yun-long: wasn't sure if `self._rgb_obs[self._ptr] = rgb_obs` will do the job, e.g., does it put a reference to the rgb_obs tensor or will copy the data.

examples/manipulation/behavior_cloning.py
yun-long: Indeed, it was an arbitrary number. It was probably not a good idea to configure the mini_batch_size because it depends on the total number of samples available. Now, adapted the function argument to directly control the total number of num_mini_batches instead.

examples/manipulation/behavior_cloning.py
yun-long: it is to make it look nicer. :)

examples/manipulation/behavior_cloning.py
yun-long: good point. removed.

examples/manipulation/behavior_cloning.py
yun-long: Made the method static. the idea is basically to make the code a bit _cleaner_ / _nicer_ ?

examples/manipulation/behavior_cloning.py
yun-long: same

examples/manipulation/grasp_env.py
yun-long: not sure if it is necessary given it is a single file example. i think it would make sense to type the output if we are providing a framework, e.g., gs-env and gs-rl

examples/manipulation/grasp_env.py
yun-long: same as previous reply

examples/manipulation/grasp_env.py
yun-long: It is the same as the genesis convention, so it should be `(w, x, y, z)`. Updated comment

examples/manipulation/grasp_env.py
yun-long: to make it more explicit. there might cases where minimum depth is also can be configured. for now, assuming minimum depth is zero.

examples/manipulation/grasp_train.py
yun-long: yes. switched to pathlib now

examples/manipulation/grasp_eval.py
yun-long: I recall there were problems with enabling both visualize camera and render camera. The visualize camera uses pyrender as backend, while the render camera uses madrona. You can not have both enabled.

examples/manipulation/grasp_eval.py
yun-long: sure. done

examples/manipulation/grasp_eval.py
yun-long: I feel it is more easy to understand or readable to use `left_img = cv2.cvtColor(left_img, cv2.COLOR_RGB2BGR)
` or?

examples/manipulation/grasp_eval.py
yun-long: good to know the PyAV package. didn't know it before.

examples/manipulation/grasp_env.py
yun-long: this should be handled automatically by the formatter right?

examples/manipulation/grasp_env.py
yun-long: the render camera (madrone) is different from the visualize_camera (pyrender backend), right? not sure if i can treat them the same.

examples/manipulation/grasp_env.py
yun-long: i guess it was removed by accident. added it back

examples/manipulation/grasp_env.py
yun-long: I see. Last time it has 4 channels. So, the alpha channel is removed now?

yun-long: Nevertheless, I guess it doesn't hurt to write in this way---just to make sure.

examples/manipulation/grasp_eval.py
yun-long: sounds good. adjusted to latter suggestion.

examples/manipulation/grasp_eval.py
yun-long: the `grasp_env` provides a function that get a stacked stereo image directly. it is more handy to separate the left and right image for converting the RGB channels. Either way requires similar workload I think.

examples/manipulation/behavior_cloning.py
duburcqa: Make sense. Would you mind adding a comment about this in the code? I would appreciate that :)

examples/manipulation/behavior_cloning.py
duburcqa: It cannot not do the job. 

First, you have a tensor `self._rgb_obs`, and `self._rgb_obs[self._ptr]` is a view which does not own any memory. As a result, you could even do:
```python
tmp = self._rgb_obs[self._ptr]
tmp[:] = rgb_obs
```
And it would work, because `tmp` is just a slice of the original memory, with on offset for the starting memory address, and different stride & shape.

Similarly, equality assignment overload the data accessor to do in-place assignment without re-allocation even in case of fancy indexing (this is true for all frameworks, numpy, torch... whatever, python, C++...). Typically you can do:
```python
self._rgb_obs[[0, 2]] = torch.stack((rgb_obs, rgb_obs), dim=0)
```
and it would just work, while storing `self._rgb_obs[[0, 2]]` in a temporary would no longer work.

------

Note that your approach corresponds to what I'm mentioning in the end, i.e. allocating a temporary and doing `copy_` on it, so you loose all the power of augmented assignment, which means that assignment with fancy indexing will not work:
```python
In [7]: x = torch.tensor([0.0, 0.0, 0.0])

In [8]: x[[0, 2]].copy_(torch.tensor([1.0, 3.0]))
Out[8]: tensor([1., 3.])

In [9]: x
Out[9]: tensor([0., 0., 0.])
```

examples/manipulation/behavior_cloning.py
duburcqa: > the idea is basically to make the code a bit cleaner / nicer ?

I have no recommendation for an example training project. If it was a RL library, I would advise to move this in generic helper module, as helpers to build standard neural network architecture seems generic and not specific to a given RL problem, but here it is a different story.

examples/manipulation/grasp_env.py
duburcqa: Unfortunately we have no formatter so far, just you and me :)

examples/manipulation/grasp_env.py
duburcqa: It is, I know because I'm the one that implemented the support of this option for Madrona :p 

But maybe you should leverage the newly added feature and add a "debug" camera? This way it would not mess up with training cameras. But both options are equally fine for me.

examples/manipulation/grasp_env.py
duburcqa: I get it, but it does make it easier for people to understand your code, and this is an example project, so I would agree its main purpose is to be read by others.

examples/manipulation/grasp_env.py
duburcqa: Yes, but I'm not sure whether it is clear for people what is Genesis convention, anyway, it does not harm I think.

examples/manipulation/grasp_env.py
duburcqa: > I see. Last time it has 4 channels. So, the alpha channel is removed now?

Yes, I changed that.

> Nevertheless, I guess it doesn't hurt to write in this way---just to make sure.

It does not harm, but it gives the feeling to the reader that it may be useful, while it is enforced by design. For something we are shipping as part of our example, I would rather suggest to fully leverage what is officially supported. I would not recommend the same for something external that only uses genesis as a source of data  like gs-rl / gs-env or such.

examples/manipulation/grasp_env.py
duburcqa: Not very convinced but ok. In that case I would rather store min, max clipping value explicitly and apply them where relevant.

examples/manipulation/grasp_eval.py
duburcqa: It is easier to understand, but this will allocate memory, while it does not in the other case.

examples/manipulation/grasp_eval.py
duburcqa: PyAV is great. It is used in torch and co. under the hood.

examples/manipulation/grasp_eval.py
duburcqa: This is no longer the case and you can now I have both, and it is fully unit tested (or maybe not, I should double check, but it is an official feature). So I would suggest to restore these options :)

examples/manipulation/behavior_cloning.py
```diff
@@ -0,0 +1,414 @@
...
+import torch.nn as nn
+import torch.nn.functional as F
+
+import numpy as np
+from collections import deque
+from collections.abc import Iterator
+
+from torch.utils.tensorboard import SummaryWriter
+
+
```
duburcqa: Import ordering (still manual for now):

```python
import os
import time
from collections import deque
from collections.abc import Iterator

import numpy as np
import torch
import torch.nn as nn
import torch.nn.functional as F
from torch.utils.tensorboard import SummaryWriter
```

examples/manipulation/behavior_cloning.py
```diff
@@ -0,0 +1,414 @@
...
+                total_action_loss += action_loss
+                total_pose_loss += pose_loss
+                num_batches += 1
+
+            end_time = time.time()
+            backward_time = end_time - start_time
+
+            # Compute average losses
+            avg_action_loss = total_action_loss / num_batches if num_batches > 0 else 0.0
+            avg_pose_loss = total_pose_loss / num_batches if num_batches > 0 else 0.0
```
duburcqa: It looks like you should rather raise an exception (at init) or consider that the average action / pose loss is undefined (`float("nan")`) if the number of batches is zero?

examples/manipulation/behavior_cloning.py
```diff
@@ -0,0 +1,414 @@
...
+
+            fps = (self._num_steps_per_env * self._env.num_envs) / (forward_time)
+            # Logging
+            if (it + 1) % 10 == 0:
```
duburcqa: Not a fan of this hard-corded `10`. Moreover, would it makes sense to log every iteration in tensorboard even if not printing all of them?

examples/manipulation/behavior_cloning.py
```diff
@@ -0,0 +1,414 @@
...
+                info_str += f" | Forward Time:  {forward_time:.2f}s\n"
+                info_str += f" | Backward Time: {backward_time:.2f}s\n"
+                info_str += f" | FPS:           {int(fps)}"
+                print(info_str)
+
+                if len(self._rewbuffer) > 0:
+                    tf_writer.add_scalar("reward/mean", np.mean(self._rewbuffer), it)
+
+            # Save checkpoints periodically
+            if (it + 1) % 50 == 0:
```
duburcqa: Same. What about some checkpoint period parameter?

examples/manipulation/behavior_cloning.py
```diff
@@ -0,0 +1,414 @@
...
+        self._action_dim = action_dim
+        self._device = device
+        self._ptr = 0
+        self._size = 0
+
+        # Buffers for data
+        self._rgb_obs = torch.empty(max_size, num_envs, *img_shape, dtype=torch.float32, device=device)
+        self._robot_pose = torch.empty(max_size, num_envs, state_dim, dtype=torch.float32, device=device)
+        self._object_poses = torch.empty(max_size, num_envs, 7, dtype=torch.float32, device=device)
+        self._actions = torch.empty(max_size, num_envs, action_dim, dtype=torch.float32, device=device)
```
duburcqa: Woud do you make sure this is consistent with policy dtype? What if someone would like to experiment with 64 bits precision during training? Does it make sense?

examples/manipulation/behavior_cloning.py
```diff
@@ -0,0 +1,414 @@
...
+        # calculate the size of each mini-batch
+        batch_size = self._size // num_mini_batches
+        for _ in range(num_epochs):
+            indices = torch.randperm(self._size)
+            for batch_idx in range(0, self._size, batch_size):
+                batch_indices = indices[batch_idx : batch_idx + batch_size]
+
+                # Yield a mini-batch of data
+                yield {
+                    "rgb_obs": self._rgb_obs[batch_indices].view(-1, *self._img_shape),
```
duburcqa: I would recommend to always use `reshape` everywhere, because `view` is just a useless legacy function that is redundant and I prefer to reduce the API coverage that I used from dependencies as a good practice to reduce maintenance burden. But it does clarify the intend for people that are not familiar with torch.

examples/manipulation/grasp_env.py
```diff
@@ -46,6 +55,9 @@ def __init__(
             ),
             profiling_options=gs.options.ProfilingOptions(show_FPS=False),
             show_viewer=show_viewer,
```
duburcqa: Style guideline that we are trying to push everywhere:
* Move non class option at the very bottom. 
* Order class options from physics to non-physics, i.e. global sim options, solver-specific options, rendering backend options, viewer options, visualiser options, then profiling options.

examples/manipulation/grasp_env.py
```diff
@@ -219,8 +239,35 @@ def rescale_action(self, action: torch.Tensor) -> torch.Tensor:
         rescaled_action = action * self.action_scales
         return rescaled_action
 
+    def get_depth_image(self, normalize: bool = True) -> torch.Tensor:
+        # Render depth image from the camera
+        _, depth, _, _ = self.batch_cam.render(rgb=False, depth=True, segmentation=False, normal=False)
```
duburcqa: I cannot find where this `batch_cam` is supposed to be defined.

examples/manipulation/behavior_cloning.py
yun-long: I agree. so, just to make sure we are on the same page, it is fine to just treat the build_cnn as a static method in the `Policy` class. 

Like this:

```python
@staticmethod
def _build_cnn(config: dict) -> nn.Sequential:
    """Build CNN encoder for grayscale images."""
    layers = []

    # Build layers from configuration
    for conv_config in config["conv_layers"]:
        layers.extend(
            [
                nn.Conv2d(
                    conv_config["in_channels"],
                    conv_config["out_channels"],
                    kernel_size=conv_config["kernel_size"],
                    stride=conv_config["stride"],
                    padding=conv_config["padding"],
                ),
                nn.BatchNorm2d(conv_config["out_channels"]),
                nn.ReLU(),
            ]
        )
```

examples/manipulation/behavior_cloning.py
yun-long: Sounds good. I have added a comment. Specifically:

```python
        # Quaternion distance: 1 - |dot(q1, q2)|
        # Note: we use this as a proxy for the actual distance between two quaternions
        # because the impact of the orientation loss (auxiliary task) is not significant
        # compared to the action loss (main task)

```

examples/manipulation/behavior_cloning.py
yun-long: Just replaced the `copy_` with `self._rgb_obs[ptr] = rgb_obs`

examples/manipulation/grasp_env.py
yun-long: I guess writing down the explicit transformation  
` goal_yaw = transform_quat_by_quat(q_yaw, q_downward)
`
is less error-prone. It reminds the user that goal orientation should be aligned with the orientation of the end-effector.

examples/manipulation/grasp_env.py
yun-long: I feel it is already quite clear what the output/interface should be given the clear variable names. On the other hand, if we are going to provide type alias for this particularly function, what about other functions, e.g.,  

```python
#class ExperienceBuffer:
    def get_batches(self, num_mini_batches: int, num_epochs: int) -> Iterator[dict[str, torch.Tensor]]:
```

or 

```python
#class Policy(nn.Module):
    def get_features(self, rgb_obs: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
```

examples/manipulation/grasp_env.py
yun-long: If it is ok for you, I will keep the current version as I feel it is more explicit. 
It is less clear to the user which camera is used if I do `scene.render_all`, unless comments/explanation is provided.

examples/manipulation/grasp_eval.py
yun-long: make sense. switching to `cv2.cvtColor(left_img, left_img, cv2.COLOR_RGB2BGR)`

examples/manipulation/behavior_cloning.py
yun-long: raise an exception `if num_batches == 0`

examples/manipulation/behavior_cloning.py
yun-long: I removed hard-coded number for logging frequency. Instead, it is defined in the `bc_cfg_dict`. The log_freq is shared by both `tensorboard` and `console` outputs. 10 is good for both.

examples/manipulation/behavior_cloning.py
yun-long: same as above reply. removed the hardcoded number and defined `save_freq`

examples/manipulation/behavior_cloning.py
yun-long: switched to `reshape`

examples/manipulation/grasp_env.py
yun-long: hmmm, why not following the default order

```python
class Scene(RBC):
    def __init__(
        self,
        sim_options: SimOptions | None = None,
        coupler_options: BaseCouplerOptions | None = None,
        tool_options: ToolOptions | None = None,
        rigid_options: RigidOptions | None = None,
        avatar_options: AvatarOptions | None = None,
        mpm_options: MPMOptions | None = None,
        sph_options: SPHOptions | None = None,
        fem_options: FEMOptions | None = None,
        sf_options: SFOptions | None = None,
        pbd_options: PBDOptions | None = None,
        vis_options: VisOptions | None = None,
        viewer_options: ViewerOptions | None = None,
        profiling_options: ProfilingOptions | None = None,
        renderer: RendererOptions | None = None,
        show_viewer: bool | None = None,
        show_FPS: bool | None = None,  # deprecated, use profiling_options.show_FPS instead
    ):
```

examples/manipulation/grasp_env.py
yun-long: my bad. initially, i was use depth image as input. I guess I will simply remove this function since it is not used now.

examples/manipulation/behavior_cloning.py
yun-long: Yes. it makes sense to keep the `dtype` of inputs same as the `dtype` of the policy. I have now made the `dtype` as an `__init__` argument and initialize it via the `dtype` from policy.

examples/manipulation/grasp_eval.py
yun-long: this is not done yet. I will address it tomorrow. have to go home now.

yun-long: There is a potential bug here. 

Description: When I add an additional `vis_cam` with a resolution of (1280, 720), it overrides the resolutions of my existing `left_cam` and `right_cam`, which are originally set to (64, 64). As a result, all camera outputs (left, right, and vis_cam) are forced to use the same resolution of `(1280, 720)`. I think this needs further investigation. For this particular example, I will just keep left and right camera. 

```python
        if self.env_cfg["visualize_camera"]:
            self.vis_cam = self.scene.add_camera(
                res=(1280, 720),
                pos=(1.5, 0.0, 0.2),
                lookat=(0.0, 0.0, 0.2),
                fov=60,
                GUI=self.env_cfg["visualize_camera"],
            )

        # == add stero camera ==
        self.left_cam = self.scene.add_camera(
            res=(self.image_width, self.image_height),
            pos=(1.25, 0.3, 0.3),
            lookat=(0.0, 0.0, 0.0),
            fov=60,
            GUI=self.env_cfg["visualize_camera"],
        )
        self.right_cam = self.scene.add_camera(
            res=(self.image_width, self.image_height),
            pos=(1.25, -0.3, 0.3),
            lookat=(0.0, 0.0, 0.0),
            fov=60,
            GUI=self.env_cfg["visualize_camera"],
        )
```

YilingQiao:
Could you please add the description in this PR as a READE.md to this folder? Such that users have an idea how to use it.

yun-long:
> Could you please add the description in this PR as a READE.md to this folder? Such that users have an idea how to use it.

Yes. I will add proper documentation this week.

examples/manipulation/behavior_cloning.py
duburcqa: Yes!

examples/manipulation/grasp_env.py
duburcqa: Ok. I would rather go for a (lengthy) comment instead of arguably "useless" math to serve as documentation since it is not free-of-cost, but not a big deal in this case because it is definitely not a bottleneck at all.

examples/manipulation/grasp_env.py
duburcqa: > I feel it is already quite clear what the output/interface should be given the clear variable names.

Variable names are not visible when the docstring is being displayed, but type annotations are.

> On the other hand, if we are going to provide type alias for this particularly function, what about other functions, e.g.,

Yes, adding aliases for all of them would be great I think! But step is probably the main entry-point for the user, so improving typing for `step` alone is already a good start.

examples/manipulation/grasp_env.py
duburcqa: > If it is ok for you, I will keep the current version as I feel it is more explicit.

Then you need to benchmark. If the difference is not significant, it is fine for me. I have no idea how much of a bottleneck this part of the code can be. Another option would be to just run some profiling to determine how much time is spent in post-processing, e.g. using `py-spy`. 

But if you already know it is not a bottleneck, I'm fine with keeping the existing version. Still, it would be nice to add a comment that mention what I'm suggesting myself, for readers that may be interested in squeeze every possible bits of performance.

examples/manipulation/grasp_env.py
duburcqa: > hmmm, why not following the default order

You can do this, it is consistent with what I said, with a few insignificant exceptions.

examples/manipulation/grasp_eval.py
duburcqa: Ok I'm fine with it.

examples/manipulation/grasp_eval.py
duburcqa: I think this bug should be fixed before merging. I will work on it.

duburcqa: I just realised the issue ! you forgot to specify `debug=True` for your `vis_cam` !!

yun-long: I see. Updated my script. Now, it uses the default camera rescoring, similar to the previous examples.

yun-long:
See this [PR](https://github.com/Genesis-Embodied-AI/genesis-doc/pull/68) for updated documentations.

duburcqa (approved):

PR #1498, opened 2025-08-04
[BUG FIX] Fix gravity setters.

Milotrince:
## Description
Allow for `set_gravity` per env

## Related Issue
`set_gravity` was first introduced in https://github.com/Genesis-Embodied-AI/Genesis/pull/1324

## Motivation and Context
Previous implementation did not properly set gravity per environment; taichi field should be set using kernel (no slicing)

Updated `test_rigid_physics.py`

## Checklist:

### DIFF
--- a/genesis/engine/solvers/base_solver.py
+++ b/genesis/engine/solvers/base_solver.py
@@ -42,7 +42,7 @@ class Solver(RBC):
     @gs.assert_built
     def set_gravity(self, gravity, envs_idx=None):
         if self._gravity is None:
-            gs.logger.warning("Gravity is not initialized, skipping set_gravity.")
+            gs.logger.debug("Gravity is not defined, skipping `set_gravity`.")
             return
         g = np.asarray(gravity, dtype=gs.np_float)
         if envs_idx is None:
--- a/genesis/engine/solvers/base_solver.py
+++ b/genesis/engine/solvers/base_solver.py
@@ -49,7 +49,14 @@ class Solver(RBC):
                 g = np.tile(g, (self._B, 1))
             self._gravity.from_numpy(g)
         else:
-            self._gravity[envs_idx] = g
+            assert g.shape == (envs_idx.shape[0], 3), "Input gravity array should match (len(envs_idx), 3)"
+            self._kernel_set_gravity(g, envs_idx)
+
+    @ti.kernel
+    def _kernel_set_gravity(self, gravity: ti.types.ndarray(), envs_idx: ti.types.ndarray()):
+        for i_b_ in range(envs_idx.shape[0]):
+            for j in ti.static(range(3)):
+                self._gravity[envs_idx[i_b_]][j] = gravity[i_b_, j]
 
     def dump_ckpt_to_numpy(self) -> dict[str, np.ndarray]:
         arrays: dict[str, np.ndarray] = {}
--- a/genesis/engine/solvers/base_solver.py
+++ b/genesis/engine/solvers/base_solver.py
@@ -42,14 +42,19 @@ class Solver(RBC):
     @gs.assert_built
     def set_gravity(self, gravity, envs_idx=None):
         if self._gravity is None:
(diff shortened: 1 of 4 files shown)


genesis/engine/solvers/base_solver.py
```diff
@@ -42,14 +42,26 @@ def build(self):
     @gs.assert_built
     def set_gravity(self, gravity, envs_idx=None):
         if self._gravity is None:
+            gs.logger.warning("Gravity is not initialized, skipping set_gravity.")
```
duburcqa: `gs.logger.debug` should be enough.

"Gravity is not defined, skipping 'set_gravity'." would be more appropriate.

tests/test_rigid_physics.py
```diff
@@ -2177,18 +2177,38 @@ def test_gravity(show_viewer, tol):
     )
 
     sphere = scene.add_entity(gs.morphs.Sphere())
-    scene.build(n_envs=2)
+    scene.build(n_envs=7)
```
duburcqa: Prefer `n_envs=2`, or `n_envs=3`, it should be enough.

tests/test_rigid_physics.py
```diff
@@ -2177,18 +2177,38 @@ def test_gravity(show_viewer, tol):
...
+                [0.0, 0.0, 0.0],
+                [0.0, 0.0, 0.0],
+                [0.0, 0.0, -1.0],
+                [0.0, 0.0, 10.0],
+                [0.0, 0.0, -0.1],
+                [-0.2, 0.0, 0.0],
+                [-0.2, 0.0, 0.0],
+            ]
+        ),
+        sphere.get_vel(),
```
duburcqa: You should probably check the acceleration instead.

tests/test_rigid_physics.py
Milotrince: I have 7 because I am asserting later that each variation of set_gravity has correctly set it for that env, ok to keep?

duburcqa: All these variations are not relevant. Just randomly initialise the gravity for one of the environment. There is no dedicated logic for each gravity component so default vs non-default should be enough.

tests/test_rigid_physics.py
```diff
@@ -2177,18 +2177,37 @@ def test_gravity(show_viewer, tol):
...
+                [0.0, 0.0, 0.0],
+                [0.0, 0.0, 0.0],
+                [0.0, 0.0, -10.0],
+                [0.0, 0.0, 100.0],
+                [0.0, 0.0, -1.0],
+                [-2.0, 0.0, 0.0],
+                [-2.0, 0.0, 0.0],
+            ]
+        ),
+        sphere.get_links_acc().squeeze(dim=1),
```
duburcqa: Are you sure squeezing is necessary?

tests/test_rigid_physics.py
```diff
@@ -2177,18 +2177,37 @@ def test_gravity(show_viewer, tol):
...
-    first_pos = sphere.get_dofs_position()[0, 2]
-    second_pos = sphere.get_dofs_position()[1, 2]
+    with np.testing.assert_raises(AssertionError):
+        scene.sim.set_gravity(torch.tensor([[0.0, 0.0, -10.0], [0.0, 0.0, -10.0]]), envs_idx=1)
+
+    scene.step()
 
-    assert_allclose(first_pos * -1, second_pos, tol=tol)
+    assert_allclose(
+        np.array(
```
duburcqa: No need to convert to array.

tests/test_rigid_physics.py
Milotrince: yes, because get_links_acc returns (batch, n_links, 3)

duburcqa: What about `sphere.get_links_acc().squeeze()`? I find it a bit more intuitive but as you want.

duburcqa (approved):

PR #1499, opened 2025-08-04
[MISC] Make collision bitmasks filter local to entity for MJCF morphs.

duburcqa:
## Description

Add extra flag to only apply collision pairs filtering based on contype / conaffinity bitmasks locally (i.e. for geometries associated with the same entity), then enforcing this flag to True for MJCF morphs, False otherwise.

## Motivation and Context

The collision bitmasks filters are ill-defined for MJCF, in a sense that they are dynamically generated, based on a the combination of some original bitmasks with an optional set of explicitly excluded collision pairs. Moreover, Mujoco does not allow loading multiple MJCF in the scene, nor does it allow to add additional entities manually afterward. This meanst that contype / conaffinity can be viewed as systemically "local". As a result, it would be error prone to use them to filter collision pairs among different MJCF entities. Still, it is relevant to have the ability to enable cross-entity contype / conaffinity for primitives and meshes, since it is perfectly well-defined in this case, and basically aligns with what is done within MJCF. 

Running the example script `examples/collision/contype.py`

## Checklist:

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_link.py
+++ b/genesis/engine/entities/rigid_entity/rigid_link.py
@@ -54,7 +54,7 @@ class RigidLink(RBC):
         self._name: str = name
         self._entity: "RigidEntity" = entity
         self._solver: "RigidSolver" = entity.solver
-        self._entity_idx_in_solver = entity.idx
+        self._entity_idx_in_solver = entity._idx_in_solver
 
         self._uid = gs.UID()
         self._idx: int = idx


YilingQiao (approved):

genesis/options/morphs.py
```diff
@@ -685,13 +684,25 @@ class MJCF(FileMorph):
...
-    MJCF file always contains a worldbody, which we will skip during loading.
-    The robots/objects in MJCF come with their own baselink pose.
-    If `pos`, `euler`, or `quat` is specified, it will override the baselink pose in the MJCF file.
+    MJCF file always contains a 'world' body. Although this body is added to the kinematic tree, it is used to define
+    the initial pose of the root link. If `pos`, `euler`, or `quat` is specified, it will override the root pose that
+    was originally specified in the MJCF file.
+
+    Note
+    ----
+    Genesis currently processes MJCF as if it describing a single entity instead of an actual scene. This means that
```
Kashu7100, line 693: "describing" -> "describes"

genesis/options/morphs.py
```diff
@@ -685,13 +684,25 @@ class MJCF(FileMorph):
...
+    there is a single gigantic kinematic chain comprising multiple physical kinematic chains connected together using
+    fee joints. The definition of kinematic chain has been stretched a bit to allow us. In particular, there must be
```
Kashu7100, line 695: "connected together using fee joints" -> "connected by free joints"

Kashu7100: "to allow us" -> "to allow for this"

Kashu7100 (approved):

PR #1518, opened 2025-08-07
[FEATURE] Add dedicated sensor manager.

Milotrince:
## Description
Add SensorManager to Simulator. Sensor classes can now have per-scene multi-sensor caching.

## Related Issue
Rigid Tactile Sensors https://github.com/Genesis-Embodied-AI/Genesis/pull/1451 will be rebased on this PR.
IMU Sensor https://github.com/Milotrince/Genesis/pull/1 is also based on this PR. (I can't do a "stacked PR" directly in Genesis repo since my branch doesn't live there, which is why the PR currently lives in my fork of Genesis)

## Motivation and Context
The current sensor implementation does not allow for caching across multiple sensors, and usage is un-Genesis-like (unintuitive, e.g. does not align with `add_camera` pattern). See new [IMU](https://github.com/Milotrince/Genesis/pull/1) for an implementation/usage example.

Will add tests per sensor.

## Checklist:

### DIFF
--- a/genesis/sensors/base_sensor.py
+++ b/genesis/sensors/base_sensor.py
@@ -1,4 +1,4 @@
-from typing import Any, List, Optional
+from typing import Any, List, Optional, TYPE_CHECKING
 
 import numpy as np
 import taichi as ti
@@ -8,7 +8,10 @@ import genesis as gs
 from genesis.options.sensors import SensorOptions
 from genesis.repr_base import RBC
 
-from .sensor_manager import SensorManager
+if TYPE_CHECKING:
+    from genesis.utils.ring_buffer import TensorRingBuffer
+    from genesis.options.sensors import SensorOptions
+    from .sensor_manager import SensorManager
 
 
 @ti.data_oriented
@@ -23,6 +26,8 @@ class Sensor(RBC):
         self._idx: int = sensor_idx
         self._manager: SensorManager = sensor_manager
         self._cache_idx: int = -1  # cache_idx is set by the SensorManager during the scene build phase
+        self._read_delay_steps: int = 0
+        self._cache_buffer_length: int = 0
 
     # =============================== implementable methods ===============================
 
@@ -32,7 +37,14 @@ class Sensor(RBC):
         This method is called by the SensorManager during the scene build phase to initialize the sensor.
         This is where any shared metadata should be initialized.
         """
-        pass
+        delay_steps_float = self._options.read_delay / self._manager._sim.dt
+        self._read_delay_steps = round(delay_steps_float)
+        if not np.isclose(delay_steps_float, self._read_delay_steps):
+            gs.logger.warn(
+                f"Read delay should be a multiple of the simulation time step. Got {self._options.read_delay} and "
+                f"{self._manager._sim.dt}. Actual read delay will be {1/self._read_delay_steps}."
+            )
+        self._cache_buffer_length = self._read_delay_steps + 1
 
     def _get_return_format(self) -> dict[str, tuple[int, int]] | None:
         """
@@ -42,6 +54,14 @@ class Sensor(RBC):
         None: the entire tensor (B, cache_length, cache_shape) will be returned.
         """
         raise NotImplementedError("Sensors must implement `return_format()`.")
+
+    @gs.assert_built
+    def _update_shared_gt_cache(self):
+        """
+        Update the shared sensor ground truth cache for all sensors of this class.
+        Shared information is stored in SensorManager
+        """
+        raise NotImplementedError("Sensors must implement `update_shared_gt_cache()`.")
 
     @gs.assert_built
     def _update_shared_cache(self):
@@ -76,7 +96,7 @@ class Sensor(RBC):
     @gs.assert_built
     def read(self, envs_idx: Optional[List[int]] = None):
         """
-        Read the sensor data.
+        Read the sensor data (with noise applied if applicable) from SensorManager._cache.
         """
         if envs_idx is None:
             envs_idx = 0 if self._manager._sim.n_envs == 0 else np.arange(self._cache.shape[0])
@@ -87,7 +107,27 @@ class Sensor(RBC):
         else:
             cache_length = self._get_cache_length()
             return {
-                key: self._cache[
+                key: self._cache.get(self._read_delay_steps)[
+                    envs_idx, self._cache_idx : self._cache_idx + cache_length, start_idx:end_idx
+                ].squeeze()
+                for key, (start_idx, end_idx) in return_format.items()
+            }
+
+    @gs.assert_built
+    def read_ground_truth(self, envs_idx: Optional[List[int]] = None):
+        """
+        Read the ground truth sensor data (without noise) from SensorManager._gt_cache.
+        """
+        if envs_idx is None:
+            envs_idx = 0 if self._manager._sim.n_envs == 0 else np.arange(self._gt_cache.shape[0])
+
+        return_format = self._get_return_format()
+        if return_format is None:
+            return self._gt_cache[envs_idx, self._cache_idx, :].squeeze()
+        else:
+            cache_length = self._get_cache_length()
+            return {
+                key: self._gt_cache[
                     envs_idx, self._cache_idx : self._cache_idx + cache_length, start_idx:end_idx
                 ].squeeze()
                 for key, (start_idx, end_idx) in return_format.items()
@@ -97,13 +137,13 @@ class Sensor(RBC):
     def is_built(self) -> bool:
         return self._manager._sim._scene._is_built
 
-    @gs.assert_built
-    def _get_cache(self) -> torch.Tensor:
-        return self._manager.get_sensor_cache(self.__class__, self._cache_idx)
+    @property
+    def _cache(self) -> "TensorRingBuffer":
+        return self._manager._cache[self.__class__]
 
     @property
-    def _cache(self) -> torch.Tensor:
-        return self._manager._cache[self.__class__]
+    def _gt_cache(self) -> torch.Tensor:
+        return self._manager._gt_cache[self.__class__]
 
     @property
     def _shared_metadata(self) -> dict[str, Any]:
--- a/genesis/sensors/base_sensor.py
+++ b/genesis/sensors/base_sensor.py
@@ -114,7 +114,7 @@ class Sensor(RBC):
 
     def _get_return_data_from_tensor(self, tensor: torch.Tensor, envs_idx: List[int] | None) -> torch.Tensor:
         if envs_idx is None:
-            envs_idx = 0 if self._manager._sim.n_envs == 0 else np.arange(tensor.shape[0])
+            envs_idx = [0] if self._manager._sim.n_envs == 0 else np.arange(tensor.shape[0])
 
         return_format = self._get_return_format()
 
@@ -122,18 +122,18 @@ class Sensor(RBC):
             return (
                 tensor[envs_idx, self._cache_idx : self._cache_idx + self._cache_size]
                 .clone()
-                .reshape(return_format)
+                .reshape(len(envs_idx), *return_format)
                 .squeeze()
             )
         elif isinstance(return_format, dict):
             data_dict = {}
+            data_tensor = tensor[envs_idx, self._cache_idx : self._cache_idx + self._cache_size].clone()
             tensor_idx = 0
             for key, data_shape in return_format.items():
                 data_size = np.prod(data_shape)
                 data_dict[key] = (
-                    tensor[envs_idx, self._cache_idx : self._cache_idx + self._cache_size]
-                    .clone()[tensor_idx : tensor_idx + data_size]
-                    .reshape(data_shape)
+                    data_tensor[0 : len(envs_idx), tensor_idx : tensor_idx + data_size]
+                    .reshape(len(envs_idx), *data_shape)
                     .squeeze()
                 )
                 tensor_idx += data_size
--- a/genesis/sensors/base_sensor.py
+++ b/genesis/sensors/base_sensor.py
@@ -21,10 +21,10 @@ class Sensor(RBC):
     A sensor must have a read() method that returns the sensor data.
     """
 
-    def __init__(self, sensor_options: SensorOptions, sensor_idx: int, sensor_manager: SensorManager):
-        self._options: SensorOptions = sensor_options
+    def __init__(self, sensor_options: "SensorOptions", sensor_idx: int, sensor_manager: "SensorManager"):
+        self._options: "SensorOptions" = sensor_options
         self._idx: int = sensor_idx
-        self._manager: SensorManager = sensor_manager
+        self._manager: "SensorManager" = sensor_manager
         self._cache_idx: int = -1  # cache_idx is set by the SensorManager during the scene build phase
         self._read_delay_steps: int = 0
         self._cache_buffer_length: int = 0
--- a/genesis/sensors/base_sensor.py
+++ b/genesis/sensors/base_sensor.py
@@ -25,17 +25,12 @@ class Sensor(RBC):
         self._options: "SensorOptions" = sensor_options
         self._idx: int = sensor_idx
         self._manager: "SensorManager" = sensor_manager
-        self._class: Type[Sensor] = self.__class__
 
         # initialized during build
-        self._cache_idx: int = -1  # set by the SensorManager during build
         self._read_delay_steps: int = 0
         # initialized by SceneManager during build
         self._shape_indices: list[tuple[int, int]] = []
-        self._cache_size: int = 0
-        self._cache: torch.Tensor | None = None
-        self._ground_truth_cache: torch.Tensor | None = None
-        self._shared_metadata: dict[str, Any] = {}
+        self._shared_metadata: dict[str, Any] | None = None
 
     # =============================== implementable methods ===============================
 
@@ -52,18 +47,6 @@ class Sensor(RBC):
                 f"Read delay should be a multiple of the simulation time step. Got {self._options.read_delay} and "
                 f"{self._manager._sim.dt}. Actual read delay will be {1/self._read_delay_steps}."
             )
-
-        return_format = self._get_return_format()
-        return_length = 0
-        if isinstance(return_format, dict):
-            return_length = sum(sum(shape) for shape in return_format.values())
-        elif isinstance(return_format, tuple):
-            return_length = sum(return_format)
-        else:
-            raise ValueError(f"Invalid sensor return format: {return_format}, should be tuple or dict")
-
-        self._cache_size = self._get_cache_length() * return_length
-        self._shared_metadata = self._manager._sensors_metadata[self._class]
 
     def _get_return_format(self) -> dict[str, tuple[int, ...]] | tuple[int, ...]:
         """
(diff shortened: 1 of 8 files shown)


YilingQiao (approved):

genesis/sensors/base_sensor.py
```diff
@@ -12,10 +17,62 @@ class Sensor(RBC):
...
         """
-        raise NotImplementedError("The Sensor subclass must implement `read()`.")
+        raise NotImplementedError("Sensors must implement `read()`.")
+
+    @property
+    def cache_length(self) -> int:
+        """
+        The length (first dimension of cache size) of the cache for this sensor.
+        """
+        return 1
```
Milotrince: 0 by default; something like lidar have `n`, and imu has `0` (special return processing)

Milotrince:
Some design decisions made:


1. Why have a `SensorManager`?

Typically in a scene, a robot may have multiple of the same type of sensor. Previously the implementation tried to cache sensor data at the class level, but this will not work when running multiple scenes. The `SensorManager` is owned by `Simulator` and thus can operate at the per-scene, per-sensor-type level. It can also automatically update all sensors (see below)

2. Why does SensorManager need to updated all sensors at `step()` instead of computing values at read?

For time delay and other types of noise, we will need to buffer the history. So either users will have to call step themselves (or otherwise face unintended/inaccurate values), or we can cleanly handle this inside of the simulator step (more aligned to Genesis's usage design).

3. Why is `read()` now handled at the base `Sensor`?

For future contributions of new sensors, we want to enforce usage of the cross-batch, cross-sensor-type caching mechanism to ensure high sensor performance in Genesis.

genesis/sensors/base_sensor.py
```diff
@@ -12,10 +21,127 @@ class Sensor(RBC):
...
+        This is where any shared metadata should be initialized.
+        """
+        delay_steps_float = self._options.read_delay / self._manager._sim.dt
+        self._read_delay_steps = round(delay_steps_float)
+        if not np.isclose(delay_steps_float, self._read_delay_steps):
+            gs.logger.warn(
+                f"Read delay should be a multiple of the simulation time step. Got {self._options.read_delay} and "
+                f"{self._manager._sim.dt}. Actual read delay will be {1/self._read_delay_steps}."
+            )
+        self._cache_buffer_length = self._read_delay_steps + 1
```
duburcqa: I don' think it is the right way to do it. The cache buffer length should be the same for all sensors, and the sensor manager should use `read_delay_step = max(sensor._read_delay_steps for sensor in self._sensors)` to determine how many data mush be stored in cache. It will make everything much easier to implement and faster, but increase the memory footprint to some extent.

genesis/sensors/base_sensor.py
```diff
@@ -12,10 +21,127 @@ class Sensor(RBC):
...
+        Data format of the read() return value.
+        dict: a dictionary with string keys and tensor values will be returned.
+              The tuple (start_idx, end_idx) will be used to get a slice of the values from the cache.
+        tuple: the slice of the cache will be returned.
+        None: the entire tensor (B, cache_length, cache_shape) will be returned.
+        """
+        raise NotImplementedError("Sensors must implement `return_format()`.")
+
+    @gs.assert_built
+    def _update_shared_gt_cache(self):
```
duburcqa: `gt` is not a great acronym because it usually means "greater than". What about 'true' or 'raw' ? Or just 'ground_truth'

genesis/sensors/base_sensor.py
```diff
@@ -12,10 +21,127 @@ class Sensor(RBC):
...
+        """
+        Update the shared sensor ground truth cache for all sensors of this class.
+        Shared information is stored in SensorManager
+        """
+        raise NotImplementedError("Sensors must implement `update_shared_gt_cache()`.")
+
     @gs.assert_built
```
duburcqa: No need to assert build for private methods I think but as you want.

genesis/sensors/base_sensor.py
```diff
@@ -12,10 +21,127 @@ class Sensor(RBC):
...
-    def read(self, envs_idx: Optional[List[int]] = None):
+    def _update_shared_cache(self):
         """
-        Read the sensor data.
-        Sensor implementations should ideally cache the data to avoid unnecessary computations.
+        Update the shared sensor cache for all sensors of this class using information stored in shared metadata.
+        """
+        raise NotImplementedError("Sensors must implement `update_shared_cache()`.")
+
+    def _get_cache_length(self) -> int:
```
duburcqa: You should describe what the cache length means, eventually with examples.

genesis/sensors/base_sensor.py
```diff
@@ -12,10 +21,127 @@ class Sensor(RBC):
...
+        delay_steps_float = self._options.read_delay / self._manager._sim.dt
+        self._read_delay_steps = round(delay_steps_float)
+        if not np.isclose(delay_steps_float, self._read_delay_steps):
+            gs.logger.warn(
+                f"Read delay should be a multiple of the simulation time step. Got {self._options.read_delay} and "
+                f"{self._manager._sim.dt}. Actual read delay will be {1/self._read_delay_steps}."
+            )
+        self._cache_buffer_length = self._read_delay_steps + 1
+
+    def _get_return_format(self) -> dict[str, tuple[int, int]] | tuple[int, int] | None:
```
duburcqa: You should provide some concrete examples.

Besides, the formatting of your docstring is not correct.

genesis/sensors/base_sensor.py
```diff
@@ -12,10 +21,127 @@ class Sensor(RBC):
...
+        tuple: the slice of the cache will be returned.
+        None: the entire tensor (B, cache_length, cache_shape) will be returned.
+        """
+        raise NotImplementedError("Sensors must implement `return_format()`.")
+
+    @gs.assert_built
+    def _update_shared_gt_cache(self):
+        """
+        Update the shared sensor ground truth cache for all sensors of this class.
+        Shared information is stored in SensorManager
```
duburcqa: Skip line between short documentation and "details".

genesis/sensors/base_sensor.py
Milotrince: yes the `SensorManager.build()` is currently doing this (taking the max for sensors of that type)-- I don't remember why I decided to store the `_cache_buffer_length` in each sensor first though, I can move it so SensorManager has all of this logic

genesis/sensors/base_sensor.py
```diff
@@ -12,10 +21,127 @@ class Sensor(RBC):
...
+
+    def _get_cache_length(self) -> int:
+        """
+        The length (first dimension of cache shape) of the cache for this sensor instance.
+        The sum of cache_length for all sensors of this type determines the length of the cache.
+        """
+        raise NotImplementedError("Sensors must implement `cache_length()`.")
+
+    @classmethod
+    def _get_cache_size(cls) -> int:
```
duburcqa: Having 'get_cache_size', 'get_cache_length', and 'get_return_format' is too much. Let's discuss this together.

genesis/sensors/sensor_manager.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class SensorManager:
+    SENSOR_TYPES_MAP: dict[Type["SensorOptions"], Type["Sensor"]] = {}
+
+    def __init__(self, sim):
+        self._sim = sim
+        self._sensors_by_type: dict[Type["Sensor"], list["Sensor"]] = {}
+        self._sensors_metadata: dict[Type["Sensor"], dict[str, Any]] = {}
+        self._gt_cache: dict[Type["Sensor"], torch.Tensor] = {}
```
duburcqa: For cache could be global, and sensor-type based. it is the big advantage of storing data flattened. But you need one cache per datatype still.

genesis/sensors/base_sensor.py
```diff
@@ -12,10 +21,138 @@ class Sensor(RBC):
...
+
+    @gs.assert_unbuilt
+    def build(self):
+        """
+        This method is called by the SensorManager during the scene build phase to initialize the sensor.
+        This is where any shared metadata should be initialized.
+        """
+        delay_steps_float = self._options.read_delay / self._manager._sim.dt
+        self._read_delay_steps = round(delay_steps_float)
+        if not np.isclose(delay_steps_float, self._read_delay_steps):
```
duburcqa: You should probably set the absolute tolerance yourself.

genesis/sensors/base_sensor.py
```diff
@@ -12,10 +21,138 @@ class Sensor(RBC):
...
+            gs.logger.warn(
+                f"Read delay should be a multiple of the simulation time step. Got {self._options.read_delay} and "
+                f"{self._manager._sim.dt}. Actual read delay will be {1/self._read_delay_steps}."
```
duburcqa: `self._read_delay_steps * self._manager._sim.dt` instead of `1/self._read_delay_steps` ?

genesis/sensors/base_sensor.py
```diff
@@ -12,10 +21,138 @@ class Sensor(RBC):
...
+            )
+
+        return_format = self._get_return_format()
+        return_length = 0
```
duburcqa: Usually the number of elements of an array is called `size`, not `length`

genesis/sensors/base_sensor.py
```diff
@@ -12,10 +21,138 @@ class Sensor(RBC):
...
+        raise NotImplementedError("Sensors must implement `get_cache_dtype()`.")
+
+    # =============================== shared methods ===============================
+
+    @gs.assert_built
+    def read(self, envs_idx: List[int] | None = None):
+        """
+        Read the sensor data (with noise applied if applicable).
+        """
+        return self._get_return_data_from_tensor(self._cache.get(self._read_delay_steps), envs_idx)
```
duburcqa: This is weird. What is `self._cache` ? It is not declared anywhere.

genesis/sensors/base_sensor.py
```diff
@@ -12,10 +21,138 @@ class Sensor(RBC):
...
-        Read the sensor data.
-        Sensor implementations should ideally cache the data to avoid unnecessary computations.
+        Read the ground truth sensor data (without noise).
         """
-        raise NotImplementedError("The Sensor subclass must implement `read()`.")
+        return self._get_return_data_from_tensor(self._ground_truth_cache, envs_idx)
+
+    def _get_return_data_from_tensor(self, tensor: torch.Tensor, envs_idx: List[int] | None) -> torch.Tensor:
+        if envs_idx is None:
+            envs_idx = 0 if self._manager._sim.n_envs == 0 else np.arange(tensor.shape[0])
```
duburcqa: Could you preallocate the full vector of environment as it is done in `Scene` ? For maximum efficiency is should be a torch tensor on device if not a scalar.

genesis/sensors/base_sensor.py
```diff
@@ -12,10 +21,138 @@ class Sensor(RBC):
...
+                data_dict[key] = (
+                    tensor[envs_idx, self._cache_idx : self._cache_idx + self._cache_size]
+                    .clone()[tensor_idx : tensor_idx + data_size]
+                    .reshape(data_shape)
+                    .squeeze()
+                )
+                tensor_idx += data_size
+            return data_dict
+        else:
+            raise ValueError(f"Invalid sensor return format: {return_format}")
```
duburcqa: Remove this. You already know that the format is valid at the point. it was checked at build time.

genesis/sensors/base_sensor.py
```diff
@@ -12,10 +21,138 @@ class Sensor(RBC):
...
+    def read_ground_truth(self, envs_idx: List[int] | None = None):
         """
-        Read the sensor data.
-        Sensor implementations should ideally cache the data to avoid unnecessary computations.
+        Read the ground truth sensor data (without noise).
         """
-        raise NotImplementedError("The Sensor subclass must implement `read()`.")
+        return self._get_return_data_from_tensor(self._ground_truth_cache, envs_idx)
+
+    def _get_return_data_from_tensor(self, tensor: torch.Tensor, envs_idx: List[int] | None) -> torch.Tensor:
```
duburcqa: This is basically `_get_unflattened` ? More common terminology.

genesis/sensors/base_sensor.py
```diff
@@ -12,10 +21,138 @@ class Sensor(RBC):
...
+        if envs_idx is None:
+            envs_idx = 0 if self._manager._sim.n_envs == 0 else np.arange(tensor.shape[0])
+
+        return_format = self._get_return_format()
+
+        if isinstance(return_format, tuple):
+            return (
+                tensor[envs_idx, self._cache_idx : self._cache_idx + self._cache_size]
+                .clone()
```
duburcqa: Just assume that the input data was cloned before calling this function. (and add this detail in the documentation)

genesis/sensors/base_sensor.py
```diff
@@ -12,10 +21,138 @@ class Sensor(RBC):
...
+                .reshape(return_format)
+                .squeeze()
```
duburcqa: You should not squeeze unconditionally.

genesis/sensors/base_sensor.py
```diff
@@ -12,10 +21,138 @@ class Sensor(RBC):
...
-        raise NotImplementedError("The Sensor subclass must implement `read()`.")
+        return self._get_return_data_from_tensor(self._ground_truth_cache, envs_idx)
+
+    def _get_return_data_from_tensor(self, tensor: torch.Tensor, envs_idx: List[int] | None) -> torch.Tensor:
+        if envs_idx is None:
+            envs_idx = 0 if self._manager._sim.n_envs == 0 else np.arange(tensor.shape[0])
+
+        return_format = self._get_return_format()
+
+        if isinstance(return_format, tuple):
```
duburcqa: Avoid code duplication by doing:
```python
return_shapes = return_format.values() if isinstance(return_format, dict) else (return_format,)
return_values = []
for shape in return_shapes:
    value = ...
    return_values.append(value)

if isinstance(return_format, dict):
    return dict(zip(return_format.keys(), return_values))
return return_values[0]
```

genesis/sensors/base_sensor.py
```diff
@@ -12,10 +21,138 @@ class Sensor(RBC):
...
+                .squeeze()
+            )
+        elif isinstance(return_format, dict):
+            data_dict = {}
+            tensor_idx = 0
+            for key, data_shape in return_format.items():
+                data_size = np.prod(data_shape)
+                data_dict[key] = (
+                    tensor[envs_idx, self._cache_idx : self._cache_idx + self._cache_size]
+                    .clone()[tensor_idx : tensor_idx + data_size]
```
duburcqa: You should precompute your exact start-stop indices for each value instead of doing it at runtime.

genesis/sensors/base_sensor.py
```diff
@@ -12,10 +21,138 @@ class Sensor(RBC):
...
+                .reshape(return_format)
+                .squeeze()
+            )
+        elif isinstance(return_format, dict):
+            data_dict = {}
+            tensor_idx = 0
+            for key, data_shape in return_format.items():
+                data_size = np.prod(data_shape)
+                data_dict[key] = (
+                    tensor[envs_idx, self._cache_idx : self._cache_idx + self._cache_size]
```
duburcqa: Start by extracting `tensor[envs_idx, self._cache_idx : self._cache_idx + self._cache_size]` once, instead of doing it repeatedly.

genesis/sensors/sensor_manager.py
```diff
@@ -0,0 +1,69 @@
...
+        sensor_cls = SensorManager.SENSOR_TYPES_MAP[type(sensor_options)]
+        sensor = sensor_cls(sensor_options, len(self._sensors), self)
+        self._sensors.append(sensor)
+        return sensor
+
+    def build(self):
+        max_cache_buf_len = 0
+        cache_size_per_dtype = {}
+        for sensor in self._sensors:
+            if sensor.__class__ not in self._sensors_metadata:
```
duburcqa: Define variable `sensor_type = type(sensor)` instead of using `sensor.__class__` everywhere.

genesis/sensors/sensor_manager.py
```diff
@@ -0,0 +1,69 @@
...
+                self._sensors_metadata[sensor.__class__] = {}
+            if sensor.__class__ not in self._first_sensor_per_type:
+                self._first_sensor_per_type[sensor.__class__] = sensor
```
duburcqa: Just use `setdefault`:
```python
self._sensors_metadata.setdefault(sensor_type, {})
self._first_sensor_per_type.setdefault(sensor_type, sensor)
```

duburcqa: By the way, you don't care about which one is the first sensor of a given type actually?

By the way about just storing `self._sensors_by_type` as you did previously? That you would get the "first" sensor per type for free, without an additional attribute. Getting all sensors from a given type efficiently may be helpful in the future.

genesis/sensors/sensor_manager.py
```diff
@@ -0,0 +1,69 @@
...
+
+            sensor.build()
+
+            if sensor._get_cache_dtype() not in cache_size_per_dtype:
+                cache_size_per_dtype[sensor._get_cache_dtype()] = 0
```
duburcqa: Same.
```python
cache_size_per_dtype.set_default(sensor._get_cache_dtype(), 0)
```

genesis/sensors/sensor_manager.py
```diff
@@ -0,0 +1,69 @@
...
+
+            sensor._cache_idx = cache_size_per_dtype[sensor._get_cache_dtype()]
```
duburcqa: Create a variable for `dtype`:
```python
dtype = sensor._get_cache_dtype()
```

genesis/sensors/sensor_manager.py
```diff
@@ -0,0 +1,69 @@
...
+            self._cache[dtype] = TensorRingBuffer(max_cache_buf_len, cache_shape, dtype=dtype)
+
+    def step(self):
+        for sensor in self._first_sensor_per_type.values():
+            sensor._update_shared_ground_truth_cache()
+            sensor._update_shared_cache()
+
+    @property
+    def sensors(self):
+        return [sensor for sensor_list in self._sensors_by_type.values() for sensor in sensor_list]
```
duburcqa: `self._sensors_by_type` is not define.

duburcqa: Prefer tuple: `tuple(sensor for sensor_list in self._sensors_by_type.values() for sensor in sensor_list)`

genesis/sensors/base_sensor.py
Milotrince: `self._cache` is `self._manager._cache[self._get_cache_dtype()]` (defined below in file); I added this property "alias" since I thought it would be used often, but i dont really think it is anymore; I can remove it since it makes things confusing/unclear

duburcqa: Why not to make it an attribute at build time ? It would be more clear I think.

genesis/sensors/base_sensor.py
Milotrince: how about `_get_formatted_data` since it can also return dict based on `self._get_return_format()`?

genesis/sensors/base_sensor.py
Milotrince: I would rather clone the tensor in the function because otherwise the entire cache tensor would be cloned instead of the slice of the cache?

genesis/sensors/base_sensor.py
duburcqa: Sounds good :)

genesis/sensors/base_sensor.py
duburcqa: > I would rather clone the tensor in the function because otherwise the entire cache tensor would be cloned instead of the slice of the cache?

I mean, yes, the entire slice of the cache should be cloned once, for all the sensors at once.

Milotrince: ah, understand

Milotrince: do you think there should be an option to clone/not clone the data from read()? assuming all the user's processing with the sensor data is performed after every scene step and not holding onto any old sensor data, it would not be necessary to clone every time

genesis/sensors/base_sensor.py
Milotrince: `sensor.build()` is inside of `SensorManager.build()` so it's a little tricky/messy. How about keeping the `@property` but renaming `sensor._cache` to `sensor._shared_cache`, that way it is more consistent with `_shared_metadata` too

duburcqa: > sensor.build() is inside of SensorManager.build() so it's a little tricky/mess

What do you mean ? I don't get it. SensorBase has its own `build` method, which is supposed to be called by the `SensorManager` after it built itself.

genesis/sensors/base_sensor.py
duburcqa: First, it should not be that much performance critical for a single clone. Second, if max delay is 0.0, then the length of the cache may be 1, so data would be overwritten at every simulation step, which would be an issue if not cloned right?

genesis/sensors/base_sensor.py
Milotrince: i think i would have to iterate through sensors twice (to first determine the cache shape, and then set the reference for each sensor). but this is at build time so I suppose this is not really an issue

duburcqa (approved):

Milotrince:
looks like unit test failed because of camera

duburcqa (approved):

PR #1521, opened 2025-08-08
[FEATURE] Add debug camera for monitoring.

Kashu7100:
## Description

* Enable to use PyRenderer camera to render along with Madrona Batched renderer
* Only add rasterizer camera if needed instead of systematically as before

## Motivation and Context

This will be useful for monitoring the training (logging eval video) in different resolution from what policy sees (e.g., HD camera for human monitoring and 64x64 for policy input) in a headless environment (like a server) where viewer is not available or when you don't want to use viewer to prevent slow down for the entire training.

```bash
python examples/rigid/single_franka_batch_render.py -r # batch_renderer only
python examples/rigid/single_franka_batch_render.py -r -d # batch_renderer + rasterizer for debug
```

## Checklist:

### DIFF
--- a/genesis/engine/scene.py
+++ b/genesis/engine/scene.py
@@ -529,6 +529,7 @@ class Scene(RBC):
         spp=256,
         denoise=None,
         env_idx=None,
+        debug=False,
     ):
         """
         Add a camera to the scene.
@@ -567,6 +568,12 @@ class Scene(RBC):
             Whether to denoise the camera's rendered image. Only available when using the RayTracer renderer. Defaults
             to True on Linux, otherwise False. If OptiX denoiser is not available in your platform, consider enabling
             the OIDN denoiser option when building the RayTracer.
+        debug : bool
+            Whether to use the debug camera. It enables to create cameras that can used to monitor / debug the
+            simulation without being part of the "sensors". Their output is rendered by the usual simple Rasterizer
+            systematically, no matter if BatchRender and RayTracer is enabled. This way, it is possible to record the
+            simulation with arbitrary resolution and camera pose, without interfering with what robots can perceive
+            from their environment. Defaults to False.
 
         Returns
         -------
@@ -576,7 +583,7 @@ class Scene(RBC):
         if denoise is None:
             denoise = sys.platform != "darwin"
         return self._visualizer.add_camera(
-            res, pos, lookat, up, model, fov, aperture, focus_dist, GUI, spp, denoise, env_idx
+            res, pos, lookat, up, model, fov, aperture, focus_dist, GUI, spp, denoise, env_idx, debug
         )
 
     @gs.assert_unbuilt
--- a/genesis/vis/batch_renderer.py
+++ b/genesis/vis/batch_renderer.py
@@ -85,19 +85,19 @@ class BatchRenderer(RBC):
         if gs.backend != gs.cuda:
             gs.raise_exception("BatchRenderer requires CUDA backend.")
 
-        cameras = self._visualizer._cameras
+        self._cameras = gs.List([camera for camera in self._visualizer._cameras if not camera.debug])
         lights = self._lights
         rigid = self._visualizer.scene.rigid_solver
         n_envs = max(self._visualizer.scene.n_envs, 1)
-        res = cameras[0].res
+        res = self._cameras[0].res
         gpu_id = gs.device.index
         use_rasterizer = self._renderer_options.use_rasterizer
 
         # Cameras
-        n_cameras = len(cameras)
-        cameras_pos = torch.stack([camera.get_pos() for camera in cameras], dim=1)
-        cameras_quat = torch.stack([camera.get_quat() for camera in cameras], dim=1)
-        cameras_fov = torch.tensor([camera.fov for camera in cameras], dtype=gs.tc_float, device=gs.device)
+        n_cameras = len(self._cameras)
+        cameras_pos = torch.stack([camera.get_pos() for camera in self._cameras], dim=1)
+        cameras_quat = torch.stack([camera.get_quat() for camera in self._cameras], dim=1)
+        cameras_fov = torch.tensor([camera.fov for camera in self._cameras], dtype=gs.tc_float, device=gs.device)
 
         # Build taichi arrays to store light properties once. If later we need to support dynamic lights, we should
         # consider storing light properties as taichi fields in Genesis.
@@ -179,8 +179,8 @@ class BatchRenderer(RBC):
         self.update_scene()
 
         # Render frame
-        cameras_pos = torch.stack([camera.get_pos() for camera in self._visualizer._cameras], dim=1)
-        cameras_quat = torch.stack([camera.get_quat() for camera in self._visualizer._cameras], dim=1)
+        cameras_pos = torch.stack([camera.get_pos() for camera in self._cameras], dim=1)
+        cameras_quat = torch.stack([camera.get_quat() for camera in self._cameras], dim=1)
         render_options = np.array((rgb_, depth_, False, False, aliasing), dtype=np.uint32)
         rgba_arr_all, depth_arr_all = self._renderer.render(
             self._visualizer.scene.rigid_solver, cameras_pos, cameras_quat, render_options
@@ -217,3 +217,7 @@ class BatchRenderer(RBC):
     @property
     def lights(self):
         return self._lights
+
+    @property
+    def cameras(self):
+        return self._cameras
--- a/genesis/vis/camera.py
+++ b/genesis/vis/camera.py
@@ -150,11 +150,10 @@ class Camera(Sensor):
 
         self._envs_offset = torch.as_tensor(self._visualizer._scene.envs_offset, dtype=gs.tc_float, device=gs.device)
 
-        self._batch_renderer = self._visualizer.batch_renderer
         self._rasterizer = self._visualizer.rasterizer
-        self._raytracer = self._visualizer.raytracer
+        self._raytracer = self._visualizer.raytracer if not self._debug else None
+        self._batch_renderer = self._visualizer.batch_renderer if not self._debug else None
 
-        self._rasterizer.add_camera(self)
         if self._batch_renderer is not None:
             self._rgb_stacked = True
             self._other_stacked = True
@@ -168,6 +167,7 @@ class Camera(Sensor):
                 self._rgb_stacked = False
                 self._other_stacked = False
             else:
+                self._rasterizer.add_camera(self)
                 self._rgb_stacked = self._visualizer._context.env_separate_rigid
                 self._other_stacked = self._visualizer._context.env_separate_rigid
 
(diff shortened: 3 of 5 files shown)


genesis/vis/camera.py
```diff
@@ -87,6 +92,7 @@ def __init__(
         far=100.0,
         transform=None,
         env_idx=None,
+        camera_type=CAMERA_TYPE.RASTERIZER,
```
duburcqa: What about defaulting to None, then replace None using the same logic as we are currently using in `render`, basically BatchRenderer if applicable, if not, RayTracer, if not Rasterizer ?

duburcqa: What about a simple boolean flag like `debug`, `force_rasterizer` or `is_sensor` ?

Do you think there is an actual use case for mixing multiple rendering backends except rasterizer + anything else ? If so, your approve is definitely the right one.

genesis/vis/visualizer.py
```diff
@@ -127,17 +127,35 @@ def destroy(self):
...
+            camera_type=camera_type,
         )
         self._cameras.append(camera)
         return camera
 
     def add_light(self, pos, dir, intensity, directional, castshadow, cutoff):
         if self._batch_renderer is not None:
             self._batch_renderer.add_light(pos, dir, intensity, directional, castshadow, cutoff)
+        else:
+            gs.logger.warning("`add_light` is specifically for batch renderer.")
```
duburcqa: Are you sure that this is a good idea? Does it pop a warning for every single simulation now or only if the user is calling `add_light` ? In the latter case, what about raising an exception instead?

genesis/vis/camera.py
Kashu7100: Maybe we want:
- rasterizer + batch renderer
- rasterizer + raytracer

duburcqa: If we don't want rasterizer + batch renderer + raytracer + ..., then I'm not sure we need something as generic as `camera_type`, that would be harder to maintain that a simple boolean flag for `rasterizer + something`.

Typically, if we want something as `camera_type`, when we should also support specifying options for multiple rendering backends at once.

Kashu7100: ok, sounds reasonable. I will adopt the `debug` flag!

genesis/engine/scene.py
```diff
@@ -566,14 +567,16 @@ def add_camera(
             Whether to denoise the camera's rendered image. Only available when using the RayTracer renderer. Defaults
             to True. If OptiX denoiser is not available in your platform, consider enabling the OIDN denoiser option
             when building the RayTracer.
+        debug : bool
+            Whether to use the debug camera (Rasterizer). Effective for BatchRenderer and RayTracer settings. Defaults to False.
```
duburcqa: Can you clarify the documentation? At this point I have no clue what it is supposed to do.

You should explain that it enables to create cameras that can used to monitor / debug the simulation without being part of the that are not officially considered as part of the "sensors". Their output is rendered by the usual simple `Rasterizer` systematically, no matter if BatchRender and RayTracer is enabled. This way, it is possible to record the simulation with arbitrary resolution and camera pose, without interfering with what robots can perceive from their environment.

genesis/vis/batch_renderer.py
```diff
@@ -85,19 +85,19 @@ def build(self):
         if gs.backend != gs.cuda:
             gs.raise_exception("BatchRenderer requires CUDA backend.")
 
-        cameras = self._visualizer._cameras
+        self._cameras = gs.List([camera for camera in self._visualizer._cameras if not camera.debug])
```
duburcqa, line 88: What about adding a `cameras` property to `visualizer`, that would filter out debug camera by default?

genesis/vis/camera.py
```diff
@@ -388,6 +395,7 @@ def render(
                     self, False, depth, segmentation, normal=normal
                 )
         else:
+            print("Camera is debug camera")
```
duburcqa: Remove this message, or using debug logging.

genesis/vis/camera.py
```diff
@@ -763,9 +771,18 @@ def is_built(self):
 
     @property
     def idx(self):
-        """The integer index of the camera."""
+        """The global integer index of the camera."""
         return self._idx
 
+    @property
+    @lru_cache(maxsize=1)
```
duburcqa: I would rather keep this inside `_batch_render` rather than make it an official property at camera level. This kind of circular dependency is not great.

Besides, I don't think caching the result is relevant for something as simple as finding an index in a list.

genesis/vis/camera.py
yuhongyi: Is the print message correct?
How about
```
if self._debug:
    print(...)
```

genesis/vis/visualizer.py
```diff
@@ -127,17 +127,33 @@ def destroy(self):
...
+            pos,
+            lookat,
+            up,
+            fov,
+            aperture,
+            focus_dist,
+            GUI,
+            spp,
+            denoise,
+            env_idx=0 if debug else env_idx,
```
duburcqa: Why are you actively ignoring env_idx. It seems weird.

genesis/vis/camera.py
Kashu7100: my bad

genesis/engine/scene.py
```diff
@@ -566,14 +567,20 @@ def add_camera(
             Whether to denoise the camera's rendered image. Only available when using the RayTracer renderer. Defaults
             to True. If OptiX denoiser is not available in your platform, consider enabling the OIDN denoiser option
             when building the RayTracer.
+        debug : bool
+            Whether to use the debug camera. It enables to create cameras that can used to monitor / debug the
+            simulation without being part of the "sensors". Their output is rendered by the usual simple Rasterizer
+            systematically, no matter if BatchRender and RayTracer is enabled.
+            This way, it is possible to record the simulation with arbitrary resolution and camera pose,
+            without interfering with what robots can perceive from their environment. Defaults to False.
```
duburcqa: Do not skip line unless it would exceed 120 linewidth otherwise.

genesis/vis/camera.py
```diff
@@ -66,6 +67,10 @@ class Camera(Sensor):
         The far plane of the camera.
     transform : np.ndarray, shape (4, 4), optional
         The transform matrix of the camera.
+    env_idx : int, optional
+        The index of the environment to track the camera.
+    camera_type : CAMERA_TYPE
+        The type of camera. Options are CAMERA_TYPE.RASTERIZER, CAMERA_TYPE.RAYTRACER, CAMERA_TYPE.BATCH_RENDERER.
```
duburcqa: Update documentation.

genesis/vis/visualizer.py
```diff
@@ -127,17 +128,36 @@ def destroy(self):
...
+            debug=debug,
         )
-        self._cameras.append(camera)
+        if debug:
+            self._debug_cameras.append(camera)
+        else:
+            self._cameras.append(camera)
```
duburcqa: I'm not a great fan of managing cameras and "debug cameras" separately. They are inherently the same things. Just the debug camera are not considered real sensors but only used to monitor the progress of the agent / debugging. Having `camera` getting hiding debug camera by default, and keeping them accessible via `_camera` private attribute sounds sufficient for me.

Kashu7100: There might be an indexing problem for batched renderer. It is using the index of the camera (which is the index in _camera) to get the camera specific image, and if there are debug camera in _camera, that will mess up the index.

genesis/vis/visualizer.py
Kashu7100: If I allow env_idx as input for debug camera, it will raise an error at `genesis/ext/pyrender/scene.py:L602` (`solid_nodes.sort(key=lambda n: -np.linalg.norm(self.get_pose(n)[:3, 3] - cam_loc))`). It seems like `cam_loc` (which is defined as `cam_loc = self.get_pose(self.main_camera_node)[:3, 3]`) is not properly sliced (maybe [...,:3,3]?) but I am not sure this should receive the batched tensor on the first hand (I think this is due to the batched renderer setup, without it, I don't get a batched tensor even if I have n_envs>1).
Also changing the slicing to `[...,:3,3]` will cause a new error:

```bash 
  File "/home/kashu/research/Genesis/examples/rigid/single_franka_batch_render.py", line 104, in <module>
    main()
  File "/home/kashu/research/Genesis/examples/rigid/single_franka_batch_render.py", line 92, in main
    debug_cam.render()
  File "/home/kashu/research/Genesis/genesis/utils/misc.py", line 146, in wrapper
    return method(self, *args, **kwargs)
  File "/home/kashu/research/Genesis/genesis/vis/camera.py", line 399, in render
    rgb_arr, depth_arr, seg_idxc_arr, normal_arr = self._rasterizer.render_camera(
  File "/home/kashu/research/Genesis/genesis/vis/rasterizer.py", line 73, in render_camera
    retval = self._renderer.render(
  File "/home/kashu/research/Genesis/genesis/ext/pyrender/offscreen.py", line 172, in render
    retval = renderer.render(scene, flags, seg_node_map)
  File "/home/kashu/research/Genesis/genesis/ext/pyrender/renderer.py", line 182, in render
    retval = self._forward_pass(scene, flags, seg_node_map=seg_node_map, env_idx=env_idx)
  File "/home/kashu/research/Genesis/genesis/ext/pyrender/renderer.py", line 422, in _forward_pass
    self.jit.forward_pass(
  File "/home/kashu/research/Genesis/genesis/ext/pyrender/jit_render.py", line 765, in forward_pass
    self._forward_pass(
  File "/home/kashu/miniconda3/envs/genesis/lib/python3.10/site-packages/numba/core/dispatcher.py", line 659, in _explain_matching_error
    raise TypeError(msg)
TypeError: No matching definition for argument type(s) array(int32, 1d, C), array(int32, 1d, C), array(float32, 3d, C), array(int32, 2d, C), array(float32, 2d, C), array(float32, 2d, C), array(int8, 2d, C), array(int32, 1d, C), array(int32, 1d, C), array(int32, 1d, C), array(float32, 2d, C), array(int32, 1d, C), array(float32, 4d, C), array(float32, 1d, C), array(float32, 3d, C), array(float32, 2d, C), array(float32, 2d, A), int64, array(float32, 2d, C), array(float32, 2d, C), int64, array(float32, 1d, C), int64, GLFunc
```

I am not sure this is worth investigating, and personally feel setting env_idx to 0 is sufficient for a debug camera.

Kashu7100: @yuhongyi Let me also know if you have any idea on this.

genesis/vis/visualizer.py
Kashu7100: After thinking twice, it should be fine.

genesis/vis/visualizer.py
duburcqa: Do you have a reproduction script?

Kashu7100: If you change `env_idx=0 if debug else env_idx` to `env_idx=env_idx` in `genesis/vis/visualizer.py`, and run `python examples/rigid/single_franka_batch_render.py -r -d` I think you will get this error.

duburcqa: > I am not sure this is worth investigating, and personally feel setting env_idx to 0 is sufficient for a debug camera.

This is not how it works. You cannot introduce "bugs" in the codebase and use "it is not important" as in excuse x)

duburcqa: fixed.

Issue #1516, opened 2025-08-06
[Bug]: draw_debug_line draws wrongly


yifan-hou:
### Bug Description

draw_debug_line(A,B) draws the line at wrong location. Also it ignores the color specification. Instead of drawing the line between A and B, it draws a line segment from B to the other direction, as shown below.
``` py
Tl = np.array([-0.12, -0.1975, 0.35  ] )
Tr = np.array([-0.12,  0.1975, 0.35  ] )
Nl = np.array([-0.12, -0.1975, 0.239 ] )
Nr = np.array([-0.12,  0.1975, 0.239 ] )

scene.draw_debug_sphere(Tl, radius=0.03, color=(1.0, 0.0, 0.0, 0.5))
scene.draw_debug_sphere(Tr, radius=0.03, color=(0.0, 1.0, 0.0, 0.5))
scene.draw_debug_sphere(Nl, radius=0.03, color=(1.0, 0.0, 0.0, 0.5))
scene.draw_debug_sphere(Nr, radius=0.03, color=(0.0, 1.0, 0.0, 0.5))

scene.draw_debug_line(Tl, Nl, radius=0.015, color=(0.0, 0.0, 1.0, 0.2))
scene.draw_debug_line(Tr, Nr, radius=0.015, color=(0.0, 1.0, 0.0, 0.2))
scene.draw_debug_line(Nl, Nr, radius=0.01, color=(1.0, 0.0, 0.0, 0.2))
scene.draw_debug_line(Tl, Tr, radius=0.01, color=(1.0, 0.0, 0.0, 0.2))
```

<img width="676" height="444" alt="Image" src="https://github.com/user-attachments/assets/1071a680-db4c-4e43-a94e-ccf6ea3d03a9" />

### Steps to Reproduce

Full script to reproduce:
```python
import numpy as np

import genesis as gs
import numpy as np

gs.init(backend=gs.cpu)

scene = gs.Scene(show_viewer=True,
        viewer_options = gs.options.ViewerOptions(
        res           = (1280, 960),
        camera_pos    = (0, -3.0, 2.5),
        camera_lookat = (0.0, 0.0, 0.5),
        camera_fov    = 40,
        max_FPS       = 60,
    ),
    vis_options = gs.options.VisOptions(
        show_world_frame = True, # visualize the coordinate frame of `world` at its origin
        world_frame_size = 1.0, # length of the world frame in meter
[...]

PR #1540, opened 2025-08-11
[BUG FIX] Correct -Z orientation  & vertex color

ceasor-mao:
## Description

1. `_np_z_up_to_R` now re-computes the orthogonal basis using the world X-axis as a fallback when the first cross-product is near-zero, guaranteeing a correct rotation matrix for all directions.
2. Pass `color` argument into `create_cylinder`

## Related Issue

Resolves Genesis-Embodied-AI/Genesis#1516

## Motivation and Context

1. `_np_z_up_to_R` in file utils/geom.py returned the identity matrix when the target vector was anti-parallel to the reference up axis, losing the required 180° rotation.
2. `create_line`in file utils/mesh.pygenerated a cylinder but never applied the requested color. 

## Checklist:

### DIFF
--- a/genesis/utils/geom.py
+++ b/genesis/utils/geom.py
@@ -1159,23 +1159,32 @@ def _np_z_up_to_R(z, up=None, out=None):
         if z_norm_i > gs.EPS:
             z /= z_norm_i
         else:
-            z[:] = 0.0, 1.0, 0.0
+            if up is None or abs(up[i][1]) < 0.5:
+                z[:] = 0.0, 1.0, 0.0
+            else:
+                z[:] = 0.0, 0.0, 1.0
 
         if up is not None:
-            a = up[i]
+            x[:] = np.cross(up[i], z)
         else:
-            a = np.array([0.0, 1.0, 0.0], dtype=z.dtype)
+            if abs(z[2]) < 1.0 - gs.EPS:
+                # up = (0.0, 0.0, 1.0)
+                x[0] = z[1]
+                x[1] = -z[0]
+                x[2] = 0.0
+            else:
+                # up = (0.0, 1.0, 0.0)
+                x[0] = z[2]
+                x[1] = 0.0
+                x[2] = -z[0]
 
-        x[:] = np.cross(a, z)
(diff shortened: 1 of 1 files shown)


ceasor-mao:
Should the same fallback-axis strategy be propagated to other rotation-matrix utilities in the codebase?

genesis/utils/geom.py
```diff
@@ -1162,18 +1162,20 @@ def _np_z_up_to_R(z, up=None, out=None):
...
         if up is not None:
-            x[:] = np.cross(up[i], z)
+            a = up[i]
         else:
-            x[0] = z[1]
-            x[1] = -z[0]
-            x[2] = 0.0
+            a = np.array([0.0, 1.0, 0.0], dtype=z.dtype)
 
+        x[:] = np.cross(a, z)
```
duburcqa: Remove this change. You just need to update the special case:
```python
        else:
            x[0] = z[2]
            x[1] = 0.0
            x[2] = -z[0]
```

genesis/utils/geom.py
```diff
@@ -1162,18 +1162,20 @@ def _np_z_up_to_R(z, up=None, out=None):
...
-        if x_norm > gs.EPS:
-            x /= x_norm
-            y[:] = np.cross(z, x)
-        else:
-            R[:] = np.eye(3, dtype=R.dtype)
+
+        if x_norm <= gs.EPS:
+            a_fallback = np.array([1.0, 0.0, 0.0], dtype=z.dtype)
+            x[:] = np.cross(a_fallback, z)
+            x_norm = np.linalg.norm(x)
```
duburcqa: You should rather adapt the special case to detect whether `x_norm < gs.EPS` before setting it. Basically, you just need to check `z[1] > math.sqrt(1.0 - gs.EPS ** 2)`, then use `axis = (1.0, 0.0, 0.0)` otherwise (virtually, without actually cross product).

duburcqa:
> Should the same fallback-axis strategy be propagated to other rotation-matrix utilities in the codebase?

Yes, this is necessary.

PR #1542, opened 2025-08-12
[FEATURE] Get hibernation compiling and working

gasnica:
## Description
Changes:
* use_contact_islands now compiles
* use_hibernation now compiles and runs
* hibernation and waking up is done per-island, instead of per-entity
  * on hibernation, a persistent island is saved; this is done by linking all entities into a circular-list via a new field in ContactIsland
* rigid_debug.py::Debug class is added to expose assertf() and validate flag (which enables additional validation code that slows sim down)
* validation functions are put into rigid_validate.py
* some rarely used functions got longer names by appending '__description_of_what_they_really_do'

Bug fix:
* this also fixes an out-of-bounds memory write, when island-building stack is overflown

Note:
* the hibernation code currently uses the original data_oriented classes, and is not migrated and ready to use ndarrays

Other things:
* num atomic_adds got reduced in some hibernation/wakeup code
* some variables were renamed to be consistent with code elsewhere -- we can revert that in this PR if you decide this adds confusion. Some such changes:
  * island -> i_island
  * ContactIsland.n_edge -> *.n_edges
  * ContactIsland.n_island -> *.n_islands
* a good bit of linting was added, some comments
* RigidSolver specifies _entities type explicitly, because it should override the more generic type of the base class. Not sure how to better do it. 

## Motivation and Context
Speedup simulation

Used runtime asserts & validation to verify all is working on a test scene of 100+ object, hibernating and unhibernating due to new collisions.

Screenshot 2025-08-11 at 8 55 05 PM

## Checklist:

### DIFF
--- a/genesis/engine/solvers/rigid/rigid_solver_decomp.py
+++ b/genesis/engine/solvers/rigid/rigid_solver_decomp.py
@@ -1274,7 +1274,6 @@ class RigidSolver(Solver):
         pass
 
     def substep_post_coupling(self, f):
-
         from genesis.engine.couplers import SAPCoupler
 
         if self.is_active() and isinstance(self.sim.coupler, SAPCoupler):
@@ -1293,6 +1292,7 @@ class RigidSolver(Solver):
                 collider_state=self.collider._collider_state,
                 rigid_global_info=self._rigid_global_info,
                 static_rigid_sim_config=self._static_rigid_sim_config,
+                contact_island=self.constraint_solver.contact_island,
             )
 
     def substep_post_coupling_grad(self, f):


genesis/engine/solvers/rigid/rigid_solver_decomp_util.py
```diff
@@ -0,0 +1,89 @@
...
+    links_state: array_class.LinksState,
+    geoms_state: array_class.GeomsState,
+    rigid_global_info: array_class.RigidGlobalInfo,
+    contact_island: ti.template(),
+):
+    # Note: Original function handled non-hibernated & fixed entities.
+    # Now, we require a properly hibernated entity to be passed in.
+    island_idx = contact_island.entity_island[i_e, i_b]
+    Debug.assertf(
+        0x7AD00008, not contact_island.island_hibernated[island_idx, i_b]
```
YilingQiao: what are those numbers?

genesis/engine/solvers/rigid/rigid_solver_decomp_util.py
```diff
@@ -0,0 +1,89 @@
...
+    )  # Island already expected to be marked as not hibernated
+
+    is_entity_fixed = island_idx == -1
+    Debug.assertf(0x7AD00007, not is_entity_fixed)  # Fixed entities are excluded from hibernation logic
```
YilingQiao: using a global variable like `Debug` might be tricky and dangerous for caching? @hughperkins

genesis/engine/solvers/rigid/rigid_validate.py
```diff
@@ -0,0 +1,65 @@
+import taichi as ti
+
+import genesis.utils.array_class as array_class
+
+from genesis.engine.solvers.rigid.rigid_debug import Debug
+from .contact_island import INVALID_NEXT_HIBERNATED_ENTITY_IDX
+
+
+@ti.func
+def validate_entity_hibernation_state_for_all_entities_in_temp_island(
```
YilingQiao: @duburcqa do you think we should introduce those Debug and assert in our kernel code?

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -67,6 +76,14 @@ def _sanitize_sol_params(
 
 @ti.data_oriented
 class RigidSolver(Solver):
+
+    # override typing
+    _entities: list[RigidEntity] = gs.List()
+
+    # created outside of __init__; added for linting
+    geoms_info: array_class.StructGeomsInfo
```
YilingQiao: I don't think it's necessary to add them. We have many other properties like links_state, links_info, etc. It stands out and looks messy.

genesis/engine/solvers/rigid/rigid_solver_decomp_util.py
gasnica: it's just an arbitrary, and preferably unique, id that identifies the assert --- we can't pass description strings in taichi, so when the asserts are enabled and the assert fails I'm printing it, and then can text-search for it in code

genesis/engine/solvers/rigid/rigid_solver_decomp_util.py
gasnica: here, the Debug class is just used to wrap the assert functionality, and assertf is a @classmethod. And I'm also referencing Debug.validate in code -- which is a class-level property, so effectively a global var

genesis/engine/solvers/rigid/rigid_solver_decomp.py
gasnica: Removing geoms_info & geoms_state; moving linting to the functions where they're initialized. 

Leaving _entities as is, since it's only initialized in the base class, but is assumed to have different type here.

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -26,30 +26,30 @@ class RigidLink(RBC):
...
+        edge_start: int,
+        verts_state_start: int,
+        vgeom_start: int,
+        vvert_start: int,
+        vface_start: int,
+        pos: ArrayLike,
+        quat: ArrayLike,
+        inertial_pos: ArrayLike,
+        inertial_quat: ArrayLike,
+        inertial_i: NDArray[np.float32] | None,  # may be None, eg. for plane; NDArray is 3x3 matrix
```
duburcqa: I don't think it has to be `np.float32`. Any dtype is supported at this point, even integers.

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -86,13 +86,13 @@ def __init__(
         if inertial_quat is not None:
             inertial_quat = np.asarray(inertial_quat, dtype=gs.np_float)
         self._inertial_quat: ArrayLike | None = inertial_quat
-        self._inertial_mass = inertial_mass
-        self._inertial_i = inertial_i
+        self._inertial_mass: float | None = inertial_mass
+        self._inertial_i: NDArray[np.float32] | None = inertial_i
```
duburcqa: Same, are you sure it only support floating point (np.float64 should probably be supported too though).

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -606,84 +606,84 @@ def quat(self) -> ArrayLike:
         return self._quat
 
     @property
-    def inertial_pos(self):
+    def inertial_pos(self) -> NDArray[np.float32]:
```
duburcqa: np.float32 ?

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -606,84 +606,84 @@ def quat(self) -> ArrayLike:
...
         """
         The initial position of the link's inertial frame.
         """
         return self._inertial_pos
 
     @property
-    def inertial_quat(self):
+    def inertial_quat(self) -> NDArray[np.float32]:
```
duburcqa: np.float32 ?

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -606,84 +606,84 @@ def quat(self) -> ArrayLike:
...
-    def inertial_mass(self):
+    def inertial_mass(self) -> float:
         """
         The initial mass of the link.
         """
         return self._inertial_mass
 
     @property
-    def inertial_i(self):
+    def inertial_i(self) -> NDArray[np.float32]:
```
duburcqa: np.float32 ?

genesis/engine/solvers/rigid/collider_decomp.py
```diff
@@ -525,7 +537,18 @@ def collider_kernel_clear(
                 ):
                     i_c_hibernated = collider_state.n_contacts_hibernated[i_b]
                     if i_c != i_c_hibernated:
-                        collider_state.contact_data[i_c_hibernated, i_b] = collider_state.contact_data[i_c, i_b]
+                        # note: copying all fields of class StructContactData:
```
duburcqa: No need for `note:` prefix. This is obviously a note. Plus start with upper case.

genesis/engine/solvers/rigid/rigid_debug.py
```diff
@@ -0,0 +1,24 @@
+import taichi as ti
+
+
+class Debug:
+
```
duburcqa: We do not skip line here.

Moreover, you should add some documentation. And `Debug` is not a proper class name. What about `TestCase` to align with standard `unittest` framework, or `DebugHelper`.

duburcqa:
```python
import taichi as ti


class Debug:
    debug = False

    id_registry = set()

    @classmethod
    def assertf(cls, id_: int, cond: Any, msg: Any) -> None:
        assert id_ not in cls.id_registry
        cls.id_registry.add(id_)

        @ti.func
        def _impl_assert():
            if ti.static(cls.debug):
                if not cond:
                    print(f"[{ti.u32(id_):#.8x}] ERROR:", msg)

        return _impl_assert


@ti.kernel
def caller(i: ti.i32, val: ti.template()):
    ti.static(Debug.assertf(0x00000001, i > 0, "i not positive"))()
    for j in range(i):
        ti.static(Debug.assertf(0x00000002, j < 5, "j >= 5"))()
    for I in ti.grouped(ti.ndrange(*val.shape)):
        ti.static(Debug.assertf(0xD0000003, (val[I] > 5).all(), f"val[I] <= 5: {val[I]}"))()


ti.init(debug=True)

Debug.debug = True

val = ti.field(ti.types.vector(3, ti.i32), shape=(2, 4))
val.fill(10)
val[0, 0][0] = 2
caller(10, val)
val[1, 0][0] = 3
caller(10, val)

```

genesis/engine/solvers/rigid/constraint_solver_decomp_island.py
```diff
@@ -1,26 +1,40 @@
...
 
+import genesis.utils.array_class as array_class
+
+from .rigid_solver_decomp_util import func_wakeup_entity_and_its_temp_island
+from .rigid_validate import (
+    validate_entity_hibernation_state_for_all_entities_in_temp_island,
+    validate_next_hibernated_entity_indices_in_entire_scene,
+)
 from .contact_island import ContactIsland
+from genesis.engine.solvers.rigid.rigid_debug import Debug
```
duburcqa: Move absolute imports before relative, except for `TYPE_CHECKING`

genesis/engine/solvers/rigid/constraint_solver_decomp_island.py
```diff
@@ -110,44 +126,46 @@ def _kernel_clear(self, envs_idx: ti.types.ndarray()):
...
-                    is_active = not self.contact_island.island_hibernated[island, i_b]
+                    is_active = not self.contact_island.island_hibernated[i_island, i_b]
                 if is_active:
-                    self.add_collision_constraints(island, i_b)
-                    self.add_joint_limit_constraints(island, i_b)
-                    self._func_init_solver(island, i_b)
-                    self._func_solve(island, i_b)
-                    self._func_update_qacc(island, i_b)
-                    self._func_update_contact_force(island, i_b)
+                    self.add_collision_constraints__and_wakeup_entities(i_island, i_b)
```
duburcqa: No double `__`

duburcqa: By the way, I'm not convinced we need to clarify in the name that it wakes up entities. This detail could be added in the documentation of the method.

genesis/engine/solvers/rigid/constraint_solver_decomp_island.py
```diff
@@ -110,44 +126,46 @@ def _kernel_clear(self, envs_idx: ti.types.ndarray()):
...
-            link_a = contact_data.link_a
-            link_b = contact_data.link_b
+            # get links indices of the contact_data
+            link_a = self._collider._collider_state.contact_data.link_a[i_col, i_b]
+            link_b = self._collider._collider_state.contact_data.link_b[i_col, i_b]
             link_a_maybe_batch = [link_a, i_b] if ti.static(self._solver._options.batch_links_info) else link_a
             link_b_maybe_batch = [link_b, i_b] if ti.static(self._solver._options.batch_links_info) else link_b
 
-            d1, d2 = gu.ti_orthogonals(contact_data.normal)
+            contact_data__normal = self._collider._collider_state.contact_data.normal[i_col, i_b]
```
duburcqa: We don't use double `__`. I would just go for `contact_normal`, there is no value in keeping `_data_` in the name.

genesis/engine/solvers/rigid/constraint_solver_decomp_island.py
```diff
@@ -110,44 +126,46 @@ def _kernel_clear(self, envs_idx: ti.types.ndarray()):
...
+            d1, d2 = gu.ti_orthogonals(contact_data__normal)
 
             invweight = self._solver.links_info.invweight[link_a_maybe_batch][0] + self._solver.links_info.invweight[
                 link_b_maybe_batch
             ][0] * (link_b > -1)
 
             for i in range(4):
                 d = (2 * (i % 2) - 1) * (d1 if i < 2 else d2)
-                n = d * contact_data.friction - contact_data.normal
+                contact_data__friction = self._collider._collider_state.contact_data.friction[i_col, i_b]
```
duburcqa: same.

genesis/engine/solvers/rigid/constraint_solver_decomp_island.py
```diff
@@ -178,7 +196,8 @@ def add_collision_constraints(self, island, i_b):
                             cdot_vel = self._solver.dofs_state.cdof_vel[i_d, i_b]
 
                             t_quat = gu.ti_identity_quat()
-                            t_pos = contact_data.pos - self._solver.links_state.COM[link, i_b]
+                            contact_data__pos = self._collider._collider_state.contact_data.pos[i_col, i_b]
```
duburcqa: Same. Remove `_data_`.

genesis/engine/solvers/rigid/constraint_solver_decomp_island.py
```diff
@@ -194,35 +213,63 @@ def add_collision_constraints(self, island, i_b):
                 if ti.static(self.sparse_solve):
                     self.jac_n_relevant_dofs[n_con, i_b] = con_n_relevant_dofs
 
+                contact_data__sol_params = self._collider._collider_state.contact_data.sol_params[i_col, i_b]
```
duburcqa: Remove `_data_`

genesis/engine/solvers/rigid/constraint_solver_decomp_island.py
```diff
@@ -194,35 +213,63 @@ def add_collision_constraints(self, island, i_b):
...
+                contact_data__penetration = self._collider._collider_state.contact_data.penetration[i_col, i_b]
```
duburcqa: Remove `_data_`

genesis/engine/solvers/rigid/constraint_solver_decomp_island.py
```diff
@@ -504,59 +551,63 @@ def _kernel_reset(self, envs_idx: ti.types.ndarray()):
...
 
-        for i_island_col in range(self.contact_island.island_col[island, i_b].n):
-            i_col_ = self.contact_island.island_col[island, i_b].start + i_island_col
+        for i_island_col in range(self.contact_island.island_col[i_island, i_b].n):
+            i_col_ = self.contact_island.island_col[i_island, i_b].start + i_island_col
             i_col = self.contact_island.constraint_id[i_col_, i_b]
 
-            contact_data = self._collider._collider_state.contact_data[i_col, i_b]
+            contact_data__normal = self._collider._collider_state.contact_data.normal[i_col, i_b]
+            contact_data__friction = self._collider._collider_state.contact_data.friction[i_col, i_b]
```
duburcqa: Remove `_data_`

genesis/engine/solvers/rigid/contact_island.py
```diff
@@ -1,65 +1,101 @@
...
 import genesis as gs
 import genesis.utils.geom as gu
+from genesis.engine.solvers.rigid.rigid_debug import Debug
+
+
+if TYPE_CHECKING:
+    from genesis.engine.solvers.rigid.collider_decomp import Collider
+    from genesis.engine.solvers.rigid.rigid_solver_decomp import RigidSolver
+
+INVALID_NEXT_HIBERNATED_ENTITY_IDX = -1
```
duburcqa, line 13: Do not mix global constant and imports.

genesis/engine/solvers/rigid/contact_island.py
```diff
@@ -1,65 +1,101 @@
...
+
+
+from genesis.engine.solvers.rigid.rigid_validate import validate_next_hibernated_entity_indices_in_entire_scene
```
duburcqa: Move this right after other `from genesis.[...]` imports.

genesis/engine/solvers/rigid/contact_island.py
```diff
@@ -1,65 +1,101 @@
...
         struct_agg_list = ti.types.struct(
             curr=gs.ti_int,
             n=gs.ti_int,
             start=gs.ti_int,
         )
 
-        self.ci_edges = ti.field(
-            dtype=gs.ti_int, shape=self.solver._batch_shape((self.collider._collider_info._max_contact_pairs[None], 2))
-        )
+        max_num_contact_pairs = self.collider._collider_info._max_contact_pairs[None]
```
duburcqa: `num_` is redundant with `max_`, remove it.

genesis/engine/solvers/rigid/contact_island.py
```diff
@@ -1,65 +1,101 @@
...
+        self.n_edges = ti.field(dtype=gs.ti_int, shape=self.solver._B)
+        self.n_islands = ti.field(dtype=gs.ti_int, shape=self.solver._B)
         self.n_stack = ti.field(dtype=gs.ti_int, shape=self.solver._B)
 
+        # description: entity_idx_to_island_idx
         self.entity_island = ti.field(dtype=gs.ti_int, shape=self.solver._batch_shape(self.solver.n_entities))
         self.stack = ti.field(dtype=gs.ti_int, shape=self.solver._batch_shape(self.solver.n_entities))
 
+        # Persistent hibernated islands:
+        #
```
duburcqa: Weird formatting.

genesis/engine/solvers/rigid/contact_island.py
```diff
@@ -1,65 +1,101 @@
...
 @ti.data_oriented
 class ContactIsland:
-    def __init__(self, collider):
-        self.solver = collider._solver
-        self.collider = collider
+    def __init__(self, collider: "Collider"):
+        self.solver: "RigidSolver" = collider._solver
+        self.collider: "Collider" = collider
+
+        # todo: mark fields as being tempoarary for island generation only?
```
duburcqa: "tempoarary"

Prefer `# TODO: ` or `# FIXME: `

genesis/engine/solvers/rigid/contact_island.py
```diff
@@ -1,65 +1,101 @@
...
+
+        # Used to make islands persist through hibernation:
+        self.entity_idx_to_next_entity_idx_in_hibernated_island = ti.field(
+            dtype=gs.ti_int, shape=self.solver._batch_shape(self.solver.n_entities)
+        )
+        self.entity_idx_to_next_entity_idx_in_hibernated_island.fill(INVALID_NEXT_HIBERNATED_ENTITY_IDX)
+
     @ti.kernel
-    def clear(self):
+    def clear__island_mapping(self):
```
duburcqa: No double `__`

genesis/engine/solvers/rigid/contact_island.py
```diff
@@ -69,57 +105,81 @@ def clear(self):
...
 
     @ti.func
     def add_edge(self, link_a, link_b, i_b):
         link_a_maybe_batch = [link_a, i_b] if ti.static(self.solver._options.batch_links_info) else link_a
         link_b_maybe_batch = [link_b, i_b] if ti.static(self.solver._options.batch_links_info) else link_b
 
-        ea = self.solver.links_info[link_a_maybe_batch].entity_idx
-        eb = self.solver.links_info[link_b_maybe_batch].entity_idx
+        ea = self.solver.links_info.entity_idx[link_a_maybe_batch]
+        eb = self.solver.links_info.entity_idx[link_b_maybe_batch]
```
duburcqa, line 101: Why to change this?

By the way, `ea` is not following our name convention. It should rather be `i_ea` or something.

genesis/engine/solvers/rigid/contact_island.py
```diff
@@ -69,57 +105,81 @@ def clear(self):
...
+        # fill in collider-info edges with indices to connected entities.
+        n_edge = self.n_edges[i_b]
         self.ci_edges[n_edge, 0, i_b] = ea
         self.ci_edges[n_edge, 1, i_b] = eb
-        self.n_edge[i_b] = n_edge + 1
+        self.n_edges[i_b] = n_edge + 1
 
     @ti.kernel
-    def add_island(self):
+    def add_island__all_contact_edges(self):
```
duburcqa: Same, no double `__`

genesis/engine/solvers/rigid/contact_island.py
```diff
@@ -69,57 +105,81 @@ def clear(self):
...
             for i_col in range(self.collider._collider_state.n_contacts[i_b]):
-                impact = self.collider._collider_state.contact_data[i_col, i_b]
-                self.add_edge(impact.link_a, impact.link_b, i_b)
+                # get links indices of the impact
+                link_a = self.collider._collider_state.contact_data.link_a[i_col, i_b]
+                link_b = self.collider._collider_state.contact_data.link_b[i_col, i_b]
+                self.add_edge(link_a, link_b, i_b)
+
+    @ti.kernel
+    def add_island__all_hibernated_island_edges(self):
```
duburcqa: No double `__`

genesis/engine/solvers/rigid/contact_island.py
```diff
@@ -162,61 +222,77 @@ def postprocess_island(self):
...
                 if self.entity_edge[i_v, i_b].n > 0 and self.solver.entities_info.n_dofs[i_v] > 0:
                     if self.entity_island[i_v, i_b] != -1:
                         continue
                     self.n_stack[i_b] = 0
                     self.stack[self.n_stack[i_b], i_b] = i_v
                     self.n_stack[i_b] = self.n_stack[i_b] + 1
+                    self.entity_island[i_v, i_b] = self.n_islands[i_b]
+                    if ti.static(Debug.validate) and self.n_stack[i_b] > self.stack.shape[0]:
+                        capacity = self.stack.shape[0]
+                        print(f"Stack overflow! capacity and size: {capacity} < {self.n_stack[i_b]}")
```
duburcqa: No print in kernels. It dramatically impedes simulation speed unfortunately. Keep it commented for now, and it could be added to our debug facility later.

genesis/engine/solvers/rigid/contact_island.py
```diff
@@ -162,61 +222,77 @@ def postprocess_island(self):
...
                         v = self.stack[self.n_stack[i_b], i_b]
-                        if self.entity_island[v, i_b] != -1:
-                            continue
-                        self.entity_island[v, i_b] = self.n_island[i_b]
 
                         for i_edge in range(self.entity_edge[v, i_b].n):
-                            _id = self.entity_edge[v, i_b].start + i_edge
-                            edge = self.edge_id[_id, i_b]
-                            next_v = self.ci_edges[edge, 0, i_b]
+                            _id = self.entity_edge[v, i_b].start + i_edge  # half-edge index
```
duburcqa, line 249: This si not following our convention.

Use `i_edge_` (with `_` suffix) for local indices, then the same name without suffix for "global" indices.

genesis/engine/solvers/rigid/contact_island.py
```diff
@@ -162,61 +222,77 @@ def postprocess_island(self):
...
+                            edge = self.edge_id[_id, i_b]  # edge index
```
duburcqa, line 250: `i_edge`

genesis/engine/solvers/rigid/contact_island.py
```diff
@@ -162,61 +222,77 @@ def postprocess_island(self):
...
+                            next_v = self.ci_edges[edge, 0, i_b]  # other entity index, connected by edge
                             if next_v == v:
                                 next_v = self.ci_edges[edge, 1, i_b]
 
-                            if self.solver.entities_info.n_dofs[next_v] > 0 and next_v != v:
+                            if (
+                                self.solver.entities_info.n_dofs[next_v] > 0
+                                and next_v != v
+                                and self.entity_island[next_v, i_b] == -1
+                            ):  # 2nd condition must not happen ?
```
duburcqa, line 259: What is this comment?

genesis/engine/solvers/rigid/contact_island.py
```diff
@@ -162,61 +222,77 @@ def postprocess_island(self):
...
                                 self.stack[self.n_stack[i_b], i_b] = next_v
                                 self.n_stack[i_b] = self.n_stack[i_b] + 1
+                                self.entity_island[next_v, i_b] = self.n_islands[i_b]
+                                if ti.static(Debug.validate) and self.n_stack[i_b] > self.stack.shape[0]:
+                                    capacity = self.stack.shape[0]
+                                    print(f"Stack overflow! capacity and size: {capacity} < {self.n_stack[i_b]}")
```
duburcqa: Same. No print for now unfortunately.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -16,16 +16,25 @@
...
+    validate_entity_hibernation_state_for_all_entities_in_temp_island,
+    validate_next_hibernated_entity_indices_in_entire_scene,
+)
 
 from ..base_solver import Solver
 from .collider_decomp import Collider
 from .constraint_solver_decomp import ConstraintSolver
 from .constraint_solver_decomp_island import ConstraintSolverIsland
+from .rigid_solver_decomp_util import func_wakeup_entity_and_its_temp_island
 from ....utils.sdf_decomp import SDF
```
duburcqa: Move depth import levels on top (`....`, then `..`, then `.`)

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -16,16 +16,25 @@
...
+from .contact_island import INVALID_NEXT_HIBERNATED_ENTITY_IDX
 
 if TYPE_CHECKING:
     from genesis.engine.scene import Scene
     from genesis.engine.simulator import Simulator
+    import genesis.engine.solvers.rigid.array_class
```
duburcqa: Move `import` before `from`

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -67,6 +76,10 @@ def _sanitize_sol_params(
 
 @ti.data_oriented
 class RigidSolver(Solver):
+
```
duburcqa: Do not skip line here.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -249,6 +262,7 @@ def build(self):
 
         self.n_equalities_candidate = max(1, self.n_equalities + self._options.max_dynamic_constraints)
 
+        # invalid inertitance of AvatarSolver <- doesn't support hibernation params
```
duburcqa: Start comments with uppercase and (ideally) make valid English sentences.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -4871,59 +4913,101 @@ def kernel_update_vgeoms(
...
-                    ][0]
-                    collider_state.sort_buffer.value[geoms_state.max_buffer_idx[i_g, i_b], i_b] = geoms_state.aabb_max[
-                        i_g, i_b
-                    ][0]
+
+            if not was_island_hibernated:
+                are_all_entities_okay_for_hibernation = True
+                entity_ref_range = ci.island_entity[island_idx, i_b]
+                for i in range(entity_ref_range.n):
+                    entity_ref = entity_ref_range.start + i
```
duburcqa: Do not use plain `i` (add some suffix).

Use additional `_` suffix to local indices, with the same name without suffix for their global counterparts.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -4871,59 +4913,101 @@ def kernel_update_vgeoms(
...
+                    entity_idx = ci.entity_id[entity_ref, i_b]
+
+                    is_entity_fixed = entities_info.n_dofs[entity_idx] == 0
```
duburcqa: There is no "better" way to check for fixed entities? Like using some boolean flag?

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -4871,59 +4913,101 @@ def kernel_update_vgeoms(
...
+                        entity_ref = entity_ref_range.start + i
+                        entity_idx = ci.entity_id[entity_ref, i_b]
+                        for i_g in range(entities_info.geom_start[entity_idx], entities_info.geom_end[entity_idx]):
+                            min_idx, min_val = geoms_state.min_buffer_idx[i_g, i_b], geoms_state.aabb_min[i_g, i_b][0]
+                            max_idx, max_val = geoms_state.max_buffer_idx[i_g, i_b], geoms_state.aabb_max[i_g, i_b][0]
+                            collider_state.sort_buffer.value[min_idx, i_b] = min_val
+                            collider_state.sort_buffer.value[max_idx, i_b] = max_val
+                else:
+                    # perform hibernation
+                    prev_entity_ref = entity_ref_range.start + entity_ref_range.n - 1
```
duburcqa, line 5027: Same, start with `i_`. And usually I tend to prefer `prev` (or `next`, `delta`...etc) at the end rather than in front, to keep words sorted by "importance".

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -4936,28 +5020,37 @@ def func_aggregate_awake_entities(
...
+    i_b: int,
+    entities_state: array_class.EntitiesState,
+    entities_info: array_class.EntitiesInfo,
+    dofs_state: array_class.DofsState,
+    links_state: array_class.LinksState,
+    geoms_state: array_class.GeomsState,
+) -> None:
+    """
+    Mark RigidEnity, individual DOFs in DofsState, RigidLinks, and RigidGeoms as hibernated.
+    Zero out DOF velocitities and accelerations.
```
duburcqa: Skip line between short description and details.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -4936,28 +5020,37 @@ def func_aggregate_awake_entities(
...
+    """
```
duburcqa, line 5095: Do not skip line after documentation.

genesis/engine/entities/rigid_entity/rigid_link.py
gasnica: changing to ArrayLike | None & resolving

genesis/engine/entities/rigid_entity/rigid_link.py
gasnica: Are you talking about _inertial_mass also ?

gasnica: Changing to NDArray[gs.np_float]

genesis/engine/entities/rigid_entity/rigid_link.py
gasnica: so those are initialized in _add_by_info(), which is called when creating terrain, primitives, and loading mujoco files. 

Loading in terrain, primitives, is clear. When loading mujoco files this probably can be anything indeed -> should we modify RigidEnitty._load_scene() to standardize & convert value types to NDArray[gs.np_float] ?

genesis/engine/entities/rigid_entity/rigid_link.py
gasnica: yeah, it can also be None, or anything loaded from mujoco..

Should we remove all typing. 

Or modify RigidEntity._load_scene() to enforce the type. And use zero arrays instead of None in other places where RigidEntity._add_by_info() is called ??

genesis/engine/solvers/rigid/rigid_solver_decomp.py
gasnica: Could I kindly ask you, Alexis, to suggest actual changes you'd be happy with -- here, and in general? That might help us to reduce ping-ping in messages. @duburcqa 

Here, what's your suggestion?
- I'm seeing we're using i in a few places, when we iterate from zero to n
- here the i is index-to-i'th-element-in-entity-ref-range, or offset-from-entity-ref-range-start

What's the thought behind not using just plain i ?

genesis/engine/solvers/rigid/rigid_solver_decomp.py
gasnica: I don't know, this is what we've been doing in the code I"m working on here. Mathematically it seems like a solid check XD

genesis/engine/solvers/rigid/rigid_solver_decomp.py
gasnica: Just browsing through the code for 'prev':
- I'm getting the impression that leading prev_ is more common
- it also is natural to read, so I'd advocate for it

Here, I'm using _ref suffix to differentiate the index into the intermediate mapping array, from _idx which is actually the index of the entity in the simulation's arrays. 

I thought it may help with readability to use separate conceptual names, rather than just i_**

--
But yeah, this is just my attempt to add my 2 cents, let's talk maybe through it

genesis/engine/solvers/rigid/rigid_debug.py
gasnica: Will change. 

But I'm just browsing through our codebase -- I'm seeing classes that have multi-line docline without blank line after the initial line -- which to me was also more intuitive if the 2nd line, is not a longer description, but actually completes the description in the first line ... hmph.

And I'm seeing spaces left after docstring, and before first element.

Here, I left the space to emphasize all elements are class-level, not instance-level. 

----

I'd like the name to be short, so dared for Debug -- and here it's important cos it's used often. I was considering SimDebug. But then maybe we'd be wanting to actually move it up outside of rigid simulation only, where I currently placed the file.

gasnica: Debug.Assert is also rather idiomatic

genesis/engine/solvers/rigid/constraint_solver_decomp_island.py
gasnica: Removing double __

Functions' name should describe what functions do. I'm not considering waking up entities a detail. I don't think that info should be hidden.

genesis/engine/solvers/rigid/contact_island.py
gasnica: okok -- are we okay with absolute path imports for same-directory modules ?? 

That's what those are. 

I advocate for full-paths, but I remember you like relative in this case... i think

genesis/engine/solvers/rigid/contact_island.py
gasnica: Is there a good way to write a comment that could visually separate a logical block of code ?? 

Anyhow -- just removing this now.

genesis/engine/solvers/rigid/contact_island.py
gasnica: removing the comment.

genesis/engine/solvers/rigid/contact_island.py
gasnica: This was added, to preserve the original name -- which might be expected when text-searching for the function -- and extend it with extra description.

Should i rename it to add_island_edges_from_all_contacts() ?

genesis/engine/solvers/rigid/contact_island.py
gasnica: Ahh, changing to Debug.assertf

gasnica: Ahh -- so this is conditioned by if ti.static(Debug.validate) -- which fixes the issue. That is, that will be the case as soon as i put the ti.static on a separate line from other conditions, like we spoke today :-)

And then I'm leaving it as is, or would you like it changed to Debug.assertf ?

genesis/engine/solvers/rigid/contact_island.py
gasnica: This is Yiling's code, I'm only adding comments without modifying. 

Will not change.

i_edge is already used and has a different meaning.

genesis/engine/solvers/rigid/contact_island.py
gasnica: Not mine. Stack-operation related comment, probably from Yiling. 

But actually -- that's the code that was causing the out-of-bound writes.

genesis/engine/solvers/rigid/contact_island.py
gasnica: Guarded by if ti.static(Debug.validate)

(will move the non-static condition to next line)

genesis/engine/solvers/rigid/rigid_solver_decomp.py
gasnica: shouldn't imports be alphabetical, according to full path ?? in this case also ??

genesis/engine/solvers/rigid/contact_island.py
gasnica: Oh -- also -- cannot do it. At least not in a straight-forward way.

rigid_validate needs the INVALID_NEXT_HIBERNATED_ENTITY_IDX constant defined right above, and must be imported after it, otherwise it does not compile.

genesis/engine/solvers/rigid/rigid_solver_decomp_util.py
duburcqa: > using a global variable like Debug might be tricky and dangerous for caching?

Still, I think @YilingQiao is getting a point. This approve won't be supported anymore I think. But I should be ok for the new approach I suggest, which evaluates function pointers at compile time?

genesis/engine/solvers/rigid/rigid_validate.py
duburcqa: > do you think we should introduce those Debug and assert in our kernel code?

I'm not against it if it works transparently and easy to maintain support for gstaichi dev team (basically Hugh for now).

genesis/engine/entities/rigid_entity/rigid_link.py
duburcqa: > should we modify RigidEnitty._load_scene() to standardize & convert value types to NDArray[gs.np_float] ?

At some point it would be nice, but may be not a job for today.

genesis/engine/entities/rigid_entity/rigid_link.py
duburcqa: yes, to some extent. Basically we should sanitize all user inputs.

genesis/engine/entities/rigid_entity/rigid_link.py
duburcqa: > Should we remove all typing.

Not necessarily, but it should be valid if typed. Maybe we should keep this for later.

genesis/engine/solvers/rigid/rigid_solver_decomp_util.py
hughperkins: re: "printing": remidner that the *mere presence* of a print in the code can double the runtime latency. We saw this for the func narrow phase kernels: I commented out some conditional prints, that weren't even being executed, and halved the speed.

hughperkins: *halved the latency

hughperkins: re: Debug.validate -> Debug.validate is a global variable, and therefore violates the constraint of making all functions pure.

If you want to make flags, you need to pass them into the kernel, as parameters, in python dataclasses, and then once we migrate to gstaichi, we can tag them with the appropriate metadata, see https://github.com/Genesis-Embodied-AI/gstaichi/blob/4063b7882ca732679a821155612167f9903d2018/tests/python/gstaichi/lang/fast_caching/test_args_hasher.py#L232-L233

genesis/engine/solvers/rigid/rigid_debug.py
duburcqa: > But I'm just browsing through our codebase -- I'm seeing classes that have multi-line docline without blank line after the initial line -- which to me was also more intuitive if the 2nd line, is not a longer description, but actually completes the description in the first line ... hmph. [...] 

Yes, the doc is ill-formatted at many places. But I do not want to format the whole thing at once and create a gigantic diff that would mess up with rebase for limited benefit.

This is the standard way of writing Python documentation. See [numpy](https://github.com/numpy/numpy/blob/7683d859f6e81735d66289bd1a8a3d3b84186d67/numpy/_core/function_base.py#L486) as a reference. Short description has to be short. One sentence. No option. If more details is needed, or as you are saying, a "longer description", you need to skip one line. This short documentation will pop automatically in many editor when your mouse is overing them, but not the full description because it would be a mess.

> And I'm seeing spaces left after docstring, and before first element.

It depends on the guidelines. It is not something you can choose as you please but it must be unified over the whole codebase. It turns out you are right, we are skipping line if there is no doctring at the moment, so nothing to change regarding this.

duburcqa: > I'd like the name to be short, so dared for Debug -- and here it's important cos it's used often. I was considering SimDebug. But then maybe we'd be wanting to actually move it up outside of rigid simulation only, where I currently placed the file.

Make sense. I'm ok with it.

genesis/engine/solvers/rigid/rigid_solver_decomp_util.py
hughperkins: re: Debug.assertf => this doesnt seem to be supported by anything allowable in pure functions.

You should probalby remove the pointer, and have a `@ti.static` if, that is conditional on one of the values coming in through the kernel parameters, as per my immediately previous comment, above.

genesis/engine/solvers/rigid/constraint_solver_decomp_island.py
duburcqa: > Functions' name should describe what functions do.

No I don't think they should. Simply because it is impossible. You cannot compress losslessly 50 lines of code in 5 words. But fine if you consider that waking up entities is not a detail. Still, we will not compromise formatting with exception rules either because of long function names.

genesis/engine/solvers/rigid/contact_island.py
duburcqa: > okok -- are we okay with absolute path imports for same-directory modules ??

No I'm not. Relative import when it comes to our own codebase. Exception for stuff guarded by `TYPE_CHECKING`.

genesis/engine/solvers/rigid/rigid_validate.py
hughperkins: @YilingQiao can you define what you mean by "our kernel code" please. I interpreted this to mean "our Genesis GPU kernels", but alexis seems to be interpreting it to mean "support natively within gstaichi".

genesis/engine/solvers/rigid/contact_island.py
duburcqa: > rigid_validate needs the INVALID_NEXT_HIBERNATED_ENTITY_IDX constant defined right above, and must be imported after it, otherwise it does not compile.

Then you need to change your code, relying on such import pattern is not ok.

genesis/engine/solvers/rigid/contact_island.py
duburcqa: What about:
```
# Persistent hibernated islands
# -----------------------------
```

It is more comment in Python.

genesis/engine/solvers/rigid/rigid_validate.py
YilingQiao: I mean `our Genesis GPU kernels`.

hughperkins:
(untagging myself from this PR, so please re-tag me if you respond to any of my comments)

genesis/engine/solvers/rigid/contact_island.py
duburcqa: > This was added, to preserve the original name -- which might be expected when text-searching for the function -- and extend it with extra description.

We don't try to preserve original names. If they need to be updated, just update them. It is ok.

> Should i rename it to add_island_edges_from_all_contacts() ?

What makes more sense to you. Although I don't think `_all_` brings any value here.

genesis/engine/solvers/rigid/contact_island.py
duburcqa: It would be great to have `Debug.print`. But what you are doing here seems a good fit for your `assertf`.

genesis/engine/solvers/rigid/contact_island.py
duburcqa: I see. We will have to refactor this at some point.

genesis/engine/solvers/rigid/contact_island.py
duburcqa: It would be great to add some @XXX if front of such comments in the future.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
duburcqa: > shouldn't imports be alphabetical, according to full path ?? in this case also ??

No. `import` and `from` as sorted by alphabetical order separately.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
duburcqa: > Could I kindly ask you, Alexis, to suggest actual changes you'd be happy with -- here, and in general

Not really. I'm just pointing out what are the rules for new additions to the codebase, but then you can do whatever you want as long as it satisfies them. I have nothing in particular I would be happy with. But Here is what you could do:
```python
for i_eref_ in range(entity_ref_range.n):
    i_eref = entity_ref_range.start + i_eref_
    [...]
```

> I'm seeing we're using i in a few places, when we iterate from zero to n. What's the thought behind not using just plain i ?

This is usually legacy code. We are progressively improving the code quality. We have higher standards for new code compared to the existing codebase. Except in places where the index does not carry any special meaning, you should be more specific. One example where it is valid:
```python
    if ti.static(constraint_type == 0):  # geometries
        ti.loop_config(serialize=static_rigid_sim_config.para_level < gs.PARA_LEVEL.PARTIAL)
        for i_g_ in range(inputs_idx.shape[0]):
            for i in ti.static(range(7)):
                geoms_info.sol_params[inputs_idx[i_g_]][i] = sol_params[i_g_, i]
    elif ti.static(constraint_type == 1):  # joints
```

genesis/engine/solvers/rigid/rigid_solver_decomp.py
duburcqa: Ok let's keep this.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
duburcqa: > Just browsing through the code

Browsing the code is not going to give you any valuable information unfortunately, since most of it is not following any guidelines.

> it also is natural to read

I don't agree. Personally I want to know what I'm dealing with before knowing it is prev. At this point I don't care if it is prev or not if I still don't know what it is all about.

> Here, I'm using _ref suffix to differentiate the index into the intermediate mapping array, from _idx which is actually the index of the entity in the simulation's arrays. I thought it may help with readability to use separate conceptual names, rather than just i_**

Let's do this. These are just general guidelines, which are not properly standardised at this point. You have some good point. You can keep what you are most comfortable with, but we should strive for unified naming conventions everything. It would be nice if you could help writing them down at some point. Here is the PR I mentioned: https://github.com/Genesis-Embodied-AI/Genesis/pull/1053

genesis/engine/solvers/rigid/rigid_solver_decomp_util.py
duburcqa: @hughperkins What do you think about https://github.com/Genesis-Embodied-AI/Genesis/pull/1542#issuecomment-3181823018 ?

genesis/engine/entities/rigid_entity/rigid_link.py
gasnica: I already converted the 4 _intertial_pos/quat/mass/i variables and params to ArrayLike | None, or float | None.

Thanks & resolving this.

genesis/engine/entities/rigid_entity/rigid_link.py
gasnica: Changed to ArrayLike | None. Resolving.

genesis/engine/entities/rigid_entity/rigid_link.py
gasnica: It's now ArrayLike | None. Resolving .

genesis/engine/entities/rigid_entity/rigid_link.py
gasnica: It's now ArrayLike | None. Resolving

genesis/engine/solvers/rigid/collider_decomp.py
gasnica: Done. Resolving.

genesis/engine/solvers/rigid/rigid_debug.py
gasnica: Thanks for all the feedback & info, Alexis. Doing nothing here for now. Resolving.

genesis/engine/solvers/rigid/constraint_solver_decomp_island.py
gasnica: Done. Resolving.

genesis/engine/solvers/rigid/constraint_solver_decomp_island.py
gasnica: Done. Resolving.

genesis/engine/solvers/rigid/constraint_solver_decomp_island.py
gasnica: Done. Resolving.

genesis/engine/solvers/rigid/constraint_solver_decomp_island.py
gasnica: Done. Resolving.

genesis/engine/solvers/rigid/constraint_solver_decomp_island.py
gasnica: Done. Resolving.

genesis/engine/solvers/rigid/contact_island.py
gasnica: Done. Resolving.

genesis/engine/solvers/rigid/contact_island.py
gasnica: Looks great, thanks for this. Will use next time.

gasnica: Removed. Resolving.

genesis/engine/solvers/rigid/contact_island.py
gasnica: resolving

genesis/engine/solvers/rigid/contact_island.py
gasnica: This is ndarray access, related to migration. Index now must go at the end, right?

I would ea, eb is not my code. Will not change.

genesis/engine/solvers/rigid/contact_island.py
gasnica: Reverted the name change. Resolving.

genesis/engine/solvers/rigid/contact_island.py
gasnica: renamed to add_contact_edges_to_islands(). Resolving.

genesis/engine/solvers/rigid/contact_island.py
gasnica: Renamed to add_hiberanted_edges_to_islands(). Resolving.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
gasnica: Done. But now we decided a line is okay when there's no docstring. Anyways--resolving.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
gasnica: Big comment added. I tried my best with English.

gasnica: Resolving.

genesis/engine/entities/rigid_entity/rigid_link.py
gasnica: resolving.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
gasnica: Done, and reordered all imports alphabetically.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
gasnica: Done resolving.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
gasnica: Changing i to i_entity_ref_offset_. But maybe let's talk through this.

- it think it might reduce readability by adding an unneeded complicated token, used in only 2 consecutive lines
- and I am strongly against things like i_eref --- inventing acronyms makes code unreadable
- also what's the full rule for applying the trailing underscore?

genesis/engine/solvers/rigid/rigid_solver_decomp.py
gasnica: Oh -- by natural, i meant, it's not how you would say things in a natural language. But I get your point too, its nice to have consistent name prefixes in complicated code.

What is the policy for using i_** --- I now assumed we use it as interators/count in loops. So that would not apply here?

Anyhow -- leaving things as they are, as you're suggesting at the end.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
gasnica: Done. Resolving.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
gasnica: Done. Resolving.

genesis/engine/solvers/rigid/rigid_solver_decomp_util.py
gasnica: Guys, I'm removing rigid_debug.py and reigid_validate.py files for now. And also removing all coding code. 

I was considering leaving calling code guarded by ti.static(False), so that is later easier to restore the code ...

But not sure if that would work for us -- removing all now for the sake of time.

hughperkins: I'm not a fan of having code in main that is not being actively used. I'm pretty strongly against it. Can you store the code in your own branch somewhere, or similar, please?

hughperkins: (having used various codebases, some with junk "just in case we might use it", and a few with anything not being used removed promptly, I *vastly* prefer the latter)

genesis/engine/solvers/rigid/constraint_solver_decomp_island.py
gasnica: Oh, what i mean -- name should describe what a function does, not how. 

But just googling pep8 here, and here are couple of hints there:
- Function names should clearly describe the action or purpose the function performs.
- Choose names that are unambiguous and avoid overly short or abbreviated names that might obscure the function's meaning.
- The name should convey what the function does, not how it does it.

Regarding length -- here, the name appears twice in code, so imho can be any length. And I prefer names that are descriptive to docstring, cos docstrings are prone to be not maintained.

genesis/engine/solvers/rigid/contact_island.py
gasnica: debug/validate removed, this is not an issue now. resolving.

genesis/engine/solvers/rigid/contact_island.py
gasnica: something to figure when we put debug/validate back in.

genesis/engine/solvers/rigid/contact_island.py
gasnica: Indeed, this should just be an assert.
Removed from this PR. Resolving.

genesis/engine/solvers/rigid/contact_island.py
gasnica: leaving as-is. resolving.

genesis/engine/solvers/rigid/contact_island.py
gasnica: Gotcha. 

This is not my code. Resolving.

genesis/engine/solvers/rigid/contact_island.py
gasnica: removed altogether from this PR. resolving.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
gasnica: Resolving.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
gasnica: resolving.

genesis/engine/solvers/rigid/contact_island.py
gasnica: Resolving.

genesis/engine/solvers/rigid/contact_island.py
gasnica: Resolving.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
gasnica: resolving.

genesis/engine/solvers/rigid/constraint_solver_decomp_island.py
gasnica: resolving.

genesis/engine/solvers/rigid/constraint_solver_decomp_island.py
gasnica: Debug removed from this PR. Resolving.

genesis/engine/solvers/rigid/rigid_solver_decomp_util.py
duburcqa: > I'm not a fan of having code in main that is not being actively used. I'm pretty strongly against it. Can you store the code in your own branch somewhere, or similar, please?

The idea would be to enable that on the CI systematically.

genesis/engine/solvers/rigid/contact_island.py
```diff
@@ -1,63 +1,89 @@
...
-            dtype=gs.ti_int, shape=self.solver._batch_shape((self.collider._collider_info._max_contact_pairs[None]))
-        )
+        # maps collider_state.contact_data index to island idx
+        self.constraint_list = ti.field(dtype=gs.ti_int, shape=self.solver._batch_shape((max_contact_pairs)))
 
+        # analogous to edge_id: maps island's constraint local-index to world's contact index
         self.constraint_id = ti.field(
             dtype=gs.ti_int,
-            shape=self.solver._batch_shape((self.collider._collider_info._max_contact_pairs[None] * 2)),
+            shape=self.solver._batch_shape((max_contact_pairs * 2)),
```
duburcqa: Keep everything in one line if it fits.

genesis/engine/solvers/rigid/rigid_solver_decomp_util.py
hughperkins: Yes, so, to be fair, this isn't my codebase, so I shouldn't really be expressing an opinion on this point concerning whether to keep the code in the codebase for now or not. I should be strictly keeping to the remit of providing consultation on the extent to which the proposed changes are compatible with future versions of GsTaichi. So yeah, @gasnica you can ignore my opinion as far as keeping the code in the Genesis codebase or not (and it sounds like Alexis is more favorable than I am to your keeping the code here for the time being :) )

genesis/engine/solvers/rigid/contact_island.py
```diff
@@ -69,57 +95,76 @@ def clear(self):
...
-        self.postprocess_island()
+        self.add_contact_edges_to_islands()
+        self.add_hiberanted_edges_to_islands()
+        self.preprocess_island_and_map_entities_to_edges()
+        self.construct_islands()
+        self.postprocess_island_and_assign_contact_data_to_islands()
 
     @ti.kernel
-    def postprocess_island(self):
+    def postprocess_island_and_assign_contact_data_to_islands(self):
```
duburcqa: `postprocess_islands_and_assign_contacts` ?

genesis/engine/solvers/rigid/contact_island.py
```diff
@@ -162,61 +207,79 @@ def postprocess_island(self):
...
             for i_v in range(self.solver.n_entities):
+                # only create islands for entities with collisions and with dofs
                 if self.entity_edge[i_v, i_b].n > 0 and self.solver.entities_info.n_dofs[i_v] > 0:
                     if self.entity_island[i_v, i_b] != -1:
                         continue
                     self.n_stack[i_b] = 0
                     self.stack[self.n_stack[i_b], i_b] = i_v
                     self.n_stack[i_b] = self.n_stack[i_b] + 1
+                    self.entity_island[i_v, i_b] = self.n_islands[i_b]
+                    if ti.static(False):  # Debug.validate
```
duburcqa: Add comment: `# FIXME: Add proper mechanism to detection overflow in Taichi-scope but raise exception in Python-scope`

genesis/engine/solvers/rigid/contact_island.py
```diff
@@ -162,61 +207,79 @@ def postprocess_island(self):
...
-                            if self.solver.entities_info.n_dofs[next_v] > 0 and next_v != v:
+                            if (
+                                self.solver.entities_info.n_dofs[next_v] > 0
+                                and next_v != v
+                                and self.entity_island[next_v, i_b] == -1
+                            ):  # 2nd condition must not happen ?
                                 self.stack[self.n_stack[i_b], i_b] = next_v
                                 self.n_stack[i_b] = self.n_stack[i_b] + 1
+                                self.entity_island[next_v, i_b] = self.n_islands[i_b]
+                                if ti.static(False):  # Debug.validate
```
duburcqa: Same.

genesis/utils/array_class.py
```diff
@@ -245,6 +249,7 @@ def __init__(self):
 
 @dataclasses.dataclass
 class StructContactData:
+    # WARNING: cannot add/remove fields here without also updeating collider_decomp.py::kernel_collider_clear
```
duburcqa: Typo `updeating`

genesis/engine/solvers/rigid/collider_decomp.py
```diff
@@ -494,14 +494,26 @@ def collider_kernel_reset(
...
 @ti.kernel
-def collider_kernel_clear(
+def kernel_collider_clear(
     envs_idx: ti.types.ndarray(),
     links_state: array_class.LinksState,
     links_info: array_class.LinksInfo,
     static_rigid_sim_config: ti.template(),
     collider_state: array_class.ColliderState,
 ):
+    cd_geom_a = ti.static(collider_state.contact_data.geom_a)
```
YilingQiao: Can we keep as it was, i.e. directly using `collider_state.contact_data.geom_a`
I'm not sure if py dataclass support this kind of alias @hughperkins

genesis/engine/solvers/rigid/rigid_solver_decomp_util.py
YilingQiao: Right now, there are only two `ti.static(False)`. Can we delete them and only add them back when we systematically add the Debug mechanism?

duburcqa: >  So yeah, @gasnica you can ignore my opinion as far as keeping the code in the Genesis codebase or not (and it sounds like Alexis is more favorable than I am to your keeping the code here for the time being :) )

First, I'm asking for your opinion. It is not "your" codebase, but it does not make your opinion less valuable. Just less binding. Seconds, I'm favorable to nothing except making life easier for maintainer and developers. I don't know if this debug feature would help actually. I'm just listening to the pros and cons and trying to make sure we are aligned with the standpoint (like running in debug all the unit tests systematically). That is all.

genesis/engine/solvers/rigid/collider_decomp.py
duburcqa: @YilingQiao Not that we are using this pattern a lot in SAP coupler I think.

YilingQiao: yeah, we need to fix that for SAP coupler. I did not touch that code otherwise it will create a lot of conflicts with Ziheng's ongoing work.

genesis/engine/solvers/rigid/rigid_solver_decomp_util.py
hughperkins: @duburcqa 

So, there are two things being debated:
- use of Debug
- allowing code that isn't being used in the codebase

You are agreeing with me that I have standing on the first point, the use of Debug in the codebase.

However, in this thread I'm stating a strong opinion against putting things in the codebase that arent being used, and I feel I don't have standing on this point.

genesis/engine/solvers/rigid/collider_decomp.py
hughperkins: yeah, assigning sub structs to variables isn't supported, though not because I've encountered any reason why it might be hard, simply because the consensus was that we don't need to do this currently. I'm not averse to investigating adding support for it, if there is a need for assigning sub structs to variables. (I dont know if it will turn out to be hard or not, since haven't looked into it)

genesis/engine/solvers/rigid/rigid_solver_decomp_util.py
gasnica: yes, Yiling, removing the ti.static(False) -- I missed those two places when removing everything else.

genesis/engine/solvers/rigid/contact_island.py
gasnica: absolutely. changing. resolving.

genesis/engine/solvers/rigid/contact_island.py
gasnica: this was written with a trailing comma, and black formatter does not reformat it then. reformatted. resolving.

genesis/engine/solvers/rigid/contact_island.py
gasnica: removed ti.static() block. addedd 2 commments. resolving.

genesis/engine/solvers/rigid/contact_island.py
gasnica: done. resolving.

genesis/utils/array_class.py
gasnica: corrected, resolving.

genesis/engine/solvers/rigid/collider_decomp.py
gasnica: reverting this change for now, 

and disabling formatting locally to avoid messy auto-formatting.

Issue #1513, opened 2025-08-06
[Feature]: IMU Sensor


Milotrince:
### What feature or enhancement are you proposing?

Implement IMU Sensor (accelerometer and gyroscope) with realistic noise.

### Motivation

IMU is the most ubiquitous sensor used in robotics.

### Potential Benefit

Easy to use IMU for all users

### What is the expected outcome of the implementation work?

- [x] ground truth IMU Sensor with no noise
- [x] sensor noise

### Additional information

We have a function in `RigidEntity` to obtain the link accelerations:
https://github.com/Genesis-Embodied-AI/Genesis/blob/efe90b988c3e4118dab2bccebfcec7ef471f0ce0/genesis/engine/entities/rigid_entity/rigid_entity.py#L1733-L1752

1. The `imu` parameter is currently unused (`mimick_imu` is always True)
2. Do we want to keep this function on `RigidEntity` if we have the IMU sensor class? the func for transforming to local frame (for both lin acc and angular) could be on the sensor 
  -  yes, move get_links_accelerometer_data to IMU sensor and remove from RigidEntity/Solver. Make sure to adjust docs, test
3. Do we want to cache/parallelize across multiple sensors, or only parallelization through batch per sensor?
  - yes, need scene cache. add SensorManager to Simulator
4. instantiate sensors with SensorOptions

PR #1551, opened 2025-08-13
[FEATURE] Add IMU sensor.

Milotrince:
## Description
Add IMU sensor and remove (or deprecate?) `get_lin_acc` which tried to mimic IMU

## Related Issue
https://github.com/Genesis-Embodied-AI/Genesis/issues/1513

## Motivation and Context
See related issue

Added test in `test_sensors.py`

## Checklist:


genesis/sensors/imu.py
```diff
@@ -0,0 +1,117 @@
...
+
+from .base_sensor import Sensor
+from .sensor_manager import register_sensor
+
+if TYPE_CHECKING:
+    from genesis.utils.ring_buffer import TensorRingBuffer
+
+
+@ti.data_oriented
+class IMU(Sensor):
```
duburcqa, line 81: No need to skip line.

genesis/sensors/imu.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def build(self):
+        self._solver = self._manager._sim.rigid_solver
+        assert self._options.link_idx >= 0 and self._options.link_idx < self._solver.n_links, "Invalid RigidLink index."
+        self._link_idx = self._options.link_idx
```
duburcqa: You should be able to check this condition when creating the sensor.

genesis/sensors/imu.py
```diff
@@ -0,0 +1,117 @@
...
+
+        quat_offset = euler_to_quat(self._options.euler_offset)
+
+        if len(self._shared_metadata) == 0:
+            self._shared_metadata["solver"] = self._solver
```
duburcqa: What about using sensor type-specific dataclasses to store metadata ? it would be better to have something more statically typed than a plain dictionary.

genesis/sensors/imu.py
```diff
@@ -0,0 +1,117 @@
...
+
+        offset_quats = transform_quat_by_quat(quats, offsets_quat.unsqueeze(0).repeat(quats.shape[0], 1, 1))
+        # acc/ang shape: (B, n_links, 3)
+        local_acc = inv_transform_by_trans_quat(acc, offsets_pos, offset_quats)
+        local_ang = inv_transform_by_trans_quat(ang, offsets_pos, offset_quats)
+
+        local_acc = local_acc - gravity.unsqueeze(1).repeat(1, local_acc.shape[1], 1)
+
+        # cache shape: (B, n_links * 6)
+        shared_ground_truth_cache.copy_(torch.cat([local_acc, local_ang], dim=2).flatten(1))
```
duburcqa: Avoid doing `cat_`. Extract the right slice for the `local_acc`, `local_ang` and update them separately.

genesis/sensors/imu.py
```diff
@@ -0,0 +1,117 @@
...
+
+    @classmethod
+    def _update_shared_cache(
+        cls, shared_metadata: dict[str, Any], shared_ground_truth_cache: torch.Tensor, shared_cache: "TensorRingBuffer"
+    ):
+        shared_cache.append(shared_ground_truth_cache)
```
duburcqa: Are you sure this function is really useful in the API? Does it make sense to customise it?

genesis/sensors/imu.py
```diff
@@ -0,0 +1,117 @@
...
+
+    @classmethod
+    def _get_cache_dtype(cls) -> torch.dtype:
+        return gs.tc_float
+
+
+@register_sensor(IMU)
```
duburcqa: It would be nice to move the options top of the file, but it is not really possible, useless to "reverse" the `registry`:
```python
@register_sensor(IMUOptions)
class IMU:
    [...]
```

genesis/sensors/imu.py
```diff
@@ -0,0 +1,117 @@
...
+
+        gravity = solver.get_gravity()
+        quats = solver.get_links_quat(links_idx=links_idx)
+        acc = solver.get_links_acc(links_idx=links_idx)
+        ang = solver.get_links_ang(links_idx=links_idx)
+        if solver.n_envs == 0:
+            gravity = gravity.unsqueeze(0)
+            quats = quats.unsqueeze(0)
+            acc = acc.unsqueeze(0)
+            ang = ang.unsqueeze(0)
```
duburcqa: I don't think this is necessary. It should be possible to have the exact same code working on batched and non-batched version.

genesis/sensors/imu.py
```diff
@@ -0,0 +1,117 @@
...
+
+        offset_quats = transform_quat_by_quat(quats, offsets_quat.unsqueeze(0).repeat(quats.shape[0], 1, 1))
```
duburcqa: If you already need to do `offsets_quat.unsqueeze(0).repeat(quats.shape[0], 1, 1)`, why not to store `offsets_quat ` directly with the right shape? You should also prefer `torch.expand` over `torch.repeat`, because it creates a view that shares memory with the original data without copy.

genesis/sensors/imu.py
```diff
@@ -0,0 +1,117 @@
...
+    ----------
+    link_idx : int
+        The global index of the RigidLink to which this IMU sensor is attached.
+    pos_offset : tuple[float, float, float]
+        The offset of the IMU sensor from the RigidLink.
+    euler_offset : tuple[float, float, float]
+        The offset of the IMU sensor from the RigidLink in euler angles.
+    """
+
+    link_idx: int
```
duburcqa: I think passing entity_idx + link_idx (local) is more intuitive for users no?

tests/test_rigid_physics.py
```diff
@@ -711,22 +711,18 @@ def test_pendulum_links_acc(gs_sim, tol):
         # Linear true acceleration:
         # * acc_classical_lin_y = sin(theta) * g (tangential angular acceleration effect)
         # * acc_classical_lin_z = - theta_dot ** 2  (radial centripedal effect)
-        acc_classical_lin_world = tensor_to_array(gs_sim.rigid_solver.get_links_acc(mimick_imu=False))
+        acc_classical_lin_world = tensor_to_array(gs_sim.rigid_solver.get_links_acc())
         assert_allclose(acc_classical_lin_world[0], 0, tol=tol)
         acc_classical_lin_local = R @ acc_classical_lin_world[2]
         assert_allclose(acc_classical_lin_local, np.array([0.0, np.sin(theta) * g, -(theta_dot**2)]), tol=tol)
-        # IMU accelerometer data:
-        # * acc_classical_lin_z = - theta_dot ** 2 - cos(theta) * g
```
duburcqa, line 719: Could you move this comment at its new location?

tests/test_sensors.py
```diff
@@ -0,0 +1,70 @@
...
+    imu_delayed = scene.add_sensor(IMUOptions(link_idx=box.base_link_idx, read_delay=DT))
+
+    scene.build()
+
+    for _ in range(10):
+        scene.step()
+
+    # freefall
+    imu_data = imu.read()
+    assert_allclose(imu_data["lin_acc"], torch.tensor([0.0, 0.0, 0.0]), tol=1e-7)
```
duburcqa: `assert_allclose(imu_data["lin_acc"], 0.0, tol=1e-7)`

tests/test_sensors.py
```diff
@@ -0,0 +1,70 @@
...
+    assert_allclose(imu_data["ang_vel"], torch.tensor([0.0, 0.0, 0.0]), tol=1e-7)
+
+    # shift COM to induce angular velocity
+    box.set_COM_shift(torch.tensor([[0.1, 0.1, 0.1]]))
+
+    for _ in range(30):
+        scene.step()
+    # collision with ground
```
duburcqa: What is this comment?

tests/test_sensors.py
```diff
@@ -0,0 +1,70 @@
...
+
+    imu_data = imu.read()
+    imu_delayed_data = imu_delayed.read()
```
duburcqa: Check the ground truth.

genesis/sensors/imu.py
Milotrince: i think black autoformatter adds this extra line

genesis/sensors/imu.py
Milotrince: added validate function in SensorOptions

genesis/sensors/imu.py
Milotrince: Yes, this is where the noise would be updated, but I'm not implementing any noise right now

genesis/sensors/imu.py
Milotrince: sure, we can reverse it

genesis/sensors/imu.py
Milotrince: I also found this annoying, but every `solver.get_links_*` does squeeze if `n_envs == 0` so I have to "undo" it:
```
    def get_links_ang(self, links_idx=None, envs_idx=None, *, unsafe=False):
        tensor = ti_field_to_torch(self.links_state.cd_ang, envs_idx, links_idx, transpose=True, unsafe=unsafe)
        return tensor.squeeze(0) if self.n_envs == 0 else tensor
```
do you think its worth adding a parameter to the `get_links_*` functions to keep batch dim? or is there a better way?

genesis/sensors/imu.py
Milotrince: updated to use `torch.expand`. see: `unsqueeze(0)` comment above first

genesis/sensors/imu.py
duburcqa: Ok let's keep it.

genesis/sensors/imu.py
duburcqa: This is what this method does? The name is confusing then.

genesis/sensors/imu.py
duburcqa: No, I just think your implementation should work the same regardless data are batched or not. Typically all the geoms helpers support N-dimensional batch shape (including empty) without having to do any squeeze or such.

genesis/sensors/imu.py
Milotrince: yeah, the current naming scheme is `ground_truth_cache` and `cache` for "with noise" (but if there's no noise implemented in the sensor, then of course there's no noise other than delay)-- any suggestions?

duburcqa: I mean, this is fine. What is confusing is the name of the method itself.

duburcqa: By the way, there should be a single function that deals with "converting" ground truth data to measured data, which includes bias + noise + delay. Here it seems that you are dealing with the delay in some other place?

genesis/sensors/imu.py
Milotrince: its a bit simplified for me if i do the unsqueeze here, otherwise i may need to add if checks elsewhere, or allow for cache to not have batch dim

genesis/sensors/imu.py
Milotrince: yes, delay was on `read()`. I updated to do this in `_update_shared_cache_with_noise` instead, but i dont think i can avoid a for loop with this

genesis/sensors/imu.py
```diff
@@ -123,20 +131,42 @@ def _update_shared_ground_truth_cache(
...
+    def _update_shared_cache_with_noise(
+        cls,
+        shared_metadata: dict[str, Any],
+        shared_ground_truth_cache: torch.Tensor,
+        shared_cache: torch.Tensor,
+        buffered_data: "TensorRingBuffer",
     ):
-        shared_cache.append(shared_ground_truth_cache)
+        # store the history of ground truth cache in the buffer
+        buffered_data.append(shared_ground_truth_cache)
```
Milotrince: I believe IMU and most sensors would need the ground truth history to do realistic noise processing, but perhaps some other sensors may actually require the previous noised data instead? so the sensor implementor has the choice here of what they need to buffer, and shared_cache/shared_ground_truth_cache are both tensors of what would be read at the current step

genesis/options/sensors.py
Milotrince, line 1: I moved options to base_sensor.py so it matches the overall pattern of `SensorOptions`, `SharedSensorMetadata`, and `Sensor` being together

genesis/sensors/imu.py
duburcqa: I see. My bad. Delay is indeed something else.

genesis/sensors/imu.py
```diff
@@ -0,0 +1,167 @@
...
+        )
+        self._shared_metadata.ang_bias = torch.cat(
+            [self._shared_metadata.ang_bias, torch.tensor([self._options.gyroscope_bias], dtype=gs.tc_float)]
+        )
+
+    def _get_return_format(self) -> dict[str, tuple[int, ...]]:
+        return {
+            "lin_acc": (3,),
+            "ang_vel": (3,),
+        }
```
Milotrince, line 111: I had `return_accelerometer : bool` in options to return lin_acc or ang_vel only, but since we infer cache size from _get_return_format, this means we can have inconsistent cache sizes per IMU sensor, meaning i cannot easily set all the values in the cache at once. So, I removed the option

genesis/sensors/imu.py
duburcqa: What about `_compute_skew_data` or just `_add_noise_and_bias` ? Something more explicit about what it does?

genesis/sensors/imu.py
duburcqa: Ok I think I will help you on this.

genesis/sensors/imu.py
duburcqa: > I believe IMU and most sensors would need the ground truth history to do realistic noise processing

It seems overkill to me, but ok.

Could you add what you are explaining to me in the documentation?

genesis/sensors/imu.py
Milotrince: i think `_update_shared_cache_with_noise` is appropriate because this is where all the noise for the sensor should be handled-- and it must modify/update `shared_cache` in place

genesis/sensors/imu.py
duburcqa: This should just affect the final data extraction (what is returned to the user). Not what is being computed internally.

genesis/sensors/imu.py
Milotrince: keep as _update_shared_cache, clarify in documentation

PR #1563, opened 2025-08-14
[FEATURE] Add New Features for Madrona Batch Rendering

ACMLCZH:
- Based on #1416 , add more features for BatchRender. See demo script `examples/rigid/single_franka_batch_render.py`.

### Render
- Use `rgb/depth/normal/segmentation` in `camera.render()` or `scene.render_all_cameras()` to control the image components you want to render.
- Use `colorize_seg` to colorize the segmentation map.
- Use `segmentation_level="geom"/"link"/"entity"` in `gs.options.VisOptions` to control the segmentation level. (Now Pyrender also supports segmentation_level="geom").
- Support rendering point clouds with BatchRender in `camera.render_pointclouds()`.
- Images are returned as `ndarray` and depth/segmentation image has shape `(..., H, W,)` (instead of `(..., H, W, 1)` ) for consistency with other renderers.

### Light
- Use `scene.add_light()` to add point/directional lights for BatchRender. Original `add_light()` function (specific for Luisa Raytracer) is renamed to `add_mesh_light()`.
- Support `color` for colored lights.
- Support `attenuation` for attenuation factor (based on squared distance) for pointed lights.

### Camera
- Support setting `near` and `far` plane distance in `scene.add_camera()`.

### Bug Fix
- Fix projection matrix in BatchRender. Now its depth images should match with Pyrender.
- Fix issue that when running in raytracer mode, the rasterization also runs. The (single-bounced) raytracer mode's speed should be slightly faster than rasterizer mode now.
- Align ambient lights as `0.05` in both rasterizer and raytracer modes. The brightness of images in the same light configuration should be the same.

### Code Base
- Add a `SegmentationManager` helper class for both Rasterizer and BatchRender.
- Remove the usage of `RigidSolver` in gs-madrona, replaced with `GeomRetriever` helper class.
- Optimize the code structure of tool class `FrameImageExporter` in `gs.utils.image_exporter.py`. Now it accepts ndarray images.

Resolves https://github.com/Genesis-Embodied-AI/Genesis/issues/1632
Resolves https://github.com/Genesis-Embodied-AI/Genesis/issues/1648

### DIFF
--- a/examples/keyboard_teleop.py
+++ b/examples/keyboard_teleop.py
@@ -62,7 +62,7 @@ def build_scene():
             enable_collision=True,
             gravity=(0, 0, -9.8),
             box_box_detection=True,
-            constraint_timeconst=0.02,
+            constraint_timeconst=0.01,
         ),
         viewer_options=gs.options.ViewerOptions(
             camera_pos=(1.5, 0.0, 0.7),
--- a/genesis/engine/solvers/rigid/collider_decomp.py
+++ b/genesis/engine/solvers/rigid/collider_decomp.py
@@ -596,7 +596,9 @@ def collider_kernel_get_contacts(
     _B = collider_state.active_buffer.shape[1]
     n_contacts_max = gs.ti_int(0)
 
-    ti.loop_config(serialize=static_rigid_sim_config.para_level < gs.PARA_LEVEL.ALL)
+    # this is a reduction operation (global max), we have to serialize it
+    # TODO: a good unittest and a better implementation from gstaichi for this kind of reduction
+    ti.loop_config(serialize=True)
     for i_b in range(_B):
         n_contacts = collider_state.n_contacts[i_b]
         if n_contacts > n_contacts_max:
--- a/genesis/engine/solvers/rigid/constraint_solver_decomp.py
+++ b/genesis/engine/solvers/rigid/constraint_solver_decomp.py
@@ -2221,7 +2221,9 @@ def kernel_get_equality_constraints(
     _B = constraint_state.ti_n_equalities.shape[0]
     n_eqs_max = gs.ti_int(0)
 
-    ti.loop_config(serialize=static_rigid_sim_config.para_level < gs.PARA_LEVEL.ALL)
+    # this is a reduction operation (global max), we have to serialize it
+    # TODO: a good unittest and a better implementation from gstaichi for this kind of reduction
+    ti.loop_config(serialize=True)
     for i_b in range(_B):
         n_eqs = constraint_state.ti_n_equalities[i_b]
         if n_eqs > n_eqs_max:
--- a/tests/test_rigid_physics.py
+++ b/tests/test_rigid_physics.py
@@ -2913,53 +2913,43 @@ def test_mesh_primitive_COM(show_viewer, tol):
 
 
 @pytest.mark.required
-@pytest.mark.parametrize("backend", [gs.cpu])
-def test_batched_aabb(show_viewer, tol):
-    scene = gs.Scene(
-        sim_options=gs.options.SimOptions(gravity=(0.0, 0.0, -10.0)),
-        show_viewer=show_viewer,
-    )
-
+def test_batched_aabb(tol):
+    scene = gs.Scene()
     plane = scene.add_entity(
-        gs.morphs.Plane(normal=(0, 0, 1), pos=(0, 0, 0)),
-        material=gs.materials.Rigid(),
+        gs.morphs.Plane(
+            normal=(0, 0, 1),
+            pos=(0, 0, 0),
+        ),
     )
     box = scene.add_entity(
-        gs.morphs.Box(size=(0.1, 0.1, 0.1), pos=(0.5, 0, 0.05)),
-        material=gs.materials.Rigid(),
+        gs.morphs.Box(
+            size=(0.1, 0.1, 0.1),
+            pos=(0.5, 0, 0.05),
+        ),
     )
     sphere = scene.add_entity(
-        gs.morphs.Sphere(radius=0.05, pos=(-0.5, 0, 0.05)),
-        material=gs.materials.Rigid(),
+        gs.morphs.Sphere(
+            radius=0.05,
+            pos=(-0.5, 0, 0.05),
+        ),
     )
-
     scene.build()
 
     all_aabbs = scene.sim.rigid_solver.get_aabb()
-    assert_allclose(torch.tensor(all_aabbs.shape), torch.tensor([3, 2, 3]), atol=tol)
-
     plane_aabb = plane.get_aabb()
     box_aabb = box.get_aabb()
     sphere_aabb = sphere.get_aabb()
 
-    assert_allclose(torch.tensor(plane_aabb.shape[-1]), torch.tensor(3), atol=tol)
-    assert_allclose(torch.tensor(box_aabb.shape[-1]), torch.tensor(3), atol=tol)
-    assert_allclose(torch.tensor(sphere_aabb.shape[-1]), torch.tensor(3), atol=tol)
+    assert_allclose(all_aabbs.shape, (3, 2, 3), atol=0)
+    assert_allclose(plane_aabb.shape[-1], 3, atol=0)
+    assert_allclose(box_aabb.shape[-1], 3, atol=0)
+    assert_allclose(sphere_aabb.shape[-1], 3, atol=0)
+    assert_allclose((plane_aabb, box_aabb, sphere_aabb), all_aabbs, atol=tol)
 
-    plane_aabb_squeezed = plane_aabb.squeeze() if plane_aabb.ndim == 3 else plane_aabb
-    box_aabb_squeezed = box_aabb.squeeze() if box_aabb.ndim == 3 else box_aabb
-    sphere_aabb_squeezed = sphere_aabb.squeeze() if sphere_aabb.ndim == 3 else sphere_aabb
+    box_aabb_min, box_aabb_max = box_aabb
+    assert_allclose(box_aabb_min, (0.45, -0.05, 0.0), atol=tol)
+    assert_allclose(box_aabb_max, (0.55, 0.05, 0.1), atol=tol)
 
-    assert_allclose(plane_aabb_squeezed, all_aabbs[0], atol=tol)
-    assert_allclose(box_aabb_squeezed, all_aabbs[1], atol=tol)
-    assert_allclose(sphere_aabb_squeezed, all_aabbs[2], atol=tol)
-
-    expected_box_min = torch.tensor([0.45, -0.05, 0.0])
-    expected_box_max = torch.tensor([0.55, 0.05, 0.1])
-    assert_allclose(box_aabb_squeezed[0], expected_box_min, atol=tol)
-    assert_allclose(box_aabb_squeezed[1], expected_box_max, atol=tol)
-
-    expected_sphere_min = torch.tensor([-0.55, -0.05, 0.0])
-    expected_sphere_max = torch.tensor([-0.45, 0.05, 0.1])
-    assert_allclose(sphere_aabb_squeezed[0], expected_sphere_min, atol=tol)
-    assert_allclose(sphere_aabb_squeezed[1], expected_sphere_max, atol=tol)
+    sphere_aabb_min, sphere_aabb_max = sphere_aabb
+    assert_allclose(sphere_aabb_min, (-0.55, -0.05, 0.0), atol=tol)
+    assert_allclose(sphere_aabb_max, (-0.45, 0.05, 0.1), atol=tol)
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -266,7 +266,7 @@ def assert_allclose(actual, desired, *, atol=None, rtol=None, tol=None, err_msg=
     if all(e.size == 0 for e in args):
         return
 
-    np.testing.assert_allclose(*args, atol=atol, rtol=rtol, err_msg=err_msg)
+    np.testing.assert_allclose(*map(np.squeeze, args), atol=atol, rtol=rtol, err_msg=err_msg)
 
 
 def assert_array_equal(actual, desired, *, err_msg=""):


.gitignore
duburcqa, line 1: Remove this change.

.gitmodules
duburcqa, line 1: Remove this change.

genesis/__init__.py
duburcqa, line 1: Remove this change.

genesis/engine/scene.py
```diff
@@ -1098,7 +1111,24 @@ def draw_debug_path(self, qposs, entity, link_idx=-1, density=0.3, frame_scaling
             )
 
     @gs.assert_built
-    def render_all_cameras(self, rgb=True, depth=False, normal=False, segmentation=False, force_render=False):
+    def get_segmentation_idx_dict(self):
```
duburcqa: Use a property instead.

genesis/ext/gs-madrona
duburcqa, line 1: Remove this change.

genesis/vis/camera.py
```diff
@@ -505,13 +492,23 @@ def render_pointcloud(self, world_frame=True):
...
+            self._rasterizer.update_scene()
+            _, depth_arr, _, _ = self._rasterizer.render_camera(
+                self, rgb=False, depth=True, segmentation=False, normal=False
+            )
+        else:
+            gs.raise_exception("Render point cloud not supported.")
 
         # Convert OpenGL projection matrix to camera intrinsics
-        P = self._rasterizer._camera_nodes[self.uid].camera.get_projection_matrix()
+        P = self.projection_matrix
```
duburcqa: Since you are computing the projection matrix manually, you could just extract `fx, fy, cx, cy` directly from camera parameters without going through the projection matrix. Could you change this?

Yet, do not remove the projection matrix property, it is nice to have it anyway.

genesis/vis/visualizer.py
```diff
@@ -8,7 +8,6 @@
 from .camera import Camera
 from .rasterizer import Rasterizer
 
-
```
duburcqa: Keep these 2 line skip before global variables.

genesis/vis/visualizer.py
```diff
@@ -143,15 +142,23 @@ def add_camera(self, res, pos, lookat, up, model, fov, aperture, focus_dist, GUI
...
         )
         self._cameras.append(camera)
         return camera
 
-    def add_light(self, pos, dir, intensity, directional, castshadow, cutoff):
+    def add_mesh_light(self, mesh, color, intensity, pos, quat, revert_dir, double_sided, cutoff):
+        if self._raytracer is not None:
+            self._raytracer.add_mesh_light(mesh, color, intensity, pos, quat, revert_dir, double_sided, cutoff)
+        else:
+            gs.raise_exception("`add_mesh_light` is specifically for raytracer renderer.")
```
duburcqa: Use the same terminology: `is only supported by`.

It would be even better to say something like `'add_mesh_light' is not supported by raytracer renderer. Please use e 'add_light'.`

genesis/vis/visualizer.py
```diff
@@ -249,6 +256,14 @@ def update_visual_states(self):
 
         self._t = self._scene._t
 
+    def colorize_seg_idxc_arr(self, seg_idxc_arr):
+        if self._batch_renderer is not None:
+            return self._batch_renderer.colorize_seg_idxc_arr(seg_idxc_arr)
+        elif self._rasterizer is not None:
+            return self._context.colorize_seg_idxc_arr(seg_idxc_arr)
+        else:
+            gs.raise_exception("Segmentation is only available with batch renderer or rasterizer.")
```
duburcqa: `self._rasterizer` is never none. No need to suppose it can be.

genesis/vis/visualizer.py
```diff
@@ -288,3 +303,32 @@ def has_display(self):
...
+        return torch.tensor([camera.far for camera in self._cameras], dtype=gs.tc_float, device=gs.device)
+
+    @property
+    def segmentation_idx_dict(self):
+        if self._batch_renderer is not None:
+            return self._batch_renderer.seg_idxc_map
+        elif self._rasterizer is not None:
+            return self._context.seg_manager.seg_idxc_map
+        else:
+            gs.raise_exception("Segmentation is only available with batch renderer or rasterizer.")
```
duburcqa: rasterizer is never none. No need to suppose it can be.

genesis/vis/visualizer.py
```diff
@@ -288,3 +303,32 @@ def has_display(self):
...
+    def camera_fov(self):
+        return torch.tensor([camera.fov for camera in self._cameras], dtype=gs.tc_float, device=gs.device)
+
+    @property
+    def camera_near(self):
+        return torch.tensor([camera.near for camera in self._cameras], dtype=gs.tc_float, device=gs.device)
+
+    @property
+    def camera_far(self):
+        return torch.tensor([camera.far for camera in self._cameras], dtype=gs.tc_float, device=gs.device)
```
duburcqa: I'm not convinced these properties should be added. Where is it used?

genesis/vis/rasterizer_context.py
```diff
@@ -13,6 +13,34 @@
 from genesis.utils.misc import tensor_to_array
 
 
+class SegmentationManager:
```
duburcqa: This class does not manage anything, and is not about "segmentation" strictly speaking. It is mapping indices to random color for use with segmentation. That is all. You should be able to come up with a better name.

duburcqa: ChatGPT suggests `SegmentationPalette` or `SegmentationColorMap`

genesis/vis/rasterizer_context.py
```diff
@@ -46,8 +74,7 @@ def __init__(self, options):
         self.dynamic_nodes = list()  # nodes that live within single frame
         self.external_nodes = dict()  # nodes added by external user
         self.seg_node_map = dict()
-        self.seg_idxc_map = {0: -1}
-        self.seg_key_map = {-1: 0}
+        self.seg_manager = SegmentationManager()
```
duburcqa: `self.seg_color_map` should be fine.

genesis/vis/rasterizer_context.py
```diff
@@ -13,6 +13,34 @@
 from genesis.utils.misc import tensor_to_array
 
 
+class SegmentationManager:
+    def __init__(self, seed=42):
+        self.seed = seed
+        self.seg_idxc_map = {0: -1}
+        self.seg_key_map = {-1: 0}
+        self.seg_idxc_to_color = None
```
duburcqa: remove `seg_` prefix, it is useless. This is already the clear based on what this class has been designed for.

genesis/utils/image_exporter.py
```diff
@@ -1,23 +1,88 @@
...
 import numpy as np
-import torch
+
 from functools import partial
 from concurrent.futures import ThreadPoolExecutor
 
 import genesis as gs
-from genesis.utils.misc import tensor_to_array
 
+RGB_S = "rgb"
```
duburcqa: Skip two lines before global variables.

genesis/vis/batch_renderer.py
```diff
@@ -3,9 +3,14 @@
...
 import numpy as np
 import torch
+from trimesh.visual.texture import TextureVisuals
+from trimesh.visual.color import ColorVisuals
+from PIL import Image
 
 import genesis as gs
 from genesis.repr_base import RBC
+from genesis.utils.misc import tensor_to_array
+from .rasterizer_context import SegmentationManager
```
duburcqa: Skip line before relative imports.

genesis/utils/image_exporter.py
duburcqa: What about using our `IMAGE_TYPE` enum ? You could convert it to a `StrEnum` to specify custom string names for each of them. Not sure it is really relevant though. What do you think?

genesis/vis/batch_renderer.py
```diff
@@ -3,9 +3,14 @@
 
 import numpy as np
 import torch
+from trimesh.visual.texture import TextureVisuals
+from trimesh.visual.color import ColorVisuals
+from PIL import Image
```
duburcqa: `PIL` should come before `trimesh` before of alphabetical ordering.

genesis/vis/batch_renderer.py
```diff
@@ -141,53 +326,73 @@ def render(self, rgb=True, depth=False, segmentation=False, normal=False, force_
...
         rgb_arr = self._data_cache.get((IMAGE_TYPE.RGB, cache_key), None)
         depth_arr = self._data_cache.get((IMAGE_TYPE.DEPTH, cache_key), None)
+        segmentation_arr = self._data_cache.get((IMAGE_TYPE.SEGMENTATION, cache_key), None)
+        normal_arr = self._data_cache.get((IMAGE_TYPE.NORMAL, cache_key), None)
 
         # Force disabling rendering whenever cached data is already available
         rgb_ = rgb and rgb_arr is None
         depth_ = depth and depth_arr is None
+        segmentation_ = segmentation and segmentation_arr is None
+        normal_ = normal and normal_arr is None
```
duburcqa: What about defining `flags = (rgb_, depth_, segmentation_, normal_)` ?

This should allow to factorize the implementation now that there is 4 options.

genesis/vis/batch_renderer.py
```diff
@@ -141,53 +326,73 @@ def render(self, rgb=True, depth=False, segmentation=False, normal=False, force_
...
 
         # Early return if there is nothing to do
-        if not (rgb_ or depth_):
-            return rgb_arr if rgb else None, depth_arr if depth else None, None, None
+        if not (rgb_ or depth_ or segmentation_ or normal_):
```
duburcqa: `if not any(flags):`

genesis/vis/batch_renderer.py
```diff
@@ -141,53 +326,73 @@ def render(self, rgb=True, depth=False, segmentation=False, normal=False, force_
...
         # Update scene
         self.update_scene()
 
         # Render frame
         cameras_pos = torch.stack([camera.get_pos() for camera in self._cameras], dim=1)
         cameras_quat = torch.stack([camera.get_quat() for camera in self._cameras], dim=1)
-        render_options = np.array((rgb_, depth_, False, False, aliasing), dtype=np.uint32)
-        rgba_arr_all, depth_arr_all = self._renderer.render(
-            self._visualizer.scene.rigid_solver, cameras_pos, cameras_quat, render_options
+        render_options = np.array((rgb_, depth_, normal_, segmentation_, antialiasing), dtype=np.uint32)
```
duburcqa: `np.array((*flags, antialiasing), dtype=np.uint32)`

genesis/vis/batch_renderer.py
```diff
@@ -141,53 +326,73 @@ def render(self, rgb=True, depth=False, segmentation=False, normal=False, force_
...
+        rgba_arr_all, depth_arr_all, segmentation_arr_all, normal_arr_all = self._renderer.render(
```
duburcqa: `buffers = [...]`

genesis/vis/batch_renderer.py
```diff
@@ -141,53 +326,73 @@ def render(self, rgb=True, depth=False, segmentation=False, normal=False, force_
...
         )
 
         # Post-processing: Remove alpha channel from RGBA, squeeze env dim if necessary, and split along camera dim
-        buffers = [rgba_arr_all[..., :3], depth_arr_all]
+        buffers = [
+            tensor_to_array(rgba_arr_all[..., :3].flip(-1)) if rgb else None,
+            tensor_to_array(depth_arr_all[..., 0]) if depth else None,
+            tensor_to_array(segmentation_arr_all[..., 0]) if segmentation else None,
+            tensor_to_array(normal_arr_all[..., :3].flip(-1)) if normal else None,
+        ]
```
duburcqa: `buffers = [tensor_to_array(buffer[..., :3].flip(-1).squeeze(-1)) if flag else None for buffer, flag in zip(buffers, flags)]`

genesis/vis/batch_renderer.py
```diff
@@ -201,8 +406,15 @@ def render(self, rgb=True, depth=False, segmentation=False, normal=False, force_
             rgb_arr = self._data_cache[(IMAGE_TYPE.RGB, cache_key)] = buffers[0]
         if depth_:
             depth_arr = self._data_cache[(IMAGE_TYPE.DEPTH, cache_key)] = buffers[1]
+        if segmentation_:
+            segmentation_arr = self._data_cache[(IMAGE_TYPE.SEGMENTATION, cache_key)] = buffers[2]
+        if normal_:
+            normal_arr = self._data_cache[(IMAGE_TYPE.NORMAL, cache_key)] = buffers[3]
```
duburcqa: Same. For loop instead.

genesis/vis/batch_renderer.py
```diff
@@ -221,3 +433,11 @@ def lights(self):
...
     def cameras(self):
         return self._cameras
+
+    @property
+    def seg_idxc_map(self):
+        return self._geom_retriever.seg_manager.seg_idxc_map
+
+    @property
+    def use_rasterizer(self):
+        return self._use_rasterizer
```
duburcqa: I'm not convinced this needs to be exposed. Is it used anywhere? Otherwise remove it.

genesis/vis/batch_renderer.py
```diff
@@ -16,18 +21,177 @@
...
+        geom_sizes = torch.ones((self.n_vgeoms, 3), dtype=torch.float32)
+        geom_sizes = geom_sizes.unsqueeze(0).repeat(num_worlds, 1, 1)
+        return geom_mat_ids, geom_rgb, geom_sizes
+
+    def retrieve_rigid_state_torch(self):
+        geom_pos = self.rigid_solver.vgeoms_state.pos.to_torch()
+        geom_rot = self.rigid_solver.vgeoms_state.quat.to_torch()
+        geom_pos = geom_pos.transpose(0, 1).contiguous()
+        geom_rot = geom_rot.transpose(0, 1).contiguous()
+        return geom_pos, geom_rot
```
duburcqa, line 190: We should expose a kernel at does exactly this, but more efficiently. Let's do this later but add some `# FIXME: ` comment to explain that this implementation is slow and should be optimized. Same for all the other helpers.

genesis/vis/batch_renderer.py
```diff
@@ -16,18 +21,177 @@
...
+        geom_rgb = geom_rgb_uint.unsqueeze(0).repeat(num_worlds, 1)
+
+        geom_mat_ids = torch.full((self.n_vgeoms,), -1, dtype=torch.int32)
+        geom_mat_ids = geom_mat_ids.unsqueeze(0).repeat(num_worlds, 1)
+
+        geom_sizes = torch.ones((self.n_vgeoms, 3), dtype=torch.float32)
+        geom_sizes = geom_sizes.unsqueeze(0).repeat(num_worlds, 1, 1)
+        return geom_mat_ids, geom_rgb, geom_sizes
+
+    def retrieve_rigid_state_torch(self):
```
duburcqa, line 185: Where is it used?

genesis/vis/batch_renderer.py
```diff
@@ -86,48 +258,61 @@ def build(self):
...
+        light_rgb = torch.tensor([light.color for light in self._lights], dtype=gs.tc_float)
         light_directional = torch.tensor([light.directional for light in self._lights], dtype=gs.tc_int)
         light_castshadow = torch.tensor([light.castshadow for light in self._lights], dtype=gs.tc_int)
         light_cutoff = torch.tensor([light.cutoffRad for light in self._lights], dtype=gs.tc_float)
+        light_attenuation = torch.tensor([light.attenuation for light in self._lights], dtype=gs.tc_float)
+        light_intensity = torch.tensor([light.intensity for light in self._lights], dtype=gs.tc_float)
 
         self._renderer = MadronaBatchRendererAdapter(
-            rigid, gpu_id, n_envs, n_cameras, n_lights, cameras_fov, *res, False, use_rasterizer
+            self._geom_retriever,
```
duburcqa: Update version requirement in pyproject ? And update version in gs-madrona accordingly.

genesis/engine/scene.py
```diff
@@ -536,6 +541,8 @@ def add_camera(
         GUI=False,
         spp=256,
         denoise=None,
+        near=0.05,
```
yuhongyi, line 561: Where are the default values from?

genesis/engine/scene.py
```diff
@@ -446,77 +446,82 @@ def link_entities(
         parent_link._child_idxs.append(child_link.idx)
 
     @gs.assert_unbuilt
-    def add_light(
+    def add_mesh_light(
```
yuhongyi, line 454: @duburcqa I think it's a good idea to separate add_mesh_light() from add_light. How do we usually communicate with our users regarding API changes?

genesis/options/surfaces.py
```diff
@@ -67,8 +67,8 @@ class Surface(Options):
     double_sided : bool | None, optional
         Whether to render both sides of the surface. Useful for non-watertight 2D objects. Defaults to True for Cloth
         material and False for others.
-    beam_angle : float
-        The beam angle of emission. Defaults to 180.0.
+    cutoff : float
```
yuhongyi, line 70: This is only used by raytracing right?
The new name is more intuitive, but we need to communicate this API change with our users somehow.

genesis/vis/batch_renderer.py
```diff
@@ -141,68 +342,88 @@ def render(self, rgb=True, depth=False, segmentation=False, normal=False, force_
...
         if force_render or self._t < self._visualizer.scene.t:
             self._data_cache.clear()
 
         # Fetch available cached data
-        cache_key = (aliasing,)
-        rgb_arr = self._data_cache.get((IMAGE_TYPE.RGB, cache_key), None)
-        depth_arr = self._data_cache.get((IMAGE_TYPE.DEPTH, cache_key), None)
+        req = [rgb, depth, segmentation, normal]
+        cache_key = (antialiasing,)
+        cached = [self._data_cache.get((t, cache_key), None) for t in range(IMAGE_TYPE.NUM_TYPES)]
```
yuhongyi: Only rgb cares about antialiasing, right?

yuhongyi: Why is antialiasing not listed in req?

ACMLCZH: `req` is specific for image components, `antialiasing` is later packed into `render_flags`.

genesis/options/surfaces.py
ACMLCZH: `beam_angle` is the functions the same as `cutoff` in Raytracer, so I changed its name, maybe we should update the version.

genesis/engine/scene.py
ACMLCZH: DEFAULT_Z_NEAR and DEFAULT_Z_FAR from `pyrender/constants.py`

yuhongyi (dismissed):

genesis/engine/scene.py
YilingQiao: Can you at least raise an exception for the original add_light function in Luisa renderer? So users know the API has changed? @ACMLCZH

ACMLCZH: OK

duburcqa: We implement hard deprecation (throwing exception) via `DeprecationError` (throwing) and soft deprecation (warning) via `gs.logger.warning`. See `RigidJoint.get_pos` and `RigidJoint.dof_idx` respectively.

Try using the exact same message, only changing the part that is must be adapted.

genesis/vis/visualizer.py
```diff
@@ -150,17 +161,25 @@ def add_camera(self, res, pos, lookat, up, model, fov, aperture, focus_dist, GUI
...
+        else:
+            gs.raise_exception("`add_mesh_light` is only supported by raytracer renderer.")
+
+    def add_light(self, pos, dir, color, intensity, directional, castshadow, cutoff, attenuation):
         if self._batch_renderer is not None:
-            self._batch_renderer.add_light(pos, dir, intensity, directional, castshadow, cutoff)
+            self._batch_renderer.add_light(pos, dir, color, intensity, directional, castshadow, cutoff, attenuation)
         else:
-            gs.raise_exception("`add_light` is specifically for batch renderer.")
+            gs.raise_exception("`add_light` is only supported by batch renderer.")
```
duburcqa: Try being as specific as possible. "only supported by" means it could be supported by other backends in the future, whereas "reserved for" / "dedicated for" / "specific to" clarifies that it is a design choice.

pyproject.toml
```diff
@@ -57,7 +57,7 @@ dependencies = [
     # Used for loading raytracing special texture images used by LuisaRender
     "OpenEXR",
     # Native batch renderer specifically designed for Genesis
-    "gs-madrona==0.0.3; platform_system == 'Linux' and (platform_machine == 'x86_64' or platform_machine == 'AMD64')",
+    "gs-madrona>=0.0.4; platform_system == 'Linux' and (platform_machine == 'x86_64' or platform_machine == 'AMD64')",
```
duburcqa: use `gs-madrona==0.0.4`. We don't know whether we will introduce breaking changes in the future.

PR #1568, opened 2025-08-15
[Fix] Minor fix for example; add requirement for release.md in PR template

YilingQiao:
## Related Issue

## Checklist:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -9,6 +9,7 @@
     - [BUG FIX] for non-breaking changes which fix an issue
     - [FEATURE] for non-breaking changes which add functionality
     - [MISC] for minor changes such as improved inline documentation or fixing typos
+    - [BEHAVIOR CHANGE] for modifications that will change the behavior (usually silently)
     - [BREAKING] **in addition to the above** for breaking changes, i.e., a fix or feature that would cause existing APIs or functionality to change
 -->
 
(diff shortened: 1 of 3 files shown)


.github/pull_request_template.md
```diff
@@ -40,6 +41,7 @@ see how your change affects other areas of the code, etc. -->
 - [ ] I read the **CONTRIBUTING** document.
 - [ ] I followed the `Submitting Code Changes` section of **CONTRIBUTING** document.
 - [ ] I tagged the title correctly (including BUG FIX/FEATURE/MISC/BREAKING)
+- [ ] I add ONE line to RELEASE_NOTES.md in the format: [breaking/behavior change/fix/feature] Short description
```
duburcqa: Why are we not following the same convention for PR tags? We should mention whether there may be multiple tags. I would prefer a deterministic number of tags instead of something that may change, but I'm fine with enforcing either 1 or 2 tags systematically.

May be we should take a bot more time to think about?

YilingQiao: Yes, we can follow the same convention for PR tags. Maybe we should add behavior change to PR tags as well? 

```
enforcing either 1 or 2 tags systematically.
```
But one commit can be feature/behavior change/fix/breaking at the same time? I'm not sure it makes much sense to have a priority over them.

Issue #1601, opened 2025-08-19
[Bug]: Robot dosent collide with the ground falls down even with gravity disabled


vybhav-ibr:
### Bug Description

I tried to load the apptronik appolo from an MJCF file and the robot doesn't collide with the ground, and falls down indefinitely even with gravity disabled.

### Steps to Reproduce

```
import genesis as gs

gs.init(logging_level="debug")

scene = gs.Scene(
    show_viewer=True,
    viewer_options=gs.options.ViewerOptions(
        res=(1280, 960),
        camera_pos=(3.5, 0.0, 2.5),
        camera_lookat=(0.0, 0.0, 0.5),
        camera_fov=40,
        max_FPS=60,
    ),
    vis_options=gs.options.VisOptions(
        show_world_frame=True,
        world_frame_size=1.0,
        show_link_frame=False,
        show_cameras=False,
        plane_reflection=True,
        ambient_light=(0.1, 0.1, 0.1),
    ),
    rigid_options=gs.options.RigidOptions(
    enable_joint_limit=False,
    enable_collision=False,
    gravity=(0, 0, 0),
),
    # renderer=gs.renderers.RayTracer(),
    renderer=gs.renderers.Rasterizer(),
)

plane = scene.add_entity(
    gs.morphs.Plane(),
)
applolo = scene.add_entity(
    gs.morphs.MJCF(file="/apptronik_appolo/apptronik_apollo.xml",
                   pos=(0,0,5),
                   collision=True),
)

cam = scene.add_camera(
    res=(640, 480),
    pos=(3.5, 0.0, 2.5),
    lookat=(0, 0, 0.5),
    fov=30,
    GUI=True,
)

scene.build()

cam.start_recording()
import numpy as np

for i in range(1200):
    scene.step()
    cam.render()
cam.stop_recording(save_to_filename="video.mp4", fps=60)


```MJCF file:
[https://github.com/google-deepmind/mujoco_menagerie/tree/main/apptronik_apollo](url)

### Expected Behavior

The robot must ragdoll on the ground

### Screenshots/Videos

_No response_

### Relevant log output

```shell
[Genesis] [20:02:33] [INFO] ╭─────────────────────────────────────────────────────────────────────────────────────╮
[...]

PR #1613, opened 2025-08-21
[BUG FIX] Clean up gravity interface.

Kashu7100:
## Description

- Removed solver specific gravity settings to make sure everything is consistent with sim (i.e., `sim.gravity` should be equal to `solver._gravity`)
- `gravity` option in each SolverOptions are now disabled. 

## Related Issue

Resolves Genesis-Embodied-AI/Genesis#1601

## Checklist:

### DIFF
--- a/tests/test_hybrid.py
+++ b/tests/test_hybrid.py
@@ -14,15 +14,14 @@ def test_rigid_mpm_muscle(show_viewer):
         sim_options=gs.options.SimOptions(
             dt=3e-3,
             substeps=10,
+            gravity=(0, 0, -9.8),
         ),
         rigid_options=gs.options.RigidOptions(
-            gravity=(0, 0, -9.8),
             constraint_timeconst=0.02,
         ),
         mpm_options=gs.options.MPMOptions(
             lower_bound=(0.0, 0.0, -0.2),
             upper_bound=(1.0, 1.0, 1.0),
-            gravity=(0.0, 0.0, 0.0),  # mimic gravity compensation
             enable_CPIC=True,
         ),
         viewer_options=gs.options.ViewerOptions(
--- a/genesis/sensors/imu.py
+++ b/genesis/sensors/imu.py
@@ -129,7 +129,7 @@ class IMU(Sensor):
         """
         Update the current ground truth values for all IMU sensors.
         """
-        gravity = shared_metadata.solver.get_gravity()
+        gravity = shared_metadata.solver.sim.get_gravity()
         quats = shared_metadata.solver.get_links_quat(links_idx=shared_metadata.links_idx)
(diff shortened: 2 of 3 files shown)


tests/test_hybrid.py
```diff
@@ -14,15 +14,14 @@ def test_rigid_mpm_muscle(show_viewer):
...
+            gravity=(0, 0, -9.8),
         ),
         rigid_options=gs.options.RigidOptions(
-            gravity=(0, 0, -9.8),
             constraint_timeconst=0.02,
         ),
         mpm_options=gs.options.MPMOptions(
             lower_bound=(0.0, 0.0, -0.2),
             upper_bound=(1.0, 1.0, 1.0),
-            gravity=(0.0, 0.0, 0.0),  # mimic gravity compensation
```
YilingQiao, line 25: sometimes users might want to have gravity compensation for one solver. How does the current implementation handle it?

Kashu7100: Actually MPMSolver is not using the gravity so this has no effect actually.
For the gravity compensation in general, we cannot do solver-wise gravity compensation at the moment.

genesis/engine/simulator.py
```diff
@@ -410,9 +418,44 @@ def get_state(self):
...
+        if gravity.ndim == 1:
+            if isinstance(gravity, torch.Tensor):
+                gravity = gravity.repeat(len(envs_idx), 1)
+            else:
+                gravity = np.tile(gravity, (len(envs_idx), 1))
+        assert gravity.shape == (len(envs_idx), 3), "Input gravity array should match (len(envs_idx), 3)"
+        self._kernel_set_gravity(gravity, envs_idx)
+
+        if self.rigid_solver.is_active():
+            if hasattr(self.rigid_solver, "_rigid_global_info"):
```
duburcqa, line 447: This is not a good approach. You should rather use `_is_built` flag. If it does not work for some reason, let's think to another solution. Note that `self.rigid_solver` should always have `_rigid_global_info` attribute. But it may be set to `None` at first. If so, you could just check whether it has been initialised or not.

genesis/options/solvers.py
```diff
@@ -213,7 +213,7 @@ class RigidOptions(Options):
     dt : float, optional
         Time duration for each simulation step in seconds. If none, it will inherit from `SimOptions`. Defaults to None.
     gravity : tuple, optional
-        Gravity force in N/kg. If none, it will inherit from `SimOptions`. Defaults to None.
+        Deprecated.
```
duburcqa, line 216: Add more details to explain what the user is supposed to do instead.

duburcqa:
Closing for now as there is no progress and several conflicts now.

duburcqa:
@Kashu7100 Feel free to reopen once to did what is necessary to get this PR to the finish line. I would be glad to review if again and ultimately merge it.

PR #1631, opened 2025-08-25
[FEATURE] Add initial experimental support of gstaichi fast cache feature.

hughperkins:


hughperkins:
outptu wiht `GS_BETA_PURE=1 GS_USE_NDARRAY=1 TI_LOG_LEVEL=trace python examples/rigid/single_franka_envs.py -n 2 --horizon 2 -c` and with taichi hacked to print non-pure kernels:
```
[Genesis] [17:11:23] [INFO] Compiling simulation kernels...
not pure fill_ndarray
[Genesis] [17:11:24] [INFO] Building visualizer...
```

genesis/engine/solvers/rigid/constraint_solver_decomp.py
```diff
@@ -9,6 +9,7 @@
 import genesis.utils.geom as gu
 import genesis.utils.array_class as array_class
 import genesis.engine.solvers.rigid.rigid_solver_decomp as rigid_solver
+from .... import maybe_pure
```
duburcqa: I would prefer using `gs.maybe_pure` everywhere if you are fine with it.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -20,6 +20,7 @@
 from genesis.utils.misc import ti_field_to_torch, DeprecationError, ALLOCATE_TENSOR_WARNING
 
 from ....utils.sdf_decomp import SDF
+from .... import maybe_pure
```
duburcqa: same.

genesis/engine/solvers/rigid/constraint_solver_decomp.py
hughperkins: Interesting. Seems that works. Updated 🙌

duburcqa (dismissed):

genesis/__init__.py
duburcqa, line 1: As discussed in DM, I will move the reloading logic in unit test, because it is too fragile and only relevant in this context.

PR #1638, opened 2025-08-27
[BUG FIX, CHANGING] Fix primitive and mesh's COM. Add 'ref' optional argument to 'get_links_pos'.

YilingQiao:
## Description

* Fix primitive and mesh's COM
* Change low-level COM's name from `COM` to `root_COM` for clarity
* Add 'ref' option for 'get_links_pos' matching 'get_links_vel' public API.

Here are the simulation videos before and after the change

https://github.com/user-attachments/assets/eb011e6c-beb9-485a-90f3-5e33fa49d281

https://github.com/user-attachments/assets/a30773f1-92cb-4fa6-98a3-025b9e9cca63

## Motivation and Context

- The initial COM for primitive and meshs are (0,0,0). We should set them to None and compute the inertial_pos based on their geometry
- The COM variable is used to represent root_COM, which is confusing. 

## Checklist:

### DIFF
--- a/genesis/engine/solvers/rigid/rigid_solver_decomp.py
+++ b/genesis/engine/solvers/rigid/rigid_solver_decomp.py
@@ -2061,8 +2061,36 @@ class RigidSolver(Solver):
             tensor = ti_field_to_torch(self.geoms_info.sol_params, geoms_idx, transpose=True, unsafe=unsafe)
         return tensor
 
-    def get_links_pos(self, links_idx=None, envs_idx=None, *, unsafe=False):
-        tensor = ti_field_to_torch(self.links_state.pos, envs_idx, links_idx, transpose=True, unsafe=unsafe)
+    @staticmethod
+    def _convert_ref_to_idx(ref: Literal["link_origin", "link_com", "root_com"]):
+        if ref == "root_com":
+            return 0
+        elif ref == "link_com":
+            return 1
+        elif ref == "link_origin":
+            return 2
+        else:
+            raise gs.raise_exception("'ref' must be either 'link_origin', 'link_com', or 'root_com'.")
+
+    def get_links_pos(
+        self,
+        links_idx=None,
+        envs_idx=None,
+        *,
+        ref: Literal["link_origin", "link_com", "root_com"] = "link_origin",
+        unsafe: bool = False,
+    ):
+        ref = self._convert_ref_to_idx(ref)
+        if ref == 0:
+            tensor = ti_field_to_torch(self.links_state.root_COM, envs_idx, links_idx, transpose=True, unsafe=unsafe)
+        elif ref == 1:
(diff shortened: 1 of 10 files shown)


duburcqa (dismissed):

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -4253,7 +4272,8 @@ def func_COM_links(
 
             i_r = links_info.root_idx[I_l]
             links_state.mass_sum[i_r, i_b] += mass
-            links_state.COM[i_r, i_b] += mass * links_state.i_pos[i_l, i_b]
+            links_state.link_COM[i_l, i_b] = links_state.i_pos[i_l, i_b]
```
duburcqa: I don't think storing `link_COM` is useful because it is basically `i_pos`. Having multiple aliases for the same quantity is a bad practice and not memory friendly.

YilingQiao: i_pos is modified later and no longer link_COM
```
links_state.i_pos[i_l, i_b] = links_state.i_pos[i_l, i_b] - links_state.root_COM[i_l, i_b]
```

YilingQiao: done

tests/test_rigid_physics.py
```diff
@@ -1686,8 +1691,9 @@ def test_mesh_repair(convexify, show_viewer, gjk_collision):
         assert all(geom.metadata["decomposed"] for geom in obj.geoms)
 
     # MPR collision detection is significantly less reliable than SDF in terms of penetration depth estimation.
-    tol_pos = 0.05 if convexify else 1e-6
-    tol_rot = 1.3 if convexify else 1e-4
+    # previous the COM is (0,0,0), closer to the plane, so more stable
+    tol_pos = 0.05 if convexify or not gjk_collision else 1e-3
+    tol_rot = 1.3  # if convexify or not gjk_collision else 1e-4
```
duburcqa: Why did you have to change the tolerance so much? It looks wrong.

duburcqa: After investigation, the right values are:
```python
    is_mpr = convexify and not gjk_collision
    tol_pos = 0.05 if is_mpr else 0.005
    tol_rot = 1.0 if is_mpr else 0.25
```
Even if the tolerance could be much higher by tweaking a bit initial condition, it would not be representative of the actually "stability" of the simulation. The tolerance must be high to cope with contact points appearing and disappearing sporadically. Hopefully this issue could be avoided once differentiable contact would be implemented @SonSang

SonSang: Yes, I'll test differentiable contact with this case later.

PR #1639, opened 2025-08-27
[FEATURE] Use FIELD_METADATA_CACHE_VALUE for static config

YilingQiao:
## Related Issue

## Checklist:


genesis/utils/array_class.py
```diff
@@ -2004,6 +2005,70 @@ def __init__(self):
         return ClassEntitiesState()
 
 
+# =========================================== StaticRigidSimConfig ===========================================
+
+
+@dataclasses.dataclass
+class StaticRigidSimDummy:
+    para_level: int = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
```
hughperkins: I can see why you're going to ask me to make it possible to set `FIELD_METADATA_CACHE_VALUE` conceptually on an entire dataclass at some point :)

genesis/utils/array_class.py
```diff
@@ -2004,6 +2005,70 @@ def __init__(self):
...
+    # dynamic properties
+    substep_dt: float = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
+    iterations: int = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
+    tolerance: float = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
+    ls_iterations: int = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
+    ls_tolerance: float = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
+    n_equalities: int = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
+    n_equalities_candidate: int = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
+    hibernation_thresh_acc: float = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
+    hibernation_thresh_vel: float = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
```
hughperkins: There was no request ot support floats, only ints and bools. Whilst it seems plausible tht it might work with floats, note that this is not tested. Please consider creating a PR onto gstaichi to add a test for floats. Something like https://github.com/Genesis-Embodied-AI/gstaichi/blob/a00b3aa5154b4b983df0c181cba3d60e20d190a8/tests/python/gstaichi/lang/fast_caching/test_args_hasher.py#L245 . (Yes, I agree that it also doesn't test bools, but floats seem more concerning to me, because of the potential for rounding issues).

YilingQiao: sure, technically, float doesn't need to be `FIELD_METADATA_CACHE_VALUE` since they are not used in ti.static. I will move those float values out of this class.

genesis/utils/array_class.py
hughperkins: If it was me, I think I might be tempted to try something like either:
```
cache_value = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})

@dataclasses.dataclass
class StaticRigidSimDummy:
    para_level: int = cache_value
    ...
```
Or, if that doesn't work, something like:
```
from functools import partial
cache_value = partial(dataclasses.field, metadata={FIELD_METADATA_CACHE_VALUE: True})

@dataclasses.dataclass
class StaticRigidSimDummy:
    para_level: int = cache_value()
    ...

```

I don't guarantee that this will work, nor that it will pass code review with other people. But just an idea to make the code more compact.

hughperkins:
```
In [1]: from functools import partial

In [2]: import dataclasses

In [3]: cache_value = partial(dataclasses.field, metadata={"foo": True})

In [4]: cache_value
Out[4]: functools.partial(<function field at 0x1031395a0>, metadata={'foo': True})

In [5]: cache_value()
Out[5]: Field(name=None,type=None,default=<dataclasses._MISSING_TYPE object at 0x103126bf0>,default_factory=<dataclasses._MISSING_TYPE object at 0x103126bf0>,init=True,repr=True,hash=None,compare=True,metadata=mappingproxy({'foo': True}),kw_only=<dataclasses._MISSING_TYPE object at 0x103126bf0>,_field_type=None)

In [7]: cache_value().metadata
Out[7]: mappingproxy({'foo': True})
```
Seems like the partial approach might potentailly work.

hughperkins:
Unit tests?

genesis/utils/array_class.py
YilingQiao: removed them

hughperkins:
(Not claiming it's feasible to write unit tests for this change, just posing the question)

genesis/utils/array_class.py
YilingQiao: I don't fully get it. Why is it better than the current implementation?

YilingQiao:
Its unit test have to run with a gstaichi that properly supports GS_BETA_PURE=1. Maybe after Alexis' patch and the new gstaichi wheel

genesis/utils/array_class.py
hughperkins: More compact. Compare:
```
@dataclasses.dataclass
class StaticRigidSimDummy:
    para_level: int = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
    use_hibernation: bool = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
    use_contact_island: bool = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
    batch_links_info: bool = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
    batch_dofs_info: bool = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
    batch_joints_info: bool = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
    enable_mujoco_compatibility: bool = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
    enable_multi_contact: bool = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
    enable_self_collision: bool = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
    enable_adjacent_collision: bool = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
    enable_collision: bool = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
    box_box_detection: bool = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
    integrator: int = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
    sparse_solve: bool = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
    solver_type: int = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
```
vs:
```
cache_value = partial(dataclasses.field, metadata={FIELD_METADATA_CACHE_VALUE: True})

@dataclasses.dataclass
class StaticRigidSimDummy:
    para_level: int = cache_value()
    use_hibernation: bool = cache_value()
    use_contact_island: bool = cache_value()
    batch_links_info: bool = cache_value()
    batch_dofs_info: bool = cache_value()
    batch_joints_info: bool = cache_value()
    enable_mujoco_compatibility: bool = cache_value()
    enable_multi_contact: bool = cache_value()
    enable_self_collision: bool = cache_value()
    enable_adjacent_collision: bool = cache_value()
    enable_collision: bool = cache_value()
    box_box_detection: bool = cache_value()
    integrator: int = cache_value()
    sparse_solve: bool = cache_value()
    solver_type: int = cache_value()
```

Not claiming it's definitely an improvement. Just an idea.

hughperkins: (I imagine @duburcqa probably has opinions on this. I cannot predict which direction his opinion will lean towards)

hughperkins:
> Its unit test have to run with a gstaichi that properly supports GS_BETA_PURE=1. Maybe after Alexis' patch and the new gstaichi wheel

GS_BETA_PURE is a Genesis concept, not a gstaichi concept. [Edit: Oh I see, a gstaichi that doesnt break when using GS_BETA_PURE :) ]

You can probably merge alexis branch https://github.com/Genesis-Embodied-AI/Genesis/pull/1645/files onto your branch, and then you will have the latest gstaichi. ~I don't rmemeber hte status of GS_BETA_PURE 🤔 Didnt we merge that already? This PR I think?~ ~github.com/Genesis-Embodied-AI/Genesis/pull/1631/files~

hughperkins:
(Note: I realize now that the latest gstaichi wheel hadn't finished building yet, but is has now :) https://pypi.org/project/gstaichi/2.1.1/#files )

YilingQiao:
I prefer merging the current PR as it is. This PR introduces the dummy variable for static_config, utilizes keyword FIELD_METADATA_CACHE_VALUE, and make sure it won't break any existing tests.

I’d prefer to systematically test "no recompilation with ndarray" and "fast cache loading" in another PR. (i.e., a PR that turns GS_USE_NDARRAY and GS_BETA_PURE from experimental environment variables into actual options in Genesis). 

Writing unit tests for cache loading is important but tricky, since these tests aren’t single-run and we need to cover many corner cases. I don't have a clear idea how to write those tests. Do you have suggestions on how to write multi-run tests and test the correctness of cache loading? @hughperkins.

hughperkins:
> I prefer merging the current PR as it is, @hughperkins.
> 
> Writing unit tests to check for "no recompilation with ndarray" and "fast cache loading" isn’t trivial, and we should come up with a proper way to test it (since these tests aren’t single-run but require several rapid runs). Do you have suggestions on how to test them? I’d prefer to handle this in another PR (e.g., a PR that turns GS_USE_NDARRAY and GS_BETA_PURE from experimental environment variables into actual options in Genesis).

In the near-term, @duburcqa was suggesting that one approach could be to run the tests multiple times, one for each combination of env vars.
- we could filter which tests we run each time, so we don't have to run the entire test set 4 times

Generally I prefer to have the tests in the same PR that needs the tests. One simple reason for this is that personally I find it much easier to review tests than to review code.

Conceptually, I'm hoping to see some tests that check that for at least 1 or 2 of the values you want to ensure change things ,that things really do change. conceptually somethign like:
```
pytest.mark.parametrize(
    "my_param,expected_value", [
         (False, 3),
          (True, 5),
    ]
)
def my_test_func(my_param: bool, expected_value: int) -> None:
    static_params = StaticParams(.... , my_param=my_param, ...)
    launch_kernel_stuff(..., static_params, ....)
    assert some_result.value == expected_value
```

Edit: note that I also added some instrumentation so you can check if things were loaded from src-ll-cache or not. See eg https://github.com/Genesis-Embodied-AI/gstaichi/blob/ce32cd7e1b636a4c414c0af775dcb715f2c747a5/tests/python/gstaichi/lang/fast_caching/test_src_ll_cache.py#L31-L34

(the instruemntation is not ideal, since requires accessing a private variable (ie `_primal`), but at least this is only for test code, not for prod code. But, yes, probably should ponder a way to make this accessible without going through private variables)

genesis/utils/array_class.py
YilingQiao: done

genesis/utils/array_class.py
```diff
@@ -2004,6 +2005,51 @@ def __init__(self):
         return ClassEntitiesState()
 
 
+# =========================================== StaticRigidSimConfig ===========================================
+
+cache_value = partial(dataclasses.field, metadata={FIELD_METADATA_CACHE_VALUE: True})
+
+
+@dataclasses.dataclass
+class StaticRigidSimDummy:
```
duburcqa: Why do we have both `StaticRigidSimConfig` and `StaticRigidSimDummy`. We should get rid of the former?

Besides, the name is not quite right. `StaticRigidSimDummy` does not mean anything. What is this class all about? If it is used for kernel caching disambiguation only, then it should rather be called `RigidSimMetadata`, or `RigidSimMetadataCache`. But `dummy` here is just misleading I think.

duburcqa: After thinking twice, I think you add to come up with this design because the caching mechanism currently is not compatible with `ti.data_oriented`. So it is fine. But nonetheless the name should be updated.

duburcqa:
> I prefer merging the current PR as it is. This PR introduces the dummy variable for static_config, utilizes keyword FIELD_METADATA_CACHE_VALUE, and make sure it won't break any existing tests. I’d prefer to systematically test "no recompilation with ndarray" and "fast cache loading" in another PR. (i.e., a PR that turns GS_USE_NDARRAY and GS_BETA_PURE from experimental environment variables into actual options in Genesis).

Why? I don't understand. It is only acceptable to merge without unit tests when we are under pressure. I thought it was not the case at the moment. If we are not, then I'm strongly in favour of adding the unit tests in this PR since the plan is to add them right after already.

duburcqa:
What about this unit test?
```python
import genesis as gs

gs.init(backend=gs.cpu, precision="32")

scenes = []
for enable_multi_contact in (False, True):
    scene = gs.Scene(
            camera_pos=(1, 1, 0.5),
            camera_lookat=(0.0, 0.0, 0.0),
        ),
            enable_multi_contact=enable_multi_contact,
        ),
        show_viewer=False,
    )
    plane = scene.add_entity(
        gs.morphs.Plane(),
    )
    cube = scene.add_entity(
        gs.morphs.Box(
            size=(0.4, 0.4, 0.4),
            pos=(0.0, 0.0, 0.18),
        )
    )
    scene.build()
    scenes.append(scene)

for scene in scenes:
    scene.rigid_solver.collider.detection()
    num_contacts = 4 if scene.rigid_options.enable_multi_contact else 1
    assert scene.rigid_solver.collider._collider_state.n_contacts.to_numpy() == num_contacts
```

hughperkins:
~@YilingQiao it looks like `enable_multi_contact` is not used in a `ti.static` conditional. What is your criteria for deciding which configuration values to flag as being part of the src-ll-cache key?~

hughperkins:
~(If the answer is "defensive programming, because it's hard to tell which parameters are in fact being used in ti.static", that's ok; just want to confirm whether this is in fact the reason?)~

hughperkins:
Note: here are all the ti.statics that I find: https://gist.github.com/hughperkins/ae70cec901e87d0602ec34ae489b1100 Not sure if I'm missing some other static conditionals?

duburcqa:
It is used in `ti.static`!

genesis/utils/array_class.py
YilingQiao: yeah, it's a duplicate for the information in `StaticRigidSimConfig`, only for cache mechanism. Maybe change its name to something like `StaticRigidSimConfigCache`?

YilingQiao: renamed to StaticRigidSimCacheKey/static_rigid_sim_cache_key

hughperkins: Great! 🙌  Much prefer `CacheKey` over simply `Cache`.

duburcqa (approved):

genesis/utils/array_class.py
duburcqa: Yes, whatever but not dummy :)

duburcqa (approved):

PR #1657, opened 2025-08-29
[MISC] Expose parameters for ground plane tiling.

yuhongyi:
## Description

Plane now can be created with specified plane size and tile size,default to 1000.0 and 1 respectively. This keeps the default behavior of plane creation unchanged.
This PR removed the hard-coded plane size and tiling from Genesis.

## Motivation and Context

This is useful when exporting the Genesis scene correctly without knowing the hidden properties of a plane primitive.
The background behind this change is that, to be able to render Genesis scene in the Apollo renderer, the Genesis scene needs to be serialized and exported to a scene description for Apollo to load.  All rendering related attributes of the plane need to be exported, thus all hidden properties need to be removed or made into parameters.

Tested with different plane sizes and tile sizes in Genesis

**plane_size**
The size of the plane geometry

**tile_size**
The size of each texture tile (not the checker tile, since the plane can be rendered with any texture)

plane_size=(3,3), tile_size=(1.5,1.5)
image

plane_size=(3,3), tile_size=(0.75,0.75)
image

## Checklist:


doc
duburcqa, line 1: Revert doc update.

genesis/options/morphs.py
```diff
@@ -429,10 +429,16 @@ class Plane(Primitive):
     conaffinity : int, optional
         The 32-bit integer bitmasks used for contact filtering of contact pairs. When the conaffinity of one geom and
         the contype of the other geom share a common bit set to 1, two geoms can collide. Defaults to 0xFFFF.
+    size: tuple, optional
+        The size of the plane in meters. Defaults to (1e3, 1e3).
+    n_tiles: tuple, optional
+        The number of tiles in the plane. If not specified, it will be set to be the same as `size`.
```
duburcqa: `tile_size` would make more sense?

genesis/options/morphs.py
```diff
@@ -429,10 +429,16 @@ class Plane(Primitive):
     conaffinity : int, optional
         The 32-bit integer bitmasks used for contact filtering of contact pairs. When the conaffinity of one geom and
         the contype of the other geom share a common bit set to 1, two geoms can collide. Defaults to 0xFFFF.
+    size: tuple, optional
+        The size of the plane in meters. Defaults to (1e3, 1e3).
```
duburcqa: `plane_size` may be? `size` is not very clear.

genesis/utils/mesh.py
```diff
@@ -889,21 +889,21 @@ def create_box(extents=None, color=(1.0, 1.0, 1.0, 1.0), bounds=None, wireframe=
...
-def create_plane(size=1e3, color=None, normal=(0.0, 0.0, 1.0)):
+def create_plane(normal=(0.0, 0.0, 1.0), size=(1e3, 1e3), n_tiles=(1e3, 1e3), color=None):
     thickness = 1e-2  # for safety
-    mesh = trimesh.creation.box(extents=[size, size, thickness])
+    mesh = trimesh.creation.box(extents=[size[0], size[1], thickness])
     mesh.vertices[:, 2] -= thickness / 2
     mesh.vertices = gu.transform_by_R(mesh.vertices, gu.z_up_to_R(np.asarray(normal, dtype=np.float32)))
 
-    half = size * 0.5
+    half = (size[0] * 0.5, size[1] * 0.5)
```
duburcqa: extract `half_x, half_y = (size[0] * 0.5, size[1] * 0.5)`

genesis/utils/mesh.py
```diff
@@ -916,11 +916,11 @@ def create_plane(size=1e3, color=None, normal=(0.0, 0.0, 1.0)):
...
                     [0, 0],
-                    [size, 0],
-                    [size, size],
+                    [n_tiles[0], 0],
+                    [n_tiles[0], n_tiles[1]],
                     [0, 0],
-                    [size, size],
-                    [0, size],
+                    [n_tiles[0], n_tiles[1]],
+                    [0, n_tiles[1]],
```
duburcqa: Same. I would rather export `x` and `y` components. More readable I think.

tests/test_render.py
```diff
@@ -777,3 +777,55 @@ def on_key_press(self, symbol: int, modifiers: int):
     # Make sure that the result is valid
     with open(IMAGE_FILENAME, "rb") as f:
         assert f.read() == png_snapshot
+
+
+@pytest.mark.required
```
duburcqa: Do not mark this test as required, it is too long:
image

tests/test_render.py
```diff
@@ -777,3 +777,55 @@ def on_key_press(self, symbol: int, modifiers: int):
...
+    [RENDERER_TYPE.BATCHRENDER_RASTERIZER],
+)
+def test_render_planes(tmp_path, png_snapshot, renderer):
+    # Manually run this test with different parameters to make sure they run sequentially
+    # Running in parallel may crash the test OOM.
+    sizes = [(1.5, 1.8), (10.0, 7.0)]
+    n_tiles = [(0.5, 0.75), (5.0, 3.0), None]
+    for test_idx, (size, n_tiles) in enumerate(itertools.product(sizes, n_tiles)):
+        CAM_RES = (512, 512)
+        NUM_STEPS = 1
```
duburcqa: Remove this parameter.

tests/test_render.py
```diff
@@ -777,3 +777,55 @@ def on_key_press(self, symbol: int, modifiers: int):
...
+
+        # Simulate
+        for i in range(NUM_STEPS):
+            # Move forward step forward in time
+            scene.step()
+            rgba, depth, _, _ = camera.render(rgb=True, depth=True)
+            exporter.export_frame_single_camera(i, test_idx, rgb=rgba, depth=depth)
+
+    # Verify that the output is correct pixel-wise over multiple simulation steps
+    for image_file in sorted(tmp_path.rglob("*.png")):
```
duburcqa: Remove all the comments. This test is simple. No need to explain what is going on.

tests/test_render.py
```diff
@@ -777,3 +777,55 @@ def on_key_press(self, symbol: int, modifiers: int):
...
+
+        # Create image exporter
+        exporter = FrameImageExporter(tmp_path)
+
+        # Simulate
+        for i in range(NUM_STEPS):
+            # Move forward step forward in time
+            scene.step()
+            rgba, depth, _, _ = camera.render(rgb=True, depth=True)
+            exporter.export_frame_single_camera(i, test_idx, rgb=rgba, depth=depth)
```
duburcqa: Remove for loop and do not call `scene.step` at all. It is not necessary and useful in this context.

tests/test_render.py
```diff
@@ -777,3 +777,55 @@ def on_key_press(self, symbol: int, modifiers: int):
     # Make sure that the result is valid
     with open(IMAGE_FILENAME, "rb") as f:
         assert f.read() == png_snapshot
+
+
+@pytest.mark.required
+@pytest.mark.parametrize(
+    "renderer_type",
+    [RENDERER_TYPE.BATCHRENDER_RASTERIZER],
```
duburcqa: Use `RENDERER_TYPE.RASTERIZER` if possible, because it works on all platforms unlike `RENDERER_TYPE.BATCHRENDER_*`

tests/test_render.py
```diff
@@ -777,3 +777,55 @@ def on_key_press(self, symbol: int, modifiers: int):
...
+)
+def test_render_planes(tmp_path, png_snapshot, renderer):
+    # Manually run this test with different parameters to make sure they run sequentially
+    # Running in parallel may crash the test OOM.
+    sizes = [(1.5, 1.8), (10.0, 7.0)]
+    n_tiles = [(0.5, 0.75), (5.0, 3.0), None]
+    for test_idx, (size, n_tiles) in enumerate(itertools.product(sizes, n_tiles)):
```
duburcqa: Are you sure you need a full grid? What about:
```
for test_idx, (size, n_tiles) in enumerate((
    ((1.5, 1.8), (0.5, 0.75)),
    ((10.0, 7.0), (5.0, 3.0)),
    ((10.0, 7.0), None),
)):
```

tests/test_render.py
```diff
@@ -777,3 +777,55 @@ def on_key_press(self, symbol: int, modifiers: int):
...
+        plane = scene.add_entity(
+            gs.morphs.Plane(size=size, n_tiles=n_tiles),
+        )
+        camera = scene.add_camera(
+            res=CAM_RES,
+            pos=(1.5, 0.5, 1.5),
+            lookat=(0.0, 0.0, 0.5),
+            fov=45,
+            GUI=False,
+        )
```
duburcqa: What about using a "perfect" top view camera angle? It would make more sense. It would even be possible to check if tiling it right based on the RGB data.

tests/test_render.py
```diff
@@ -777,3 +777,55 @@ def on_key_press(self, symbol: int, modifiers: int):
...
+    "renderer_type",
+    [RENDERER_TYPE.BATCHRENDER_RASTERIZER],
+)
+def test_render_planes(tmp_path, png_snapshot, renderer):
+    # Manually run this test with different parameters to make sure they run sequentially
+    # Running in parallel may crash the test OOM.
+    sizes = [(1.5, 1.8), (10.0, 7.0)]
+    n_tiles = [(0.5, 0.75), (5.0, 3.0), None]
+    for test_idx, (size, n_tiles) in enumerate(itertools.product(sizes, n_tiles)):
+        CAM_RES = (512, 512)
```
duburcqa: `CAM_RES = (256, 256)` would be faster to render and good enough for this.

tests/test_render.py
duburcqa, line 814: This is not good enough I think. You should ideally check that the pattern is correct, or at least that all the images are different. (this is done in advanced render api test).

duburcqa (changes requested):

doc
yuhongyi: Yes, I'm trying to remove the update to doc and LuisaRender

genesis/options/morphs.py
yuhongyi: tile_size will be confusing, since n_tiles here means number of tiles.
tile_size == plane_size / n_tiles

tests/test_render.py
yuhongyi: With a top-down shot, it should be easy to tell whether the patterns in the snapshots are correct.

genesis/options/morphs.py
yuhongyi: Update

genesis/utils/mesh.py
yuhongyi: Update

genesis/utils/mesh.py
yuhongyi: Updated

tests/test_render.py
yuhongyi: Removed required

tests/test_render.py
yuhongyi: Remove. One step now

tests/test_render.py
yuhongyi: Removed comments

tests/test_render.py
yuhongyi: Removed for-loop. Calling step() once.

tests/test_render.py
yuhongyi: Should be enough.  Updated

tests/test_render.py
yuhongyi: Updated to 256

tests/test_render.py
yuhongyi: Good idea.  I'll also remove depth output, since it's not useful for this test.

tests/test_render.py
yuhongyi: Updated

tests/test_render.py
yuhongyi: Now patterns can be verified with top-down shots

genesis/options/morphs.py
yuhongyi: Renamed it to texrepeat, which is the same attribute name used by Mujuco.

tests/test_render.py
```diff
@@ -777,3 +777,42 @@ def on_key_press(self, symbol: int, modifiers: int):
...
+            pos=(0.0, 0.0, 5),
+            lookat=(0.0, 0.0, 0.0),
+            fov=45,
+            GUI=False,
+        )
+        scene.build()
+
+        exporter = FrameImageExporter(tmp_path)
+
+        scene.step()
```
duburcqa: Remove step. It is useless.

yuhongyi: Removed

Issue #1685, opened 2025-09-05
[Bug]: Terrain not working when batching enabled.


ax-anoop:
### Bug Description

Supposedly related to: 

" The error occurs at line 445 in array_class.py:

  entity = solver._entities[solver.links_info.entity_idx.to_numpy()[links_idx[0]]]

  The issue is that when batch_links_info=True, the solver.links_info.entity_idx.to_numpy() returns a numpy array, and indexing it with links_idx[0] (which is likely also an array) causes the TypeError.
"
 - Claude 

but effectively I can't set terrains if 
        batch_dofs_info=True,
        batch_links_info=True,

```

### Steps to Reproduce

import genesis as gs
import torch
import numpy as np
import time
import sys

########################## pyramid terrain ##########################
def add_pyramid_terrain(scene: gs.Scene,
                        total_length: float = 8.0,
                        total_width: float = 2.0,
                        platform_width: float = 3.0,
                        step_width: float = 0.30,
                        step_height: float = 0.04,
                        center: tuple = (0.0, 0.0),
                        friction: float = 0.8):
    cx, cy = float(center[0]), float(center[1])
    half_len = total_length * 0.5
    side_len = max(0.0, half_len - platform_width * 0.5)
    num_steps = int(max(1, side_len // step_width))
    entities = []
    for s in range(1, num_steps + 1):
        h = s * step_height
        x_off = platform_width * 0.5 + (s - 0.5) * step_width
        size = (step_width, total_width, h)
        pos_p = (cx + x_off, cy, h * 0.5)
        pos_n = (cx - x_off, cy, h * 0.5)
        try:
            box_p = scene.add_entity(gs.morphs.Box(size=size, pos=pos_p))
            box_n = scene.add_entity(gs.morphs.Box(size=size, pos=pos_n))
            try:
                box_p.set_friction(friction)
                box_n.set_friction(friction)
            except Exception:
                pass
            entities.extend([box_p, box_n])
        except Exception:
            pass
    return entities

########################## build ##########################
def add_simple_terrain(scene):
    hf = np.zeros((40, 40), dtype=np.int16)
    hf[10:30, 10:30] = 200 * np.hanning(20)[:, None] * np.hanning(20)[None, :]

    horizontal_scale = 0.25  # metres between grid points
    vertical_scale   = 0.005  # metres per height-field unit

    scene.add_entity(
        morph=gs.morphs.Terrain(
            height_field=hf,
[...]

PR #1691, opened 2025-09-05
[BUG FIX] Fix entity idx to be always int when loading terrain with batch_links_info.

Kashu7100:
## Related Issue

Resolves Genesis-Embodied-AI/Genesis#1685

## Checklist:


YilingQiao:
Can you write a test? Everytime we resolved a bug, we need to add a test for it.

tests/test_rigid_benchmarks.py
```diff
@@ -684,3 +684,20 @@ def test_box_pyramid(factory_logger, request, n_cubes, solver, enable_island, n_
         }
     ) as logger:
         logger.write(request.getfixturevalue("box_pyramid"))
+
+
+@pytest.mark.parametrize("batch_links_info", [False, True])
```
duburcqa: * Why are you adding this in benchmarks?
* You should mark this test as required
* You should explicitly check for something, or document why it is not necessary
* Are you sure doing one step is necessary? Build is not enough?

Kashu7100: - to test if batching options (batch_links_info, batch_joints_info, batch_dofs_info) work correctly for terrains.
- yes
- this is now required
- step is not necessary so removed

YilingQiao: - "Why are you adding this in benchmarks? " → moved to test_rigid_physics
- "explicitly check for something" → yes, checked for shapes

PR #1701, opened 2025-09-10
[FEATURE] Differentiable contact detection, Part 1

SonSang:
## Description

Added the main differentiable contact detection algorithm and unit test based on the comparison to the finite difference. There are several issues that should be addressed in the following PRs, including:

1) Unstable Newton solver for `pyramid.py` example
2) Potentially incorrect & unstable backward pass (e.g. sometimes `nan` gradient arises in 32 precision)
3) Hyperparameter tuning for controlling the degree of differentiability

You can switch to the differentiable contact detection algorithm by setting `requires_grad=True` in `gs.options.SimOptions` as shown below. Also, this algorithm requires to use `GJK` contact detection algorithm. 

```
scene = gs.Scene(
            # Turn on differentiable contact
            requires_grad=True,
        ),
            box_box_detection=False,
        ),
)
```

## Related Issue

## Motivation and Context

Added a unit test for validating the backward pass. You can run the test as follows:

```
pytest "tests/test_diff_rigid.py::test_diff_contact"
```

## Checklist:


tests/test_diff_rigid.py
```diff
@@ -0,0 +1,164 @@
+import numpy as np
+import pytest
+import torch
+
+import genesis as gs
+from scipy.spatial.transform import Rotation as R
+
+
+@pytest.mark.precision("64")
+@pytest.mark.parametrize("backend", [gs.cpu])
```
duburcqa: Why this is test not marked as required? Is it too slow?

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,164 @@
+import numpy as np
+import pytest
+import torch
+
+import genesis as gs
+from scipy.spatial.transform import Rotation as R
```
duburcqa: Do not use scipy for this. We have all the helpers you need in `gs.utils.geoms`

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,164 @@
...
+
+    scene = gs.Scene(
+        sim_options=gs.options.SimOptions(
+            dt=0.01,
+            # Turn on differentiable mode
+            requires_grad=True,
+        ),
+        rigid_options=gs.options.RigidOptions(
+            box_box_detection=False,
+        ),
```
duburcqa: Why are you doing this? If box-box is non-differentiable, you should make the default "dynamic", in a way that it can take into account whether `requires_grad` is True or not. If `requires_grad=True`, the default should be False, otherwise True.

In practice, set the default value to None, and define its set value when creating the scene in `_validate_options` for instance. There has other options that behaves like this already.

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,164 @@
...
+        show_viewer=False,
```
duburcqa: Is there anything to see that could be interesting to debug this unit test in case of issue? If so, you should rather do `show_viewer=show_viewer` and pass the fixture parameter `show_viewer` as input argument of the unit test.

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,164 @@
...
+        ),
+    )
+
+    scene.build()
+    solver = scene.sim.rigid_solver
+    collider = solver.collider
+
+    # Set up initial configuration
+    x_ang = 0.03
+    y_ang = 0.03
```
duburcqa: I would suggest doing: `x_ang, y_ang = 0.03, 0.03` because it is conceptually a pair, just unpacked.

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,164 @@
...
+    box_pos_offset = (0.0, 0.0, 0.0) + 0.5 * box_size * vec_one
+
+    box0 = scene.add_entity(
+        gs.morphs.Box(size=box_size * vec_one, pos=box_pos_offset, fixed=True),
+    )
+    box1 = scene.add_entity(
+        gs.morphs.Box(
+            size=box_size * vec_one, pos=box_pos_offset + 0.8 * box_spacing * np.array([0, 0, 1]), fixed=True
+        ),
+    )
```
duburcqa: I would suggest to split all input arguments in a single line in this context, no matter the linewidth, because it greatly improve readability.

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,164 @@
...
+    # Set up initial configuration
+    x_ang = 0.03
+    y_ang = 0.03
+    box1.set_quat(R.from_euler("xy", [np.deg2rad(x_ang), np.deg2rad(y_ang)]).as_quat(scalar_first=True))
+
+    box0_init_pos = box0.get_pos()
+    box1_init_pos = box1.get_pos()
+    box0_init_quat = box0.get_quat()
+    box1_init_quat = box1.get_quat()
+
```
duburcqa: Do not skip line here, it is the same block.

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,164 @@
...
+    ### Compute the initial loss and compute gradients using differentiable contact detection
+    # Detect contact
+    collider.detection()
+
+    # Get contact outputs and their grads
+    contacts = collider.get_contacts(as_tensor=True, to_torch=True, keep_batch_dim=True)
+    normal = contacts["normal"].clone().requires_grad_()
+    position = contacts["position"].clone().requires_grad_()
+    penetration = contacts["penetration"].clone().requires_grad_()
```
duburcqa: No need to clone. It is already "a copy" by design.

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,164 @@
...
+
+    loss = ((normal * position).sum(dim=-1) * penetration).sum()
+    dL_dnormal = torch.autograd.grad(loss, normal, retain_graph=True)[0]
+    dL_dposition = torch.autograd.grad(loss, position, retain_graph=True)[0]
+    dL_dpenetration = torch.autograd.grad(loss, penetration)[0]
+
+    # Compute analytical gradients of the geoms position and quaternion
+    collider.backward(dL_dposition, dL_dnormal, dL_dpenetration)
+    dL_dpos = torch.from_numpy(solver.geoms_state.pos.grad.to_numpy())
+    dL_dquat = torch.from_numpy(solver.geoms_state.quat.grad.to_numpy())
```
duburcqa: You should rather use `ti_to_torch(solver.geoms_state.pos)`, which is designed exactly for this.

Another option would be to do `ti_to_torch(solver.geoms_state.pos.grad)`, but I don't like it as much because it exposes taichi internal gradient attribute.

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,164 @@
...
+    def compute_loss(pos, quat):
+        box0_pos = pos[0]
+        box1_pos = pos[1]
+        box0_quat = quat[0]
+        box1_quat = quat[1]
+
+        box0.set_pos(box0_pos)
+        box1.set_pos(box1_pos)
+        box0.set_quat(box0_quat)
+        box1.set_quat(box1_quat)
```
duburcqa: You are packing data for the two boxes, then unpacking them again. I'm not sure to see the point of packing them in the first place. You keep them as `box0_pos, box1_pos, box0_quat, box1_quat` all along.

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,164 @@
...
+
+        # Re-detect contact, we need to manually reset the contact counter as we are not running the whole sim step
+        collider._collider_state.n_contacts.fill(0)
+        collider._collider_state.n_diff_contacts.fill(0)
+        collider.detection()
+        contacts = collider.get_contacts(as_tensor=True, to_torch=True, keep_batch_dim=True)
+
```
duburcqa: No need to skip line here. Basically the same block.

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,164 @@
...
+        normal = contacts["normal"]
+        position = contacts["position"]
+        penetration = contacts["penetration"]
```
duburcqa: Extract values in one line:
```python
normal, position, penetration = contacts["normal"], contacts["position"], contacts["penetration"]
```

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,164 @@
...
+        num_contacts = len(normal[0])
+
+        if num_contacts != num_init_contacts:
+            raise ValueError(f"Number of contacts changed from {num_init_contacts} to {num_contacts}")
+
+        loss = ((normal * position).sum(dim=-1) * penetration).sum()
+
```
duburcqa: No need to skip line here.

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,164 @@
...
+        return loss
+
+    ### Compute directional derivatives along random directions
+    FD_EPS = 1e-5
+    TRIALS = 100
+
+    # dL_dpos
+    dL_dpos_error = 0
```
duburcqa: Use floats for floats.

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,164 @@
...
+        try:
+            # 1 * eps
+            input_pos = init_pos + rand_dpos.squeeze(1) * FD_EPS
+            lossP1 = compute_loss(input_pos, init_quat)
+
+            # -1 * eps
+            input_pos = init_pos - rand_dpos.squeeze(1) * FD_EPS
+            lossP2 = compute_loss(input_pos, init_quat)
+        except ValueError as e:
+            continue
```
duburcqa: Add a comment to explain why this exception can be safely ignored.

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,164 @@
...
+        dL = (rand_dpos * dL_dpos).sum()
+        try:
+            # 1 * eps
+            input_pos = init_pos + rand_dpos.squeeze(1) * FD_EPS
+            lossP1 = compute_loss(input_pos, init_quat)
+
+            # -1 * eps
+            input_pos = init_pos - rand_dpos.squeeze(1) * FD_EPS
+            lossP2 = compute_loss(input_pos, init_quat)
+        except ValueError as e:
```
duburcqa: Just do `except ValueError:` if you are not doing anything with the exception.

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,164 @@
...
+            input_quat = torch.nn.functional.normalize(input_quat, dim=-1)
+            lossP2 = compute_loss(init_pos, input_quat)
+        except ValueError as e:
+            continue
+        valid_trials += 1
+        dL_fd = (lossP1 - lossP2) / (2 * FD_EPS)
+        dL_error = (dL - dL_fd).abs() / max(dL.abs(), dL_fd.abs(), gs.EPS)
+        dL_dquat_error += dL_error
+    dL_dquat_error /= valid_trials
+    assert dL_dquat_error < rtol, f"Relative error is too large: {dL_dquat_error * 100.0:.4g}% < {rtol * 100.0:.4g}%"
```
duburcqa: There is too much duplication wrt the previous block. It would be great to factorize your test a bit if possible.

genesis/utils/array_class.py
```diff
@@ -292,6 +292,60 @@ def __init__(self):
...
+        "localpos1a": V_VEC(3, dtype=gs.ti_float, shape=(_B, max_contacts_per_pair)),
+        "localpos1b": V_VEC(3, dtype=gs.ti_float, shape=(_B, max_contacts_per_pair)),
+        "localpos1c": V_VEC(3, dtype=gs.ti_float, shape=(_B, max_contacts_per_pair)),
+        "localpos2a": V_VEC(3, dtype=gs.ti_float, shape=(_B, max_contacts_per_pair)),
+        "localpos2b": V_VEC(3, dtype=gs.ti_float, shape=(_B, max_contacts_per_pair)),
+        "localpos2c": V_VEC(3, dtype=gs.ti_float, shape=(_B, max_contacts_per_pair)),
+        "vid1": V_VEC(3, dtype=gs.ti_int, shape=(_B, max_contacts_per_pair)),
+        "vid2": V_VEC(3, dtype=gs.ti_int, shape=(_B, max_contacts_per_pair)),
+        "w_localpos1": V_VEC(3, dtype=gs.ti_float, shape=(_B, max_contacts_per_pair)),
+        "w_localpos2": V_VEC(3, dtype=gs.ti_float, shape=(_B, max_contacts_per_pair)),
```
duburcqa: These parameters are not following our guidelines in terms of naming conventions:
https://github.com/Genesis-Embodied-AI/Genesis/pull/1053

It should be: `local_pos1_a`

genesis/utils/array_class.py
```diff
@@ -292,6 +292,60 @@ def __init__(self):
...
+    localpos2b: V_ANNOTATION
+    localpos2c: V_ANNOTATION
+    vid1: V_ANNOTATION
+    vid2: V_ANNOTATION
+    w_localpos1: V_ANNOTATION
+    w_localpos2: V_ANNOTATION
+    normal: V_ANNOTATION
+    ref_id: V_ANNOTATION
+    # Differentiable data
+    ref_penetration: V_ANNOTATION
```
duburcqa, line 381: Could you document what is stored in each of these attributes individually? (one line max should be fine)

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -1043,6 +1044,8 @@ def _func_constraint_clear(self):
         self.constraint_solver.constraint_state.n_constraints_equality.fill(0)
         self.constraint_solver.constraint_state.n_constraints_frictionloss.fill(0)
         self.collider._collider_state.n_contacts.fill(0)
+        if hasattr(self.collider._collider_state, "n_diff_contacts"):
+            self.collider._collider_state.n_diff_contacts.fill(0)
```
duburcqa: Using `hasattr` is a bad practice. You should rather using the condition for which this attribute should exist. I guess : `if self._static_rigid_sim_config.requires_grad:` ?

tests/test_diff_rigid.py
SonSang: It does not have rendering, so I'll just keep it as False (this is only gradient comparison to the finite difference).

YilingQiao:
@SonSang can you address the conficts and merge the latest main (after your ddl)? I think the CI error can be resolved using the latest main.

SonSang:
@YilingQiao Sure, I'll do that!

YilingQiao (approved):

YilingQiao (approved):

YilingQiao:
ah, maybe I shouldn't commit auto pilot's commit as it doesn't check format.

genesis/utils/array_class.py
```diff
@@ -356,6 +356,62 @@ def __init__(self):
...
+    # Local positions of the 3 vertices from the two geometries that define the face on the Minkowski difference
+    local_pos1_a: V_ANNOTATION
+    local_pos1_b: V_ANNOTATION
+    local_pos1_c: V_ANNOTATION
+    local_pos2_a: V_ANNOTATION
+    local_pos2_b: V_ANNOTATION
+    local_pos2_c: V_ANNOTATION
+    # Local positions of the 1 vertex from the two geometries that define the support point for the face above
+    w_local_pos1: V_ANNOTATION
+    w_local_pos2: V_ANNOTATION
```
duburcqa, line 374: You are mixing `_a`,`_b` and `1`,`2` to refer to the same information, and you even recycle `_a`,`_b`,`_c` for the vertices of the Minkowski difference. This is confusing.

What about:
```python
    geom_a: V_ANNOTATION
    geom_b: V_ANNOTATION
    local_pos1_a: V_ANNOTATION
    local_pos2_a: V_ANNOTATION
    local_pos3_a: V_ANNOTATION
    local_pos1_b: V_ANNOTATION
    local_pos2_b: V_ANNOTATION
    local_pos3_b: V_ANNOTATION
    w_local_pos_a: V_ANNOTATION
    w_local_pos_b V_ANNOTATION
```

genesis/utils/array_class.py
```diff
@@ -356,6 +356,62 @@ def __init__(self):
         return ClassContactData()
 
 
+@dataclasses.dataclass
+class StructDiffContactInput:
```
duburcqa, line 360: I would suggest to stick to `Info` vs `State` suffix (or maybe `Cache` but I don't like it much in this context as it is not exactly what it is) as naming conventions for "constant" data vs "dynamic" data. It is easier to understand what it is about at first glance, even if it may be slightly abusive grammatically-speaking.

tests/test_diff_rigid.py
duburcqa, line 1: What about fusing this file with `test_grad.py` ?

Issue #1716, opened 2025-09-12
[Bug]: An gs.recorders error occurred when running the sensor sample


XXLiu-HNU:
### Bug Description

```~/Genesis/Genesis$ python examples/sensors/imu.py 
[Genesis] [11:03:46] [INFO] ╭─────────────────────────────────────────────────────────────────────────────────────╮
[Genesis] [11:03:46] [INFO] │┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉ Genesis ┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉│
[Genesis] [11:03:46] [INFO] ╰─────────────────────────────────────────────────────────────────────────────────────╯
[Genesis] [11:03:47] [INFO] Consider setting 'performance_mode=True' in production to maximise runtime speed, if significantly increasing compilation time is not a concern.
[Genesis] [11:03:47] [INFO] Running on [NVIDIA GeForce RTX 4060 Laptop GPU] with backend gs.cuda. Device memory: 7.61 GB.
[Genesis] [11:03:47] [INFO] 🚀 Genesis initialized. 🔖 version: 0.3.3, 🌱 seed: None, 📏 precision: '32', 🐛 debug: False, 🎨 theme: 'dark'.
[Genesis] [11:03:47] [INFO] Scene <ef05f9e> created.
[Genesis] [11:03:47] [INFO] Adding <gs.RigidEntity>. idx: 0, uid: <dbf9ffd>, morph: <gs.morphs.Plane>, material: <gs.materials.Rigid>.
[Genesis] [11:03:47] [INFO] Adding <gs.RigidEntity>. idx: 1, uid: <28fa41b>, morph: <gs.morphs.MJCF(file='/home/xingxun/Genesis/Genesis/genesis/assets/xml/franka_emika_panda/panda.xml')>, material: <gs.materials.Rigid>.

[Genesis] [11:03:47] [INFO] Applying offset to base link's pose with user provided value in morph.
Traceback (most recent call last):
  File "/home/xingxun/Genesis/Genesis/genesis/options/options.py", line 31, in __init__
    super().__init__(**data)
[...]

PR #1718, opened 2025-09-13
[BUG FIX][FEATURE] Add test_recorders.py and add saving plot to video file

Milotrince:
## Related Issue
Resolves https://github.com/Genesis-Embodied-AI/Genesis/issues/1716

## Checklist:

### DIFF
--- a/genesis/recorders/plotters.py
+++ b/genesis/recorders/plotters.py
@@ -22,6 +22,7 @@ except ImportError:
 
 IS_MATPLOTLIB_AVAILABLE = False
 try:
+    import matplotlib as mpl
     import matplotlib.pyplot as plt
 
     IS_MATPLOTLIB_AVAILABLE = True
@@ -47,22 +48,28 @@ class BasePlotterOptions(RecorderOptions):
 
     Parameters
     ----------
+    title: str
+        The title of the plot.
     labels: tuple[str] | dict[str, tuple[str]] | None
         The labels for the plot. The length of the labels should match the length of the data.
         If a dict is provided, the data should also be a dict of tuples of strings that match the length of the data.
         The keys will be used as subplot titles and the values will be used as labels within each subplot.
     window_size: tuple[int, int]
-        The size of the window.
+        The size of the window in pixels.
     history_length: int
-        The length of the history to store.
-    title: str
-        The title of the plot.
+        The maximum number of previous data to store.
+    save_to_filename: str | None
+        If provided, the animation will be saved to a file with the given filename.
+    show_window: bool
+        Whether to show the window. Defaults to True, set to False for headless mode.
     """
 
-    labels: tuple[str] | dict[str, tuple[str]] | None = None
+    title: str = ""
+    labels: tuple[str, ...] | dict[str, tuple[str, ...]] | None = None
     window_size: tuple[int, int] = (800, 600)
     history_length: int = 100
-    title: str = ""
+    save_to_filename: str | None = None
+    show_window: bool = True
 
 
 class BasePlotter(Recorder):
@@ -89,17 +96,17 @@ class BasePlotter(Recorder):
         """Set up the plot structure based on labels or first data sample."""
         if isinstance(labels_or_data, dict):
             self.is_dict_data = True
-            first_value = next(iter(labels_or_data.values()))
+            next_dict_value = next(iter(labels_or_data.values()))
 
-            if isinstance(first_value, list):
-                # labels were provided
-                self.subplot_structure = {k: tuple(v) for k, v in labels_or_data.items()}
-            elif isinstance(first_value, (torch.Tensor, np.ndarray)):
+            if isinstance(next_dict_value, (torch.Tensor, np.ndarray)):
                 # data was provided
                 self.subplot_structure = {}
                 for key, values in labels_or_data.items():
                     values = _data_to_array(values)
                     self.subplot_structure[key] = tuple(f"{key}_{i}" for i in range(len(values)))
+            elif isinstance(next_dict_value, Sequence) and isinstance(next_dict_value[0], str):
+                # labels were provided
+                self.subplot_structure = {k: tuple(v) for k, v in labels_or_data.items()}
             else:
                 gs.raise_exception(f"[{type(self).__name__}] Unsupported input argument type: {type(labels_or_data)}")
         else:
@@ -162,16 +169,20 @@ class PyQtPlotterOptions(BasePlotterOptions):
 
     Parameters
     ----------
+    title: str
+        The title of the plot.
     labels: tuple[str] | dict[str, tuple[str]] | None
         The labels for the plot. The length of the labels should match the length of the data.
         If a dict is provided, the data should also be a dict of tuples of strings that match the length of the data.
         The keys will be used as subplot titles and the values will be used as labels within each subplot.
     window_size: tuple[int, int]
-        The size of the window.
+        The size of the window in pixels.
     history_length: int
-        The length of the history to store.
-    title: str
-        The title of the plot.
+        The maximum number of previous data to store.
+    save_to_filename: str | None
+        If provided, the animation will be saved to a file with the given filename.
+    show_window: bool
+        Whether to show the window. Defaults to True, set to False for headless mode.
     """
 
     pass
@@ -203,8 +214,14 @@ class PyQtPlotter(BasePlotter):
         else:
             self.app = pg.QtWidgets.QApplication.instance()
 
-        self.widget = pg.GraphicsLayoutWidget(show=True, title=self._options.title)
+        self.widget = pg.GraphicsLayoutWidget(show=self._options.show_window, title=self._options.title)
         self.widget.resize(*self._options.window_size)
+
+        if self._options.save_to_filename:
+            self.video_writer = self._manager.add_recorder(
+                data_func=self._video_data_func,
+                rec_options=gs.recorders.VideoFile(filename=self._options.save_to_filename, hz=self._options.hz),
+            )
 
         gs.logger.info("[PyQtPlotter] created PyQtGraph window")
 
@@ -260,6 +277,17 @@ class PyQtPlotter(BasePlotter):
     def run_in_thread(self) -> bool:
         return True
 
+    def _video_data_func(self):
+        """Capture the plot image as a video frame."""
+        pixmap = self.widget.grab()
+        qimage = pixmap.toImage()
+
+        qimage = qimage.convertToFormat(pg.QtGui.QImage.Format_RGB888)
+        ptr = qimage.bits()
+        ptr.setsize(qimage.byteCount())
+
+        return np.array(ptr).reshape(qimage.height(), qimage.width(), 3)
+
 
 class MPLPlotterOptions(BasePlotterOptions):
     """
@@ -268,6 +296,15 @@ class MPLPlotterOptions(BasePlotterOptions):
     Parameters
     ----------
     labels: tuple[str] | dict[str, tuple[str]] | None
+        The labels for the plot. The length of the labels should match the length of the data.
+        If a dict is provided, the data should also be a dict of tuples of strings that match the length of the data.
+        The keys will be used as subplot titles and the values will be used as labels within each subplot.
+    window_size: tuple[int, int]
+        The size of the window.
+    history_length: int
+        The length of the history to store.
+    title: str
+        The title of the plot.   labels: tuple[str] | dict[str, tuple[str]] | None
         The labels for the plot. The length of the labels should match the length of the data.
         If a dict is provided, the data should also be a dict of tuples of strings that match the length of the data.
         The keys will be used as subplot titles and the values will be used as labels within each subplot.
@@ -306,13 +343,18 @@ class MPLPlotter(BasePlotter):
 
         # create figure and subplots
         n_subplots = len(self.subplot_structure)
+        dpi = mpl.rcParams.get("figure.dpi", 100)
+        # matplotlib figsize uses inches
+        figsize = (
+            self._options.window_size[0] / dpi,
+            self._options.window_size[1] / dpi,
+        )
+
         if n_subplots == 1:
-            self.fig, ax = plt.subplots(figsize=self._options.window_size)
+            self.fig, ax = plt.subplots(figsize=figsize)
             self.axes = [ax]
         else:
-            self.fig, axes = plt.subplots(
-                n_subplots, 1, figsize=self._options.window_size, sharex=True, constrained_layout=True
-            )
+            self.fig, axes = plt.subplots(n_subplots, 1, figsize=figsize, sharex=True, constrained_layout=True)
             self.axes = axes if isinstance(axes, (list, tuple, np.ndarray)) else [axes]
         self.fig.suptitle(self._options.title)
 
@@ -342,11 +384,18 @@ class MPLPlotter(BasePlotter):
             ax.set_xlim(0, 10)
             ax.set_ylim(-1, 1)
 
-        self.fig.show()
+        if self._options.show_window:
+            self.fig.show()
         self.fig.canvas.draw()
 
         for ax in self.axes:
             self.backgrounds.append(self.fig.canvas.copy_from_bbox(ax.bbox))
+
+        if self._options.save_to_filename:
+            self.video_writer = self._manager.add_recorder(
+                data_func=self._video_data_func,
+                rec_options=gs.recorders.VideoFile(filename=self._options.save_to_filename, hz=self._options.hz),
+            )
 
     def process(self, data, cur_time):
         """Process new data point and update plot using blitting."""
@@ -419,3 +468,8 @@ class MPLPlotter(BasePlotter):
     @property
     def run_in_thread(self) -> bool:
         return gs.platform != "macOS"
+
+    def _video_data_func(self):
+        """Capture the plot image as a video frame."""
+        np_buffer = np.frombuffer(self.fig.canvas.tostring_argb(), dtype="uint8")
+        return np_buffer.reshape(*reversed(self.fig.canvas.get_width_height()), 4)[..., 1:]
--- a/tests/test_recorders.py
+++ b/tests/test_recorders.py
@@ -1,0 +1,166 @@
+import matplotlib as mpl
+import pytest
+
+mpl.use("Agg")  # headless mode for plt
+import csv
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+import genesis as gs
+
+
+@pytest.mark.required
+def test_plotter(tmp_path):
+    """Test if the plotter recorders works."""
+    DT = 0.01
+    STEPS = 10
+
+    scene = gs.Scene(
+        sim_options=gs.options.SimOptions(dt=DT),
+        show_viewer=False,
+        show_FPS=False,
+    )
+
+    scene.add_entity(
+        morph=gs.morphs.Box(size=(0.1, 0.1, 0.1), pos=(0.0, 0.0, 0.5)),
+        material=gs.materials.Rigid(rho=1000.0),
+    )
+
+    call_count = 0
+
+    def dummy_data_func():
+        nonlocal call_count
+        call_count += 1
+        return {
+            "a": [call_count * 0.1, call_count * 0.2, call_count * 0.3],
+            "b": [call_count * 0.01, call_count * 0.02],
+        }
+
+    filepath = tmp_path / "test_plot.mp4"
+    plotter = gs.recorders.MPLPlot(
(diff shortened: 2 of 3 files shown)


genesis/recorders/plotters.py
```diff
@@ -260,6 +277,17 @@ def cleanup(self):
     def run_in_thread(self) -> bool:
         return True
 
+    def _video_data_func(self):
```
duburcqa: It is more like `_image_array_func`? This could be useful beyond the scope of recording videos, i.e. just take "screenshots".

genesis/recorders/plotters.py
```diff
@@ -275,6 +303,15 @@ class MPLPlotterOptions(BasePlotterOptions):
         The size of the window.
     history_length: int
         The length of the history to store.
+    title: str
+        The title of the plot.   labels: tuple[str] | dict[str, tuple[str]] | None
+        The labels for the plot. The length of the labels should match the length of the data.
```
duburcqa: Broken formatting.

genesis/recorders/plotters.py
```diff
@@ -306,13 +343,18 @@ def build(self):
 
         # create figure and subplots
         n_subplots = len(self.subplot_structure)
+        dpi = mpl.rcParams.get("figure.dpi", 100)
+        # matplotlib figsize uses inches
+        figsize = (
+            self._options.window_size[0] / dpi,
+            self._options.window_size[1] / dpi,
+        )
```
duburcqa: one line is good enough.

genesis/recorders/plotters.py
```diff
@@ -162,16 +169,20 @@ class PyQtPlotterOptions(BasePlotterOptions):
...
+        The size of the window in pixels.
     history_length: int
-        The length of the history to store.
-    title: str
-        The title of the plot.
+        The maximum number of previous data to store.
+    save_to_filename: str | None
+        If provided, the animation will be saved to a file with the given filename.
+    show_window: bool
+        Whether to show the window. Defaults to True, set to False for headless mode.
```
duburcqa: The default should be True if machine has a graphical server running / display capability, False otherwise.

duburcqa: Do detect this, you can use this pattern: https://github.com/Genesis-Embodied-AI/Genesis/blob/76f33f09370282e6acc3841a60307c31c2b850a7/genesis/vis/visualizer.py#L47

I would suggest moving this in misc utils to share the code between visualiser and plotter.

genesis/recorders/plotters.py
```diff
@@ -203,9 +214,15 @@ def build(self):
...
             self.app = pg.QtWidgets.QApplication.instance()
 
-        self.widget = pg.GraphicsLayoutWidget(show=True, title=self._options.title)
+        self.widget = pg.GraphicsLayoutWidget(show=self._options.show_window, title=self._options.title)
         self.widget.resize(*self._options.window_size)
 
+        if self._options.save_to_filename:
+            self.video_writer = self._manager.add_recorder(
+                data_func=self._video_data_func,
+                rec_options=gs.recorders.VideoFile(filename=self._options.save_to_filename, hz=self._options.hz),
```
duburcqa: Multiple lines for input arguments for readability.

genesis/recorders/plotters.py
```diff
@@ -342,12 +384,19 @@ def build(self):
...
+            self.fig.show()
         self.fig.canvas.draw()
 
         for ax in self.axes:
             self.backgrounds.append(self.fig.canvas.copy_from_bbox(ax.bbox))
 
+        if self._options.save_to_filename:
+            self.video_writer = self._manager.add_recorder(
+                data_func=self._video_data_func,
+                rec_options=gs.recorders.VideoFile(filename=self._options.save_to_filename, hz=self._options.hz),
```
duburcqa: Multiple lins.

genesis/recorders/plotters.py
```diff
@@ -419,3 +468,8 @@ def cleanup(self):
     @property
     def run_in_thread(self) -> bool:
         return gs.platform != "macOS"
+
+    def _video_data_func(self):
+        """Capture the plot image as a video frame."""
+        np_buffer = np.frombuffer(self.fig.canvas.tostring_argb(), dtype="uint8")
+        return np_buffer.reshape(*reversed(self.fig.canvas.get_width_height()), 4)[..., 1:]
```
duburcqa: Just extract width / height. it would be more readable.
```
width, height = self.fig.canvas.get_width_height()
return np_buffer.reshape((height, width, 4))[..., 1:]
```

duburcqa: I would recommend using `savefig`, which is backend-agnostic (`tostring_argb` only supports AGG-based backends) and exposes many nice conveniences options.

```python
from matplotlib.backends.backend_agg import FigureCanvasAgg

if isinstance(fig.canvas, FigureCanvasAgg):
    width, height = self.fig.canvas.get_width_height()
    rgba_array_flat = np.frombuffer(fig.canvas.buffer_rgba(), dtype=np.uint8)
    return rgba_array_flat.reshape((height, width, 4))[..., :3]
else:
    # Slower but more generic fallback only if necessary
    buffer = io.BytesIO()
    fig.savefig(buffer, format='png', dpi='figure', transparent=False, bbox_inches='tight')
    buf.seek(0)
    img = Image.open(buffer)
    return np.asarray(img.convert('RGB'))
```

tests/test_recorders.py
```diff
@@ -0,0 +1,166 @@
+import matplotlib as mpl
+import pytest
+
+mpl.use("Agg")  # headless mode for plt
```
duburcqa: Matplotlib is not smart enough to pick the most appropriate backend automatically?

tests/test_recorders.py
```diff
@@ -0,0 +1,166 @@
...
+import csv
```
duburcqa, line 1: csv is part of standard library. It should be move on the very top import group.

tests/test_recorders.py
```diff
@@ -0,0 +1,166 @@
...
+
+import matplotlib.pyplot as plt
```
duburcqa: Do not import `matplotlib.pyplot` here, it is useful?

duburcqa: By the way, could to update this part? https://github.com/Genesis-Embodied-AI/Genesis/blob/76f33f09370282e6acc3841a60307c31c2b850a7/genesis/recorders/plotters.py#L23

It slipped through the review but it is not correct. You should just do
```python
IS_MATPLOTLIB_AVAILABLE = False
try:
    import matplotlib

    IS_MATPLOTLIB_AVAILABLE = True
except ImportError:
    pass
```

And import `import matplotlib.pyplot as plt` on top of every functions that requires it. This will allow changing matplotlib backend after importing genesis, which is much nicer, but also greatly speed-up genesis import on MacOS in particular.

tests/test_recorders.py
```diff
@@ -0,0 +1,166 @@
...
+
+    for _ in range(STEPS):
+        scene.step()
+
+    assert call_count == STEPS // 2
+
+    scene.stop_recording()
+
+    assert filepath.exists(), "Recorded video file should exist"
+    assert filepath.stat().st_size > 0, "Recorded video file should not be empty"
```
duburcqa: You should use your fixture `png_snapshot` to add pixel-matching condition to the unit test. See `test_batch_deformable_render` for instance.

Do "save" the reference snapshot, add `--update-snapshot` pytest argument when running the tests sequentially via `-n 0`. Then, copy the image in our huggingface repo (https://huggingface.co/datasets/Genesis-Intelligence/snapshots) and finally, update the commit hash in `tests/utils.py` (HUGGINGFACE_SNAPSHOT_REVISION).

tests/test_recorders.py
```diff
@@ -0,0 +1,166 @@
...
+
+    filepath = tmp_path / "test_plot.mp4"
+    plotter = gs.recorders.MPLPlot(
+        labels={"a": ("x", "y", "z"), "b": ("u", "v")},
+        title="Test MPLPlotter",
+        history_length=50,
+        window_size=(400, 300),
+        hz=1.0 / DT / 2,  # half of the simulation frequency, so every other step
+        show_window=False,
+        save_to_filename=str(filepath),
```
duburcqa: You already have a test checking video recording, so you should not check again this feature in this test. Just call `_video_data_func` manually to get a screenshot and do your stuff on it.

tests/test_recorders.py
Milotrince: Ah, yes it is https://matplotlib.org/stable/users/explain/figure/backends.html#selecting-a-backend

tests/test_recorders.py
Milotrince: PR to the hf https://huggingface.co/datasets/Genesis-Intelligence/snapshots/discussions/3

tests/utils.py
```diff
@@ -1014,3 +1015,10 @@ def simulate_and_check_mujoco_consistency(gs_sim, mj_sim, qpos=None, qvel=None,
         gs_sim.scene.step()
         # if gs_sim.scene.visualizer:
         #     gs_sim.scene.visualizer.update()
+
+
+def rgb_array_to_buffer(rgb_arr: np.ndarray) -> bytes:
```
duburcqa: `rgb_array_to_png_bytes` would be more appropriate.

tests/test_recorders.py
```diff
@@ -37,31 +35,28 @@ def dummy_data_func():
...
-        window_size=(400, 300),
-        hz=1.0 / DT / 2,  # half of the simulation frequency, so every other step
-        show_window=False,
-        save_to_filename=str(filepath),
+    plotter = scene.start_recording(
+        data_func=dummy_data_func,
+        rec_options=gs.recorders.MPLPlot(
+            labels={"a": ("x", "y", "z"), "b": ("u", "v")},
+            title="Test MPLPlotter",
+            history_length=50,
```
duburcqa: Could you add a check to ensure `history_length` is taken into account? Probably by inspecting the low-level data rolling buffer because screenshot is not designed for this kind of fine-grained assessment.

tests/test_recorders.py
```diff
@@ -37,31 +35,28 @@ def dummy_data_func():
...
+            window_size=(400, 300),
+            hz=1.0 / DT / 2,  # half of the simulation frequency, so every other step
+            show_window=False,
```
duburcqa, line 76: You should pass `show_viewer` fixture parameter as input argument of the test and use it here.

genesis/vis/visualizer.py
```diff
@@ -37,29 +34,16 @@ def __init__(self, scene, show_viewer, vis_options, viewer_options, renderer_opt
...
-                display = pyglet.display.get_display()
-                screen = display.get_default_screen()
-                try:
-                    screen_scale = screen.get_scale()
-                except NotImplementedError:
-                    # Probably some headless screen
-                    screen_scale = 1.0
-            screen_height, screen_width = screen.height, screen.width
-            self._has_display = True
+            self._has_display, screen_height, _, screen_scale = gs.utils.try_get_display_size()
```
duburcqa: Remove `self._has_display` from the output values, and just set it to true if not raising an exception.

genesis/utils/misc.py
```diff
@@ -360,6 +361,53 @@ def _default_factory():
...
+            except NotImplementedError:
+                # Probably some headless screen
+                screen_scale = 1.0
+        screen_height, screen_width = screen.height, screen.width
+        has_display = True
+    except Exception:
+        if raise_exception:
+            raise
+        has_display = False
+    return has_display, screen_height, screen_width, screen_scale
```
duburcqa: I would rather return `screen_height = None, screen_width = None, screen_scale = None` in case of failure and remove `has_display` output argument.

duburcqa: I'm not convinced the entire `raise_exception` logics is really useful. I would just request the users to catch the exception themselves. Less obfuscated. You can still expose some `has_display` method that does this automatically.

genesis/recorders/plotters.py
```diff
@@ -469,7 +478,27 @@ def cleanup(self):
...
+        from matplotlib.backends.backend_agg import FigureCanvasAgg
+
+        if isinstance(self.fig.canvas, FigureCanvasAgg):
+            print("Using FigureCanvasAgg")
+            width, height = self.fig.canvas.get_width_height()
+            rgba_array_flat = np.frombuffer(self.fig.canvas.buffer_rgba(), dtype=np.uint8)
+            return rgba_array_flat.reshape((height, width, 4))[..., :3]
+        else:
+            # Slower but more generic fallback only if necessary
+            print("Using fallback")
```
duburcqa: Remove these debug prints.

genesis/recorders/plotters.py
```diff
@@ -334,6 +337,10 @@ def build(self):
             )
         super().build()
 
+        if not self._options.show_window:
+            mpl.use("Agg")
```
duburcqa: I this really necessary? I don't like messing up with the rendering backend inside modules. In unit tests and example files it is fine, but not in global scope. It would interfere with the plotting logics of the user.

genesis/recorders/plotters.py
```diff
@@ -182,10 +183,10 @@ class PyQtPlotterOptions(BasePlotterOptions):
     save_to_filename: str | None
         If provided, the animation will be saved to a file with the given filename.
     show_window: bool
-        Whether to show the window. Defaults to True, set to False for headless mode.
+        Whether to show the window. Defaults to True if a display is connected, False otherwise.
     """
 
-    pass
+    show_window: bool = has_display()
```
duburcqa: You should delay this logics at init and set it too default here, otherwise "has_display" while be executed at init, which is not only slow, but also quite invasive as it will initialise the rendering backend (similar to `import matplotlib.pyplot`).

genesis/recorders/plotters.py
Milotrince: Removed

tests/test_recorders.py
```diff
@@ -0,0 +1,164 @@
...
+
+    for _ in range(STEPS):
+        scene.step()
+
+    assert call_count == STEPS // 2
+    assert len(plotter.x_data) == HISTORY_LENGTH
+    assert np.isclose(plotter.x_data[-1], STEPS * DT, atol=gs.EPS)
+    assert rgb_array_to_png_bytes(plotter.get_image_array()) == png_snapshot
+
+    scene.stop_recording()
```
duburcqa: This should not be necessary but ok.

duburcqa (dismissed):

genesis/recorders/base_recorder.py
```diff
@@ -148,25 +151,57 @@ def start(self):
...
+        """Wait for the processor thread to finish."""
+        if self._processor_thread is None:
+            self._data_queue = queue.Queue(maxsize=self._options.buffer_size)
+            self._processor_thread = threading.Thread(target=self._process_data_loop)
+            self._processor_thread.start()
+        else:
+            gs.logger.warning(f"[{type(self).__name__}] start_thread(): Processor thread already exists.")
+
+    @gs.assert_built
+    def sync(self, timestep: float = 0.1, timeout: float | None = None):
```
duburcqa: Are you sure exposing timestep is relevant? It is usually never the case. I would use `timestep = min(0.1, timeout)`  for simplicity.

genesis/recorders/base_recorder.py
```diff
@@ -148,25 +151,57 @@ def start(self):
...
+            The maximum time to wait for the data queue to be empty. If None, wait indefinitely.
+            If the timeout is reached, an exception is raised.
+        """
+        if self._data_queue is not None:
+            if timeout is not None:
+                start_time = time.time()
+            while not self._data_queue.empty():
+                time.sleep(timestep)
+                if timeout is not None and time.time() - start_time > timeout:
+                    gs.raise_exception(f"[{type(self).__name__}] sync(): Timeout waiting for data queue to be empty.")
```
duburcqa: This pattern is a bit broken. You wait then you raise right away without checking first if the condition is meat? This is not cool! Start by checking the condition first, then sleep :)

duburcqa: By the way, for the sleep duration, you should do:
```python
dt = min(timestep, (start_time + timeout) - time.time())
if dt > 0.0:
    time.sleep(dt)
```

duburcqa (approved):

XXLiu-HNU:
There are still problems when running the code
```(genesis) xingxun@xingxun:~/Genesis/Genesis$ /home/xingxun/miniconda3/envs/genesis/bin/python /home/xingxun/Genesis/Genesis/examples/sensors/imu.py
[Genesis] [10:24:26] [INFO] ╭─────────────────────────────────────────────────────────────────────────────────────╮
[Genesis] [10:24:26] [INFO] │┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉ Genesis ┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉│
[Genesis] [10:24:26] [INFO] ╰─────────────────────────────────────────────────────────────────────────────────────╯
[Genesis] [10:24:26] [INFO] Consider setting 'performance_mode=True' in production to maximise runtime speed, if significantly increasing compilation time is not a concern.
[Genesis] [10:24:26] [INFO] Running on [NVIDIA GeForce RTX 4060 Laptop GPU] with backend gs.cuda. Device memory: 7.61 GB.
[Genesis] [10:24:26] [INFO] 🚀 Genesis initialized. 🔖 version: 0.3.3, 🌱 seed: None, 📏 precision: '32', 🐛 debug: False, 🎨 theme: 'dark'.
[Genesis] [10:24:26] [INFO] Scene <7fae28f> created.
[Genesis] [10:24:26] [INFO] Adding <gs.RigidEntity>. idx: 0, uid: <e114647>, morph: <gs.morphs.Plane>, material: <gs.materials.Rigid>.
[Genesis] [10:24:26] [INFO] Adding <gs.RigidEntity>. idx: 1, uid: <ab2eae8>, morph: <gs.morphs.MJCF(file='/home/xingxun/Genesis/Genesis/genesis/assets/xml/franka_emika_panda/panda.xml')>, material: <gs.materials.Rigid>.
[Genesis] [10:24:27] [WARNING] (MJCF) Approximating tendon by joint actuator for `finger_joint1`
[Genesis] [10:24:27] [WARNING] (MJCF) Actuator control gain and bias parameters cannot be reduced to a unique PD control position gain. Using max between gain and bias for joint `finger_joint1`.
[Genesis] [10:24:27] [WARNING] (MJCF) Approximating tendon by joint actuator for `finger_joint2`
[Genesis] [10:24:27] [WARNING] (MJCF) Actuator control gain and bias parameters cannot be reduced to a unique PD control position gain. Using max between gain and bias for joint `finger_joint2`.
[Genesis] [10:24:27] [INFO] Applying offset to base link's pose with user provided value in morph.
[Genesis] [10:24:27] [INFO] Building scene <7fae28f>...
[Genesis] [10:24:28] [WARNING] Reference robot position exceeds joint limits.
[Genesis] [10:24:28] [WARNING] Constraint solver time constant should be greater than 2*substep_dt. timeconst is changed from `0.005` to `0.02`). Decrease simulation timestep or increase timeconst to avoid altering the original value.
[Genesis] [10:24:32] [INFO] Compiling simulation kernels...
[Genesis] [10:24:51] [INFO] Building visualizer...
[Genesis] [10:24:52] [INFO] Viewer created. Resolution: 1066×800, max_FPS: 60.
QObject::moveToThread: Current thread (0x1c9915f0) is not the object's thread (0x58a8b960).
Cannot move to target thread (0x1c9915f0)

qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "/home/xingxun/miniconda3/envs/genesis/lib/python3.11/site-packages/cv2/qt/plugins" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: xcb, eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, wayland-egl, wayland, wayland-xcomposite-egl, wayland-xcomposite-glx, webgl.

Aborted (core dumped)
```

Milotrince:
@XXLiu-HNU Please make a new issue and include the script you are running as well as what OS youre running on. afaict this looks like an issue with PyQt installation? have you tried MPLPlotter instead?

XXLiu-HNU:
> @XXLiu-HNU Please make a new issue and include the script you are running as well as what OS youre running on. afaict this looks like an issue with PyQt installation? have you tried MPLPlotter instead?

https://github.com/Genesis-Embodied-AI/Genesis/issues/1746#issue-3428487902

Issue #1722, opened 2025-09-13
[Bug]: Build crashes on adding multiple sensors


jgleyze:
### Bug Description

When adding a second sensor to a scene, the build crashes throwing : 
``` File "/home/veram/.test/atalante_gs/test.py", line 74, in main
    scene.build()
  File "/genesis/utils/misc.py", line 138, in wrapper
    return method(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/genesis/engine/scene.py", line 783, in build
    self._sim.step()
  File "/genesis/engine/simulator.py", line 287, in step
    self._sensor_manager.step()
  File "/genesis/sensors/sensor_manager.py", line 80, in step
    sensor_cls._update_shared_cache(
  File "/genesis/sensors/contact_force.py", line 293, in _update_shared_cache
    cls._add_noise_drift_bias(shared_metadata, shared_cache)
  File "/genesis/sensors/base_sensor.py", line 527, in _add_noise_drift_bias
    output += shared_metadata.bias
  File "/torch/utils/_device.py", line 104, in __torch_function__
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
RuntimeError: The size of tensor a (6) must match the size of tensor b (2) at non-singleton dimension 1
```

Maybe I do things the wrong way !

### Steps to Reproduce

Take the hand force sensor example and add two lines below https://github.com/Genesis-Embodied-AI/Genesis/blob/76f33f09370282e6acc3841a60307c31c2b850a7/examples/sensors/force.py#L55-L56 
```python
wrist = hand.get_link("wrist")
force_sensor_2 = scene.add_sensor(gs.sensors.ContactForce(entity_idx=hand.idx, link_idx_local=wrist.idx_local))
```

### Expected Behavior

/

### Relevant log output

```shell

```

### Environment

/

### Release version or Commit ID

76f33f09370282e6acc3841a60307c31c2b850a7

duburcqa:
Whoops, we will fix it and add unit test coverage for this.

PR #1726, opened 2025-09-15
[BUG FIX] Add tests for adding multiple sensors and envs.

Milotrince:
## Description
Remove some code duplication, add at least two of each kind of sensors in tests and add fixture for 0 and 2 envs for coverage.

## Related Issue
Resolves https://github.com/Genesis-Embodied-AI/Genesis/issues/1722

## Motivation and Context
Bug :(

`pytest tests/test_sensors.py`

## Checklist:

### DIFF
--- a/genesis/sensors/contact_force.py
+++ b/genesis/sensors/contact_force.py
@@ -44,13 +44,13 @@ def _kernel_get_contacts_forces(
         if contact_data_link_a == sensors_link_idx[i_s] or contact_data_link_b == sensors_link_idx[i_s]:
             j_s = i_s * 3  # per-sensor output dimension is 3
 
-            quat_a = ti.Vector.zero(ti.f32, 4)
-            quat_b = ti.Vector.zero(ti.f32, 4)
+            quat_a = ti.Vector.zero(gs.ti_float, 4)
+            quat_b = ti.Vector.zero(gs.ti_float, 4)
             for j in ti.static(range(4)):
                 quat_a[j] = links_quat[i_b, contact_data_link_a, j]
                 quat_b[j] = links_quat[i_b, contact_data_link_b, j]
 
-            force_vec = ti.Vector.zero(ti.f32, 3)
+            force_vec = ti.Vector.zero(gs.ti_float, 3)
             for j in ti.static(range(3)):
                 force_vec[j] = contact_forces[i_b, i_c, j]
 
@@ -89,7 +89,7 @@ class ContactSensorMetadata(SharedSensorMetadata):
     """
 
     solver: RigidSolver | None = None
-    expanded_links_idx: torch.Tensor = make_tensor_field((0, 1, 0), dtype_factory=lambda: gs.tc_int)
+    expanded_links_idx: torch.Tensor = make_tensor_field((0,), dtype_factory=lambda: gs.tc_int)
 
 
 @register_sensor(ContactSensorOptions, ContactSensorMetadata)
@@ -104,11 +104,12 @@ class ContactSensor(Sensor):
         if self._shared_metadata.solver is None:
             self._shared_metadata.solver = self._manager._sim.rigid_solver
 
+        link_start = self._shared_metadata.solver.entities[self._options.entity_idx].link_start
         self._shared_metadata.expanded_links_idx = concat_with_tensor(
             self._shared_metadata.expanded_links_idx,
-            self._options.entity_idx + self._options.link_idx_local,
-            expand=(self._manager._sim._B, 1, -1),
-            dim=-1,
+            link_start + self._options.link_idx_local,
+            expand=(1,),
+            dim=0,
         )
 
     def _get_return_format(self) -> tuple[int, ...]:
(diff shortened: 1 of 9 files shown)


genesis/utils/ring_buffer.py
```diff
@@ -67,7 +67,8 @@ def at(self, idx: int | torch.Tensor) -> torch.Tensor:
         rel_idx = (self._idx_ptr.value - idx) % self.N
         if isinstance(idx, torch.Tensor):
             batch_size = len(idx) if idx.ndim > 0 else 1
-            return self.buffer[rel_idx, :batch_size]
+            batch_indices = torch.arange(batch_size, device=gs.device, dtype=gs.tc_int)
+            return self.buffer[rel_idx, batch_indices]
```
Milotrince: I need to figure this out still.. It works when i do arange.. When using :batch_size and reshapes or squeeze(1) or having ifs i could not get it to work for both batched/unbatched case. I was tired when I tried last night though. Help is appreciated 🥲

genesis/utils/ring_buffer.py
```diff
@@ -67,7 +67,8 @@ def at(self, idx: int | torch.Tensor) -> torch.Tensor:
         rel_idx = (self._idx_ptr.value - idx) % self.N
         if isinstance(idx, torch.Tensor):
             batch_size = len(idx) if idx.ndim > 0 else 1
-            return self.buffer[rel_idx, :batch_size]
+            # using [tensor_indices, slice] results in an extra dimension
+            return self.buffer[rel_idx, :batch_size][:, 0]
```
Milotrince: I think there is still something funky going on with the indexing and/or delay handling. I will try to add another test tomorrow.

genesis/sensors/contact_force.py
```diff
@@ -127,9 +123,12 @@ def _update_shared_ground_truth_cache(
         cls, shared_metadata: ContactSensorMetadata, shared_ground_truth_cache: torch.Tensor
     ):
         all_contacts = shared_metadata.solver.collider.get_contacts(as_tensor=True, to_torch=True)
-        contact_links = torch.cat([all_contacts["link_a"], all_contacts["link_b"]], dim=-1)
-        is_contact = (contact_links.unsqueeze(-2) == shared_metadata.expanded_links_idx).any(dim=-1)
-        shared_ground_truth_cache.copy_(is_contact)
+        if all_contacts["link_a"].numel() == 0:
```
duburcqa, line 128: DELETE COMMENT (I was wrong)

genesis/sensors/contact_force.py
```diff
@@ -127,9 +123,12 @@ def _update_shared_ground_truth_cache(
...
+            shared_ground_truth_cache.fill_(False)
+        else:
+            contact_links = torch.cat([all_contacts["link_a"], all_contacts["link_b"]], dim=-1)
+            is_contact = (contact_links[..., -1, None] == shared_metadata.expanded_links_idx).any(dim=-1)
+            shared_ground_truth_cache.copy_(is_contact)
```
duburcqa: Concatenating is probably less efficient that loop over the two arrays separately I think.

genesis/sensors/contact_force.py
duburcqa: Just do `is_contact = any(torch.isin(links_idx, shared_metadata.expanded_links_idx).any(dim=0) for links_idx in (all_contacts["link_a"], all_contacts["link_b"]))`

genesis/utils/ring_buffer.py
duburcqa: I let you in the wrong direction here sorry. Now I understand what you are trying to achieve. In this specific case, you MUST use a tensor for the second index. But it is not an issue here, since you are already doing fancy indexing in the first place so allocating memory for the extracted buffer is expected... 

If you pass two tensors of the same length `indices_1, indices_2`, it will be interpreted as a tuple fancy index `(indices_1, indices_2)`, which will return a 1D vector (whatever the dimensionality of the input tensor) that corresponds to:
```
out[0] = in[indices_1[0], indices_2[0]]
out[1] = in[indices_1[1], indices_2[1]]
[...]
out[N] = in[indices_1[N], indices_2[N]]
```

If you mix a single tensor index with slices, it is completely different. It is a same as extracted sub-matrices that so stack by together, followed (or preceded, it is the same) by slicing on the other dimensions:
```
out = torch.stack([in[indices_1[i]] for i in indices_1], dim=0)[:, :N]
# Or equivalently
out[0, :] = in[indices_1[0], :N]
out[1, :] = in[indices_1[1], :N]
[...]
out[N, :] = in[indices_1[1], :N]
```

Not much in common!

duburcqa: You should update the doc to clarify what this method is supposed to do when a tensor index is specified.

By the way, the doc could be improved "Index of the element to get, where 0 is the latest element, 1 is the second latest, etc." --> "Index of the element to get, ordered from most recent to oldest [...]"

duburcqa (approved):

genesis/utils/ring_buffer.py
duburcqa: Fixed!

genesis/sensors/contact_force.py
duburcqa: fixed!

genesis/sensors/contact_force.py
duburcqa: fixed!

duburcqa (dismissed):

duburcqa (approved):

PR #1730, opened 2025-09-16
[FEATURE] Enable pure function in CI

YilingQiao:
## Description

Add additional variable to cache key to make sure pure function works.

TODO:
move some of the dynamic properyies into field / ndarray

## Related Issue

## Checklist:

### DIFF
--- a/.github/workflows/generic.yml
+++ b/.github/workflows/generic.yml
@@ -45,6 +45,7 @@ jobs:
       PY_COLORS: 1
       GS_CACHE_FILE_PATH: ".cache/genesis"
       GS_USE_NDARRAY: ${{ matrix.GS_USE_NDARRAY }}
+      GS_BETA_PURE: ${{ matrix.GS_USE_NDARRAY }} # we turn on beta pure for ndarray
       GS_TORCH_FORCE_CPU_DEVICE: "1"
       TI_OFFLINE_CACHE: "1"
       TI_OFFLINE_CACHE_CLEANING_POLICY: "never"
--- a/genesis/utils/array_class.py
+++ b/genesis/utils/array_class.py
@@ -2032,6 +2032,7 @@ cache_value = partial(dataclasses.field, metadata={FIELD_METADATA_CACHE_VALUE: T
 @dataclasses.dataclass
 class StaticRigidSimCacheKey:
     para_level: int = cache_value()
+    use_gjk_collision: bool = cache_value()
     use_hibernation: bool = cache_value()
     batch_links_info: bool = cache_value()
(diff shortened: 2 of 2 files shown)


.github/workflows/generic.yml
```diff
@@ -45,6 +45,7 @@ jobs:
       PY_COLORS: 1
       GS_CACHE_FILE_PATH: ".cache/genesis"
       GS_USE_NDARRAY: ${{ matrix.GS_USE_NDARRAY }}
+      GS_BETA_PURE: ${{ matrix.GS_USE_NDARRAY }} # we turn on beta pure for ndarray
```
hughperkins, line 48: When on Mac, and using any gpu (vulkan or Metal), and using ndarray, this should be forced to 0, per my understanding.

I'm unclear whether other vulkan gpus have the same issue, since we've only ever tested on Mac Metal GPU, and CUDA GPU. Seems very plausible though.

genesis/utils/array_class.py
```diff
@@ -2044,11 +2045,37 @@ class StaticRigidSimCacheKey:
     integrator: int = cache_value()
     sparse_solve: bool = cache_value()
     solver_type: int = cache_value()
+    # dynamic properties
```
hughperkins, line 2048: what do you mean by 'dynamic'? Do you mean, you intend to move this from static config into ndarray/fields later? Maybe add `# TODO`, and a short descriptin?

.github/workflows/generic.yml
hughperkins: oh, this is just for CI anyway, not general purpose user usage 🤔 Well, I suppose if CI passes then we are good :)

hughperkins (approved):
Great!

genesis/utils/array_class.py
YilingQiao: The comments are below
```
# TODO: we should store those properties into field/ndarray to avoid recompilation when they are changed
```

.github/workflows/generic.yml
YilingQiao: the CI fails though. Now sure how it is related to my change

hughperkins: I was assuming it was something that we didn't fix yet, rather than somethign that this PR broke.

Maybe there is some other primitive value that is being access from a data oriented, for example?

hughperkins: (though I dont see what kernel they are calling 🤔 )

duburcqa: Working on the CI issue.

PR #1733, opened 2025-09-17
[FEATURE] Differentiable constraint solver

SonSang:
## Description

Added the backward pass of the constraint solver. Also added a unit test based on the comparison to the finite difference.

## Related Issue

## Motivation and Context

Added a unit test for validating the backward pass. You can run the test as follows:

```
pytest "tests/test_diff_rigid.py::test_diff_contact"
```

## Checklist:

### DIFF
--- a/genesis/engine/solvers/rigid/constraint_solver_decomp.py
+++ b/genesis/engine/solvers/rigid/constraint_solver_decomp.py
@@ -449,6 +449,278 @@ class ConstraintSolver:
             self._solver._static_rigid_sim_cache_key,
         )
 
+    def backward(self, dL_dqacc):
+        # Copy upstream gradients
+        self.constraint_state.dL_dqacc.from_numpy(dL_dqacc)
+
+        # 1. We first need to find a solution to A^T * u = g system.
+        constraint_solver_backward_kernel_solve_Au_g(
+            entities_info=self._solver.entities_info,
+            rigid_global_info=self._solver._rigid_global_info,
+            constraint_state=self.constraint_state,
+            static_rigid_sim_config=self._solver._static_rigid_sim_config,
+        )
+
+        # 2. Using the solution u, we can compute the gradients of the input variables.
+        constraint_solver_backward_kernel_compute_gradients(
+            entities_info=self._solver.entities_info,
+            constraint_state=self.constraint_state,
+            static_rigid_sim_config=self._solver._static_rigid_sim_config,
+        )
+
+
+@ti.func
+def func_backward_compute_Ap(
+    entities_info: array_class.EntitiesInfo,
+    constraint_state: array_class.ConstraintState,
+    rigid_global_info: array_class.RigidGlobalInfo,
+    static_rigid_sim_config: ti.template(),
+    i_b,
+):
+    """
+    Compute Ap = (M + J^T * diag(D) * J) * p, where
+
+    M = mass matrix, J = Jacobian, D = diagonal matrix of efc_D
+    p = search direction
+    """
+    n_dofs = constraint_state.bw_p.shape[0]
+    for i_d in range(n_dofs):
+        constraint_state.bw_Ap[i_d, i_b] = 0.0
+
+    # Mp: Block multiplication
+    n_entities = entities_info.n_links.shape[0]
+    for i_e in range(n_entities):
+        for i_d1 in range(entities_info.dof_start[i_e], entities_info.dof_end[i_e]):
+            acc = gs.ti_float(0.0)
+            for i_d2 in range(entities_info.dof_start[i_e], entities_info.dof_end[i_e]):
+                acc += rigid_global_info.mass_mat[i_d1, i_d2, i_b] * constraint_state.bw_p[i_d2, i_b]
+            constraint_state.bw_Ap[i_d1, i_b] += acc
+
+    # tmp = J v
+    for i_c in range(constraint_state.n_constraints[i_b]):
+        jv = 0.0
+        if ti.static(static_rigid_sim_config.sparse_solve):
+            for k in range(constraint_state.jac_n_relevant_dofs[i_c, i_b]):
+                i_d = constraint_state.jac_relevant_dofs[i_c, k, i_b]
+                jv += constraint_state.jac[i_c, i_d, i_b] * constraint_state.bw_p[i_d, i_b]
+        else:
+            for i_d in range(n_dofs):
+                jv += constraint_state.jac[i_c, i_d, i_b] * constraint_state.bw_p[i_d, i_b]
+        # only active constraints contribute
+        jv *= constraint_state.efc_D[i_c, i_b] * constraint_state.active[i_c, i_b]
+        # out += J^T (D * J v)
+        if ti.static(static_rigid_sim_config.sparse_solve):
+            for k in range(constraint_state.jac_n_relevant_dofs[i_c, i_b]):
+                i_d = constraint_state.jac_relevant_dofs[i_c, k, i_b]
+                constraint_state.bw_Ap[i_d, i_b] += constraint_state.jac[i_c, i_d, i_b] * jv
+        else:
+            for i_d in range(n_dofs):
+                constraint_state.bw_Ap[i_d, i_b] += constraint_state.jac[i_c, i_d, i_b] * jv
+
+
+@ti.kernel
+def constraint_solver_backward_kernel_solve_Au_g(
+    entities_info: array_class.EntitiesInfo,
+    rigid_global_info: array_class.RigidGlobalInfo,
+    constraint_state: array_class.ConstraintState,
+    static_rigid_sim_config: ti.template(),
+):
+    """
+    Solve A * u = g, where
+    A = M + J^T * diag(D) * J,  (M = mass matrix, J = Jacobian, D = diagonal matrix of efc_D)
+    g = dL_dqacc
+    Note that A is SPD, so we solve A * u = g. Here we use CG solver.
+    TODO: Implement Cholesky decomposition if we used Newton solver for forward pass.
+    """
+    n_dofs = constraint_state.bw_u.shape[0]
+    _B = constraint_state.bw_u.shape[1]
+
+    # Initialize u
+    for i_d, i_b in ti.ndrange(n_dofs, _B):
+        constraint_state.bw_u[i_d, i_b] = 0.0
+
+    if ti.static(static_rigid_sim_config.solver_type == gs.constraint_solver.Newton):
+        """
+        Since we already have the Cholesky decomposition of A (= L * L^T), we can use it to solve A * u = g.
+        """
+        for i_b in range(_B):
+            # z = L^{-1} g  (forward substitution)
+            # Save solution to bw_r
+            for i_d in range(n_dofs):
+                z = constraint_state.dL_dqacc[i_d, i_b]
+                for j_d in range(i_d):
+                    z -= constraint_state.nt_H[i_d, j_d, i_b] * constraint_state.bw_r[j_d, i_b]
+                z /= constraint_state.nt_H[i_d, i_d, i_b]
+                constraint_state.bw_r[i_d, i_b] = z
+
+            # u = L^{-T} z  (back substitution)
+            for i_d_ in range(n_dofs):
+                i_d = n_dofs - 1 - i_d_
+                u = constraint_state.bw_r[i_d, i_b]
+                for j_d in range(i_d + 1, n_dofs):
+                    u -= constraint_state.nt_H[j_d, i_d, i_b] * constraint_state.bw_u[j_d, i_b]
+                u /= constraint_state.nt_H[i_d, i_d, i_b]
+                constraint_state.bw_u[i_d, i_b] = u
+    else:
+        """
+        Use CG solver for solving A * u = g.
+        """
+        # 2. Local buffers for solving A * u = g
+        # Initialize r, p with dL_dqacc
+        for i_d, i_b in ti.ndrange(n_dofs, _B):
+            # Residual: g - A * 0 (u = 0)
+            constraint_state.bw_r[i_d, i_b] = constraint_state.dL_dqacc[i_d, i_b]
+            # Search direction: p = r
+            constraint_state.bw_p[i_d, i_b] = constraint_state.bw_r[i_d, i_b]
+
+        # 3. Solve A * u = g, parallelized over batch dimension
+        for i_b in range(_B):
+            # Compute Ap for the current search direction
+            for it in range(static_rigid_sim_config.iterations):
+                func_backward_compute_Ap(
+                    entities_info=entities_info,
+                    rigid_global_info=rigid_global_info,
+                    constraint_state=constraint_state,
+                    static_rigid_sim_config=static_rigid_sim_config,
+                    i_b=i_b,
+                )
+
+                # alpha = (r,r)/(p,Hp)
+                num = gs.ti_float(0.0)
+                den = gs.ti_float(0.0)
+                for i_d in range(n_dofs):
+                    num += constraint_state.bw_r[i_d, i_b] * constraint_state.bw_r[i_d, i_b]
+                    den += constraint_state.bw_p[i_d, i_b] * constraint_state.bw_Ap[i_d, i_b]
+                alpha = num / ti.max(den, gs.EPS)
+
+                # print(f"[Au = g] iteration {it}: residual norm^2 = {num:.20g}")
+
+                # u += alpha p ; r -= alpha Hp
+                for i_d in range(n_dofs):
+                    constraint_state.bw_u[i_d, i_b] += alpha * constraint_state.bw_p[i_d, i_b]
+                    constraint_state.bw_r[i_d, i_b] -= alpha * constraint_state.bw_Ap[i_d, i_b]
+
+                # check tol (optional: per-batch)
+                # TODO: Might need lower tolerance?
+                if num < gs.EPS:
+                    break
+
+                # beta = (r_new,r_new)/(r_old,r_old)
+                num_new = gs.ti_float(0.0)
+                for i_d in range(n_dofs):
+                    num_new += constraint_state.bw_r[i_d, i_b] * constraint_state.bw_r[i_d, i_b]
+                beta = num_new / ti.max(num, gs.EPS)
+
+                # p = r + beta p
+                for i_d in range(n_dofs):
+                    constraint_state.bw_p[i_d, i_b] = (
+                        constraint_state.bw_r[i_d, i_b] + beta * constraint_state.bw_p[i_d, i_b]
+                    )
+
+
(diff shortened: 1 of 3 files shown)


genesis/engine/solvers/rigid/constraint_solver_decomp.py
```diff
@@ -451,6 +451,278 @@ def delete_weld_constraint(self, link1_idx, link2_idx, envs_idx=None, *, unsafe=
...
+        # 2. Using the solution u, we can compute the gradients of the input variables.
+        constraint_solver_backward_kernel_compute_gradients(
+            entities_info=self._solver.entities_info,
+            constraint_state=self.constraint_state,
+            static_rigid_sim_config=self._solver._static_rigid_sim_config,
+        )
+
+
+@ti.func
+def func_backward_compute_Ap(
```
YilingQiao: can we move those backward functions to a new file (e.g. `backward_constraint_solver.py`)

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,216 @@
...
+
+        dL_error /= TRIALS
+        return dL_error
+
+    dL_dforce_error = compute_dL_error(dL_dforce, "force")
+    dL_daref_error = compute_dL_error(dL_daref, "aref")
+    dL_defc_D_error = compute_dL_error(dL_defc_D, "efc_D")
+    dL_djac_error = compute_dL_error(dL_djac, "jac")
+    dL_dM_error = compute_dL_error(dL_dM, "mass")
+
```
YilingQiao: can you use 
```
assert_allclose(x, 0)
```

genesis/engine/solvers/rigid/constraint_solver_decomp.py
```diff
@@ -451,6 +451,278 @@ def delete_weld_constraint(self, link1_idx, link2_idx, envs_idx=None, *, unsafe=
...
+    for i_e in range(n_entities):
+        for i_d1 in range(entities_info.dof_start[i_e], entities_info.dof_end[i_e]):
+            acc = gs.ti_float(0.0)
+            for i_d2 in range(entities_info.dof_start[i_e], entities_info.dof_end[i_e]):
+                acc += rigid_global_info.mass_mat[i_d1, i_d2, i_b] * constraint_state.bw_p[i_d2, i_b]
+            constraint_state.bw_Ap[i_d1, i_b] += acc
+
+    # tmp = J v
+    for i_c in range(constraint_state.n_constraints[i_b]):
+        jv = 0.0
```
duburcqa: prefer `gs.ti_float(0.0)`

genesis/engine/solvers/rigid/constraint_solver_decomp.py
```diff
@@ -451,6 +451,278 @@ def delete_weld_constraint(self, link1_idx, link2_idx, envs_idx=None, *, unsafe=
...
+    rigid_global_info: array_class.RigidGlobalInfo,
+    constraint_state: array_class.ConstraintState,
+    static_rigid_sim_config: ti.template(),
+):
+    """
+    Solve A * u = g, where
+    A = M + J^T * diag(D) * J,  (M = mass matrix, J = Jacobian, D = diagonal matrix of efc_D)
+    g = dL_dqacc
+    Note that A is SPD, so we solve A * u = g. Here we use CG solver.
+    TODO: Implement Cholesky decomposition if we used Newton solver for forward pass.
```
duburcqa: You actually implemented this feature!

genesis/engine/solvers/rigid/constraint_solver_decomp.py
```diff
@@ -451,6 +451,278 @@ def delete_weld_constraint(self, link1_idx, link2_idx, envs_idx=None, *, unsafe=
...
+    entities_info: array_class.EntitiesInfo,
+    rigid_global_info: array_class.RigidGlobalInfo,
+    constraint_state: array_class.ConstraintState,
+    static_rigid_sim_config: ti.template(),
+):
+    """
+    Solve A * u = g, where
+    A = M + J^T * diag(D) * J,  (M = mass matrix, J = Jacobian, D = diagonal matrix of efc_D)
+    g = dL_dqacc
+    Note that A is SPD, so we solve A * u = g. Here we use CG solver.
```
duburcqa: * Could you describe with a short sentence what does  "Solve A * u = g" means in practice? Because for someone that is not very familiar with Rigid Body constraint solver it is unfathomable. Same remark for the previous method `constraint_solver_backward_kernel_solve_Au_g`.
* Skip not between short description and details. Also avoid acronyms if possible (or just specified in parentheses), i.e. "Note that A is Semi-Positive Definite (SPD), [...]"
* Note is not very clear.

genesis/engine/solvers/rigid/constraint_solver_decomp.py
```diff
@@ -451,6 +451,278 @@ def delete_weld_constraint(self, link1_idx, link2_idx, envs_idx=None, *, unsafe=
...
+    _B = constraint_state.bw_u.shape[1]
+
+    # Initialize u
+    for i_d, i_b in ti.ndrange(n_dofs, _B):
+        constraint_state.bw_u[i_d, i_b] = 0.0
+
+    if ti.static(static_rigid_sim_config.solver_type == gs.constraint_solver.Newton):
+        """
+        Since we already have the Cholesky decomposition of A (= L * L^T), we can use it to solve A * u = g.
+        """
```
duburcqa: Inline comment using `#` would do fine. Keep this formatting for docstring.

genesis/engine/solvers/rigid/constraint_solver_decomp.py
```diff
@@ -451,6 +451,278 @@ def delete_weld_constraint(self, link1_idx, link2_idx, envs_idx=None, *, unsafe=
...
+                i_d = n_dofs - 1 - i_d_
+                u = constraint_state.bw_r[i_d, i_b]
+                for j_d in range(i_d + 1, n_dofs):
+                    u -= constraint_state.nt_H[j_d, i_d, i_b] * constraint_state.bw_u[j_d, i_b]
+                u /= constraint_state.nt_H[i_d, i_d, i_b]
+                constraint_state.bw_u[i_d, i_b] = u
+    else:
+        """
+        Use CG solver for solving A * u = g.
+        """
```
duburcqa: Same. You can skip space if you want it to look like a title.

genesis/engine/solvers/rigid/constraint_solver_decomp.py
```diff
@@ -451,6 +451,278 @@ def delete_weld_constraint(self, link1_idx, link2_idx, envs_idx=None, *, unsafe=
...
+
+                # alpha = (r,r)/(p,Hp)
+                num = gs.ti_float(0.0)
+                den = gs.ti_float(0.0)
+                for i_d in range(n_dofs):
+                    num += constraint_state.bw_r[i_d, i_b] * constraint_state.bw_r[i_d, i_b]
+                    den += constraint_state.bw_p[i_d, i_b] * constraint_state.bw_Ap[i_d, i_b]
+                alpha = num / ti.max(den, gs.EPS)
+
+                # print(f"[Au = g] iteration {it}: residual norm^2 = {num:.20g}")
```
duburcqa: Remove this line.

genesis/engine/solvers/rigid/constraint_solver_decomp.py
```diff
@@ -451,6 +451,278 @@ def delete_weld_constraint(self, link1_idx, link2_idx, envs_idx=None, *, unsafe=
...
+
+    # clear grads
+    constraint_state.dL_dforce.fill(0.0)
+    constraint_state.dL_daref.fill(0.0)
+    constraint_state.dL_defc_D.fill(0.0)
+    constraint_state.dL_djac.fill(0.0)
+    constraint_state.dL_dM.fill(0.0)
+    constraint_state.bw_Ju.fill(0.0)
+    constraint_state.bw_y.fill(0.0)
+    constraint_state.bw_w.fill(0.0)
```
duburcqa: I like this syntax but I heard it was more efficient to use a single for-loop to reset all variables, because otherwise it is create multiple independent parallel context for each `fill`, and overheat will add up.

genesis/utils/array_class.py
```diff
@@ -277,6 +292,35 @@ def get_constraint_state(constraint_solver, solver):
...
+            shape=f_batch(
+                len_constraints_,
+            ),
+        ),
+        "bw_w": V(
+            dtype=gs.ti_float,
+            shape=f_batch(
+                len_constraints_,
+            ),
+        ),
```
duburcqa: One line. Do not spread declarations it makes everything unreadable.

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,216 @@
...
+import genesis as gs
+from genesis.utils.misc import ti_to_torch
+from genesis.engine.solvers.rigid.constraint_solver_decomp import func_init_solver, func_solve
+from genesis.engine.solvers.rigid.rigid_solver_decomp import kernel_step_1
+
+
+@pytest.mark.required
+@pytest.mark.field_only
+@pytest.mark.precision("64")
```
duburcqa: Why do you require 64bits precision?

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,216 @@
...
+@pytest.mark.parametrize("backend", [gs.cpu, gs.gpu])
```
duburcqa: Why do you require running on both CPU and GPU? Do you think one could fail but not the other? If one of the two is strictly more likely to fail than the other, then just test this one but not both?

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,216 @@
...
+def test_diff_solver(backend):
+    torch.manual_seed(0)
```
duburcqa: This is useless, it is already enforced automatically by some fixture.

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,216 @@
...
+        rigid_options=gs.options.RigidOptions(
+            # We use Newton's method because it converges faster than CG, and therefore gives better gradient estimation
+            # when using finite difference method
+            constraint_solver=gs.constraint_solver.Newton,
+        ),
+        show_viewer=False,
+    )
+
+    plane = scene.add_entity(gs.morphs.Plane(pos=(0, 0, 0)))
+    box = scene.add_entity(gs.morphs.Box(size=(1, 1, 1), pos=(10, 10, 0.49)))
```
duburcqa: Prefer multiple lines in this context even if it fits in one line. It improves readability.

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,216 @@
...
+    franka = scene.add_entity(
+        gs.morphs.MJCF(file="xml/franka_emika_panda/panda.xml"),
+    )
+
+    scene.build()
+    rigid_solver = scene._sim.rigid_solver
+    constraint_solver = rigid_solver.constraint_solver
+
+    qpos = np.array([-1.0124, 1.5559, 1.3662, -1.6878, -1.5799, 1.7757, 1.4602, 0.04, 0.04])
+    franka.set_qpos(qpos)
```
duburcqa: Just do `franka.set_qpos([-1.0124, 1.5559, 1.3662, -1.6878, -1.5799, 1.7757, 1.4602, 0.04, 0.04])`.

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,216 @@
...
+        updated_acc_smooth = torch.linalg.solve(input_mass.squeeze(-1), input_force.squeeze(-1))
+        input_acc_smooth = updated_acc_smooth.unsqueeze(-1).detach().cpu().numpy()
+
+        rigid_solver.dofs_state.acc_smooth.from_numpy(input_acc_smooth)
+
+        constraint_solver_resolve()
+        output_qacc = constraint_solver.qacc.to_numpy()
+        th_output_qacc = torch.from_numpy(output_qacc).to(device=target_qacc.device)
+        loss = ((th_output_qacc - target_qacc) ** 2).mean()
+        return loss
```
duburcqa: What about leveraging built-in pytest monkey patching mechanism?

We are already doing it at several places, it seems to fit perfectly for this use case. See the unit tests `test_plotter`, `test_interactive_viewer_key_press`, or `test_batch_deformable_render`. The official doc is available [here](https://docs.pytest.org/en/stable/how-to/monkeypatch.html).

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,216 @@
...
+    init_input_aref = ti_to_torch(constraint_solver.constraint_state.aref)
+    init_input_efc_D = ti_to_torch(constraint_solver.constraint_state.efc_D)
+    init_input_force = ti_to_torch(rigid_solver.dofs_state.force)
+
+    # [acc_smooth] is dependent on [force], not an independent input ---> No need to compute gradient for it
+    init_input_acc_smooth = ti_to_torch(rigid_solver.dofs_state.acc_smooth)
+
+    # Initial output of the constraint solver
+    init_output_qacc = ti_to_torch(constraint_solver.qacc)
+    target_qacc = torch.rand_like(init_output_qacc) * init_output_qacc.abs().mean()
```
duburcqa: You should use numpy random instead of torch rand, because torch rand is machine specific and backend specific (generated numbers for the same seed are different between CPU, CUDA, MPS...etc), which is undesirable as it increases flakiness. See `test_render_api_advanced` as an example.

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,216 @@
...
+
+    # Number of constraints
+    n_constraints = ti_to_torch(constraint_solver.n_constraints)
```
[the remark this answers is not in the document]
duburcqa: This looks about right!

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,216 @@
...
+
+    # Solve the constraint solver and get the output
+    output_qacc = constraint_solver.qacc.to_numpy()
```
duburcqa: I recently added (yesterday) `ti_to_numpy` if you want!

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,216 @@
...
+    th_output_qacc = torch.from_numpy(output_qacc).to(device=target_qacc.device).requires_grad_(True)
```
duburcqa: I would prefer `gs.device`. We are trying to enforce consistent device everywhere, it is a good way to assert everything is right as the test may fail if `target_qacc.device != gs.device`

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,216 @@
...
+
+    # Compute loss and gradient of the output
+    loss = ((th_output_qacc - target_qacc) ** 2).mean()
+    dL_dqacc = torch.autograd.grad(loss, th_output_qacc)[0].cpu().numpy()
```
duburcqa: `torch_to_array(torch.autograd.grad(loss, th_output_qacc)[0])`

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,216 @@
...
+    ### Compute directional derivatives along random directions
+    FD_EPS = 1e-4
+    TRIALS = 100
+    errors = {}
+
+    def compute_dL_error(dL_dx, x_type):
+        dL_error = 0.0
+        for trial in range(TRIALS):
+            rand_dx = torch.randn_like(dL_dx)
+            rand_dx = torch.nn.functional.normalize(rand_dx, dim=0 if x_type in ["force", "aref", "efc_D"] else [0, 1])
```
duburcqa: Prefer tuple over list for immutable sequences as a general good practice, even if it is not a big deal here.

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,216 @@
...
+
+    ### Compute directional derivatives along random directions
+    FD_EPS = 1e-4
+    TRIALS = 100
+    errors = {}
+
+    def compute_dL_error(dL_dx, x_type):
+        dL_error = 0.0
+        for trial in range(TRIALS):
+            rand_dx = torch.randn_like(dL_dx)
```
duburcqa: Use numpy random.

tests/test_diff_rigid.py
duburcqa: Beyond that, I would suggest normalizing yourself using `ti.maximum(torch.linalg.norm(rand_dx, dim=[...]), gs.EPS)`instead of relying on `torch.nn.functional.normalize`, which is confusing in this context I think because fairly ML-oriented.

tests/test_diff_rigid.py
duburcqa: No, you should! Because it will print the value if case of failure automatically, therefore you should also get rid of the error message that is no longer relevant after this, i.e.:
```python
assert_allclose(dL_dforce_error, 0.0, rtol=rtol)
```

tests/test_diff_rigid.py
```diff
@@ -0,0 +1,216 @@
...
+            dL_error += (dL - dL_fd).abs() / max(dL.abs(), dL_fd.abs(), gs.EPS)
+
+        dL_error /= TRIALS
+        return dL_error
+
+    dL_dforce_error = compute_dL_error(dL_dforce, "force")
+    dL_daref_error = compute_dL_error(dL_daref, "aref")
+    dL_defc_D_error = compute_dL_error(dL_defc_D, "efc_D")
+    dL_djac_error = compute_dL_error(dL_djac, "jac")
+    dL_dM_error = compute_dL_error(dL_dM, "mass")
```
duburcqa: I would recommend NOT defining an inline function `compute_dL_error` and just do a for loop. Besides, just evaluate the assert on-the-fly instead of computing all the errors first, i.e.:
```python
for dL_dx, x_type in ((dL_dforce, "force"), (dL_daref, "aref"), (dL_defc_D, "efc_D"), (dL_djac, "jac"), (dL_dM, "mass")):
    dL_error = 0.0
    for trial in range(TRIALS):
        rand_dx = torch.randn_like(dL_dx)
        [...]
    assert_allclose(dL_error, 0.0, rtol=rtol, err_msg=f"Error is too large for '{mass}'.")
```

tests/test_diff_rigid.py
SonSang: This is because we use small perturbation (`1e-4`) for computing reliable gradient using finite difference. If we want to use 32bits, the perturbation should be larger than this but then the gradient from finite difference becomes unreliable. But we can solve optimization problem with 32bits, I'll add another test for this.

duburcqa: If it is only for the unit test, this is perfectly fine. But you should document it. I was afraid it was a limitation of the simulator itself, similar to SAPCoupler that strictly requires 64bits precision computations at the moment.

YilingQiao:
@duburcqa can we merge it?

duburcqa (approved):

PR #1620, opened 2025-08-22
[FEATURE] Animate cloth by attaching its particles to RigidLinks

gasnica:
## Description
* attach (animate) cloth by rigid links
  * rigid link is not affected by cloth's mass
* detaching cloth sometimes causes the cloth to disappear. This has not been debugged yet.
* attachment force is clipped to avoid super-high corrective velocities at smaller time steps (via clamped_inv_dt in LegacyCoupler)

## Known issues
* this is one-way interaction, while cloth-link collisions are two-way interactions, hence unrealistic spinning or explosion can occur if the particles are attached in an overlapping/colliding configuration
  * when attaching cloth to a robot, this may cause unexpected forces on joints

## How Has This Been / Can This Be Tested?
Manual/visual testing only, with the attached demo that teleports the body around.

## Screenshots (if appropriate):

https://github.com/user-attachments/assets/47c14bf5-5189-49df-b9f3-ed3e199f3caa

## Checklist:

- [x] I read the **CONTRIBUTING** document.
- [x] I followed the `Submitting Code Changes` section of **CONTRIBUTING** document.
- [x] I tagged the title correctly (including BUG FIX/FEATURE/MISC/BREAKING)
- [ ] I updated the [documentation](https://github.com/Genesis-Embodied-AI/genesis-doc) accordingly or no change is needed.
- [ ] I tested my changes and added instructions on how to test it for reviewers.

- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1767, opened 2025-09-22
[FEATRE] Merge and Add a Test for #1620 (Animate cloth by attaching its particles to RigidLinks)

YilingQiao:
## Description

Merge the latest main and add a unit test for #1620 

I cannot modify that PR so created a new PR for it.

Try 
```
python examples/coupling/cloth_attached_to_rigid.py -v
# or 
pytest --vis -v tests/test_fem.py::test_cloth_attachment

```

## Related Issue

## Checklist:


genesis/engine/solvers/pbd_solver.py
```diff
@@ -912,6 +922,33 @@ def _kernel_set_particles_vel(
                 self.particles[i_p, i_b].vel[i] = vels[i_b_, i_p_, i]
 
     @ti.kernel
+    def _kernel_set_particle_velocity(self, particle_idx: ti.i32, vel: ti.types.ndarray(), i_b: ti.i32):
+        for i in range(3):
+            self.particles[particle_idx, i_b].vel[i] = vel[i]
+        self.particles[particle_idx, i_b].free = 0
```
duburcqa: This method should be removed, it is redundant with `_kernel_set_particles_vel`.

tests/test_pbd.py
```diff
@@ -41,3 +43,58 @@ def test_maxvolume(pbd_material, show_viewer, box_obj_path):
         f"Mesh with maxvolume=0.01 generated {pbd2.n_elems} elements; "
         f"expected more than {pbd1.n_elems} elements without a volume limit."
     )
+
+
+@pytest.mark.required
+def test_cloth_attachment(show_viewer):
```
duburcqa: Could you move the existing unit test `test_attach_cloth` along with this one, and rename in `test_cloth_attach_fixed_point`. Similarly, you should rename yours `test_cloth_attach_rigid_link`.

examples/coupling/cloth_attached_to_rigid.py
```diff
@@ -0,0 +1,117 @@
...
+
+import genesis.utils.geom as gu
+
+from genesis.utils.misc import to_gs_tensor
+
+if TYPE_CHECKING:
+    from genesis.engine.entities.rigid_entity.rigid_entity import RigidEntity
+    from genesis.engine.materials.base import Material
+    from genesis.options.surfaces import Surface
+    from genesis.options.morphs import Morph
```
duburcqa: I think enforcing strict typing in example file is just making them harder to read without significant advantage.

examples/coupling/cloth_attached_to_rigid.py
```diff
@@ -0,0 +1,117 @@
...
+import argparse
+
+import numpy as np
+import torch
+
+import genesis as gs
+
+import genesis.utils.geom as gu
+
+from genesis.utils.misc import to_gs_tensor
```
duburcqa: Remove this import, it is not used.

examples/coupling/cloth_attached_to_rigid.py
```diff
@@ -0,0 +1,117 @@
+from typing import TYPE_CHECKING
+
+import argparse
+
+import numpy as np
+import torch
```
duburcqa: Remove unused import.

examples/coupling/cloth_attached_to_rigid.py
```diff
@@ -0,0 +1,117 @@
...
+    cloth_morph: "Morph" = gs.morphs.Mesh(pos=cloth_pos, scale=cloth_scale, file="meshes/cloth.obj")
+    cloth_material: "Material" = gs.materials.PBD.Cloth()
+    cloth_surface: "Surface" = gs.surfaces.Default(color=(0.2, 0.4, 0.8, 1.0))  # , vis_mode="particle")
+    scene.add_entity(cloth_morph, cloth_material, cloth_surface)
+
+    ########################## build ##########################
+    scene.build(n_envs=0)
+
+    particles_idx = np.array([0, 1, 2, 3, 4, 5, 6, 7])
+    box_link_idx = box.links[0].idx
```
duburcqa: You should rather do `box_link_idx = box.link_start` I think.

examples/coupling/cloth_attached_to_rigid.py
```diff
@@ -0,0 +1,117 @@
...
+    scene.pbd_solver.set_animate_particles_by_link(particles_idx, box_link_idx, scene.rigid_solver.links_state)
+
+    vel = np.array([-0.0, 1.0, 0.0], dtype=np.float32)
+    box.set_dofs_velocity(vel, dofs_idx_local=np.arange(3))
```
duburcqa: Remove numpy, it is useless here:
```python
    particles_idx = [0, 1, 2, 3, 4, 5, 6, 7]
    box_link_idx = box.link_start
    scene.pbd_solver.set_animate_particles_by_link(particles_idx, box_link_idx, scene.rigid_solver.links_state)

    box.set_dofs_velocity([-0.0, 1.0, 0.0], dofs_idx_local=[0, 1, 2])
```

examples/coupling/cloth_attached_to_rigid.py
```diff
@@ -0,0 +1,117 @@
...
+
+    horizon = 50
```
duburcqa: Make this parameter an input argument with argparse.

examples/coupling/cloth_attached_to_rigid.py
```diff
@@ -0,0 +1,117 @@
...
+    for i in range(horizon):
+        scene.step()
+
+    num_teleports = 5
+    for j in range(num_teleports):
```
duburcqa: Same.

examples/coupling/cloth_attached_to_rigid.py
```diff
@@ -0,0 +1,117 @@
...
+        if j == 2:
+            if False:
+                # Clearing attachment causes the cloth to vanish. Haven't debugged.
+                # All is okay when clearing attachment immeidately, or after 1st teleport.
+                # Not okay when clearng attachment after 2+ teleports:
+                # scene.pbd_solver.clear_animate_particles_by_link(particles_idx)
+                pass
+        new_pos = (0.2 * np.sin(j), 0.2 * np.cos(j), 0.5)
```
duburcqa: The use `import math` module instead of numpy for scalar math.

examples/coupling/cloth_attached_to_rigid.py
```diff
@@ -0,0 +1,117 @@
...
+
+    num_teleports = 5
+    for j in range(num_teleports):
+        if j == 2:
+            if False:
+                # Clearing attachment causes the cloth to vanish. Haven't debugged.
+                # All is okay when clearing attachment immeidately, or after 1st teleport.
+                # Not okay when clearng attachment after 2+ teleports:
+                # scene.pbd_solver.clear_animate_particles_by_link(particles_idx)
+                pass
```
duburcqa: This is a serious concern I feel. I'm not comfortable with merging as is. We should at least create a dedicated issue with reproduction script to track this new bug.

examples/coupling/cloth_attached_to_rigid.py
```diff
@@ -0,0 +1,117 @@
...
+        new_pos = (0.2 * np.sin(j), 0.2 * np.cos(j), 0.5)
+        new_rot = gu.euler_to_quat((-30 * j, -40 * j, 70 * j))
+        box.set_pos(new_pos)
+        box.set_quat(new_rot)
+        for i in range(horizon):
+            scene.step()
+
+        # validation of particle positions: todo: move to unit test:
+        unused_f = 0
+        pbd_state: "PBDSolverState" = scene.pbd_solver.get_state(unused_f)
```
duburcqa: What is this?

examples/coupling/cloth_attached_to_rigid.py
```diff
@@ -0,0 +1,117 @@
...
+
+        link_pos = scene.rigid_solver.links[box_link_idx].get_pos()
+        link_quat = scene.rigid_solver.links[box_link_idx].get_quat()
```
duburcqa: This is not the correct way to access link pos / quat. You should rather do:
```python
link = box.links[0]
link_idx = link.idx

link_pos, link_quat = link.get_pos(), link.get_quat()
```

examples/coupling/cloth_attached_to_rigid.py
```diff
@@ -0,0 +1,117 @@
...
+
+        for i_p_ in range(particles_idx.shape[0]):
+            i_p = particles_idx[i_p_]
+            i_env = 0
+            local_pos = scene.pbd_solver.particle_animation_info.local_pos[i_p, i_env].to_numpy()
```
duburcqa: This is overly complicated. Just do:
```python
for i_p in particles_idx:
    local_pos = scene.pbd_solver.particle_animation_info.local_pos[i_p, 0].to_numpy()
```

Anyway, this has nothing to do in an example as it exposes private API and does nothing with it.

genesis/engine/couplers/legacy_coupler.py
```diff
@@ -593,8 +596,98 @@ def pbd_rigid(self, f: ti.i32):
                             i_b,
                         )
 
+    @ti.kernel
+    def kernel_pbd_rigid_set_animate_particles_by_link(
+        self,
+        particles_idx: ti.types.ndarray(),  # 1d array
+        link_idx: ti.i32,
+        links_state: LinksState,
+        envs_idx: ti.types.ndarray(),  # 1d array
```
duburcqa: I would suggest swapping these two input arguments to better align with other kernels.

genesis/engine/couplers/legacy_coupler.py
```diff
@@ -593,8 +596,98 @@ def pbd_rigid(self, f: ti.i32):
                             i_b,
                         )
 
+    @ti.kernel
+    def kernel_pbd_rigid_set_animate_particles_by_link(
+        self,
+        particles_idx: ti.types.ndarray(),  # 1d array
```
duburcqa: I think this can be enforced using `ti.types.ndarray(ndim=1)`

duburcqa: For the record, all kernels that accept `particles_idx` as input argument of particle-based solvers now expect a 2D array with shape `(len(envs_idx), num_particles)`

genesis/engine/couplers/legacy_coupler.py
```diff
@@ -593,8 +596,98 @@ def pbd_rigid(self, f: ti.i32):
                             i_b,
                         )
 
+    @ti.kernel
+    def kernel_pbd_rigid_set_animate_particles_by_link(
```
duburcqa: It does not feel right to access and modify internals of PDB solver inside the coupler. The coupler should rather call some public or private method that is exposed in PDB solver itself.

genesis/engine/couplers/legacy_coupler.py
```diff
@@ -603,58 +696,61 @@ def _func_pbd_collide_with_rigid_geom(self, i, pos_world, vel, mass, normal_prev
...
+        new_vel = vel
         if signed_dist < self.pbd_solver.particle_size / 2:  # skip non-penetration particles
 
             rvel = vel - vel_rigid
-            rvel_normal_magnitude = rvel.dot(normal_rigid)  # negative if inward
-            rvel_tan = rvel - rvel_normal_magnitude * normal_rigid
-            rvel_tan_norm = rvel_tan.norm(gs.EPS)
+            rvel_normal_magnitude = rvel.dot(contact_normal)  # negative if inward
+            _unused_rvel_tan = rvel - rvel_normal_magnitude * contact_normal
+            _unused_rvel_tan_norm = _unused_rvel_tan.norm(gs.EPS)
```
duburcqa: We don't use `_unused_` prefix for unused variables in Genesis, just `_`.

Moreover, avoid doing unused computations, this is just confusing.

genesis/engine/couplers/legacy_coupler.py
```diff
@@ -603,58 +696,61 @@ def _func_pbd_collide_with_rigid_geom(self, i, pos_world, vel, mass, normal_prev
...
 
-            delta_v_norm = delta_normal_magnitude * normal_rigid
-            vel = v_norm
+            # vel = new_vel
 
             #################### particle -> rigid ####################
-            delta_mv = mass * delta_v_norm
-            force = (-delta_mv / self.rigid_solver._substep_dt) * (1 - energy_loss)
+            delta_mv = mass * (new_vel - vel)
+            force = (-delta_mv / self.rigid_solver._substep_dt) * (1 - _unused_energy_loss)
```
duburcqa: Why some `_unused_energy_loss` is used in computations? It makes no sense.

genesis/engine/couplers/legacy_coupler.py
```diff
@@ -603,58 +696,61 @@ def _func_pbd_collide_with_rigid_geom(self, i, pos_world, vel, mass, normal_prev
...
 
             self.rigid_solver._func_apply_external_force(
-                pos_world,
+                pos,
```
duburcqa: `pos_world` was a better name.

genesis/engine/solvers/pbd_solver.py
```diff
@@ -148,26 +159,23 @@ def init_particle_fields(self):
...
-        )
+        self.particles = struct_particle_state.field(shape=batched_shape, layout=ti.Layout.SOA)
+        self.particles_reordered = struct_particle_state.field(shape=batched_shape, layout=ti.Layout.SOA)
 
-        self.particles_render = struct_particle_state_render.field(
-            shape=self._batch_shape(self._n_particles), layout=ti.Layout.SOA
-        )
+        self.particle_animation_info = self.ParticleAnimateByLinkInfo.field(shape=batched_shape, layout=ti.Layout.SOA)
+        self.particle_animation_info.link_idx.fill(-1)
+        self.particle_animation_info.local_pos.fill(gs.ti_vec3(0.0, 0.0, 0.0))
```
duburcqa: I don't think this should be part of PDB solver, since it is completely externally managed by the legacy coupler.

tests/test_pbd.py
```diff
@@ -41,3 +43,58 @@ def test_maxvolume(pbd_material, show_viewer, box_obj_path):
...
+    """Attach 8 cloth particles to a cube, batched (n_envs=2), verify attachment with leftward motion."""
+
+    particle_size = 1e-2
```
duburcqa: `PARTICLE_SIZE = 0.01`, and skip line after defining all test constants.

tests/test_pbd.py
```diff
@@ -41,3 +43,58 @@ def test_maxvolume(pbd_material, show_viewer, box_obj_path):
...
+    scene = gs.Scene(
+        sim_options=gs.options.SimOptions(dt=2e-2, substeps=10),
+        pbd_options=gs.options.PBDOptions(particle_size=particle_size),
+        show_viewer=show_viewer,
+    )
+    rigid_material = gs.materials.Rigid(needs_coup=True, coup_friction=0.0)
+    scene.add_entity(gs.morphs.Plane(), rigid_material)
+
+    box = scene.add_entity(
+        morph=gs.morphs.Box(pos=(0.25, 0.25, 2.25), size=(0.25, 0.25, 0.25)),
```
duburcqa: In this context, I suggest using multiple line for all arguments systematically, even if it fits on one line. All of them are important, putting them on separated lines improves readability.

tests/test_pbd.py
```diff
@@ -41,3 +43,58 @@ def test_maxvolume(pbd_material, show_viewer, box_obj_path):
...
+def test_cloth_attachment(show_viewer):
+    """Attach 8 cloth particles to a cube, batched (n_envs=2), verify attachment with leftward motion."""
+
+    particle_size = 1e-2
+    scene = gs.Scene(
+        sim_options=gs.options.SimOptions(dt=2e-2, substeps=10),
+        pbd_options=gs.options.PBDOptions(particle_size=particle_size),
+        show_viewer=show_viewer,
+    )
+    rigid_material = gs.materials.Rigid(needs_coup=True, coup_friction=0.0)
```
duburcqa: I'm not convinced it is worth defining a variable for this, just copy this twice where it is used, because there is no requirement to use the exact same material twice for the test to pass, and recycling materials is error-prone as they could be modified in-place when building the scene.

tests/test_pbd.py
```diff
@@ -41,3 +43,58 @@ def test_maxvolume(pbd_material, show_viewer, box_obj_path):
...
+
+    scene.build(n_envs=2)
+
+    particles_idx = np.array([0, 1, 2, 3, 4, 5, 6, 7], dtype=np.int32)
+    box_link_idx = box.links[0].idx
+    scene.pbd_solver.set_animate_particles_by_link(particles_idx, box_link_idx, scene.rigid_solver.links_state)
+
+    # leftward velocity for both envs on base linear DOFs [3,4,5]
+    vel = np.array([[-0.0, 5.0, 0.0], [-5.0, 0.0, 0.0]], dtype=np.float32)
+    box.set_dofs_velocity(vel, dofs_idx_local=np.arange(3))
```
duburcqa: See my comment for the example. No need to use numpy arrays here.

tests/test_pbd.py
```diff
@@ -41,3 +43,58 @@ def test_maxvolume(pbd_material, show_viewer, box_obj_path):
...
+
+    cloth_pos0 = cloth.get_particles_pos().cpu().numpy()[:, particles_idx]
+    link_pos0 = scene.rigid_solver.links[box_link_idx].get_pos().cpu().numpy()
```
duburcqa: Remove `.cpu().numpy()` in both case. It is useless.

tests/test_pbd.py
```diff
@@ -41,3 +43,58 @@ def test_maxvolume(pbd_material, show_viewer, box_obj_path):
...
+
+    for _ in range(25):
+        scene.step()
+
+    # Check that the attached particles followed the link displacement per env
+    cloth_pos1 = cloth.get_particles_pos().cpu().numpy()[:, particles_idx]
+    link_pos1 = scene.rigid_solver.links[box_link_idx].get_pos().cpu().numpy()
```
duburcqa: Same.

tests/test_pbd.py
```diff
@@ -41,3 +43,58 @@ def test_maxvolume(pbd_material, show_viewer, box_obj_path):
...
+
+    cloth_disp = cloth_pos1 - cloth_pos0
+    link_disp = link_pos1 - link_pos0
+    # broadcast link_disp to match cloth_disp shape
+    link_disp = np.tile(link_disp[:, None, :], (1, len(particles_idx), 1))
```
duburcqa: Remove tiling. Just do `unsqueeze(1)`. `assert_allclose` is strong enough to understand what to do.

tests/test_pbd.py
```diff
@@ -41,3 +43,58 @@ def test_maxvolume(pbd_material, show_viewer, box_obj_path):
...
+
+    assert_allclose(cloth_disp, link_disp, atol=1e-1)
```
duburcqa: This tolerance is very bad, it is not even clear that it is enforcing anything at this point.

tests/test_pbd.py
duburcqa: This test is not good enough. It should check that the position / velocity of a rigid body with predefined motion (or at least non-zero initial velocity) is properly tracked.

genesis/engine/solvers/pbd_solver.py
YilingQiao: done

tests/test_pbd.py
YilingQiao: done

examples/coupling/cloth_attached_to_rigid.py
YilingQiao: done

examples/coupling/cloth_attached_to_rigid.py
YilingQiao: done

examples/coupling/cloth_attached_to_rigid.py
YilingQiao: done

examples/coupling/cloth_attached_to_rigid.py
YilingQiao: done

examples/coupling/cloth_attached_to_rigid.py
YilingQiao: done

examples/coupling/cloth_attached_to_rigid.py
YilingQiao: done

examples/coupling/cloth_attached_to_rigid.py
YilingQiao: done

examples/coupling/cloth_attached_to_rigid.py
YilingQiao: done, removed those unused variables / lines

tests/test_pbd.py
YilingQiao: The rigid body has a non-zero initial velocity
```
vel = np.array([[-0.0, 5.0, 0.0], [-5.0, 0.0, 0.0]], dtype=np.float32)
box.set_dofs_velocity(vel, dofs_idx_local=np.arange(3))
```

And I'm checking the displacement of the cloth is same as the rigid body. Not checking velocity though. Do you mean adding one more check on velocity? (although I think the displacement is almost enough)

tests/test_pbd.py
YilingQiao: done

tests/test_pbd.py
YilingQiao: done

tests/test_pbd.py
YilingQiao: done

tests/test_pbd.py
YilingQiao: no, it's not. If we input two arrays of 
`(2, 8, 3)` and `(2, 1, 3)`
In `assert_allclose` we have
```
    args = np.broadcast_arrays(*map(np.squeeze, args))
```

And it will throw an error
```
E       ValueError: shape mismatch: objects cannot be broadcast to a single shape.  Mismatch is between arg 0 with shape (2, 8, 3) and arg 1 with shape (2, 3).
```

tests/test_pbd.py
YilingQiao: done

tests/test_pbd.py
YilingQiao: done

tests/test_pbd.py
YilingQiao: done

genesis/engine/couplers/legacy_coupler.py
YilingQiao: done

examples/coupling/cloth_attached_to_rigid.py
YilingQiao: done

examples/coupling/cloth_attached_to_rigid.py
YilingQiao: fixed

examples/coupling/cloth_attached_to_rigid.py
YilingQiao: done

examples/coupling/cloth_attached_to_rigid.py
YilingQiao: done

genesis/engine/couplers/legacy_coupler.py
YilingQiao: done

genesis/engine/couplers/legacy_coupler.py
YilingQiao: done

tests/test_pbd.py
YilingQiao: This is not a hard constraint and the cube moves fast. I reduced the initial speed a bit to use smaller tolerance.

genesis/engine/couplers/legacy_coupler.py
YilingQiao: done

tests/test_pbd.py
duburcqa: It should work. The error that you are pointing out is something different:
```python
arg 0 with shape (2, 8, 3) and arg 1 with shape (2, 3).
```

You are trying to broadcast `(2, 8, 3)` and `(2, 3)`, which does not work indeed.

tests/test_pbd.py
duburcqa: We should have a setup where the tolerance can be around 1e-3, otherwise it is not testing much. Maybe stop the motion at the end and wait for a few steps?

tests/test_pbd.py
duburcqa: If you are checking the deplacement along the entire trajectory it should be fine.

examples/coupling/cloth_attached_to_rigid.py
```diff
@@ -0,0 +1,98 @@
...
+        import time
+
+        for i_horizon in range(args.horizon):
+            scene.step()
+            cloth_pos1 = cloth.get_particles_pos()[particles_idx]
+            # if j == 2:
+            #     from IPython import embed; embed()
+            #     # exit()
+            #     # cloth_pos1 = cloth.get_particles_pos()[:, particles_idx]
+            #     time.sleep(0.5)
```
duburcqa: Can you remove these debug comments?

examples/coupling/cloth_attached_to_rigid.py
```diff
@@ -0,0 +1,98 @@
...
+
+    ########################## entities ##########################
+    rigid_material = gs.materials.Rigid(needs_coup=True, coup_friction=0.0)
+
+    # create ground plane
+    scene.add_entity(gs.morphs.Plane(), rigid_material)
+
+    # create box
+    box_pos = [0.25, 0.25, 0.25]
+    box_size = [0.25, 0.25, 0.25]
```
duburcqa: No need to define variable for this. They are not used anywhere and it is already crystal clear what it stores.

examples/coupling/cloth_attached_to_rigid.py
```diff
@@ -0,0 +1,98 @@
...
+    cloth_scale = 0.5
+    cloth_morph = gs.morphs.Mesh(pos=cloth_pos, scale=cloth_scale, file="meshes/cloth.obj")
+    cloth_material = gs.materials.PBD.Cloth()
+    cloth_surface = gs.surfaces.Default(color=(0.2, 0.4, 0.8, 1.0))  # , vis_mode="particle")
+    cloth = scene.add_entity(cloth_morph, cloth_material, cloth_surface)
+
+    ########################## build ##########################
+    scene.build(n_envs=0)
+
+    particles_idx = np.array([0, 1, 2, 3, 4, 5, 6, 7])
```
duburcqa: Are you sure using a numpy array is necessary here? I would just use a list of indices.

genesis/engine/couplers/legacy_coupler.py
```diff
@@ -672,7 +762,12 @@ def couple(self, f):
 
         # PBD <-> Rigid
         if self._rigid_pbd and self.rigid_solver.is_active():
-            self.pbd_rigid(f)
+            self.kernel_pbd_rigid_collide()
+
+            # 1-way: animate particles by links
+            full_step_inv_dt = 1.0 / self.pbd_solver._dt
+            clamped_inv_dt = min(full_step_inv_dt, 50.0)
```
duburcqa: `50.0`  should be made a global constant on top of this file.

genesis/engine/solvers/pbd_solver.py
```diff
@@ -30,6 +32,15 @@ class MATERIAL(gs.IntEnum):
...
         PARTICLE = 3  # non-physics particles
 
+    @ti.dataclass
+    class ParticleAnimateByLinkInfo:
+        """
+        Index to and offset from the RigidLink that keyframe-animates this particle.
+        """
+
+        link_idx: ti.i32  # gs.ti_int not available before gs.init()
+        local_pos: ti.math.vec3  # gs.ti_vec3 causes uncomprehensible error !!
```
duburcqa: Remove these annotations, they are invalid. And the comments are also not right.

genesis/engine/solvers/pbd_solver.py
```diff
@@ -30,6 +32,15 @@ class MATERIAL(gs.IntEnum):
         LIQUID = 2
         PARTICLE = 3  # non-physics particles
 
+    @ti.dataclass
+    class ParticleAnimateByLinkInfo:
```
duburcqa: I would much prefer use an inlined data structure as we are doing everywhere else in this file. This would allow specifying the correct datatype.

genesis/engine/solvers/pbd_solver.py
```diff
@@ -393,13 +401,15 @@ def _kernel_apply_external_force(self, f: ti.i32, t: ti.f32):
...
+
+            # attached particles are not free but still need to update position to follow the link
+            self.particles[i_p, i_b].pos = (
+                self.particles[i_p, i_b].pos + self.particles[i_p, i_b].vel * self._substep_dt
+            )
 
     @ti.kernel
     def _kernel_solve_stretch(self, f: ti.i32):
-        for _ in ti.static(range(self._max_stretch_solver_iterations)):
+        for i_iter in ti.static(range(self._max_stretch_solver_iterations)):
```
duburcqa: For this, we usually using `it` as an exception.

tests/test_pbd.py
duburcqa: Ok I did some testing, `assert_allclose` is indeed broken. I will fix it.

duburcqa: Here is the patched implementation:
```python
def assert_allclose(actual, desired, *, atol=None, rtol=None, tol=None, err_msg=None):
    # Determine absolute and relative tolerance from input arguments
    assert (tol is not None) ^ (atol is not None or rtol is not None)
    if tol is not None:
        atol = tol
        rtol = tol
    if rtol is None:
        rtol = 0.0
    if atol is None:
        atol = 0.0

    # Convert input arguments as numpy arrays
    args = [actual, desired]
    for i, arg in enumerate(args):
        if isinstance(arg, torch.Tensor):
            arg = tensor_to_array(arg)
        elif isinstance(arg, (tuple, list)):
            arg = [tensor_to_array(val) for val in arg]
        args[i] = np.asanyarray(arg)

    # Early return without checking anything is both arrays are empty (0D arrays have size 1).
    if all(e.size == 0 for e in args):
        return

    # Try to make sure both arrays have the exact same shape.
    # First, try to broadcast both matrices. Then it is does not work, squeeze them before trying again.
    try:
        args = np.broadcast_arrays(*args)
    except ValueError:
        args = np.broadcast_arrays(*map(np.squeeze, args))

    np.testing.assert_allclose(*args, atol=atol, rtol=rtol, err_msg=err_msg)


def assert_array_equal(actual, desired, *, err_msg=None):
    assert_allclose(actual, desired, atol=0.0, rtol=0.0, err_msg=err_msg)
```

genesis/engine/solvers/pbd_solver.py
YilingQiao: done

genesis/engine/solvers/pbd_solver.py
YilingQiao: done

genesis/engine/solvers/pbd_solver.py
YilingQiao: done

genesis/engine/couplers/legacy_coupler.py
YilingQiao: done

examples/coupling/cloth_attached_to_rigid.py
YilingQiao: done

examples/coupling/cloth_attached_to_rigid.py
YilingQiao: done

examples/coupling/cloth_attached_to_rigid.py
YilingQiao: done

examples/coupling/cloth_attached_to_rigid.py
YilingQiao: done

examples/coupling/cloth_attached_to_rigid.py
YilingQiao: done

tests/test_pbd.py
YilingQiao: They might not be the exactly same within the trajectory because of constraints. Now the displacement on the  position is small enough

```
assert_allclose(cloth_disp, link_disp, atol=2e-5)
```

tests/test_pbd.py
YilingQiao: sure now it's 2e-5

genesis/engine/solvers/pbd_solver.py
YilingQiao: done

genesis/engine/couplers/legacy_coupler.py
YilingQiao: that's what coupler has been doing... It has very fine-grained operation on different solver's attributes. Hard to wrap them all as API calls.

genesis/engine/couplers/legacy_coupler.py
YilingQiao: done

PR #1742, opened 2025-09-17
[FEATURE] Add Raycaster, Lidar, DepthCamera and sensor.draw_debug methods.

Milotrince:
## Description
Add Raycaster sensor, which is used for Lidar and DepthCamera.
Based on Zifan's work https://github.com/aCodeDog/OmniPerception

Also added `debug_draw` option to base Sensor and implement for each existing sensor.

In a future PR, will add common lidar configurations and dynamic Livox pattern

## Related Issue
 https://github.com/Genesis-Embodied-AI/Genesis/issues/1111 https://github.com/Genesis-Embodied-AI/Genesis/issues/419 https://github.com/Genesis-Embodied-AI/Genesis/discussions/45

## Motivation and Context
Lidar has been a long awaited feature!

## How Has This Been / Can This Be Tested?
`examples/sensors/lidar_teleop.py`
Need to add tests.

## Screenshots (if appropriate):

`python examples/sensors/lidar_teleop.py --pattern spherical`
Screenshot 2025-09-22 at 1 34 49 PM

`python examples/sensors/lidar_teleop.py --pattern depth`

https://github.com/user-attachments/assets/f5f1dada-3f7e-4ded-8fc7-5656af3a2041

`python examples/sensors/lidar_teleop.py --pattern grid`
Screenshot 2025-09-22 at 1 36 11 PM

`python examples/sensors/contact_force_go2.py --no-force`

Screenshot 2025-09-22 at 1 00 35 PM

`python examples/sensors/contact_force_go2.py --force`
Screenshot 2025-09-22 at 1 27 45 PM

## Checklist:

- [x] I read the **CONTRIBUTING** document.
- [x] I followed the `Submitting Code Changes` section of **CONTRIBUTING** document.
- [x] I tagged the title correctly (including BUG FIX/FEATURE/MISC/BREAKING)
- [ ] I updated the [documentation](https://github.com/Genesis-Embodied-AI/genesis-doc) accordingly or no change is needed.
- [x] I tested my changes and added instructions on how to test it for reviewers.

PR #1770, opened 2025-09-24
[FEATURE] Add Sensor.draw_debug method.

Milotrince:
## Related Issue
Splits https://github.com/Genesis-Embodied-AI/Genesis/pull/1742 into smaller PRs for manageability

## Motivation and Context
Allow for easily visually debugging sensor data.

see example scripts

`python examples/sensors/contact_force_go2.py --no-force`

Screenshot 2025-09-22 at 1 00 35 PM

`python examples/sensors/contact_force_go2.py --force`
Screenshot 2025-09-22 at 1 27 45 PM

## Checklist:


examples/sensors/imu_franka.py
Milotrince, line 1: renamed example files for clarity (`genesis/sensors/imu.py` != `examples/sensors/imu.py`)

examples/sensors/contact_force_go2.py
```diff
@@ -0,0 +1,98 @@
...
+                    gs.recorders.MPLLinePlot(
+                        title="Force Sensor Data",
+                        labels=["force_x", "force_y", "force_z"],
+                    )
+                )
+            else:
+                print("matplotlib or pyqtgraph not found, skipping real-time plotting.")
+        else:
+            contact_sensor = scene.add_sensor(
+                gs.sensors.Contact(
```
duburcqa: There is no rendering for this type of sensor?

examples/sensors/contact_force_go2.py
```diff
@@ -0,0 +1,98 @@
...
+            file="urdf/go2/urdf/go2.urdf",
+            pos=(0.0, 0.0, 0.2),
+            links_to_keep=foot_link_names,
+        )
+    )
+
+    for link_name in foot_link_names:
+        if args.force:
+            force_sensor = scene.add_sensor(
+                gs.sensors.ContactForce(
```
duburcqa: What about replacing these 2 branches by just:
```python
sensor_type = gs.sensors.ContactForce if args.force else gs.sensors.Contact
```

One other option would be to do:
```python
if args.force:
    sensor_options = gs.sensors.ContactForce(
        [...]
    )
else:
    sensor_options = gs.sensors.Contact(
        [...]
    )
sensor = scene.add_sensor(sensor_options)
```

examples/sensors/contact_force_go2.py
```diff
@@ -0,0 +1,98 @@
...
+                )
+            )
+
+    scene.build()
+
+    try:
+        steps = int(args.seconds / args.timestep)
+        for _ in tqdm(range(steps)):
+            scene.step()
+
```
duburcqa: We don't skip space between try and except clauses.

genesis/sensors/base_sensor.py
```diff
@@ -386,10 +397,10 @@ def build(self):
 
         batch_size = self._manager._sim._B
 
-        link_start = self._shared_metadata.solver.entities[self._options.entity_idx].link_start
-        self._shared_metadata.links_idx = concat_with_tensor(
-            self._shared_metadata.links_idx, self._options.link_idx_local + link_start
-        )
+        entity = self._shared_metadata.solver.entities[self._options.entity_idx]
+        self.link_idx = self._options.link_idx_local + entity.link_start
```
duburcqa: This should be protected as a read-only property. See `RigidEntity` for instance.

Moreover, you should declare (i.e. initialise to None) in `__init__.py` all the attributes that will be set later in any other method (likely in `build` in our case). We are not doing this anywhere in old legacy code, but it is definitely a bad practice. Let's do better for new code :) The reasoning is that the set of attributes (and methods) that can be called/get/set for a given class must be static and never change. This makes debugging / understanding of what is happening much easier. Note that this property is enforced by design in compiled object-oriented languages such as C++.

genesis/sensors/contact_force.py
```diff
@@ -79,8 +80,17 @@ class ContactSensorOptions(RigidSensorOptionsMixin, SensorOptions):
         The delay in seconds before the sensor data is read.
     update_ground_truth_only : bool
         If True, the sensor will only update the ground truth data, and not the measured data.
+    draw_debug : bool, optional
+        If True and the rasterizer visualization is active, a sphere will be drawn at the sensor's position.
```
duburcqa: We are referring to the "rasterizer visualization" as the "interactive viewer", which is more intuitive for the user.

genesis/sensors/contact_force.py
```diff
@@ -104,14 +114,17 @@ def build(self):
         if self._shared_metadata.solver is None:
             self._shared_metadata.solver = self._manager._sim.rigid_solver
 
-        link_start = self._shared_metadata.solver.entities[self._options.entity_idx].link_start
+        entity = self._shared_metadata.solver.entities[self._options.entity_idx]
+        self.link_idx = self._options.link_idx_local + entity.link_start
+        self.link = entity.links[self._options.link_idx_local]
```
duburcqa: As before, should be declared in `__init__.py`

genesis/sensors/contact_force.py
```diff
@@ -143,6 +156,25 @@ def _update_shared_cache(
...
+
+        pos = self.link.get_pos(envs_idx=envs_idx).squeeze(0)
+        is_contact = self.read(envs_idx=envs_idx).squeeze(0).item()
+
+        if self.debug_object is not None:
+            context.clear_debug_object(self.debug_object)
+
+        if is_contact:
+            self.debug_object = context.draw_debug_sphere(
+                pos=pos, radius=self._options.debug_sphere_radius, color=self._options.debug_color
```
duburcqa: Instead of deleting and recreating primitive debug geometries at every rendering steps by hands, in theory you should just specify `persistent=False`, which is the exact target use-case for this option. However, I'm afraid this feature may be broken at the moment, so if it does not work, let me know and I may investigate.

Besides that, creating and deleting geometries repeatedly is very inefficient and should be avoided in possible. Here, if the debug object has already been added previously and should only be moved at another location. I just created a PR to add this feature: https://github.com/Genesis-Embodied-AI/Genesis/pull/1781/

genesis/sensors/contact_force.py
```diff
@@ -299,7 +343,25 @@ def _update_shared_cache(
         cls._add_noise_drift_bias(shared_metadata, shared_cache)
         shared_cache_per_sensor = shared_cache.reshape(shared_cache.shape[0], -1, 3)  # B, n_sensors * 3
         # clip for max force
-        shared_cache_per_sensor.clamp_(max=shared_metadata.max_force)
+        shared_cache_per_sensor.clamp_(min=-shared_metadata.max_force, max=shared_metadata.max_force)
```
duburcqa, line 356: Maybe you should update the doc to reflect this change?

genesis/sensors/contact_force.py
```diff
@@ -299,7 +343,25 @@ def _update_shared_cache(
...
         shared_cache_per_sensor[torch.abs(shared_cache_per_sensor) < shared_metadata.min_force] = 0.0
         cls._quantize_to_resolution(shared_metadata.resolution, shared_cache)
+
+    def _draw_debug(self, context: "RasterizerContext"):
+        """
+        Draw debug arrow representing the contact force.
+
+        Only draws for first environment.
+        """
+        envs_idx = 0 if self._manager._sim.n_envs > 0 else None
```
duburcqa: You should not do this. You should rendering for the selected rendering env, ie
```python
env_idx = context.rendered_envs_idx[0]
```

Note that it should be possible to add batching support to all `draw_debug_*` methods, but let's ignore this for now.

genesis/recorders/plotters.py
```diff
@@ -34,56 +35,87 @@
...
     def build(self):
         super().build()
-
         self.show_window = self._options.show_window if self._options.show_window is not None else has_display()
 
+    def process(self, data, cur_time):
+        pass  # allow super() calls
+
+    def cleanup(self):
+        pass  # allow super() calls
```
duburcqa: Could you clarify what you mean by "allow super() calls" ?

Why is this part of the public API of `BasePlotter`, while I have the feeling nobody is supposed to store abstract references to `BasePlotter` and call `process`, `cleanup` manually. You should only expose whatever is strictly necessary as part of the (public and private) API when doing inheritance. `BasePyQtPlotter` and `BaseMPLPlotter ` while just define their own, completely independent `process`, `cleanup` methods and call them internally without any contract assuming it should be part of `BasePlotter` API, which is perfectly fine. And none of them will call `super().*` since there is none.

genesis/recorders/plotters.py
```diff
@@ -93,13 +125,13 @@ def build(self):
 
         if self._options.labels is not None:
             self._setup_plot_structure(self._options.labels)
+        else:
+            self._setup_plot_structure(self._data_func())
```
duburcqa: I'm not a fan of this logics. I would rather have `_setup_plot_structure` taking two input arguments, with data being specified systematically, which is good to implement some asserts checking consistency between labels and data at build-time.
```python
self._setup_plot_structure(self._options.labels, self._data_func())
```

genesis/recorders/plotters.py
```diff
@@ -224,71 +254,117 @@ def get_image_array(self):
...
+            The RGB image as a numpy array.
+        """
+        pixmap = self.widget.grab()
+        qimage = pixmap.toImage()
+
+        qimage = qimage.convertToFormat(pg.QtGui.QImage.Format_RGB888)
+        ptr = qimage.bits()
+        ptr.setsize(qimage.byteCount())
+
+        return np.array(ptr).reshape((qimage.height(), qimage.width(), 3))
```
duburcqa: Prefer using public API if possible, i.e. `pg.imageToArray(qimage, copy=False, transpose=True)`. Note that it is up to you to set the optional argument properly. You need to do a copy if the image may be garbage collected before the array view is no longer useful, which seems to be the case here.


Edit: It seems that `pg.imageToArray` cannot be used because you are enforcing format `Format_RGB888`. Why this format? Yous should document this choice, and add a comment stating that `pg.imageToArray` cannot be used because of this.

genesis/sensors/contact_force.py
```diff
@@ -143,6 +156,25 @@ def _update_shared_cache(
         buffered_data.append(shared_ground_truth_cache)
         cls._apply_delay_to_shared_cache(shared_metadata, shared_cache, buffered_data)
 
+    def _draw_debug(self, context: "RasterizerContext"):
+        """
+        Draw debug sphere when the sensor detects contact.
+
+        Only draws for first environment.
+        """
+        envs_idx = 0 if self._manager._sim.n_envs > 0 else None
```
duburcqa: `env_idx = context.rendered_envs_idx[0]` (see comment below)

genesis/sensors/imu.py
```diff
@@ -310,3 +327,28 @@ def _update_shared_cache(
...
         cls._add_noise_drift_bias(shared_metadata, shared_cache)
         cls._quantize_to_resolution(shared_metadata.resolution, shared_cache)
+
+    def _draw_debug(self, context: "RasterizerContext"):
+        """
+        Draw debug arrow for the IMU acceleration.
+
+        Only draws for first environment.
+        """
+        envs_idx = 0 if self._manager._sim.n_envs > 0 else None
```
duburcqa: `env_idx = context.rendered_envs_idx[0]`

genesis/sensors/imu.py
```diff
@@ -310,3 +327,28 @@ def _update_shared_cache(
...
+
+        quat = self.link.get_quat(envs_idx=envs_idx).squeeze(0)
+        pos = self.link.get_pos(envs_idx=envs_idx).squeeze(0) + transform_by_quat(self.pos_offset, quat)
+
+        data = self.read(envs_idx=envs_idx)
+        acc_vec = data["lin_acc"].squeeze(0) * self._options.debug_acc_scale
+        gyro_vec = data["ang_vel"].squeeze(0) * self._options.debug_gyro_scale
```
duburcqa: All these `squeeze(0)` are painful... what about just do `pos=pose[0], vec=acc_vec[0]` when calling all `context.draw_debug_*` methods? It seems more readable no? This comment holds for all `sensor._draw_debug` implementations.

examples/sensors/contact_force_go2.py
Milotrince: its boolean data so i did not think it was as interesting to include, but the plotter does support it-- added!

duburcqa: Boolean data also desires love!

genesis/sensors/contact_force.py
Milotrince: when there is no contact and i want to not show the sphere, should i update the object to some faraway position (e.g. 50,50,-1) or delete it? also for arrows, i will have to create/delete to show a new arrow length/orientation, right?

duburcqa: Delete if it should be deleted, move if it should be add whereas it is already there. That should cover 90% of the cases already.

genesis/recorders/plotters.py
Milotrince: this is due to the mixin/multiple inheritance pattern. The base `Recorder` class throws `NotImplementedError`. `MPLLinePlotter` inherits from both `BasePlotter` and `LinePlotterMixin`.  `MPLLinePlotter` needs to call `super()`. If the `def process(...): _pass` is omitted then it will throw the error. I think the proper solution would be to use ABCs but we dont use ABC in our codebase

duburcqa: Can you remind me why you need mixin pattern for something else than option classes ? It is rarely useful and multiple inheritance is tricky to get right in all object-oriented languages, but especially in Python.

Milotrince: MPLPlotter + LinePlotter = MPLLinePlotter, PyQtPlotter + LinePlotter = PyQtLinePlotter

genesis/recorders/plotters.py
Milotrince: i didnt actually know `pg.imageToArray` existed. [docs](https://pyqtgraph.readthedocs.io/en/latest/_modules/pyqtgraph/functions.html#imageToArray) it always outputs `(b,g,r,a)`. An argument for keeping `Format_RGB888` is to keep the output of `get_image_array` consistent across the two platforms. of course, one could also still use `get_image_array` and reformat the output.

but using imageToArray is cleaner. i included this change in the latest commit https://github.com/Genesis-Embodied-AI/Genesis/pull/1770/commits/8fdbdd65e15860b6c67f31829765daa681eaa45b

genesis/recorders/plotters.py
Milotrince: I opted to use composition instead so MPLLinePlotter inherits from only MPLPlotter, and has `self.line_plot = LinePlotHelper(...)` in the latest commit

duburcqa: Yes, I think this is a good idea indeed. Keep mixins only for options. Composition is often more convenient to work with, easier to tweak, and less error prone :)

genesis/sensors/contact_force.py
```diff
@@ -143,6 +162,28 @@ def _update_shared_cache(
...
+        pos = self._link.get_pos(envs_idx=env_idx)[0]
+        is_contact = self.read(envs_idx=env_idx if self._manager._sim.n_envs > 0 else None).item()
+
+        if is_contact:
+            if self.debug_object is None:
+                self.debug_object = context.draw_debug_sphere(
+                    pos=pos, radius=self._options.debug_sphere_radius, color=self._options.debug_color
+                )
+            else:
+                context.update_debug_objects([self.debug_object], trans_to_T(pos).unsqueeze(0))
```
duburcqa, line 182: [trans_to_T(pos)] is more intuitive but no big deal.

Kashu7100:
@Milotrince 
[BUG REPORT] `python examples/sensors/imu_franka.py`

```bash
  File "/Users/kashu/research/Genesis/examples/sensors/imu_franka.py", line 129, in <module>
    main()
  File "/Users/kashu/research/Genesis/examples/sensors/imu_franka.py", line 83, in main
    scene.build()
  File "/Users/kashu/research/Genesis/genesis/utils/misc.py", line 140, in wrapper
    return method(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kashu/research/Genesis/genesis/engine/scene.py", line 797, in build
    self._visualizer.build()
  File "/Users/kashu/research/Genesis/genesis/vis/visualizer.py", line 192, in build
    self.reset()
  File "/Users/kashu/research/Genesis/genesis/vis/visualizer.py", line 169, in reset
    self._viewer.update(auto_refresh=True)
  File "/Users/kashu/research/Genesis/genesis/vis/viewer.py", line 153, in update
    self.context.update()
  File "/Users/kashu/research/Genesis/genesis/vis/rasterizer_context.py", line 952, in update
    self.update_sensors()
  File "/Users/kashu/research/Genesis/genesis/vis/rasterizer_context.py", line 807, in update_sensors
    self.sim._sensor_manager.draw_debug(self)
  File "/Users/kashu/research/Genesis/genesis/sensors/sensor_manager.py", line 92, in draw_debug
    sensor._draw_debug(context)
  File "/Users/kashu/research/Genesis/genesis/sensors/imu.py", line 342, in _draw_debug
    data = self.read(envs_idx=env_idx)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kashu/research/Genesis/genesis/utils/misc.py", line 150, in wrapper
    return method(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kashu/research/Genesis/genesis/sensors/base_sensor.py", line 225, in read
    return self._get_formatted_data(self._manager.get_cloned_from_cache(self), envs_idx)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kashu/research/Genesis/genesis/sensors/base_sensor.py", line 331, in _get_formatted_data_dict
    return dict(zip(self._return_format.keys(), self._get_return_values(tensor, envs_idx)))
                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kashu/research/Genesis/genesis/sensors/base_sensor.py", line 315, in _get_return_values
    envs_idx = self._sanitize_envs_idx(envs_idx)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kashu/research/Genesis/genesis/sensors/base_sensor.py", line 338, in _sanitize_envs_idx
    return self._manager._sim._scene._sanitize_envs_idx(envs_idx)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kashu/research/Genesis/genesis/engine/scene.py", line 1350, in _sanitize_envs_idx
    gs.raise_exception("`envs_idx` is not supported for non-parallelized scene.")
  File "/Users/kashu/research/Genesis/genesis/utils/misc.py", line 42, in raise_exception
    raise gs.GenesisException(msg)
genesis.GenesisException: `envs_idx` is not supported for non-parallelized scene.
```

`python examples/sensors/contact_force_go2.py --force`
```bash
  File "/Users/kashu/research/Genesis/examples/sensors/contact_force_go2.py", line 94, in <module>
    main()
  File "/Users/kashu/research/Genesis/examples/sensors/contact_force_go2.py", line 84, in main
    scene.step()
  File "/Users/kashu/research/Genesis/genesis/utils/misc.py", line 150, in wrapper
    return method(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kashu/research/Genesis/genesis/engine/scene.py", line 925, in step
    self._visualizer.update(force=False, auto=refresh_visualizer)
  File "/Users/kashu/research/Genesis/genesis/vis/visualizer.py", line 199, in update
    self._viewer.update(auto_refresh=auto)
  File "/Users/kashu/research/Genesis/genesis/vis/viewer.py", line 153, in update
    self.context.update()
  File "/Users/kashu/research/Genesis/genesis/vis/rasterizer_context.py", line 952, in update
    self.update_sensors()
  File "/Users/kashu/research/Genesis/genesis/vis/rasterizer_context.py", line 807, in update_sensors
    self.sim._sensor_manager.draw_debug(self)
  File "/Users/kashu/research/Genesis/genesis/sensors/sensor_manager.py", line 92, in draw_debug
    sensor._draw_debug(context)
  File "/Users/kashu/research/Genesis/genesis/sensors/contact_force.py", line 377, in _draw_debug
    self.debug_object = context.draw_debug_arrow(pos=pos[0], vec=vec[0], color=self._options.debug_color)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kashu/research/Genesis/genesis/vis/rasterizer_context.py", line 829, in draw_debug_arrow
    gu.z_up_to_R(tensor_to_array(vec).astype(np.float32), out=pose[0, :3, :3])
  File "/Users/kashu/research/Genesis/genesis/utils/geom.py", line 1290, in z_up_to_R
    return _np_z_up_to_R(z, up, out)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kashu/research/Genesis/.venv/lib/python3.12/site-packages/numba/np/arrayobj.py", line 2134, in normalize_reshape_value
    raise ValueError("total size of new array must be unchanged")
ValueError: total size of new array must be unchanged

[Genesis] [01:04:50] [ERROR] ValueError: total size of new array must be unchanged
```

Kashu7100:
image

Also the marker seems off? This is on Mac M4.

Milotrince:
@Kashu7100 
> [ERROR] ValueError: total size of new array must be unchanged

I'm aware of this. I have a local fix, will push soon to the next PR https://github.com/Genesis-Embodied-AI/Genesis/pull/1772

duburcqa:
Thank you for the report @Kashu7100 Some unit tests are missing and will be added!

Issue #1111, opened 2025-05-06
[Feature]: Simple lidar, raycast, or scandots sensor


nootnootpengupengu:
### What feature or enhancement are you proposing?
Is it possible to simulate a basic distance-based sensor, such as LIDAR, scan dots, or even a GPU-accelerated raycast?

### Motivation
As a robotics engineer, I want to simulate distance-based sensors within my application in order to test perception algorithms, obstacle avoidance, and navigation behaviors in a virtual environment.

### Potential Benefit
Distance-based sensors are ubiquitous in robotics applications. Integrating support for them in Genesis would significantly enhance its capability as a robotics simulation platform. It would allow developers to prototype and validate perception stacks more realistically.

### What is the expected outcome of the implementation work?
A performant, GPU-accelerated distance sensor simulation (e.g., LIDAR, raycast, or scan dots) that can provide real-time range data in a virtual environment.

### Additional Information
This feature would ideally include customizable parameters such as range, field of view, resolution, and noise models to closely mimic real-world sensor behavior.

Issue #419, opened 2024-12-31
Does it support LiDAR detection?


wdnmd15968:
Does it support LiDAR?  OR it is on the way？I am conducting simulation research related to radar ranging
请问支持激光雷达吗，或者是否在开发了，我做的是与雷达测距相关的仿真研究。

oscar-lima:
Related issue: #335

PR #1772, opened 2025-09-25
[FEATURE] Add Raycaster sensor (Lidar and DepthCamera)

Milotrince:
## Description
Add Raycaster sensor, which is used for Lidar and DepthCamera.
Based on Zifan's work https://github.com/aCodeDog/OmniPerception

## Related Issue
Resolves https://github.com/Genesis-Embodied-AI/Genesis/issues/1111 https://github.com/Genesis-Embodied-AI/Genesis/issues/419 https://github.com/Genesis-Embodied-AI/Genesis/discussions/45

## Motivation and Context
Lidar has been a long awaited feature!

`examples/sensors/lidar_teleop.py`
`tests/test_sensors.py`

`python examples/sensors/lidar_teleop.py --pattern spherical`
Screenshot 2025-09-22 at 1 34 49 PM

`python examples/sensors/lidar_teleop.py --pattern depth`

https://github.com/user-attachments/assets/f5f1dada-3f7e-4ded-8fc7-5656af3a2041

`python examples/sensors/lidar_teleop.py --pattern grid`
Screenshot 2025-09-22 at 1 36 11 PM

## Checklist:


Milotrince:
Note: Review after https://github.com/Genesis-Embodied-AI/Genesis/pull/1770 is merged

examples/sensors/lidar_teleop.py
duburcqa, line 1: I'm not convinced splitting this example file in multiple methods is worth it. I tends to make it harder to debug by scoping variables, requiring using a debugger.

examples/sensors/lidar_teleop.py
```diff
@@ -0,0 +1,222 @@
+import argparse
+
+import numpy as np
+from pynput import keyboard
```
duburcqa: `pynput` is not part of your dependencies. You should add some ImportError handling with proper error message, and add the deps in 'dev' optional deps group.

genesis/utils/keyboard.py
duburcqa, line 1: I don't think this should be part of the library itself. It is only used in examples for now, and keyword handling in the interactive viewer should leverage the existing listener capabilities of pyrender itself, which does not require any external dependency.

examples/sensors/lidar_teleop.py
```diff
@@ -0,0 +1,222 @@
...
+
+import genesis as gs
+from genesis.sensors.raycaster.patterns import DepthCameraPattern, GridPattern, SphericalPattern
+from genesis.utils.geom import euler_to_quat
+from genesis.utils.keyboard import KeyboardDevice
+
```
duburcqa, line 10: Skip 2 lines before (not just after) global variables.

examples/sensors/lidar_teleop.py
```diff
@@ -0,0 +1,222 @@
...
+            camera_pos=(6.0, 6.0, 4.0),
+            camera_lookat=(0.0, 0.0, 0.5),
+            camera_fov=60,
+            max_FPS=60,
+        ),
+        show_viewer=show_viewer,
+        show_FPS=False,
+    )
+
+    scene.add_entity(gs.morphs.Plane(is_free=is_free))
```
duburcqa: What does 'is_free' do? It is the first time that this feature is used in the entire codebase. There must be something "wrong" in our design for this feature to be so seldomly relevant.

examples/sensors/lidar_teleop.py
```diff
@@ -0,0 +1,222 @@
...
+
+    # create ring of obstacles to visualize raycaster sensor hits
+    inner_radius = 3.0
```
duburcqa: What about defining these two radius as global variables ?

examples/sensors/lidar_teleop.py
```diff
@@ -0,0 +1,222 @@
...
+
+    target_pos = init_pos.copy()
+    target_euler = init_euler.copy()
+
+    def apply_pose_to_all_envs(pos_np: np.ndarray, quat_np: np.ndarray):
+        if n_envs > 0:
+            pos_np = np.expand_dims(pos_np, axis=0).repeat(n_envs, axis=0)
+            quat_np = np.expand_dims(quat_np, axis=0).repeat(n_envs, axis=0)
+        robot.set_pos(pos_np, zero_velocity=False)
+        robot.set_quat(quat_np, zero_velocity=False)
```
duburcqa: Do you really care about zero-ing the velocity?

examples/sensors/lidar_teleop.py
```diff
@@ -0,0 +1,222 @@
...
+        help="Load obstacles as fixed and cast only against fixed objects (is_free=False)",
+        default=True,
+    )
+    parser.add_argument(
+        "-nf",
+        "--no-fixed",
+        action="store_false",
+        dest="fixed",
+        help="Load obstacles as dynamic (is_free=True), raycaster will update BVH every step",
+    )
```
duburcqa: I'm not convinced being able to select between these two modes is relevant. Why would someone want to cast only against fixed objects?

genesis/sensors/raycaster/base_pattern.py
```diff
@@ -0,0 +1,91 @@
...
+        cur_t : float
+            Current simulation time in seconds.
+        """
+        raise NotImplementedError(f"{type(self).__name__} must implement `update_ray_directions()`.")
+
+    def get_ray_directions(self) -> torch.Tensor:
+        """
+        Get the local direction vectors of the rays.
+        """
+        if self._cached_ray_directions.isnan().any():
```
duburcqa: This condition is weird, kind of hacky, and not very efficient to evaluate. Moreover, not that it is not initialised to `float("nan")` by default apparently.

Why do you need this logics? What is the criteria to invalidate the cache? I think it would be better to just call directly `self.update_ray_directions` whether it is necessary. If not possible, then move the logics inside each concrete pattern generators implementations. And in last resort, maybe add some `_is_cache_valid` attribute.

genesis/sensors/raycaster/base_pattern.py
```diff
@@ -0,0 +1,91 @@
...
+
+RAYCAST_PATTERN_NAME_TO_OPTIONS_MAP: dict[str, type[RaycastPattern]] = {}
+RAYCAST_PATTERN_OPTIONS_TO_GENERATOR_MAP: dict[type[RaycastPattern], type[RaycastPatternGenerator]] = {}
+
+
+def register_pattern(pattern_type: type[RaycastPattern], name: str):
+    def _impl(generator_type: type[RaycastPatternGenerator]):
+        RAYCAST_PATTERN_NAME_TO_OPTIONS_MAP[name] = pattern_type
+        RAYCAST_PATTERN_OPTIONS_TO_GENERATOR_MAP[pattern_type] = generator_type
+        return generator_type
```
duburcqa: The correct typing should be:
```python
from typing import TypeVar

T = TypeVar("T", bound=type[RaycastPatternGenerator])

def _impl(generator_type: T) -> T:
    RAYCAST_PATTERN_NAME_TO_OPTIONS_MAP[name] = pattern_type
    RAYCAST_PATTERN_OPTIONS_TO_GENERATOR_MAP[pattern_type] = generator_type
    return generator_type
```

Because this methods returns in output an object of the exact same type than the one provided as input.

genesis/sensors/raycaster/depth_camera.py
```diff
@@ -0,0 +1,64 @@
...
+        option for the **first** lidar sensor will be the behavior for **all** Lidar sensors.
+    delay : float, optional
+        The delay in seconds before the sensor data is read.
+    update_ground_truth_only : bool, optional
+        If True, the sensor will only update the ground truth cache, and not the measured cache.
+    draw_debug : bool, optional
+        If True and the rasterizer visualization is active, spheres will be drawn at every hit point.
+    debug_sphere_radius: float, optional
+        The radius of each debug hit point sphere drawn in the scene. Defaults to 0.02.
+    """
```
duburcqa, line 17: Docstring is already inherited. If you do not add any parameter, then I don't think it is necessary to copy-paste the original doc.

duburcqa: Ok apparently you changed the type of `pattern` in the description... Not ideal but ok.

I guess you should also re-declare the attribute with the correct type, not just update validate, otherwise it will not be sufficient for pyright to infer the correct type.

genesis/sensors/raycaster/depth_camera.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def validate(self, scene):
+        super().validate(scene)
+        if not isinstance(self.pattern, DepthCameraPattern):
+            gs.raise_exception("DepthCamera pattern must be an instance of DepthCameraPattern")
```
duburcqa: Why this is not part of the `RaycasterOptions` ? I think you should better leverage to power of dataclass to do this automatically for you. Moreover, why not use built-in `__post_init__` method for validation of dataclass attributes instead of using some ad-hoc method `validate`. See official doc for details: https://docs.python.org/3/library/dataclasses.html#dataclasses.__post_init__

Here is an example:

```python
@dataclass
class Example:
    field_1: int
    field_2: np.ndarray

    def __post_init__(self):
        for f in fields(self):
            field_name, field_type = f.name, f.type
            value = getattr(self, field_name)
            if issubclass(field_type, (np.ndarray, torch.tensor)):
                setattr(self, field_name, sanitize_tensor(value))
            elif issubclass(field_type, (numbers.Real, np.floating)):
                setattr(self, field_name, float(field_type))
            elif issubclass(field_type, (numbers.Integer, np.integer)):
                setattr(self, field_name, int(field_type))
            elif not isinstance(value, field_type):
                raise TypeError(f"Field {f.name} expected {f.type}, got {value!r}")
```

genesis/sensors/raycaster/depth_camera.py
```diff
@@ -0,0 +1,64 @@
...
+        Read the depth image from the sensor.
+
+        This method uses the hit distances from the underlying RaycasterSensor.read() method and reshapes into image.
+
+        Returns
+        -------
+        torch.Tensor
+            The depth image with shape (height, width).
+        """
+        return self.read()["hit_ranges"].reshape(self._options.pattern.height, self._options.pattern.width)
```
duburcqa: It seems that read should return a `TypedDict` instead of a plain dict, if you rely on the existence of some keys (eventually combined with `ReadOnly` specifier to clearly state that the result of read should not be altered). See official documentation for details: https://typing.python.org/en/latest/spec/typeddict.html

In this context, `NamedTuple` may be an even better option, because it is not just pure typing but an actual type while still being extremely lightweight and efficient compared to plain dict. Therefore, it can also be used in conjunction with `isinstance` and the likes. See official documentation for details: https://docs.python.org/3/library/collections.html#collections.namedtuple

genesis/sensors/raycaster/lidar.py
```diff
@@ -0,0 +1,48 @@
...
+    debug_sphere_radius: float, optional
+        The radius of each debug hit point sphere drawn in the scene. Defaults to 0.02.
+    """
+
+    pass
+
+
+@register_sensor(LidarOptions, RaycasterSharedMetadata)
+class LidarSensor(RaycasterSensor):
+    pass
```
duburcqa: Class you elaborate while you don't think to implement anything specific for Lidar? This is weird. Why not just `RaycasterSensor` sensor directly then, and eventually define some alias in `__init__.py` so that it can be imported as either `RaycasterSensor` or `LidarSensor`? Or I may be missing something here.

genesis/sensors/raycaster/patterns.py
```diff
@@ -0,0 +1,297 @@
...
+
+
+def _generate_uniform_angles(
+    n: int | None = None,
+    fov: float | tuple[float, float] | None = None,
+    res: float | None = None,
+    angles: Sequence[float] | None = None,
+    use_degrees: bool = True,
+) -> torch.Tensor:
+
```
duburcqa, line 112: Do not skip space here unless you add docstring. By the way, could you add a short on-line docstring to explain how what this method does and in which context it is used?

Is this method performance-critical? (i.e. called very frequently, inside the step update loop)

genesis/sensors/raycaster/patterns.py
```diff
@@ -0,0 +1,297 @@
...
+    if use_degrees:
+        angles = torch.deg2rad(angles)
+
+    return angles
+
+
+def _compute_focal_lengths(
+    width: int, height: int, fov_horizontal: float | None, fov_vertical: float | None
+) -> tuple[float, float]:
```
duburcqa, line 204: Same questions than `_generate_uniform_angles`.

genesis/sensors/raycaster/patterns.py
```diff
@@ -0,0 +1,297 @@
...
+        fv_rad = math.radians(fov_vertical)
+
+    fx = width / (2.0 * math.tan(fh_rad / 2.0))
+    fy = height / (2.0 * math.tan(fv_rad / 2.0))
+
+    return fx, fy
+
+
+# ============================== Generic Patterns ==============================
+@dataclass
```
duburcqa: Maybe skip line between comment separator and actual python code? It will look nicer.

genesis/sensors/raycaster/patterns.py
```diff
@@ -0,0 +1,297 @@
...
+    direction : tuple[float, float, float]
+        Ray direction vector.
+    ordering : str
+        Point ordering, either "xy" or "yx".
+    """
+
+    resolution: float = 0.1
+    size: tuple[float, float] = (2.0, 2.0)
+    direction: tuple[float, float, float] = (0.0, 0.0, -1.0)
+    ordering: str = "xy"
```
duburcqa: Your typing is wrong. It should be `tp.Literal["xy", "yx"]`.

In general, using string to encode a finite set of options is discouraged. You should prefer StrEnum (or IntEnum) if there are more than 2 values, otherwise I would recommend a simple boolean flag `use_yx_ordering: bool = False`. In both cases, there is nothing to validate manually anymore. Using Enum may be slightly more painful if something needs to be imported manually, but it could be exposed as part of `gs.options.*` if you pick this approach, although I would advice using the boolean flag in this context for simplicity.

genesis/sensors/raycaster/patterns.py
```diff
@@ -0,0 +1,297 @@
...
+
+    def validate(self):
```
duburcqa: As mentioned before, what about `__post_init__` ?

genesis/sensors/raycaster/patterns.py
```diff
@@ -0,0 +1,297 @@
...
+        if self.ordering not in ["xy", "yx"]:
+            raise ValueError(f"Ordering must be 'xy' or 'yx'. Received: '{self.ordering}'.")
+        if self.resolution <= 0:
+            raise ValueError(f"Resolution must be greater than 0. Received: '{self.resolution}'.")
```
duburcqa: "strictly positive" ?

genesis/sensors/raycaster/patterns.py
```diff
@@ -0,0 +1,297 @@
...
+
+    resolution: float = 0.1
+    size: tuple[float, float] = (2.0, 2.0)
+    direction: tuple[float, float, float] = (0.0, 0.0, -1.0)
+    ordering: str = "xy"
+
+    def validate(self):
+        if self.ordering not in ["xy", "yx"]:
+            raise ValueError(f"Ordering must be 'xy' or 'yx'. Received: '{self.ordering}'.")
+        if self.resolution <= 0:
```
duburcqa: Is there a more meaningful lower bound? `gs.EPS` is probably also invalid. Maybe something like 1mm?

genesis/sensors/raycaster/patterns.py
```diff
@@ -0,0 +1,297 @@
...
+            grid_x, grid_y = torch.meshgrid(self.x_coords, self.y_coords, indexing="xy")
+        else:
+            grid_x, grid_y = torch.meshgrid(self.x_coords, self.y_coords, indexing="ij")
+
+        starts = torch.empty((*self._return_shape, 3), dtype=gs.tc_float, device=gs.device)
+        starts[..., 0] = grid_x
+        starts[..., 1] = grid_y
+        starts[..., 2] = 0.0
+
+        return starts
```
duburcqa: You always do the exact same processing for x and y, I would recommend to store such quantities as a tuple to be able to loop over each axis:
```python

    def __init__(self, options: GridPattern):
        super().__init__(options)
        self.coords = [torch.arange(
            -size / 2, size / 2 + gs.EPS, options.resolution, dtype=gs.tc_float, device=gs.device
        ) for size in options.size]
        self.direction = torch.tensor(options.direction, dtype=gs.tc_float, device=gs.device)

    def get_ray_directions(self) -> torch.Tensor:
        return self.direction.expand((*self._return_shape, 3))

    def get_ray_starts(self) -> torch.Tensor:
        grid_x, grid_y = torch.meshgrid(*self.coords, indexing="xy" if self._options.ordering == "xy" else "ij")
        starts = torch.zeros((*self._return_shape, 3), dtype=gs.tc_float, device=gs.device)
        starts[..., 0] = grid_x
        starts[..., 1] = grid_y
        return starts
```

genesis/sensors/raycaster/patterns.py
```diff
@@ -0,0 +1,297 @@
...
+
+
+@register_pattern(GridPattern, "grid")
+class GridPatternGenerator(RaycastPatternGenerator):
+    """Generator for 2D grid ray patterns."""
+
+    def __init__(self, options: GridPattern):
+        super().__init__(options)
+        self.x_coords = torch.arange(
+            -options.size[0] / 2, options.size[0] / 2 + gs.EPS, options.resolution, dtype=gs.tc_float, device=gs.device
```
duburcqa: Maybe you should check that size is a multiple of resolution? Or it is not important if it gets cut?

genesis/sensors/raycaster/patterns.py
```diff
@@ -0,0 +1,297 @@
...
+
+    fov_vertical: float | tuple[float, float] = 30.0
+    fov_horizontal: float | tuple[float, float] = 360.0
+    n_vertical: int = 64
+    n_horizontal: int = 128
+    res_vertical: float | None = None
+    res_horizontal: float | None = None
+    vertical_angles: Sequence[float] | None = None
+    horizontal_angles: Sequence[float] | None = None
+    use_degrees: bool = True
```
duburcqa: I don't like this kind of option that increases the maintenance burden and the probability of bugs with limited benefits for the end-user. Let's use whatever you think is the most common use-case in practice, and ask the user to do the conversion manually if necessary. Similarly to what we are doing for `euler` optional argument of `Morph` that only expects degrees.

genesis/sensors/raycaster/patterns.py
```diff
@@ -0,0 +1,297 @@
...
+            if (isinstance(fov, float) and (fov < 0 or fov > full_rotation + gs.EPS)) or (
+                isinstance(fov, tuple) and (fov[1] - fov[0] > full_rotation + gs.EPS)
+            ):
+                gs.raise_exception(f"[{type(self).__class__}] FOV should not be <0 or >{full_rotation}. Got: {fov}.")
+
+    def get_return_shape(self) -> tuple[int, ...]:
+        if self.vertical_angles is not None and self.horizontal_angles is not None:
+            return (len(self.vertical_angles), len(self.horizontal_angles))
+        else:
+            return (self.n_vertical, self.n_horizontal)
```
duburcqa: You should probably sanitize all user-arguments before at init, to ensure that all attributes have a unique type after init. This way, you could remove all these branches that just increase the maintenance burden for no benefit. Note that this is a common practice.

duburcqa: Note that it would be a better to enforce a unique type when defining the dataclass, and implement a custom `__init__` method to accept non-canonical types and only specify default arguments of the unique canonical type.
```python
@dataclass
class SphericalPattern(RaycastPattern):
    fov_vertical: tuple[float, float] = (30.0, 30.0)
    n_vertical: int = 64

    def __init__(self, fov_vertical: float | tuple[float, float], n_vertical: Any):
        self.fov_vertical = tuple(fov_vertical) if isinstance(fov_vertical, tp.Iterable) else (fov_vertical,  fov_vertical)
        self.n_vertical = int(n_vertical)
```
However, this pattern is not very standard and kind of partially defeating the benefit of dataclass. But it is not very standard to support non-canonical arguments that would be sanitazed in the first place.

genesis/sensors/raycaster/patterns.py
```diff
@@ -0,0 +1,297 @@
...
+            use_degrees=self._options.use_degrees,
+        )
+        h_angles = _generate_uniform_angles(
+            n=self._options.n_horizontal,
+            fov=self._options.fov_horizontal,
+            res=self._options.res_horizontal,
+            angles=self._options.horizontal_angles,
+            use_degrees=self._options.use_degrees,
+        )
+
```
duburcqa: Same. I'm not a fan of duplicating everything. I would recommend giving tuple for x, y quantities everywhere, and just loop systematically:
```python
    def get_ray_directions(self) -> torch.Tensor:
        angles = _generate_uniform_angles(
            n=self._options.n,
            fov=self._options.fov,
            res=self._options.res,
            angles=self._options.vertical,
            use_degrees=self._options.use_degrees,
        )
        grid = torch.meshgrid(*angles, indexing="ij")
        return spherical_to_cartesian(*grid)
```

genesis/sensors/raycaster/patterns.py
```diff
@@ -0,0 +1,297 @@
...
+        if W <= 0 or H <= 0:
+            raise ValueError("Image dimensions must be positive")
+
+        fx, fy, cx, cy = self._options.fx, self._options.fy, self._options.cx, self._options.cy
+        if fx is None or fy is None:
+            fx, fy = _compute_focal_lengths(W, H, self._options.fov_horizontal, self._options.fov_vertical)
+        if cx is None:
+            cx = W * 0.5
+        if cy is None:
+            cy = H * 0.5
```
duburcqa: All of this should be part of `__init__` and `__post_init__`. You are doing the same post-processing repeatedly while it could be done once and for all. Same logic: try to enforce a unique canonical type and value after init, so that, once initialised, many branches for edge-cases can be removed.

genesis/sensors/raycaster/patterns.py
```diff
@@ -0,0 +1,297 @@
...
+
+        u = torch.arange(0, W, dtype=gs.tc_float, device=gs.device) + 0.5
+        v = torch.arange(0, H, dtype=gs.tc_float, device=gs.device) + 0.5
+        uu, vv = torch.meshgrid(u, v, indexing="xy")
+
+        # standard camera frame coordinates
+        x_c = (uu - cx) / fx
+        y_c = (vv - cy) / fy
+        z_c = torch.ones_like(x_c, dtype=gs.tc_float, device=gs.device)
```
duburcqa: `torch.*_like` already "inherit" dtype and device. No need to specify it once again.

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+    h = ray_dir.cross(edge2)
+    a = edge1.dot(h)
+
+    # Check all conditions in sequence without early returns
+    valid = True
+
+    t = 0.0
+    u = 0.0
+    v = 0.0
+    f = 0.0
```
duburcqa: gs.ti_float(0.0)

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+    s = ti.math.vec3(0.0, 0.0, 0.0)
+    q = ti.math.vec3(0.0, 0.0, 0.0)
```
duburcqa: `ti.Vector.zero(gs.ti_float, 3)`

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+
+
+@ti.func
+def ray_triangle_intersection(ray_start, ray_dir, v0, v1, v2):
+    """
+    Möller-Trumbore ray-triangle intersection.
+
+    Returns: vec4(t, u, v, hit) where hit=1.0 if intersection found, 0.0 otherwise
+    """
+    result = ti.math.vec4(0.0, 0.0, 0.0, 0.0)
```
duburcqa: `ti.Vector.zero(gs.ti_float, 4)`

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+    Fast ray-AABB intersection test.
+    Returns the t value of intersection, or -1.0 if no intersection.
+    """
+    result = -1.0
+
+    # Use the slab method for ray-AABB intersection
+    inv_dir = 1.0 / ray_dir
+
+    # Handle potential division by zero with large values
+    large_value = ti.static(1 / gs.EPS)
```
duburcqa: Remove static here. We do not want to precompile statically for the value of EPS.

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+    if ti.abs(ray_dir.z) < gs.EPS:
+        inv_dir.z = large_value if ray_dir.z >= 0.0 else -large_value
+
+    t1 = (aabb_min - ray_start) * inv_dir
+    t2 = (aabb_max - ray_start) * inv_dir
+
+    tmin = ti.min(t1, t2)
+    tmax = ti.max(t1, t2)
+
+    t_near = ti.max(ti.max(tmin.x, tmin.y), tmin.z)
```
duburcqa: `t_near = ti.max(tmin.x, tmin.y, tmin.z)`

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+    t_far = ti.min(ti.min(tmax.x, tmax.y), tmax.z)
```
duburcqa: `t_far = ti.min(tmax.x, tmax.y, tmax.z)`

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+
+    # Check if ray intersects AABB
+    if t_near <= t_far and t_far >= 0.0:
+        result = ti.max(t_near, 0.0)
```
duburcqa: This could be simplified as:
```python
    t_near = ti.max(tmin.x, tmin.y, tmin.z, 0.0)
    t_far = ti.min(tmax.x, tmax.y, tmax.z)

    if t_near <= t_far:
        result = t_near
```

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+    inv_dir = 1.0 / ray_dir
+
+    # Handle potential division by zero with large values
+    large_value = ti.static(1 / gs.EPS)
+    if ti.abs(ray_dir.x) < gs.EPS:
+        inv_dir.x = large_value if ray_dir.x >= 0.0 else -large_value
+    if ti.abs(ray_dir.y) < gs.EPS:
+        inv_dir.y = large_value if ray_dir.y >= 0.0 else -large_value
+    if ti.abs(ray_dir.z) < gs.EPS:
+        inv_dir.z = large_value if ray_dir.z >= 0.0 else -large_value
```
duburcqa: This is too complicated for something that should be very simple:
```python
    sign = ti.select(ray_dir >= 0.0, 1.0, -1.0)
    ray_dir = sign * ti.max(ti.abs(ray_dir), gs.EPS)
    inv_dir = 1.0 / ray_dir
```

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+    is_world_frame: ti.types.ndarray(ndim=1),  # [n_sensors]
+    output_hits: ti.types.ndarray(ndim=2),  # [n_env, n_points_per_sensor * 4 * n_sensors]
+):
+    """
+    Taichi kernel for ray casting, accelerated by a Bounding Volume Hierarchy (BVH).
+
+    The result `output_hits` will be a 2D array of shape (n_env, n_points * 4) where in the second dimension,
+    the first n_points * 3 are hit points and the last n_points is hit ranges.
+    """
+    STACK_SIZE = ti.static(64)
```
duburcqa: I would define this as a global constant right after imports.

duburcqa: Could you also document the effect of this parameter? What is the impact is the value is very large or very small (performance, memory footprint...etc)? This parameter may be tuned for it is irrelevant / already just right?

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+        ray_direction_world = ti_normalize(ti_transform_by_quat(ray_dir_local, link_quat))
+
+        # --- 2. BVH Traversal ---
+        max_range = max_ranges[i_s]
+        hit_face = -1
+
+        # Stack for non-recursive traversal
+        stack = ti.Vector.zero(ti.i32, STACK_SIZE)
+        stack[0] = 0  # Start traversal at the root node (index 0)
+        stack_ptr = 1
```
duburcqa: This is just `stack_idx`. Nothing to do with a pointer from what I understand?

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+
+        ray_dir_local = ti.math.vec3(ray_directions[i_p, 0], ray_directions[i_p, 1], ray_directions[i_p, 2])
+        ray_direction_world = ti_normalize(ti_transform_by_quat(ray_dir_local, link_quat))
+
+        # --- 2. BVH Traversal ---
+        max_range = max_ranges[i_s]
+        hit_face = -1
+
+        # Stack for non-recursive traversal
+        stack = ti.Vector.zero(ti.i32, STACK_SIZE)
```
duburcqa: This should be called `node_stack` because `stack` is very generic and not specific enough.

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+                    assert node.left >= 0
+                    assert node.right >= 0
+                    assert node.left < bvh_nodes.shape[1]
+                    assert node.right < bvh_nodes.shape[1]
+                    # Push children onto the stack for further traversal
+                    # Make sure stack doesn't overflow
+                    if stack_ptr < ti.static(STACK_SIZE - 2):
+                        stack[stack_ptr] = node.left
+                        stack[stack_ptr + 1] = node.right
+                        stack_ptr += 2
```
duburcqa: Based on this, it seems that your "stack" should be a 2D tensor, or just two 1D tensors. This kind of flattened buffer  is irrelevant in this context and just makes the code harder to read.

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+                    if hit_result.w > 0.0 and hit_result.x < max_range and hit_result.x >= 0.0:
+                        max_range = hit_result.x
+                        hit_face = i_f
+                        # hit_u, hit_v could be stored here if needed
+
+                else:  # It's an INTERNAL node
+                    assert node.left >= 0
+                    assert node.right >= 0
+                    assert node.left < bvh_nodes.shape[1]
+                    assert node.right < bvh_nodes.shape[1]
```
duburcqa: Do not use assert in taichi kernel, it does not work reliably.

Could these asserts be triggered in practice? It seems to be asserting something that is prevented by design, so I would just remove them entirely as they are just confusing if not checking anything.

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+                output_hits[i_b, i_p * 3 + 1] = hit_point.y
+                output_hits[i_b, i_p * 3 + 2] = hit_point.z
+            else:
+                # Local frame output along provided local ray direction
+                hit_point = dist * ti_normalize(
+                    ti.math.vec3(
+                        ray_directions[i_p, 0],
+                        ray_directions[i_p, 1],
+                        ray_directions[i_p, 2],
+                    )
```
duburcqa: It does not fit in one line?

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+    draw_debug : bool, optional
+        If True and the rasterizer visualization is active, spheres will be drawn at every hit point.
+    debug_sphere_radius: float, optional
+        The radius of each debug hit point sphere drawn in the scene. Defaults to 0.02.
+    """
+
+    pattern: RaycastPattern | str
+    min_range: float = 0.0
+    max_range: float = 20.0
+    no_hit_value: float | None = None
```
duburcqa: As mentioned before, ideally you would like to enforce strict typing, with runtime default eventually.

duburcqa: For instance:
```python
@dataclass(init=False)
class Example:
    a: int = 1
    b: int

    def __init__(self, a: int, b: int | None = None):
        self.a = a
        # allow user to override b, else default to a
        self.b = b if b is not None else a
```

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+
+    debug_sphere_radius: float = 0.02
+
+
+@dataclass
+class RaycasterSharedMetadata(RigidSensorMetadataMixin, SharedSensorMetadata):
+    bvh: LBVH | None = None
+    aabb: AABB | None = None
+    only_cast_fixed: bool = False
+    map_lidar_faces: Any | None = None
```
duburcqa: Same issue.

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+            np_map_lidar_faces = np.arange(n_lidar_faces)
+            if self._shared_metadata.only_cast_fixed:
+                # count the number of faces in a fixed geoms
+                geom_is_fixed = np.logical_not(self._shared_metadata.solver.geoms_info.is_free.to_numpy())
+                faces_geom = self._shared_metadata.solver.faces_info.geom_idx.to_numpy()
+                n_lidar_faces = np.sum(geom_is_fixed[faces_geom])
+                if n_lidar_faces == 0:
+                    gs.raise_exception(
+                        "No fixed geoms found in the scene. To use only_cast_fixed, "
+                        "at least some entities should have is_free=False."
```
duburcqa: Only skip line when reaching max line width.

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+        # first lidar sensor initialization: build aabb and bvh
+        if self._shared_metadata.bvh is None:
+            self._shared_metadata.only_cast_fixed = self._options.only_cast_fixed  # set for first only
+
+            n_lidar_faces = self._shared_metadata.solver.faces_info.geom_idx.shape[0]
+            np_map_lidar_faces = np.arange(n_lidar_faces)
+            if self._shared_metadata.only_cast_fixed:
+                # count the number of faces in a fixed geoms
+                geom_is_fixed = np.logical_not(self._shared_metadata.solver.geoms_info.is_free.to_numpy())
+                faces_geom = self._shared_metadata.solver.faces_info.geom_idx.to_numpy()
```
duburcqa: `ti_to_numpy` has been added, which has been optimised for performance.

But it make be more efficient to keep everything in torch from the start? Not sure, this would require profiling performance.

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+            )
+
+            print("RaycasterSensor::build preupdate -----")
+            print("aabb n_batches")
+            print(self._shared_metadata.aabb.n_batches)
+            print("aabb n_aabbs")
+            print(self._shared_metadata.aabb.n_aabbs)
+            print("aabbs")
+            print(self._shared_metadata.aabb.aabbs)
+
```
duburcqa: Remove prints.

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+            print("aabb n_aabbs")
+            print(self._shared_metadata.aabb.n_aabbs)
+            print("aabbs")
+            print(self._shared_metadata.aabb.aabbs)
+            print()
+            print("bvh nodes")
+            print(self._shared_metadata.bvh.nodes)
+            print("bvh morton codes")
+            print(self._shared_metadata.bvh.morton_codes)
+
```
duburcqa: Remove prints.

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+        num_rays = math.prod(self.pattern_generator._return_shape)
+        self._shared_metadata.sensors_ray_start_idx.append(self._shared_metadata.total_n_rays)
+        self._shared_metadata.total_n_rays += num_rays
+
+        self._shared_metadata.points_to_sensor_idx = concat_with_tensor(
+            self._shared_metadata.points_to_sensor_idx,
+            [self._idx] * num_rays,
+            flatten=True,
+        )
```
duburcqa, line 481: One line.

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,589 @@
...
+        print("aabb n_aabbs")
+        print(shared_metadata.aabb.n_aabbs)
+        print("aabbs")
+        print(shared_metadata.aabb.aabbs)
+        print()
+        print("bvh nodes")
+        print(shared_metadata.bvh.nodes)
+        print("bvh morton codes")
+        print(shared_metadata.bvh.morton_codes)
+
```
duburcqa: Remove prints.

genesis/engine/scene.py
```diff
@@ -787,11 +794,15 @@ def build(
 
             self._is_built = True
 
-        if compile_kernels:
+        # This check should be removed when deprecating `compile_kernels`.
+        if compile_kernels or len(self._sim._sensor_manager._sensors) > 0:
```
duburcqa: No condition, just do this systematically.

genesis/recorders/plotters.py
```diff
@@ -466,12 +463,6 @@ def get_image_array(self):
         """
         Capture the plot image as a video frame.
 
-        Returns
-        -------
-        image_array : np.ndarray
-            The RGB image as a numpy array.
-        """
-        from matplotlib.backends.backend_agg import FigureCanvasAgg
```
duburcqa, line 474: Your rebase is broken.

genesis/sensors/contact_force.py
```diff
@@ -131,6 +131,9 @@ def build(self):
             self._shared_metadata.expanded_links_idx, link_idx, expand=(1,), dim=0
         )
 
+        if self._options.draw_debug:
+            self.debug_object = None
```
duburcqa: Just reset to None systematically. This extra check does not bring any benefit.

genesis/sensors/contact_force.py
```diff
@@ -298,6 +301,9 @@ def build(self):
             _to_tuple(self._options.max_force, length_per_value=3),
         )
 
+        if self._options.draw_debug:
+            self.debug_object = None
```
duburcqa: same.

tests/test_sensors.py
```diff
@@ -248,3 +248,63 @@ def test_rigid_tactile_sensors_gravity_force(show_viewer, tol, n_envs):
...
+    EXPECTED_DISTANCE = 1.5  # Distance from sensor to floor
+    NUM_RAYS_XY = 4  # Total number of rays is NUM_RAYS_XY ** 2
+    BOX_SIZE = 0.1  # Size of the box to which the LiDAR is attached
+
+    scene = gs.Scene(
+        profiling_options=gs.options.ProfilingOptions(show_FPS=False),
+        show_viewer=show_viewer,
+    )
+
+    plane = scene.add_entity(gs.morphs.Plane(is_free=not is_fixed))
```
duburcqa: Multiple lines for input arguments in unit tests even if it fits on one line. It improves readability I think.

tests/test_sensors.py
```diff
@@ -248,3 +248,63 @@ def test_rigid_tactile_sensors_gravity_force(show_viewer, tol, n_envs):
...
+    scene.build(n_envs=n_envs)
+
+    scene.step()
+
+    sensor_data = lidar.read()
+    hit_points = sensor_data["hit_points"]
+    distances = sensor_data["hit_ranges"]
+
+    expected_shape = (NUM_RAYS_XY, NUM_RAYS_XY) if n_envs == 0 else (n_envs, NUM_RAYS_XY, NUM_RAYS_XY)
+    assert distances.shape == expected_shape, f"Expected distances shape {expected_shape}, got {distances.shape}"
```
duburcqa: Remove message, it is useless, the error should be formatted properly by pytest already I think.

tests/test_sensors.py
```diff
@@ -248,3 +248,63 @@ def test_rigid_tactile_sensors_gravity_force(show_viewer, tol, n_envs):
...
+
+    assert_allclose(hit_points[..., 2], 0.0, tol=tol, err_msg="LiDAR hit point should be at ground level (z≈0)")
+    assert_allclose(
+        distances,
+        EXPECTED_DISTANCE,
+        tol=tol,
+        err_msg=f"LiDAR should measure distance {EXPECTED_DISTANCE}m to ground plane",
+    )
```
duburcqa, line 343: What if you put the lidar inside a sphere? Does not work or it would become completely crazy?

tests/test_sensors.py
```diff
@@ -248,3 +248,63 @@ def test_rigid_tactile_sensors_gravity_force(show_viewer, tol, n_envs):
         tol=NOISE * 10,
         err_msg="ContactForceSensor should read bias and noise and -gravity (normal) force clipped by max_force.",
     )
+
+
+@pytest.mark.required
+@pytest.mark.parametrize("n_envs", [0, 2])
+@pytest.mark.parametrize("is_fixed", [False, True])
+def test_lidar_grid_pattern_ground_distance(show_viewer, tol, n_envs, is_fixed):
```
duburcqa: Could you add a box on the scene and check that you are measuring the correct analytical distance? I'm not sure a plane is sufficient.

It may be more realistic and improve coverage to add a fixed plane and a moving cube on the scene, and just toggle `only_cast_fixed` based on test parameter `is_fixed`.

tests/test_sensors.py
```diff
@@ -248,3 +248,63 @@ def test_rigid_tactile_sensors_gravity_force(show_viewer, tol, n_envs):
...
+    hit_points = sensor_data["hit_points"]
+    distances = sensor_data["hit_ranges"]
+
+    expected_shape = (NUM_RAYS_XY, NUM_RAYS_XY) if n_envs == 0 else (n_envs, NUM_RAYS_XY, NUM_RAYS_XY)
+    assert distances.shape == expected_shape, f"Expected distances shape {expected_shape}, got {distances.shape}"
+
+    assert_allclose(hit_points[..., 2], 0.0, tol=tol, err_msg="LiDAR hit point should be at ground level (z≈0)")
+    assert_allclose(
+        distances,
+        EXPECTED_DISTANCE,
```
duburcqa, line 340: Why is the measured distance from the plane supposed to be constant? I'm confused.

examples/sensors/lidar_teleop.py
Milotrince: `pynput` is used in other examples, so zifan originally wrote this teleop example based on that. im not really sure if it should be used at all. pyrender visualizer has its own key handling, itd maybe be better to use that

examples/sensors/lidar_teleop.py
Milotrince: it may be relevant in scenarios where the environment is static, e.g. a robot mapping a scan of a building

Milotrince: i will try to do a simple benchmark

examples/sensors/lidar_teleop.py
Milotrince: there are still some examples that do this, but agreed, i dont like it as much

examples/sensors/lidar_teleop.py
Milotrince: sorry, my autoformatter does this. do you have a setup?

examples/sensors/lidar_teleop.py
Milotrince: is_free : bool, optional
        Whether the entity is free to move. Defaults to True. **This is only used for RigidEntity.**
        This determines whether the entity's geoms have their vertices put into StructFreeVertsState or
        StructFixedVertsState, and effectively whether they're stored per batch-element, or stored once and shared
        for the entire batch. That affects correct processing of collision detection.
        
    fixed : bool, optional
        Whether the baselink of the entity should be fixed. Defaults to False. **This is only used for RigidEntity.**
        
        
I also find the naming of `is_free` and `fixed` confusing.

Milotrince: @YilingQiao thoughts? I would suggest `is_free` -> `is_static` and `fixed` -> `fix_base`

examples/sensors/lidar_teleop.py
Milotrince: nope!

genesis/sensors/raycaster/base_pattern.py
Milotrince: my bad, dynamic patterns should not even be included in this PR. Removed. it can be implemented properly in the future if someone requests it.

genesis/sensors/raycaster/depth_camera.py
Milotrince: ah i didnt know docstrings are automatically inherited, pyright does this? i dont have pyright active because there are too many not-properly-typed things in our codebase

Milotrince: in that case i will also update other docstrings to not be redundant

Milotrince: will it show properly in autogenerated api docs?

genesis/sensors/raycaster/raycaster.py
Milotrince: does not fit on one line!

genesis/sensors/raycaster/raycaster.py
Milotrince: yes, and that is the case here

genesis/sensors/raycaster/raycaster.py
Milotrince: i did not write this kernel so i am not sure either behind the exact reasons for this choice. @aCodeDog ?

examples/sensors/lidar_teleop.py
duburcqa: I don't understand this documentation. It seems that `is_free` means at least one geometry of the entity is free to move, while `fixed` means that the base link of the entity is fixed.

If this is correct, I don't understand why `is_free` is user-specified instead of automatically deduced at runtime. If `is_free` is no longer user-specified, then I think `fixed` is good enough, because `fix_base` (or `is_root_fixed`...) is more accurate but also more confusing for some users as it may require more expertise to understand what it means I feel. But I'm ok with renaming this option.

YilingQiao: 1. To minimize the inconvenience to user change, can we keep `fixed` argument as a lot users are already use that
2. yeah, can we try deprecating is_free from the option and automatically deduce it during runtime? (i.e. all fixed 0-dof objects are static)

Milotrince: the only potential argument against this that i can think of is if someone had a rigged/multi-link object they wanted to load and have the entire thing fixed, without having `is_free` as an api option they would have to modify their object file. but this is not a common use case at all.

I'll go ahead with deprecating `is_free`

genesis/sensors/raycaster/depth_camera.py
Milotrince: thanks, didnt know about post_init!

genesis/sensors/raycaster/depth_camera.py
duburcqa: > will it show properly in autogenerated api docs?

Yes it should, or at least it is easy to configure sphinx to do this:
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_inherit_docstrings

genesis/sensors/raycaster/raycaster.py
Milotrince: ms difference in performance is not critical at build time, but i will do in torch for consistency

duburcqa: For sure, I must have skimmed through the code a bit too quickly and missed that this part was only executed at build time.

examples/sensors/lidar_teleop.py
duburcqa: No big deal.

examples/sensors/lidar_teleop.py
duburcqa: > pynput is used in other examples, so zifan originally wrote this teleop example based on that.

What is your point exactly?

> im not really sure if it should be used at all. pyrender visualizer has its own key handling, itd maybe be better to use that

It would be better, but we are missing the necessary facilities to easily register user-specified keyboard and mouse controls. But I think supporting this feature would be great and useful for many projects, including RL policy evaluation. I was supporting this feature in Jiminy.

Milotrince: made a backlog task in linear for this. will do ImportError handling as we did for matplotlib/pyqtgraph for now.

genesis/sensors/raycaster/patterns.py
Milotrince: sure. although it's not as if sub millimeter resolution would result in error, I can't really think of valid use cases of this

genesis/sensors/raycaster/patterns.py
Milotrince: not important if clipped

genesis/sensors/raycaster/patterns.py
Milotrince: I think the options/generator pattern is actually not appropriate in the case of RaycastPattern. Will turn into a plain ol' class.

examples/sensors/imu_franka.py
```diff
@@ -59,26 +59,34 @@ def main():
...
             scene.start_recording(
-                imu.read_ground_truth,
+                data_func,
                 gs.recorders.MPLLinePlot(title="IMU Ground Truth Data", labels=labels),
             )
         else:
             print("matplotlib or pyqtgraph not found, skipping real-time plotting.")
 
-    imu.start_recording(gs.recorders.NPZFile(filename="imu_data.npz"))
-
```
duburcqa, line 81: Why did you removed this? It was interesting as documentation no? Maybe just comment it out?

examples/sensors/lidar_teleop.py
```diff
@@ -0,0 +1,236 @@
...
+    def on_press(self, key: "keyboard.Key"):
+        with self.lock:
+            self.pressed_keys.add(key)
+
+    def on_release(self, key: "keyboard.Key"):
+        with self.lock:
+            self.pressed_keys.discard(key)
+
+    def get_cmd(self):
+        return self.pressed_keys
```
duburcqa, line 60: This is now part of `from examples.util.keyboard import KeyboardDevice` ?

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -87,7 +86,11 @@ def __init__(
 
         self._visualize_contact: bool = visualize_contact
 
-        self._is_free: bool = morph.is_free
+        # This determines whether the entity's geoms have their vertices put into StructFreeVertsState or
+        # StructFixedVertsState, and effectively whether they're stored per batch-element, or stored once and shared
+        # for the entire batch. That affects correct processing of collision detection.
```
duburcqa: I think this comment is very misleading. You should first state what being "free" means in this context:
```
This attribute specifies whether at least one of the geometries of the entity is free to move, as opposed to all of 
them being fixed in world (i.e. static). In practice, this attributes determines whether the geometry vertices 
associated with the entity are stored per batch-element and updated at every simulation step, or computed once at 
build time and shared among the entire batch. This affects correct processing of collision detection and sensor 
raycasting as a side-effect.
```

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -87,7 +86,11 @@ def __init__(
...
+        self._is_free: bool = False
+        # self._is_free: bool = any(tuple(not link.is_fixed for link in self._links))
```
duburcqa: I will update the current logics to make it possible to do this computation at init time instead of having to postpone it at runtime.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -87,7 +86,11 @@ def __init__(
 
         self._visualize_contact: bool = visualize_contact
 
-        self._is_free: bool = morph.is_free
```
duburcqa: Could you deprecate `is_free` property from rigid entities / links / geoms ...etc?

genesis/sensors/raycaster/raycaster.py
Milotrince: upgraded pydantic version to be able use Field default_factory with values from other parameters!

Milotrince: https://docs.pydantic.dev/latest/concepts/fields/#default-values
https://pydantic.dev/articles/pydantic-v2-11-release

tests/test_sensors.py
Milotrince: because GridPattern casts from a 2d plane instead of a single origin with multiple directions

tests/test_sensors.py
Milotrince: It should work.

genesis/sensors/raycaster/raycaster.py
Milotrince: I know the `| None` is not nice to have, but these are sensor shared metadata, the `SensorManager` instantiates `*SharedMetadata` on init and these values have to set by the sensor in `build` time. what would you recommend here?

genesis/sensors/raycaster/raycaster.py
Milotrince: I believe it is the right use of stack ? It is processing each node at a time. Adding left/right node happens in the case the current node is an internal node

genesis/sensors/raycaster/raycaster.py
Milotrince: https://madmann91.github.io/2021/01/06/bvhs-part-2.html
> The maximum size of that stack is given as the parameter max_stack_size. It can either be a “large enough” constant

https://forums.developer.nvidia.com/t/thinking-parallel-part-ii-tree-traversal-on-the-gpu/148342
>     NodePtr stack[64];

pyproject.toml
```diff
@@ -11,7 +11,7 @@ requires-python = ">=3.10,<3.14"
 dependencies = [
     "psutil",
     "gstaichi==2.4.0",
-    "pydantic>=2.7.1",
+    "pydantic>=2.11.0",
```
Milotrince, line 14: https://pydantic.dev/articles/pydantic-v2-11-release

primary motivation for bumping version: using default factory https://docs.pydantic.dev/latest/concepts/fields/#default-values

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,586 @@
...
+    links_quat: ti.types.ndarray(ndim=3),  # [n_env, n_sensors, 4]
+    ray_starts: ti.types.ndarray(ndim=2),  # [n_points, 3]
+    ray_directions: ti.types.ndarray(ndim=2),  # [n_points, 3]
+    max_ranges: ti.types.ndarray(ndim=1),  # [n_sensors]
+    no_hit_values: ti.types.ndarray(ndim=1),  # [n_sensors]
+    is_world_frame: ti.types.ndarray(ndim=1),  # [n_sensors]
+    points_to_sensor_idx: ti.types.ndarray(ndim=1),  # [n_points]
+    sensor_cache_offsets: ti.types.ndarray(ndim=1),  # [n_sensors] - cache start index for each sensor
+    sensor_point_offsets: ti.types.ndarray(ndim=1),  # [n_sensors] - point start index for each sensor
+    sensor_point_counts: ti.types.ndarray(ndim=1),  # [n_sensors] - number of points for each sensor
```
Milotrince, line 186: yes, all these are to get correct cache arrangement per sensor. this is not ideal, but it only uses a little bit more memory. this kind of came with the sensor data format changes i made. suggestions? (see lines 268-271)

genesis/sensors/raycaster/raycaster.py
duburcqa: If it is only defined at build time, then it is a different issue. There is no "solution" in this case, because it may really be none after initialisation.

genesis/sensors/raycaster/raycaster.py
duburcqa: Great! Could you document this and provide the link(s) as a reference?

genesis/sensors/raycaster/raycaster.py
duburcqa: > this is not ideal, but it only uses a little bit more memory.

Why is this not ideal? It looks ideal to me!

genesis/engine/entities/rigid_entity/rigid_entity.py
duburcqa: Taking care of this myself.

genesis/sensors/raycaster/raycaster.py
duburcqa: I see. This is fine.

examples/sensors/lidar_teleop.py
Milotrince: I was thinking it's nicer to have each example script fully self contained. would you prefer `examples.util.keyboard`?

duburcqa: Sounds good !

examples/sensors/lidar_teleop.py
Milotrince: yeah its negligible, and i think it is better to do the not-update-aabbs-if-everythings-fixed logic automatically

genesis/sensors/raycaster/raycaster.py
```diff
@@ -0,0 +1,601 @@
...
+            points_to_sensor_idx=shared_metadata.points_to_sensor_idx,
+            sensor_cache_offsets=shared_metadata.sensor_cache_offsets,
+            sensor_point_offsets=shared_metadata.sensor_point_offsets,
+            sensor_point_counts=shared_metadata.sensor_point_counts,
+            output_hits=(
+                shared_ground_truth_cache if shared_ground_truth_cache.is_contiguous() else shared_metadata.output_hits
+            ),
+        )
+        if not shared_ground_truth_cache.is_contiguous():
+            shared_ground_truth_cache[:] = shared_metadata.output_hits
```
Milotrince, line 554: `shared_ground_truth_cache` may not be contiguous since it is shape `(B, cache_size)` and we slice `[:, cache_slice]`.

duburcqa: In this case, the cache slice should be the first dimension.

examples/sensors/lidar_teleop.py
duburcqa: Yes, that would be great. I implemented this optimisation in rigid body in the PR refactoring is_fixed vs is_free.

genesis/sensors/raycaster/raycaster.py
Milotrince: this is the last change i need to make i think but i've been having trouble since it affects many sensor functions

duburcqa (dismissed):

duburcqa (dismissed):

duburcqa (dismissed):

duburcqa (approved):

PR #1789, opened 2025-09-29
[FEATURE] Added benchmark comparison job for production workflow

SonSang:
## Description

Added a benchmark comparison job for production workflow. After the benchmark is done, we compare this benchmark results against `5` previous benchmark results stored in artifacts. If either `compile_time` or `runtime_fps` is slower than the average of previous benchmarks by 10%, we make a PR comment to alarm the authors. The values are configurable in `production.yaml` as follows.

```
- name: Compare benchamrk stats vs last main runs and comment (no fail on regression)
        env:
          ...
          # Number of main runs to use as baselines
          REGRESSION_BASELINES: "5"
          # Regression tolerance (%)
          RUNTIME_REGRESSION_TOLERANCE_PCT: "10"
          COMPILE_REGRESSION_TOLERANCE_PCT: "10"
```

## Related Issue

## Checklist:

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -111,3 +111,26 @@ jobs:
         with:
           name: speed-test-results
           path: "/mnt/data/artifacts/speed_test_${{ env.SLURM_JOB_NAME }}.txt"
+
+      - name: Compare benchamrk stats vs last main runs and comment (no fail on regression)
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          WORKFLOW_NAME: Production
+          # Number of main runs to use as baselines
+          REGRESSION_BASELINES: "5"
+          # Regression tolerance (%)
+          RUNTIME_REGRESSION_TOLERANCE_PCT: "10"
+          COMPILE_REGRESSION_TOLERANCE_PCT: "10"
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          CURRENT_SPEED_TXT_PATH: "/mnt/data/artifacts/speed_test_${{ env.SLURM_JOB_NAME }}.txt"
+          OUTPUT_DEBUG_JSON: "mnt/data/artifacts/benchmark_compare_result.json"
+        run: |
+          python -m pip install --upgrade requests
+          python tests/compare_benchmarks.py
+
+      - name: Upload benchmark compare report artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: benchmark-compare-report
+          path: |
+            mnt/data/artifacts/benchmark_compare_result.json
--- a/tests/compare_benchmarks.py
+++ b/tests/compare_benchmarks.py
@@ -1,0 +1,281 @@
+#!/usr/bin/env python3
+import os, sys, io, re, zipfile, json
+from statistics import mean
+import requests
+
+GITHUB_API = "https://api.github.com"
+IGNORE_KEYS = {"compile_time", "runtime_fps", "realtime_factor"}
+
+# ---------- parsing ----------
+def parse_speed_txt_lines(lines):
+    """
+    Parse speed_test.txt lines into a map of { test_id: {"runtime_fps": float|None, "compile_time": float|None} }
+    """
+    out = {}
+    for line in lines:
+        if not line.strip():
+            continue
+        parts = [p.strip() for p in line.strip().split('|') if '=' in p]
+        kv = {}
+        for p in parts:
+            k, v = p.split('=', 1)
+            kv[k.strip()] = v.strip()
+        # test_id construction (exclude keys that are not needed for comparison, sort)
+        items = [(k, kv[k]) for k in kv.keys() if k not in IGNORE_KEYS]
+        items.sort()
+        test_id = '|'.join(f'{k}={v}' for k, v in items)
+        # parse values
+        rt = kv.get('runtime_fps')
+        ct = kv.get('compile_time')
+        try:
+            rt = float(rt) if rt is not None else None
+        except: rt = None
+        try:
+            ct = float(ct) if ct is not None else None
+        except: ct = None
+        out[test_id] = {"runtime_fps": rt, "compile_time": ct}
+    return out
+
+def parse_speed_txt_from_zip_bytes(zip_bytes):
+    """Parse speed_test*.txt files from artifact zip and merge into a dict"""
+    merged = {}
+    with zipfile.ZipFile(io.BytesIO(zip_bytes)) as zf:
+        for name in zf.namelist():
+            if re.search(r"speed_test.*\.txt$", name):
+                with zf.open(name) as f:
+                    lines = io.TextIOWrapper(f, encoding='utf-8', errors='ignore').read().splitlines()
+                    cur = parse_speed_txt_lines(lines)
+                    merged.update(cur)
+    return merged
+
+# ---------- GitHub helpers ----------
+def gh_json(session, url, params=None, headers=None):
+    r = session.get(url, params=params, headers=headers or {})
+    if r.status_code != 200:
+        raise RuntimeError(f"GET {url} failed: {r.status_code} {r.text}")
+    return r.json()
+
+def gh_bin(session, url):
+    # Accept is not forced: API redirects to zip
+    r = session.get(url, allow_redirects=True)
+    if r.status_code == 415:
+        r = session.get(url, headers={"Accept": "application/vnd.github+json"}, allow_redirects=True)
+    ct = (r.headers.get("Content-Type") or "").lower()
+    if r.status_code == 200 and ("zip" in ct or "octet-stream" in ct or "binary" in ct):
+        return r.content
+    try:
+        j = r.json()
+        raise RuntimeError(f"GET(bin) {url} failed: {r.status_code} {j}")
+    except ValueError:
+        raise RuntimeError(f"GET(bin) {url} failed: {r.status_code} {r.text[:300]}")
+
+# ---------- main ----------
+def main():
+    # Required/optional ENV
+    repo = os.environ.get("GITHUB_REPOSITORY")
+    run_id = os.environ.get("GITHUB_RUN_ID")
+    token = os.environ.get("GITHUB_TOKEN") or os.environ.get("GH_TOKEN")
+    pr_number = os.environ.get("PR_NUMBER")  # ${{ github.event.pull_request.number }}
+    workflow_name = os.environ.get("WORKFLOW_NAME", "Production")
+
+    # Regression threshold (%) — runtime is regression if decrease compared to mean is greater than threshold, 
+    # compile is regression if increase is greater than threshold
+    tol_runtime = float(os.environ.get("RUNTIME_REGRESSION_TOLERANCE_PCT", "10"))
+    tol_compile = float(os.environ.get("COMPILE_REGRESSION_TOLERANCE_PCT", "10"))
+
+    baselines_k = int(os.environ.get("REGRESSION_BASELINES", "5"))
+    current_txt_path = os.environ.get("CURRENT_SPEED_TXT_PATH")
+
+    # Debug output files (for artifact upload)
+    out_json_path = os.environ.get("OUTPUT_DEBUG_JSON", "benchmark_compare_result.json")
+
+    # 1) Fail if token is missing (policy)
+    if not token:
+        print("Missing GITHUB_TOKEN (or GH_TOKEN). Failing by policy.", file=sys.stderr)
+        sys.exit(2)
+
+    session = requests.Session()
+    session.headers.update({
+        "Authorization": f"Bearer {token}",
+        "X-GitHub-Api-Version": "2022-11-28",
+        "Accept": "application/vnd.github+json",
+    })
+
+    # 2) Parse speed_test.txt of current PR run (local file first)
+    current_map = {}
+    if current_txt_path and os.path.exists(current_txt_path):
+        with open(current_txt_path, "r", encoding="utf-8", errors="ignore") as f:
+            current_map = parse_speed_txt_lines(f.read().splitlines())
+    else:
+        if not (repo and run_id):
+            print("No CURRENT_SPEED_TXT_PATH and missing GITHUB_REPOSITORY/GITHUB_RUN_ID.", file=sys.stderr)
+            sys.exit(2)
+        owner, repo_name = repo.split("/", 1)
+        arts = gh_json(session, f"{GITHUB_API}/repos/{owner}/{repo_name}/actions/runs/{run_id}/artifacts")
+        for a in arts.get("artifacts", []):
+            if a.get("name") == "speed-test-results" and not a.get("expired", False):
+                zip_bytes = gh_bin(session, a["archive_download_url"])
+                current_map = parse_speed_txt_from_zip_bytes(zip_bytes)
+                break
+
+    if not current_map:
+        print("No current speed_test data found.", file=sys.stderr)
+        sys.exit(2)
+
+    # 3) Collect baselines from recent N runs of main branch in the same workflow
+    if not repo:
+        print("Missing GITHUB_REPOSITORY.", file=sys.stderr)
+        sys.exit(2)
+    owner, repo_name = repo.split("/", 1)
+
+    # Workflow ID
+    wfs = gh_json(session, f"{GITHUB_API}/repos/{owner}/{repo_name}/actions/workflows")
+    wf_id = None
+    for wf in wfs.get("workflows", []):
+        if wf.get("name") == workflow_name:
+            wf_id = wf["id"]; break
+    if not wf_id:
+        print(f"Workflow '{workflow_name}' not found.", file=sys.stderr)
+        sys.exit(2)
+
+    baseline_rt = {}  # test_id -> [runtime_fps...]
+    baseline_ct = {}  # test_id -> [compile_time...]
+    page = 1
+    collected = 0
+    while collected < baselines_k and page <= 10:
+        runs = gh_json(session, f"{GITHUB_API}/repos/{owner}/{repo_name}/actions/workflows/{wf_id}/runs",
+                       params={"branch": "main", "status": "success", "per_page": 50, "page": page})
+        page += 1
+        for r in runs.get("workflow_runs", []):
+            if str(r.get("id")) == str(run_id):
+                continue
+            r_arts = gh_json(session, r["artifacts_url"])
+            got = False
+            for a in r_arts.get("artifacts", []):
+                if a.get("name") == "speed-test-results" and not a.get("expired", False):
+                    zip_bytes = gh_bin(session, a["archive_download_url"])
+                    base_map = parse_speed_txt_from_zip_bytes(zip_bytes)
+                    for tid, vals in base_map.items():
+                        if (v := vals.get("runtime_fps")) is not None:
+                            baseline_rt.setdefault(tid, []).append(float(v))
+                        if (c := vals.get("compile_time")) is not None:
+                            baseline_ct.setdefault(tid, []).append(float(c))
+                    got = True
+                    break
+            if got:
+                collected += 1
+                if collected >= baselines_k:
+                    break
+
+    # 4) Compare (runtime_fps ↓, compile_time ↑ each based on mean)
+    runtime_regs = []  # (tid, curr, base_mean, delta_pct)
+    compile_regs = []  # (tid, curr, base_mean, delta_pct)
+    results = []       # per test combined result
+
+    for tid, vals in current_map.items():
+        curr_rt = vals.get("runtime_fps")
+        curr_ct = vals.get("compile_time")
+
+        base_rt_mean = None
+        base_ct_mean = None
+        d_rt = None
+        d_ct = None
+        st_rt = "no-baseline"
+        st_ct = "no-baseline"
+
+        # runtime comparison (baseline mean)
+        bases_rt = baseline_rt.get(tid, [])
+        if bases_rt and curr_rt is not None:
+            base_rt_mean = mean(bases_rt)
+            if base_rt_mean > 0:
+                d_rt = (curr_rt - base_rt_mean) / base_rt_mean * 100.0
+                st_rt = "ok" if d_rt >= -tol_runtime else "regressed"
+                if st_rt == "regressed":
+                    runtime_regs.append((tid, curr_rt, base_rt_mean, d_rt))
+
+        # compile comparison (baseline mean)
+        bases_ct = baseline_ct.get(tid, [])
+        if bases_ct and curr_ct is not None:
+            base_ct_mean = mean(bases_ct)
+            if base_ct_mean > 0:
+                d_ct = (curr_ct - base_ct_mean) / base_ct_mean * 100.0
+                st_ct = "ok" if d_ct <= tol_compile else "regressed"
+                if st_ct == "regressed":
+                    compile_regs.append((tid, curr_ct, base_ct_mean, d_ct))
+
+        results.append({
+            "id": tid,
+            "runtime": {"current": curr_rt, "baseline_mean": base_rt_mean, "delta_pct": d_rt, "status": st_rt},
+            "compile": {"current": curr_ct, "baseline_mean": base_ct_mean, "delta_pct": d_ct, "status": st_ct},
+        })
+
+    # 5) Create comment body (if there is a regression, comment)
+    comment_body = None
+    if pr_number and (runtime_regs or compile_regs):
+        def trunc(s, n=140): return (s[:n] + "…") if len(s) > n else s
+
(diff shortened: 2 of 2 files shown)


Kashu7100:
Do you think this should be against certain commit? There could be a chance that small degeneration (let's say -5% each time) pile up without getting caught by this check if we do against last 5 commits.

SonSang:
> Do you think this should be against certain commit? There could be a chance that small degeneration (let's say -5% each time) pile up without getting caught by this check if we do against last 5 commits.

Yes I agree. If we have a reliable standard, we can use it too. For now it is an initial stage and we need to adjust it. @YilingQiao How do you think this idea?

YilingQiao:
Yeah, this PR could be an initial point. We could introduce tests with longer horizon in the future. How to best benchmark regression might be a problem.

SonSang:
We need to first merge this PR to `main` to test the functionality of adding PR comment for alarming, because it is done by new workflow `alarm`. Right now it is not triggered because it is not in the `main` branch. I'll test the functionality and make another PR after this one is merged.

YilingQiao:
sure, can you first create another tiny PR for that alarm.yml? which we can merge first

SonSang:
Made another PR(#1791 ) for `alarm` workflow. Once it is merged, I'll test the PR comment functionality in this PR.

SonSang:
Confirmed PR commenting works in my fork, the comment is fired when there is regression as shown below.
Screenshot 2025-09-30 at 11 25 15 AM

.github/workflows/production.yml
```diff
@@ -15,6 +15,7 @@ concurrency:
 
 jobs:
   linux-gpu:
+    if: github.event_name != 'pull_request_target'
```
duburcqa: Could you add a one or two line comment to explain what this does, and add a URL to some source that explains it more in details.

Here is what I found about `pull_request_target`:
```
Instead of running against the workflow and code from the merge commit, the event runs against the workflow and code from the base of the PR. 
This means the workflow is running from a trusted source and is given access to a read/write token as well as secrets. 
For reference, see https://github.blog/news-insights/product-news/github-actions-improvements-for-fork-and-pull-request-workflows
```
Based on this, it is still not clear to me why you would ignore this event.

.github/workflows/production.yml
```diff
@@ -65,11 +66,20 @@ jobs:
...
+          # Regression tolerance (%)
+          RUNTIME_REGRESSION_TOLERANCE_PCT: "10"
+          COMPILE_REGRESSION_TOLERANCE_PCT: "10"
+          PR_NUMBER: ${{ github.event.pull_request.number }}
         run: |
           SLURM_JOB_NAME="$(uuidgen)_$(date +%Y%m%d_%H%M%S)"
           echo "SLURM_JOB_NAME=${SLURM_JOB_NAME}" >> $GITHUB_ENV
 
-          SLURM_ENV_VARS="NVIDIA_DRIVER_CAPABILITIES=all,HF_TOKEN"
+          SLURM_ENV_VARS="NVIDIA_DRIVER_CAPABILITIES=all,HF_TOKEN,GITHUB_TOKEN,GITHUB_REPOSITORY,GITHUB_RUN_ID,WORKFLOW_NAME,REGRESSION_BASELINES,RUNTIME_REGRESSION_TOLERANCE_PCT,COMPILE_REGRESSION_TOLERANCE_PCT,PR_NUMBER"
```
duburcqa: Split variable assignment in multiple lines for readability:
```bash
SLURM_ENV_VARS="\
NVIDIA_DRIVER_CAPABILITIES=all,\
HF_TOKEN,\
GITHUB_TOKEN,\
GITHUB_REPOSITORY,\
GITHUB_RUN_ID,\
WORKFLOW_NAME,\
REGRESSION_BASELINES,\
RUNTIME_REGRESSION_TOLERANCE_PCT,\
COMPILE_REGRESSION_TOLERANCE_PCT,\
PR_NUMBER"
```

.github/workflows/production.yml
```diff
@@ -91,7 +101,16 @@ jobs:
               tmate -S /tmp/tmate.sock display -p '#{tmate_ssh}'
               pip install -e '.[dev,render]' && \
               pytest --print -x -m 'benchmarks' ./tests && \
-              cat speed_test*.txt > '/mnt/data/artifacts/speed_test_${SLURM_JOB_NAME}.txt'
+              cat speed_test*.txt > '/mnt/data/artifacts/speed_test_${SLURM_JOB_NAME}.txt' && \
+              python -m pip install --upgrade requests && \
```
duburcqa: Is this really necessary? We should probably install this package directly in the image, but it is ok to do this for now.

tests/compare_benchmarks.py
duburcqa, line 1: I don't link this approach. Downloading artefacts using `requests` packages is very fragile and somehow abusing GitHub API. We are storing logs on WanDB precisely to avoid this.

If you want to download the last few runs, you can just do this:
```python
import wandb
from tqdm import tqdm
from datetime import datetime

api = wandb.Api()

max_num_runs = int(os.environ["REGRESSION_BASELINES"])
records = []
for run in api.runs("genesis-ai-company/genesis-benchmarks", order="-created_at"):
    if len(records) == max_num_runs:
        break
    if not run.state == "finished":
        continue

    config = run.config
    summary = run.summary
    records.append(
        {
            "benchmark_id": config["benchmark_id"],
            "timestamp": datetime.fromtimestamp(config["timestamp"]),
            "revision": config["revision"],
            "runtime_fps": summary.get("runtime_fps", float("nan")),
            "compile_time": summary.get("compile_time", float("nan")),
        }
    )

```

.github/workflows/alarm.yml
```diff
@@ -9,18 +9,76 @@ permissions:
...
+              print("No markdown body -> skip"); sys.exit(0)
+
+          tag = f"bench-guard-run:{wr['id']}"
+          cs = sess.get(f"{target_repo_api}/issues/{pr_num}/comments").json()
+          if any(tag in (c.get('body') or '') for c in cs):
+              print("Already commented -> skip"); sys.exit(0)
+
+          r = sess.post(f"{target_repo_api}/issues/{pr_num}/comments", json={'body': body})
+          print("Comment status:", r.status_code, r.text[:200])
+          PY
```
duburcqa: You should leverage existing GitHub Actions to easily write comments on Pull Requests instead of writing raw HTTP requests:
```bash
- uses: actions/github-script@v7
  env:
    COMMENT_BODY: ${{env.SCRIPT_OUTPUT}}
  with:
    script: |
      github.rest.issues.createComment({
        issue_number: context.issue.number,
        owner: context.repo.owner,
        repo: context.repo.repo,
        body: process.env.COMMENT_BODY
      })
```

SonSang:
@duburcqa @YilingQiao Now it only uses `alarm.yml` to check if there is regression and put PR comment if there is any. It is self-contained and does not checkout our codebase. If there is any regression, the report is shown as:
Screenshot 2025-09-30 at 11 26 03 PM

hughperkins:
Random question: why arent we seeing the benchmark scores reported into this PR itself too?

hughperkins:
Oh I see:
Screenshot 2025-09-30 at 11 52 29 PM

SonSang:
@hughperkins Yeah, and if there is no regression we do not post the report.

hughperkins:
Out of curiosity, why not trigger on pr too/instead? Something like:
```
on:
  pull_request:
    branches:
      - main
```
?

hughperkins:
> @hughperkins Yeah, and if there is no regression we do not post the report.

I would prefer we *always* post the report. What do you see an advantage of not posting the report?

hughperkins:
By the way, on the whole, I feel the ideal might be that we trigger the benchmark report by adding a specific comment to the pr, like 'benchmark me'. I've seen this appraoch before, and works fairly well. I confess I don't know how to configure that though. But maybe AI can help here? Thoughts?

SonSang:
> Out of curiosity, why not trigger on pr too/instead? Something like:
> 
> ```
> on:
>   pull_request:
>     branches:
>       - main
> ```
> 
> ?

@hughperkins We cannot post PR comment if the workflow is triggered by PR

hughperkins:
> @hughperkins We cannot post PR comment if the workflow is triggered by PR

Why? Is your concern taht there would be an infinite loop? Note that PR triggers only fire for new commits, not on new comments.

SonSang:
> > @hughperkins Yeah, and if there is no regression we do not post the report.
> 
> I would prefer we _always_ post the report. What do you see an advantage of not posting the report?

Hmm because we get notified when we get a comment in PR? I thought of this functionality as warning (if it is fine, we get nothing). But I can also post success if there is no regression. @YilingQiao Can I ask your opinion about this?

SonSang:
> > @hughperkins We cannot post PR comment if the workflow is triggered by PR
> 
> Why? Is your concern taht there would be an infinite loop? Note that PR triggers only fire for new commits, not on new comments.

No, it is github policy. We cannot write PR comment in the workflow triggered by PR. That is the reason why we added alarm.yml.

SonSang:
> By the way, on the whole, I feel the ideal might be that we trigger the benchmark report by adding a specific comment to the pr, like 'benchmark me'. I've seen this appraoch before, and works fairly well. I confess I don't know how to configure that though. But maybe AI can help here? Thoughts?

I feel it would be possible, but what’s the advantage of doing this? Because it is relatively cheap to read in the benchmark results and compare them, I think the cost is negligible

hughperkins:
> > By the way, on the whole, I feel the ideal might be that we trigger the benchmark report by adding a specific comment to the pr, like 'benchmark me'. I've seen this appraoch before, and works fairly well. I confess I don't know how to configure that though. But maybe AI can help here? Thoughts?
> 
> I feel it would be possible, but what’s the advantage of doing this? Because it is relatively cheap to read in the benchmark results and compare them, I think the cost is negligible

Because I would like the full report posted into the PR each time it runs. And if it runs for every commit, that's a lot of reports.

SonSang:
> > > By the way, on the whole, I feel the ideal might be that we trigger the benchmark report by adding a specific comment to the pr, like 'benchmark me'. I've seen this appraoch before, and works fairly well. I confess I don't know how to configure that though. But maybe AI can help here? Thoughts?
> > 
> > 
> > I feel it would be possible, but what’s the advantage of doing this? Because it is relatively cheap to read in the benchmark results and compare them, I think the cost is negligible
> 
> Because I would like the full report posted into the PR each time it runs. And if it runs for every commit, that's a lot of reports.

I see, that makes sense. @YilingQiao How about this?

YilingQiao:
I think it's already triggered by every PR? And it will post a report if there is a reguress @hughperkins 
```
on:
  workflow_run:
    workflows: ["Production"]
    types: [completed]
```
Every PR will run `production.yml` and this workflow is after that

SonSang:
> I think it's already triggered by every PR? And it will post a report if there is a reguress @hughperkins
> 
> ```
> on:
>   workflow_run:
>     workflows: ["Production"]
>     types: [completed]
> ```
> 
> Every PR will run `production.yml` and this workflow is after that

@hughperkins Oh yeah, this workflow is technically triggered at every PR. I thought the question was why don't we use
```
on:
  pull_requests: ...
```

Sorry if I misunderstood.

hughperkins:
> Every PR will run production.yml and this workflow is after that

Question: why isnt the workflow running on this PR too?

> And it will post a report if there is a reguress

Any reason for not posting the report even if there is not an obvious regression? I assume we are using some threshold, like say 10%? It might be useful to be able to see smaller regressions perhaps, by having the report always?

SonSang:
> > Every PR will run production.yml and this workflow is after that
> 
> Question: why isnt the workflow running on this PR too?

This workflow (`alarm.yml`) runs based on the yaml file in the main branch. Right now this workflow is dummy in the main branch, so this PR is not affected. If this PR is merged to the main branch, the future PR will be affected and this workflow will run.

.github/workflows/alarm.yml
```diff
@@ -18,9 +19,319 @@ jobs:
...
+          # ----- Read event / find PR -----
+          API = os.environ.get("GITHUB_API_URL", "https://api.github.com")
+          ev  = json.load(open(os.environ["GITHUB_EVENT_PATH"], "r", encoding="utf-8"))
+          wr  = ev["workflow_run"]
+          repo= ev["repository"]
+          owner, name = repo["full_name"].split("/", 1)
+          head_sha = wr.get("head_sha")
+          wr_id = wr["id"]
+
+          s = requests.Session()
```
YilingQiao: oh, nvm

SonSang: Oh my bad, I missed it.

YilingQiao (dismissed):

YilingQiao:
```
 I assume we are using some threshold, like say 10%? It might be useful to be able to see smaller regressions perhaps, by having the report always?
```
We could reduce the threshold to 0 if you prefer but just that every commit will have a report and it might be noisy. I don't have strong opinion though.

YilingQiao:
Isn't this workflow already triggered?
image

SonSang:
> Isn't this workflow already triggered? image

@hughperkins This workflow does nothing because this PR has not been merged yet. It just checkout the code and quit.

Screenshot 2025-10-01 at 2 46 58 AM

After this PR is merged, it will work as shown below
Screenshot 2025-10-01 at 2 48 32 AM

hughperkins:
> ```
>  I assume we are using some threshold, like say 10%? It might be useful to be able to see smaller regressions perhaps, by having the report always?
> ```
> 
> We could reduce the threshold to 0 if you prefer but just that every commit will have a report and it might be noisy. I don't have strong opinion though.

Yes, so what I'm thinking is that maybe:
- the workflow only runs when we comment in the PR, something like "run performance check", or some phrase we choose
- the workflow always posts the results into the PR, as a comment

YilingQiao:
This PR serves as an automatic alarm to notify people about regressions.
```
The workflow only runs when we comment on the PR, something like "run performance check"
```
This feels more like a tool, but I don’t think it can fully replace an automatic guard. Maybe we can add the tool in a future PR, while this PR remains as a safeguard?

hughperkins:
> This PR serves as an automatic alarm to notify people about regressions.
> 
> ```
> The workflow only runs when we comment on the PR.
> ```
> 
> This feels more like a tool, but I don’t think it can fully replace an automatic guard. Maybe we can add the tool in a future PR, while this PR remains as a safeguard?

Yeah I don't have a strong opinion on this. In my previous job, the regression tests were very expensive to run, so we only ran them once or twice on a PR - rather than every commit - to keep costs down.

YilingQiao:
Yeah, now we already run benchmark every commit. This PR is to detect regression based on the results. When we can a more expensive regression test, maybe we can set the mechanism you suggested, i.e. reviewer call the tool for the PR that might cause regression.

YilingQiao (approved):

hughperkins:
Question: how do we know when the check has run?

hughperkins:
Cool. And will this show the check in this list?

Screenshot 2025-10-02 at 10 50 26 AM

Or, if not, how do we find the URL of the table?

SonSang:
@hughperkins Now we can view the full comparison table by clicking the "View details" button on the check (even when there is no regression):
Screenshot 2025-10-02 at 10 48 28 AM

The full comparison table would be shown like:
Screenshot 2025-10-02 at 10 50 42 AM

SonSang:
> Cool. And will this show the check in this list?
> 
> Screenshot 2025-10-02 at 10 50 26 AM
> Or, if not, how do we find the URL of the table?

It will be visible for the future PRs after this PR is merged

hughperkins:
> > Cool. And will this show the check in this list?
> > Screenshot 2025-10-02 at 10 50 26 AM
> > Or, if not, how do we find the URL of the table?
> 
> It will be visible for the future PRs after this PR is merged

woo, fantastic! 💪🔥🔥🔥

.github/workflows/alarm.yml
```diff
@@ -9,6 +9,9 @@ permissions:
   contents: read
   actions: read
   issues: write
```
duburcqa: Why writing permission of issues is necessary? Only on Pull Request matters no?

.github/workflows/alarm.yml
```diff
@@ -18,9 +21,364 @@ jobs:
...
+          def fmt_pct(v, highlight=False):
+              if v is None or (isinstance(v,float) and math.isnan(v)): return "—"
+              s = f"{v:+.2f}%"
+              return f"**{s}**" if highlight else s
+
+          # ----- thresholds -----
+          tol_rt = float(os.environ.get("RUNTIME_REGRESSION_TOLERANCE_PCT","10"))
+          tol_ct = float(os.environ.get("COMPILE_REGRESSION_TOLERANCE_PCT","10"))
+          MAX_REVISIONS = int(os.environ.get("MAX_REVISIONS","5"))
+          NO_CHANGE_PATIENCE = int(os.environ.get("NO_CHANGE_PATIENCE","100"))
```
duburcqa: * Remove default values.
* Use lower case for Python variable.

.github/workflows/alarm.yml
```diff
@@ -18,9 +21,364 @@ jobs:
...
+
+          # ----- load artifact (current results) -----
+          artifacts_path = os.path.abspath(os.environ.get("ARTIFACTS_DIR", "./artifacts"))
+          if not os.path.exists(artifacts_path):
+              open(os.environ.get("PR_COMMENT_PATH","pr_comment.md"),"w").close()
+              open(os.environ.get("CHECK_BODY_PATH","check_output.md"),"w").close()
+              sys.exit(0)
```
duburcqa: * Remove default.
* Use pathlib to create file:
```python
from pathlib import Path

Path('path/to/file.txt').touch()
```

.github/workflows/alarm.yml
```diff
@@ -18,9 +21,364 @@ jobs:
...
+
+          current_txt_path = None
+          for root, _, files in os.walk(artifacts_path):
```
duburcqa: Use `pathlib` instead of `os`.

You should load all these path arguments from env variables as pathlib.Path objects directly.

.github/workflows/alarm.yml
```diff
@@ -18,9 +21,364 @@ jobs:
...
+              for fname in files:
+                  if fname.startswith("speed_test") and fname.endswith(".txt"):
+                      current_txt_path = os.path.join(root, fname); break
```
duburcqa: * No inline breaks
* Migrate this to pathlib:

```python
for path in files:
    if path.stem.startswith("speed_test") and path.suffix == ".txt":
        current_txt_path = root / path
        break
```

.github/workflows/alarm.yml
```diff
@@ -18,9 +21,364 @@ jobs:
...
+              if current_txt_path: break
```
duburcqa: No inline break.

.github/workflows/alarm.yml
```diff
@@ -18,9 +21,364 @@ jobs:
...
+          if current_txt_path is None:
```
duburcqa, line 142: You could use `for-else` pattern:

```python
for i in range(10):
    pass
else:
    print("Reached the end with breaking")
```


In this case, just replace `if current_txt_path is None:` by `else:`

.github/workflows/alarm.yml
```diff
@@ -18,9 +21,364 @@ jobs:
...
+              open(os.environ.get("PR_COMMENT_PATH","pr_comment.md"),"w").close()
+              open(os.environ.get("CHECK_BODY_PATH","check_output.md"),"w").close()
```
duburcqa: pathlib.Path.touch

.github/workflows/alarm.yml
```diff
@@ -18,9 +21,364 @@ jobs:
...
+              sys.exit(0)
+
+          with open(current_txt_path, "r", encoding="utf-8") as f:
+              current_bm = artifacts_parse_speed_txt_lines(f.readlines())
+
+          # ----- W&B baselines -----
+          if not os.getenv("WANDB_API_KEY"):
+              try: wandb.login(anonymous="allow", relogin=True)
+              except Exception: pass
```
duburcqa: * Is this fallback relevant? I think we could just skip alarm if WANDB_API_KEY is not properly configured no?
* No inline try-else.

.github/workflows/alarm.yml
```diff
@@ -18,9 +21,364 @@ jobs:
...
+          ENTITY = os.environ.get("WANDB_ENTITY","")
+          PROJECT= os.environ.get("WANDB_PROJECT","")
```
duburcqa: * No need for default.
* Wrong PROJECT= spacing.

.github/workflows/alarm.yml
```diff
@@ -18,9 +21,364 @@ jobs:
...
+                      no_change += 1
+                      if no_change >= NO_CHANGE_PATIENCE: break
+                  continue
+              if selected_revs is not None and rev not in selected_revs:
+                  no_change += 1
+                  if no_change >= NO_CHANGE_PATIENCE: break
+                  continue
+              if rev not in by_rev:
+                  by_rev[rev]={}; rev_order.append(rev)
+                  if len(rev_order) >= MAX_REVISIONS: selected_revs = set(rev_order)
```
duburcqa: Avoid inline statements.

duburcqa: I don't understand why this is so complicated. On my side, I can just read the summary. Which information are you missing?
```python
records = []
for run in tqdm(runs, desc="Fetching run summaries"):
    if not run.state == "finished":
        continue
    config = run.config
    summary = run.summary

    records.append(
        {
            "benchmark_id": config["benchmark_id"],
            "timestamp": pd.to_datetime(config["timestamp"], unit="s"),
            "revision": config["revision"],
            "runtime_fps": summary.get("runtime_fps", float("nan")),
            "compile_time": summary.get("compile_time", float("nan")),
        }
    )

```

.github/workflows/alarm.yml
```diff
@@ -18,9 +21,364 @@ jobs:
...
+                      no_change += 1
+                      if no_change >= NO_CHANGE_PATIENCE: break
+
+          def mean_of(metric, bid):
+              vals=[]
+              for r in by_rev.keys():
+                  v = by_rev.get(r,{}).get(bid,{}).get(metric)
+                  if isinstance(v,(int,float)) and not (isinstance(v,float) and math.isnan(v)):
+                      vals.append(float(v))
+              return stats.mean(vals) if vals else None
```
duburcqa: Just do `sum(vals) / len(vals)` for a simple mean.

.github/workflows/alarm.yml
```diff
@@ -18,9 +21,364 @@ jobs:
...
+          check_parts += ["### Runtime FPS", *table_rt, ""]
+          check_parts += ["### Compile Time", *table_ct, ""]
+          check_body = "\n".join(check_parts)
+
+          # ----- COMMENT body (only if regressions) -----
+          if reg_found:
+              comment_parts = [
+                  ":warning: **Benchmark Regression Detected**",
+                  *baseline_block,
+                  "",
```
duburcqa: Replace these `""` by `\n` when stating title lines `### ...`

.github/workflows/alarm.yml
```diff
@@ -18,9 +21,364 @@ jobs:
...
+              ]
+              comment_body = "\n".join(comment_parts)
+          else:
+              comment_body = ""
+
+          # write files
+          with open(os.environ.get("CHECK_BODY_PATH","check_output.md"), "w", encoding="utf-8") as f:
+              f.write(check_body+"\n")
+          with open(os.environ.get("PR_COMMENT_PATH","pr_comment.md"), "w", encoding="utf-8") as f:
+              f.write(comment_body+"\n")
```
duburcqa: * Use `pathlib.Path.write_text`

.github/workflows/alarm.yml
```diff
@@ -18,9 +21,364 @@ jobs:
...
+
+          # flag for next steps
+          open("HAS_REGRESSIONS.txt","w").write("true\n" if reg_found else "false\n")
```
duburcqa: This approach is very hacky. You should rather leverage the exit code `sys.exit(1)`.

```bash
{ python - <<'PY'; has_regression =$?; } || true
    [...]
PY
```

It will store the return code in variable `has_regression`

.github/workflows/alarm.yml
SonSang: When running this code on my local machine, I can also access `run.summary`. However, when running on CI, `run.summary` is not working and I could not find the reason. The alternative was to use `run.scan_history(keys=["runtime_fps","compile_time"])`. And this entire block became longer because I wanted to make it robust (e.g. several runs from different PRs come in mixed order). If we don't have to care about such things, I can make it much shorter. For now, I'll make it more concise for readability. Please let me know if we want to make it shorter.

duburcqa: Ok let me have a look.

PR #1795, opened 2025-10-01
[MISC] Remove 'is_free' that was confusing and partially redundant with 'is_fixed'.

duburcqa:
## Description

* Minor cleanup.
* Cleanup gstaichi and examples unit tests.
* Fix interactive viewer deadlock when failing to initialize the viewer on background thread.
* Remove 'is_free' that was confusing and partially redundant with 'is_fixed'.
* Only update vertices if necessary.

## Related issues

Resolves Genesis-Embodied-AI/Genesis/issues/1794

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -22,7 +22,7 @@ from genesis.utils.misc import ALLOCATE_TENSOR_WARNING, DeprecationError, ti_to_
 
 from ..base_entity import Entity
 from .rigid_equality import RigidEquality
-from .rigid_geom import RigidGeom, _kernel_get_free_verts, _kernel_get_fixed_verts
+from .rigid_geom import RigidGeom
 from .rigid_joint import RigidJoint
 from .rigid_link import RigidLink
 
@@ -56,7 +56,8 @@ class RigidEntity(Entity):
         geom_start=0,
         cell_start=0,
         vert_start=0,
-        verts_state_start=0,
+        free_verts_state_start=0,
+        fixed_verts_state_start=0,
         face_start=0,
         edge_start=0,
         vgeom_start=0,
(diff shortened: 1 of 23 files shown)


genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -59,9 +61,23 @@ def __init__(
...
         self._child_idxs: list[int] = list()
-        self._invweight: float | None = invweight
+
+        # 'is_fixed' attribute specifies whether the link is free to move.
+        # In practice, this attributes determines whether the geometry vertices associated with the entity are stored
+        # per batch-element and updated at every simulation step, or computed once at build time and shared among the
+        # entire batch. This affects correct processing of collision detection and sensor raycasting as a side-effect.
+        is_fixed = True
+        link = self
+        while True:
```
YilingQiao, line 72: Will this logic only check the path from the root to the current link?

duburcqa: Yes exactly! And since we are building the tree from root to leaves, and the parent links are guaranteed to be properly initialised already.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1970,32 +1975,38 @@ def set_quat(self, quat, envs_idx=None, *, relative=False, zero_velocity=True, u
...
+                n_free_verts += link.n_verts
+        n_verts = n_free_verts + n_fixed_verts
+
+        tensor = torch.empty(self._solver._batch_shape((n_verts, 3), True), dtype=gs.tc_float, device=gs.device)
+        _kernel_get_fixed_verts(
+            tensor[0, :n_fixed_verts], self._fixed_verts_state_start, n_fixed_verts, self._solver.fixed_verts_state
+        )
+        tensor[1:, :n_fixed_verts] = tensor[0, :n_fixed_verts]
+        _kernel_get_free_verts(
+            tensor[:, n_fixed_verts:], self._free_verts_state_start, n_free_verts, self._solver.free_verts_state
```
YilingQiao: why is fixed verts always before free verts? Is this consistent with face index?

duburcqa: Probably not. I will update the PR to preserve geometry ordering (and therefore face index).

duburcqa: done

Kashu7100:
Just to note: if I remember correctly, `is_free` was introduced to avoid the redundant geometry copy of the terrain to minimize the memory overhead. We did this to distinguish from `fixed` option, which articulated entities also use to indicate if the base if fixed to the world or not.

YilingQiao:
@Kashu7100 yes, now this PR automatically detects `is_free` so it will minimize memory automatically. Do you observe anything off? Or this purpose is not satisfied after this change?

duburcqa:
I can confirm that this behavior should be preserved, and even further extended to geometries of rigid entities.

YilingQiao (approved):

PR #1800, opened 2025-10-03
[MISC] Some Bug Fixes for Mesh Related Issues

ACMLCZH:
- Fixes EXR path return.
- Update `tex_offset` to `int64`. See gs-madrona(https://github.com/Genesis-Embodied-AI/gs-madrona/pull/40)
- Fixes bug that using `Raytracer` and `Rasterizer` simultaneously to render will cause error.
- Update shader parsing for USD.

### DIFF
--- a/genesis/vis/camera.py
+++ b/genesis/vis/camera.py
@@ -150,11 +150,11 @@ class Camera(RBC):
             if self._env_idx is not None:
                 gs.raise_exception("Binding a camera to one specific environment index not supported by BatchRender.")
         else:
+            self._rasterizer.add_camera(self)
             if self._raytracer is not None:
                 self._raytracer.add_camera(self)
                 self._is_batched = False
             else:
-                self._rasterizer.add_camera(self)
                 self._is_batched = self._visualizer.scene.n_envs > 0 and self._visualizer._context.env_separate_rigid
                 if self._is_batched:
                     gs.logger.warning(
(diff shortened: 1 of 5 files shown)


genesis/vis/camera.py
```diff
@@ -657,7 +657,7 @@ def set_pose(self, transform=None, pos=None, lookat=None, up=None, envs_idx=None
         # Refresh rendering backend to taken into account updated camera pose
         if self._raytracer is not None:
             self._raytracer.update_camera(self)
-        elif self._batch_renderer is None:
+        if self._batch_renderer is None:
```
duburcqa, line 660: Why this change? It is not possible to have both at the same time.

ACMLCZH: It is possible to use Raytracer to render RGB and Rasterizer to render normal or seg.

duburcqa: Yeah of course, but here it is about raytracer vs batchrender.

duburcqa:
Hum… we do not test the RayTracer for now… no need to right unit tests for this specific part.

genesis/vis/camera.py
ACMLCZH: Wait, isn't here when `BatchRender` is not set, we `update_camera` for the rasterizer? (when `BatchRender` is set, it does not need explicitly call update).

duburcqa: You are right... sorry about that.

YilingQiao (approved):

ACMLCZH:
Added unittest for the other parts. "EXR path" and "Raytracer" should wait for the Luisa unittest.

duburcqa (approved):

duburcqa (approved):

PR #1589, opened 2025-08-18
[FEATURE] Heterogeneous Simulation

YilingQiao:
## Description

This PR enables the rigid solver to simulate different objects in different environments. In other words, an entity can have different morphs in different environments. 

Currently, it only supports primitive and mesh morphs, and only the Pyrender rasterizer supports this feature. We plan to support variable articulation and additional rendering backends in the future.

Try 
```
python examples/rigid/heterogeneous_simulation.py -v
```

https://github.com/user-attachments/assets/f862d51e-b3f7-4f5b-8b2f-9e900542c8be

## Related Issue

Resolves Genesis-Embodied-AI/Genesis#1449
Resolves Genesis-Embodied-AI/Genesis/issues/333

## Checklist:

- [x] I read the **CONTRIBUTING** document.
- [x] I followed the `Submitting Code Changes` section of **CONTRIBUTING** document.
- [x] I tagged the title correctly (including BUG FIX/FEATURE/MISC/BREAKING)
- [x] I updated the [documentation](https://github.com/Genesis-Embodied-AI/genesis-doc) accordingly or no change is needed.
- [x] I tested my changes and added instructions on how to test it for reviewers.

- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1808, opened 2025-10-07
[FEATURE] Differentiable forward dynamics for rigid body sim.

SonSang:
## Description

This PR has following changes:

- Unify hibernation / non-hibernation code in rigid_solver_decomp.py
- Unify dynamic and static inner loops based on `is_backward` (static when `is_backward=True`)
- Differentiable formulations for functions in rigid_solver_decomp.py

We do not add additional dimension to the states in the rigid body simulation for the frames, because it incurs too much code change.

## Related Issue

Resolves Genesis-Embodied-AI/Genesis/issues/1022

## Motivation and Context

There is a unit test to verify by solving an optimization problem `tests/test_grad.py::test_differentiable_rigid`.

## Checklist:

### DIFF
--- a/genesis/engine/solvers/rigid/rigid_solver_decomp.py
+++ b/genesis/engine/solvers/rigid/rigid_solver_decomp.py
@@ -1794,7 +1794,7 @@ class RigidSolver(Solver):
         if gs.use_zerocopy:
             mask = (0, *indices_to_mask(qs_idx)) if self.n_envs == 0 else indices_to_mask(envs_idx, qs_idx)
             data = ti_to_torch(self._rigid_global_info.qpos, transpose=True, copy=False)
-            assign_indexed_tensor(data, mask, qpos, gs.tc_float)
+            assign_indexed_tensor(data, mask, qpos)
             if mask and isinstance(mask[0], torch.Tensor):
                 envs_idx = mask[0].reshape((-1,))
         else:
@@ -1910,7 +1910,7 @@ class RigidSolver(Solver):
             data = ti_to_torch(getattr(self.dofs_info, name), transpose=True, copy=False)
             num_values = len(tensor_list)
             for j, mask_j in enumerate(((*mask, ..., j) for j in range(num_values)) if num_values > 1 else (mask,)):
-                assign_indexed_tensor(data, mask_j, tensor_list[j], gs.tc_float)
+                assign_indexed_tensor(data, mask_j, tensor_list[j])
             return
 
         tensor_list = list(tensor_list)
@@ -1992,7 +1992,7 @@ class RigidSolver(Solver):
                 if velocity is None:
                     vel[mask] = 0.0
                 else:
-                    assign_indexed_tensor(vel, mask, velocity, gs.tc_float)
+                    assign_indexed_tensor(vel, mask, velocity)
                 if mask and isinstance(mask[0], torch.Tensor):
                     envs_idx = mask[0].reshape((-1,))
                 elif not isinstance(envs_idx, torch.Tensor):
@@ -2073,7 +2073,7 @@ class RigidSolver(Solver):
             ctrl_mode = ti_to_torch(self.dofs_state.ctrl_mode, transpose=True, copy=False)
             ctrl_mode[mask] = gs.CTRL_MODE.FORCE
             ctrl_force = ti_to_torch(self.dofs_state.ctrl_force, transpose=True, copy=False)
-            assign_indexed_tensor(ctrl_force, mask, force, gs.tc_float)
+            assign_indexed_tensor(ctrl_force, mask, force)
             return
 
         force, dofs_idx, envs_idx = self._sanitize_io_variables(
@@ -2092,7 +2092,7 @@ class RigidSolver(Solver):
             ctrl_pos = ti_to_torch(self.dofs_state.ctrl_pos, transpose=True, copy=False)
             ctrl_pos[mask] = 0.0
             ctrl_vel = ti_to_torch(self.dofs_state.ctrl_vel, transpose=True, copy=False)
-            assign_indexed_tensor(ctrl_vel, mask, velocity, gs.tc_float)
+            assign_indexed_tensor(ctrl_vel, mask, velocity)
             return
 
         velocity, dofs_idx, envs_idx = self._sanitize_io_variables(
@@ -2109,7 +2109,7 @@ class RigidSolver(Solver):
             ctrl_mode = ti_to_torch(self.dofs_state.ctrl_mode, transpose=True, copy=False)
             ctrl_mode[mask] = gs.CTRL_MODE.POSITION
             ctrl_pos = ti_to_torch(self.dofs_state.ctrl_pos, transpose=True, copy=False)
-            assign_indexed_tensor(ctrl_pos, mask, position, gs.tc_float)
+            assign_indexed_tensor(ctrl_pos, mask, position)
             ctrl_vel = ti_to_torch(self.dofs_state.ctrl_vel, transpose=True, copy=False)
             ctrl_vel[mask] = 0.0
             return
@@ -2128,9 +2128,9 @@ class RigidSolver(Solver):
             ctrl_mode = ti_to_torch(self.dofs_state.ctrl_mode, transpose=True, copy=False)
             ctrl_mode[mask] = gs.CTRL_MODE.POSITION
             ctrl_pos = ti_to_torch(self.dofs_state.ctrl_pos, transpose=True, copy=False)
-            assign_indexed_tensor(ctrl_pos, mask, position, gs.tc_float)
+            assign_indexed_tensor(ctrl_pos, mask, position)
             ctrl_vel = ti_to_torch(self.dofs_state.ctrl_vel, transpose=True, copy=False)
-            assign_indexed_tensor(ctrl_vel, mask, velocity, gs.tc_float)
+            assign_indexed_tensor(ctrl_vel, mask, velocity)
             return
 
         position, dofs_idx, _ = self._sanitize_io_variables(
--- a/examples/rendering/demo.py
+++ b/examples/rendering/demo.py
@@ -151,10 +151,8 @@ def main():
         spp=512,
     )
     scene.build()
-
-    ########################## forward + backward twice ##########################
     scene.reset()
-    horizon = 2000
+    horizon = 10
 
     for i in range(horizon):
         scene.step()
--- a/examples/sensors/camera_as_sensor.py
+++ b/examples/sensors/camera_as_sensor.py
@@ -1,0 +1,253 @@
+"""
+Example demonstrating camera sensors with different rendering backends.
+
+Creating cameras as sensors using add_sensor() with three backends
+Rasterizer, Raytracer and BatchRenderer.
+Test the attachment, add light, batch rendering functionalities.
+"""
+
+import os
+import matplotlib.pyplot as plt
+import genesis as gs
+from genesis.utils.misc import tensor_to_array
+from genesis.options.sensors import RasterizerCameraOptions, RaytracerCameraOptions, BatchRendererCameraOptions
+
+########################## init ##########################
+gs.init(seed=0, precision="32", backend=gs.gpu, logging_level="info")
+
+########################## check dependencies ##########################
+# Try to import LuisaRenderPy to determine if raytracer is available
+try:
+    import LuisaRenderPy
+
+    ENABLE_RAYTRACER = True
+    print("✓ LuisaRenderPy available - Raytracer will be enabled")
+except ImportError:
+    ENABLE_RAYTRACER = False
+    print("⊘ LuisaRenderPy not available - Raytracer will be disabled")
+
+try:
+    import gs_madrona
+
+    ENABLE_MADRONA = True
+    print("✓ gs_madrona available - BatchRenderer will be enabled")
+except ImportError:
+    ENABLE_MADRONA = False
+    print("⊘ gs_madrona not available - BatchRenderer will be disabled")
+ENABLE_MADRONA = ENABLE_MADRONA and (gs.backend == gs.cuda)
+########################## create a scene ##########################
+# Choose renderer based on raytracer availability
+if ENABLE_RAYTRACER:
+    renderer = gs.renderers.RayTracer(
+        env_surface=gs.surfaces.Emission(
+            emissive_texture=gs.textures.ColorTexture(color=(0.2, 0.3, 0.5)),
+        ),
+        env_radius=20.0,
+    )
+else:
+    # Use Rasterizer as fallback renderer
+    renderer = gs.renderers.Rasterizer()
+
+scene = gs.Scene(
+    rigid_options=gs.options.RigidOptions(
+        enable_collision=True,
+        gravity=(0, 0, -9.8),
+    ),
+    renderer=renderer,
+    show_viewer=False,
+)
+
+########################## entities ##########################
+plane = scene.add_entity(
+    morph=gs.morphs.Plane(),
+    surface=gs.surfaces.Rough(color=(0.4, 0.4, 0.4)),
+)
+
+sphere = scene.add_entity(
+    morph=gs.morphs.Sphere(pos=(0.0, 0.0, 2.0), radius=0.5),
+    surface=gs.surfaces.Smooth(color=(1.0, 0.5, 0.5)),
+)
+
+box = scene.add_entity(
+    morph=gs.morphs.Box(pos=(1.0, 1.0, 1.0), size=(0.3, 0.3, 0.3)),
+    surface=gs.surfaces.Rough(color=(0.5, 1.0, 0.5)),
+)
+
+########################## Camera Configurations ##########################
+# Define common camera parameters
+CAMERA_COMMON_KWARGS = dict(
+    {
+        "up": (0.0, 0.0, 1.0),
+        "near": 0.1,
+        "far": 100.0,
+    }
+)
+
+CAMERA_SENSORS_KWARGS = [
+    {
+        "name": "cam0",
+        "pos": (3.0, 0.0, 2.0),
+        "lookat": (0.0, 0.0, 1.0),
+        "fov": 60.0,
+        "attachment": None,  # No attachment
+        "lights": [{"pos": (2.0, 2.0, 5.0), "color": (1.0, 1.0, 1.0), "intensity": 1.0}],
+    },
+    {
+        "name": "cam1",
+        "pos": (0.0, 3.0, 2.0),
+        "lookat": (0.0, 0.0, 1.0),
+        "fov": 60.0,
+        "attachment": None,
+        "lights": [],
+    },
+    {
+        "name": "cam_attached",
+        "pos": (0.0, 0.0, 1.0),
+        "lookat": (0.0, 0.0, 0.0),
+        "fov": 70.0,
+        "attachment": {
+            "entity_idx": None,
+            "link_idx_local": 0,
+            "pos_offset": (0.0, 0.0, 0.0),
+            "euler_offset": (0.0, 0.0, 0.0),
+        },
+        "lights": [],
+    },
+]
+
+
+# Create camera configurations for all backends
+backends = [
+    ("raster", RasterizerCameraOptions, True),  # Always enabled
+    ("raytrace", RaytracerCameraOptions, ENABLE_RAYTRACER),
+    ("batch", BatchRendererCameraOptions, ENABLE_MADRONA),
+]
+
+backend_configs = {}
+for backend_name, options_class, enabled in backends:
+    if not enabled:
+        continue
+
+    configs = []
+    for camera_config in CAMERA_SENSORS_KWARGS:
+        name = f"{backend_name}_{camera_config['name']}"
+        res = (500, 600)
+
+        # Create options with common and backend-specific parameters
+        options_kwargs = {
+            "res": res,
+            "pos": camera_config["pos"],
+            "lookat": camera_config["lookat"],
+            "up": CAMERA_COMMON_KWARGS["up"],
+            "fov": camera_config["fov"],
+            "lights": camera_config["lights"],
+        }
+
+        # Handle attachment
+        attachment = camera_config["attachment"]
+        if attachment is not None:
+            # For attached cameras, set the entity_idx to the sphere's index
+            options_kwargs.update(
+                {
+                    "entity_idx": sphere.idx,
+                    "link_idx_local": attachment["link_idx_local"],
+                    "pos_offset": attachment["pos_offset"],
+                    "euler_offset": attachment["euler_offset"],
+                }
+            )
+
+        # Add backend-specific parameters
+        if backend_name == "raster":
+            options_kwargs.update({"near": CAMERA_COMMON_KWARGS["near"], "far": CAMERA_COMMON_KWARGS["far"]})
+        elif backend_name == "raytrace":
+            options_kwargs.update(
+                {
+                    "model": "pinhole",
+                    "spp": 64,
+                    "denoise": False,
+                }
+            )
+            if attachment is None:  # Only add env surface for non-attached cameras
+                options_kwargs.update(
+                    {
+                        "env_surface": gs.surfaces.Emission(
+                            emissive_texture=gs.textures.ColorTexture(color=(0.2, 0.3, 0.5)),
+                        ),
+                        "env_radius": 20.0,
+                    }
+                )
+        elif backend_name == "batch":
+            options_kwargs.update({"use_rasterizer": True})
+            if camera_config["lights"]:
+                adjusted_lights = [{**light, "directional": False} for light in camera_config["lights"]]
+                options_kwargs["lights"] = adjusted_lights
+
+        # Adjust lights for raytracer (different intensity/color)
+        if backend_name == "raytrace" and camera_config["lights"]:
+            adjusted_lights = [
+                {**light, "color": (10.0, 10.0, 10.0), "intensity": 1.0} for light in camera_config["lights"]
+            ]
+            options_kwargs["lights"] = adjusted_lights
+
+        options = options_class(**options_kwargs)
+        configs.append(
+            {
+                "name": name,
+                "options": options,
+                "attachment": camera_config["attachment"],
+            }
+        )
+
+    backend_configs[backend_name] = configs
+
+
+########################## Create Cameras ##########################
+cameras = {}
+
+for group_name, configs in backend_configs.items():
+    print(f"\n=== {group_name} Cameras ===")
+    for config in configs:
+        camera = scene.add_sensor(config["options"])
+        cameras[config["name"]] = camera
+
+    print(f"✓ Created {len(configs)} {group_name.lower()} cameras")
+
+
+########################## build ##########################
+n_envs = 1
+scene.build(n_envs=n_envs)
+
+########################## identify attached cameras ##########################
+print("\n=== Identifying Attached Cameras ===")
+
+# Identify cameras that are configured to be attached
+attached_cameras = []
+for group_name, configs in backend_configs.items():
+    for config in configs:
+        if config["attachment"] is not None:
+            camera = cameras[config["name"]]
+            attached_cameras.append(camera)
+            print(f"✓ {config['name']} is attached to sphere")
+
+print(f"✓ Identified {len(attached_cameras)} attached cameras")
+
+########################## simulate and render ##########################
+os.makedirs("camera_sensor_output", exist_ok=True)
+
+for i in range(100):
+    scene.step()
+    # Render every 10 steps
+    if i % 10 == 0:
+        print(f"\n--- Step {i} ---")
+
+        camera_data = {}
+        for cam_name, camera in cameras.items():
+            data = camera.read()
+            camera_data[cam_name] = data
+            print(f"  {cam_name.replace('_', ' ').title()} RGB shape: {data.rgb.shape}")
+
+        for cam_name, data in camera_data.items():
+            rgb_data = data.rgb[0] if data.rgb.ndim > 3 else data.rgb
+            suffix = "_env0" if n_envs > 1 else ""
+            filename = f"camera_sensor_output/{cam_name}{suffix}_step{i:03d}.png"
+            plt.imsave(filename, tensor_to_array(rgb_data))
--- a/genesis/engine/sensors/__init__.py
+++ b/genesis/engine/sensors/__init__.py
@@ -3,4 +3,5 @@ from . import contact_force
 from . import imu
 from . import raycaster
 from . import depth_camera
+from . import camera
 from .sensor_manager import SensorManager
--- a/genesis/engine/sensors/base_sensor.py
+++ b/genesis/engine/sensors/base_sensor.py
@@ -339,6 +339,11 @@ class RigidSensorMixin(Generic[RigidSensorMetadataMixinT]):
 
         batch_size = self._manager._sim._B
 
+        # If entity_idx is < 0, this is a static sensor (not attached to any link)
+        if self._options.entity_idx is None or self._options.entity_idx < 0:
+            self._link = None
+            return
+
         entity = self._shared_metadata.solver.entities[self._options.entity_idx]
         self._link = entity.links[self._options.link_idx_local]
         self._shared_metadata.links_idx = concat_with_tensor(
--- a/genesis/engine/sensors/camera.py
+++ b/genesis/engine/sensors/camera.py
@@ -1,0 +1,843 @@
+"""
+Camera sensors for rendering: Rasterizer, Raytracer, and Batch Renderer.
+"""
+
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any, Dict, List, NamedTuple, Optional, Type
+
+import numpy as np
+import torch
+
+import genesis as gs
+from genesis.options.sensors import (
+    RasterizerCameraOptions,
+    RaytracerCameraOptions,
+    BatchRendererCameraOptions,
+    SensorOptions,
+)
+
+from genesis.utils.geom import pos_lookat_up_to_T, trans_quat_to_T
+from genesis.utils.misc import tensor_to_array
+from genesis.vis.batch_renderer import BatchRenderer
+from genesis.options.renderers import BatchRenderer as BatchRendererOptions
+from genesis.options.vis import VisOptions
+from genesis.vis.rasterizer_context import RasterizerContext
+from .base_sensor import Sensor, SharedSensorMetadata, RigidSensorMixin, RigidSensorMetadataMixin
+from .sensor_manager import register_sensor
+
+
+if TYPE_CHECKING:
+    from genesis.utils.ring_buffer import TensorRingBuffer
+    from genesis.vis.rasterizer import Rasterizer
+    from genesis.vis.rasterizer_context import RasterizerContext
+    from genesis.vis.batch_renderer import BatchRenderer
+    from genesis.vis.raytracer import Raytracer
+    from genesis.morphs.rigid_link import RigidLink
+
+
+# ========================== Data Class ==========================
+
+
+class CameraData(NamedTuple):
+    """Camera sensor return data."""
+
+    rgb: torch.Tensor
+
+
+class MinimalVisualizerWrapper:
+    """
+    Minimal visualizer wrapper for BatchRenderer camera sensors.
+
+    BatchRenderer requires a visualizer-like object to provide camera information and context,
+    but camera sensors don't need the full visualizer functionality (viewer, UI, etc.).
+    This wrapper provides just the minimal interface expected by BatchRenderer while
+    avoiding the overhead of creating a full visualizer instance.
+    """
+
+    def __init__(self, scene, sensors, vis_options):
+        self.scene = scene
+        self._cameras = []  # Will be populated with camera wrappers
+        self._sensors = sensors  # Keep reference to sensors
+
+        # Create a minimal rasterizer context for camera frustum visualization
+        # (required by BatchRenderer even though cameras don't render frustums)
+        self._context = RasterizerContext(vis_options)
+        self._context.build(scene)
+        self._context.reset()
+
+
+class BaseCameraWrapper:
+    """Base class for camera wrappers to reduce code duplication."""
+
+    def __init__(self, sensor):
+        self.sensor = sensor
+        self.uid = sensor._idx
+        self.res = sensor._options.res
+        self.fov = sensor._options.fov
+        self.near = sensor._options.near
+        self.far = sensor._options.far
+
+
+class RasterizerCameraWrapper(BaseCameraWrapper):
+    """Lightweight wrapper object used by the rasterizer backend."""
+
+    def __init__(self, sensor: "RasterizerCameraSensor"):
+        super().__init__(sensor)
+        self.aspect_ratio = self.res[0] / self.res[1]
+
+
+class BatchRendererCameraWrapper(BaseCameraWrapper):
+    """Wrapper object used by the batch renderer backend."""
+
+    def __init__(self, sensor: "BatchRendererCameraSensor"):
+        super().__init__(sensor)
+        self.idx = len(sensor._shared_metadata.sensors)  # Camera index in batch
+        self.debug = False
+
+        # Initial pose
+        pos = torch.tensor(sensor._options.pos, dtype=gs.tc_float, device=gs.device)
+        lookat = torch.tensor(sensor._options.lookat, dtype=gs.tc_float, device=gs.device)
+        up = torch.tensor(sensor._options.up, dtype=gs.tc_float, device=gs.device)
+
+        # Store pos/lookat/up for later updates
+        self._pos = pos
+        self._lookat = lookat
+        self._up = up
+        self.transform = pos_lookat_up_to_T(pos, lookat, up)
+
+    def get_pos(self):
+        """Get camera position (for batch renderer)."""
+        n_envs = self.sensor._manager._sim._B
+        if n_envs == 0:
+            return self._pos.unsqueeze(0)
+        else:
+            return self._pos.unsqueeze(0).expand(n_envs, -1)
+
+    def get_quat(self):
+        """Get camera quaternion (for batch renderer)."""
+        from genesis.utils.geom import T_to_trans_quat
+
+        _, quat = T_to_trans_quat(self.transform.unsqueeze(0))
+        n_envs = self.sensor._manager._sim._B
+        if n_envs > 0:
+            return quat.expand(n_envs, -1)
+        return quat
+
+
+# ========================== Shared Metadata ==========================
+
+
+@dataclass
+class RasterizerCameraSharedMetadata(RigidSensorMetadataMixin, SharedSensorMetadata):
+    """Shared metadata for all Rasterizer cameras."""
+
+    # Rasterizer instance
+    renderer: Optional["Rasterizer"] = None
+    # RasterizerContext instance
+    context: Optional["RasterizerContext"] = None
+    # List of light dictionaries
+    lights: Optional[List[Dict[str, Any]]] = None
+    # List of RasterizerCameraSensor instances
+    sensors: Optional[List["RasterizerCameraSensor"]] = None
+    # {sensor_idx: np.ndarray with shape (B, H, W, 3)}
+    image_cache: Optional[Dict[int, np.ndarray]] = None
+    # Track when rasterizer cameras were last updated
+    last_render_timestep: int = -1
+
+
+@dataclass
+class RaytracerCameraSharedMetadata(RigidSensorMetadataMixin, SharedSensorMetadata):
+    """Shared metadata for all Raytracer cameras."""
+
+    # Raytracer instance
+    renderer: Optional["Raytracer"] = None
+    # List of light objects
+    lights: Optional[List[Any]] = None
+    # List of RaytracerCameraSensor instances
+    sensors: Optional[List["RaytracerCameraSensor"]] = None
+    # {sensor_idx: np.ndarray with shape (B, H, W, 3)}
+    image_cache: Optional[Dict[int, np.ndarray]] = None
+    # Track when raytracer cameras were last updated
+    last_render_timestep: int = -1
+
+
+@dataclass
+class BatchRendererCameraSharedMetadata(RigidSensorMetadataMixin, SharedSensorMetadata):
+    """Shared metadata for all Batch Renderer cameras."""
+
+    # BatchRenderer instance
+    renderer: Optional["BatchRenderer"] = None
+    # gs.List of lights
+    lights: Optional[Any] = None
+    # List of BatchRendererCameraSensor instances
+    sensors: Optional[List["BatchRendererCameraSensor"]] = None
+    # {sensor_idx: np.ndarray with shape (B, H, W, 3)}
+    image_cache: Optional[Dict[int, np.ndarray]] = None
+    # Track when batch was last rendered
+    last_render_timestep: int = -1
+    # MinimalVisualizerWrapper instance
+    visualizer_wrapper: Optional["MinimalVisualizerWrapper"] = None
+
+
+# ========================== Base Camera Sensor ==========================
+
+
+class BaseCameraSensor(RigidSensorMixin, Sensor[SharedSensorMetadata]):
+    """
+    Base class for camera sensors that render RGB images into an internal image_cache.
+
+    This class centralizes:
+    - Attachment handling via RigidSensorMixin
+    - The _stale flag used for auto-render-on-read
+    - Common Sensor cache integration (shape/dtype)
+    - Shared read() method returning torch tensors
+    """
+
+    def __init__(self, options: "SensorOptions", idx: int, data_cls: Type[CameraData], manager: "gs.SensorManager"):
+        super().__init__(options, idx, data_cls, manager)
+        self._stale: bool = True
+
+    # ========================== Cache Integration (shared) ==========================
+
+    def _get_return_format(self) -> tuple[tuple[int, ...], ...]:
+        w, h = self._options.res
+        return ((h, w, 3),)
+
+    @classmethod
+    def _get_cache_dtype(cls) -> torch.dtype:
+        return torch.uint8
+
+    @classmethod
+    def _update_shared_ground_truth_cache(
+        cls, shared_metadata: SharedSensorMetadata, shared_ground_truth_cache: torch.Tensor
+    ):
+        pass
+
+    @classmethod
+    def _update_shared_cache(
+        cls,
+        shared_metadata: SharedSensorMetadata,
+        shared_ground_truth_cache: torch.Tensor,
+        shared_cache: torch.Tensor,
+        buffered_data: "TensorRingBuffer",
+    ):
+        # No per-step measured-cache update for cameras (handled lazily on read()).
+        pass
+
+    def _draw_debug(self, context: "RasterizerContext", buffer_updates: dict[str, np.ndarray]):
+        """No debug drawing for cameras."""
+        pass
+
+    # ========================== Attachment handling ==========================
+
+    @gs.assert_built
+    def move_to_attach(self):
+        """
+        Move the camera to follow the currently attached rigid link.
+
+        Uses a shared transform computation and delegates to _apply_camera_transform().
+        """
+        if self._link is None:
+            gs.raise_exception("Camera not attached to any rigid link.")
+
+        # Use pos directly as offset from link
+        pos_offset = torch.tensor(self._options.pos, dtype=gs.tc_float, device=gs.device)
+        offset_T = trans_quat_to_T(pos_offset, torch.tensor([1.0, 0.0, 0.0, 0.0], dtype=gs.tc_float, device=gs.device))
+
+        camera_T = _camera_compute_T_from_link(self._link, offset_T)
+        self._apply_camera_transform(camera_T)
+
+    # ========================== Hooks for subclasses ==========================
+
+    def _apply_camera_transform(self, camera_T: torch.Tensor):
+        """Apply the computed camera transform to the backend-specific camera representation."""
+        pass
+
+    def _render_current_state(self):
+        """Perform the actual render for the current state; subclasses must implement."""
+        pass
+
+    # ========================== Shared read() ==========================
+
+    def _get_image_cache_entry(self):
+        """Return this sensor's entry in the shared image cache."""
+        return self._shared_metadata.image_cache[self._idx]
+
+    def _ensure_rendered_for_current_state(self):
+        """Ensure this camera has an up-to-date render before reading.
+
+        Base handles staleness and timestamps; subclasses implement _render_current_state().
+        """
+        scene = self._manager._sim.scene
+
+        # If the scene time advanced, mark all cameras as stale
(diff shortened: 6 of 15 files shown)


examples/collision/pyramid.py
```diff
@@ -23,6 +23,7 @@ def main():
     scene = gs.Scene(
         sim_options=gs.options.SimOptions(
             dt=0.01,
+            requires_grad=True,
```
duburcqa: Are you sure about that?

examples/collision/pyramid.py
```diff
@@ -8,7 +8,7 @@ def main():
     parser.add_argument("--pile_type", type=str, default="falling", choices=["static", "falling"])
     parser.add_argument("--num_cubes", type=int, default=5, choices=range(5, 11))
     parser.add_argument("--cpu", action="store_true", help="Use CPU backend instead of GPU")
-    parser.add_argument("--steps", type=int, default=100)
+    parser.add_argument("--steps", type=int, default=10000)
```
duburcqa: Are you sure about that? If so, you should rather use `10000 if "PYTEST_VERSION" not in os.environ else 5`

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1440,6 +1440,7 @@ def _kernel_forward_kinematics(
                 self._solver.entities_info,
                 self._solver._rigid_global_info,
                 self._solver._static_rigid_sim_config,
+                False,
```
duburcqa: Use keyword argument when passing unnamed variables.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1466,6 +1467,7 @@ def _kernel_forward_kinematics(
...
```
duburcqa: Use keyword argument when passing unnamed variables.

genesis/utils/path_planning.py
```diff
@@ -429,6 +429,7 @@ def _kernel_rrt_step1(
                         entities_info,
                         rigid_global_info,
                         self._solver._static_rigid_sim_config,
+                        False,
```
duburcqa: Use keyword argument when passing unnamed variables.

genesis/utils/path_planning.py
```diff
@@ -798,6 +799,7 @@ def _kernel_rrt_connect_step1(
...
```
duburcqa: Use keyword argument when passing unnamed variables.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -36,6 +36,8 @@
 # Minimum ratio between simulation timestep `_substep_dt` and time constant of constraints
 TIME_CONSTANT_SAFETY_FACTOR = 2.0
 
+USE_BACKWARD_LOGIC_IN_FORWARD = False
```
duburcqa: Add a comment to explain what this does.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -95,6 +96,23 @@ def __init__(
...
+        self.init_tgt_vars()
+        self.init_ckpt()
+
+    def init_tgt_keys(self):
+
+        self._tgt_keys = ["pos", "quat", "qpos", "dofs_velocity"]
+
+    def init_tgt_vars(self):
+
+        # temp variable to store targets for next step
```
duburcqa: Do not skip line on top of undocumented methods.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1620,6 +1640,128 @@ def plan_path(
...
+        _tgt = self._tgt.copy()
+        for tgt in _tgt:
+            k = tgt["key"]
+            assert k in self._tgt_keys, f"Invalid target key: {k} not in {self._tgt_keys}"
+
+            # We do not need zero velocity here because if it was true, [set_dofs_velocity] from zero_velocity would
+            # be in [tgt]
+            zero_velocity = False
+            if k == "pos":
+                _pos = tgt["pos"]
```
duburcqa: This pattern is weird. It looks like `self._tgt` should just be a dictionary of dictionary, using `tgt["key"]` as top-level key.

Note that built-in python `dict` is preserving insertion order as per official python standard, just like sequence containers `list` and `tuple`.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1620,6 +1640,128 @@ def plan_path(
...
+                self.set_qpos(
+                    _qpos, qs_idx_local=_qs_idx_local, envs_idx=_envs_idx, zero_velocity=zero_velocity, unsafe=_unsafe
+                )
+            elif k == "dofs_velocity":
+                _velocity = tgt["velocity"]
+                _dofs_idx_local = tgt["dofs_idx_local"]
+                _envs_idx = tgt["envs_idx"]
+                _unsafe = tgt["unsafe"]
+
+                self.set_dofs_velocity(_velocity, dofs_idx_local=_dofs_idx_local, envs_idx=_envs_idx, unsafe=_unsafe)
```
duburcqa: This seems to be a perfect use case for `match-case`!

```python
for key, tgt in self._tgt.items():
    match key:
        case "pos":
            [...]
        case "quat":
            [...]
        case _:
            gs.raise_exception(f"Invalid target key: {k} not in {self._tgt_keys}")
```

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1620,6 +1640,128 @@ def plan_path(
...
+                    raise NotImplementedError("Backward pass for set_qpos_grad is not implemented yet.")
+
+            elif k == "dofs_velocity":
+                _velocity = tgt["velocity"]
+                _dofs_idx_local = tgt["dofs_idx_local"]
+                _envs_idx = tgt["envs_idx"]
+                _unsafe = tgt["unsafe"]
+
+                if _velocity is not None and _velocity.requires_grad:
+                    _velocity._backward_from_ti(self.set_dofs_velocity_grad, _dofs_idx_local, _envs_idx, _unsafe)
```
duburcqa: Why are you prefixing with `_` ? What is the meaning of this? This usually means private variable, which obviously is not applicable here.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1620,6 +1640,128 @@ def plan_path(
...
+
+                self.set_dofs_velocity(_velocity, dofs_idx_local=_dofs_idx_local, envs_idx=_envs_idx, unsafe=_unsafe)
+
+        self._tgt = []
+
+    def process_input_grad(self):
+        index = self._sim.cur_step_local - self._sim._steps_local
+        _tgt = self._tgt_buffer[index].copy()
+
+        for tgt in reversed(_tgt):
```
duburcqa: Using `_tgt` vs `tgt` to distinguish top-level container from element is far from ideal. What about something like `tgt` vs `data_kwargs` or something like this?

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1620,6 +1640,128 @@ def plan_path(
...
+            k = tgt["key"]
+            assert k in self._tgt_keys, f"Invalid target key: {k} not in {self._tgt_keys}"
+            if k == "pos":
+                _pos = tgt["pos"]
+                _envs_idx = tgt["envs_idx"]
+                _relative = tgt["relative"]
+                _unsafe = tgt["unsafe"]
+
+                if _pos is not None and _pos.requires_grad:
+                    _pos._backward_from_ti(self.set_pos_grad, _envs_idx, _relative, _unsafe)
```
duburcqa: Why are you unpacking all these arguments to pass them back? It would be more clear and future proof to just do:
```python
_pos._backward_from_ti(self.set_pos_grad, **tgt)
```

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1620,6 +1640,128 @@ def plan_path(
...
+        for tgt in reversed(_tgt):
+            k = tgt["key"]
+            assert k in self._tgt_keys, f"Invalid target key: {k} not in {self._tgt_keys}"
+            if k == "pos":
+                _pos = tgt["pos"]
+                _envs_idx = tgt["envs_idx"]
+                _relative = tgt["relative"]
+                _unsafe = tgt["unsafe"]
+
+                if _pos is not None and _pos.requires_grad:
```
duburcqa: Why are you guarding for `_pos is not None`? It is you that added this `None` value in `tgt` in the first place? If so, it would make more sense to simple skip adding keys that are irrelevant.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1620,6 +1640,128 @@ def plan_path(
...
+
+    def reset_grad(self):
+        self._tgt_buffer.clear()
+
+    @gs.assert_built
+    def get_state(self):
+        state = RigidEntityState(self, self._sim.cur_step_global)
+
+        solver_state = self._solver.get_state()
+        pos = solver_state.links_pos[:, self._base_links_idx].squeeze(-2)
```
duburcqa: Only use negative squeeze index if it is necessary, i.e. to support optional starting dimensional. In this case, `1` ?


Beyond that, not that you could extract unsqueeze pos directly if you want:
```python
pos = solver_state.links_pos[:, self._base_links_idx : self._base_links_idx + 1]
```

Note that `pos = solver_state.links_pos[:, [self._base_links_idx]]` would also work but it is MUCH slower because, first, the list of indices must be converted to a torch tensor on device, then it will rely on fancy indexing and return a copy instead of a view.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1980,6 +2134,18 @@ def set_pos(self, pos, envs_idx=None, *, relative=False, zero_velocity=True, uns
         if zero_velocity:
             self.zero_all_dofs_velocity(envs_idx, unsafe=unsafe)
 
+    @gs.assert_built
+    def set_pos_grad(self, envs_idx, relative, unsafe, pos_grad):
+        tmp_pos_grad = pos_grad.unsqueeze(-2).clone()
+        self._solver.set_base_links_pos_grad(
```
duburcqa: Could you explain why cloning is necessary in this context? If it is not, just remove it. But you need to know whatever it is guaranteed to be continuous or discontinuous by design.

genesis/engine/solvers/rigid/constraint_noslip.py
```diff
@@ -36,10 +36,12 @@ def kernel_build_efc_AR_b(
             rigid_solver.func_solve_mass_batched(
                 constraint_state.Mgrad,
                 constraint_state.Mgrad,
+                constraint_state.Mgrad,  # this will not be used anyway because is_backward is False
```
duburcqa: I don't get why specifying a different out variable for the value and its gradient is an option in the first place. It seems that it is never the case in your PR. If so, I would suggest removing this option that is adding complexity (and therefore increasing maintenance burden because in theory it should be unit-tested) without any benefit.

genesis/engine/states/entities.py
```diff
@@ -188,3 +188,49 @@ def vel(self):
...
     def active(self):
         return self._active
+
+
+class RigidEntityState(RBC):
+    """
+    Dynamic state queried from a genesis RigidEntity.
+    """
+
+    def __init__(self, entity, s_global):
```
duburcqa, line 198: `s_global` is not a good variable name.

genesis/engine/states/entities.py
```diff
@@ -188,3 +188,49 @@ def vel(self):
...
+        self._entity = entity
+        self._s_global = s_global
+
+        num_batch = self._entity._solver._B
+        self._pos = gs.zeros(
+            (num_batch, 3),
+            dtype=float,
+            requires_grad=self._entity.scene.requires_grad,
```
duburcqa: What about extracting `requires_grad`, just like you did for `num_batch`, so that everything can fit on one line.

genesis/utils/geom.py
```diff
@@ -83,8 +83,11 @@ def ti_rotvec_to_R(rotvec):
 def ti_rotvec_to_quat(rotvec):
     quat = ti.Vector.zero(gs.ti_float, 4)
 
-    theta = rotvec.norm()
-    if theta > gs.EPS:
+    # We need to use [norm_sqr] instead of [norm] to avoid nan gradients in the backward pass. Even when theta = 0,
+    # the gradient of [norm] operation is computed and used (it is multiplied by 0 but still results in nan gradients).
+    thetasq = rotvec.norm_sqr()
+    if thetasq > (gs.EPS * gs.EPS):
```
duburcqa: Just use `gs.EPS ** 2`

genesis/utils/geom.py
```diff
@@ -83,8 +83,11 @@ def ti_rotvec_to_R(rotvec):
 def ti_rotvec_to_quat(rotvec):
     quat = ti.Vector.zero(gs.ti_float, 4)
 
-    theta = rotvec.norm()
-    if theta > gs.EPS:
+    # We need to use [norm_sqr] instead of [norm] to avoid nan gradients in the backward pass. Even when theta = 0,
+    # the gradient of [norm] operation is computed and used (it is multiplied by 0 but still results in nan gradients).
```
duburcqa: Could you elaborate? Do you have a reproduction script? This is weird...

tests/test_grad.py
```diff
@@ -223,3 +223,93 @@ def test_differentiable_push(precision, show_viewer):
     for v_i in v_list[:-1]:
         assert (v_i.grad.abs() > gs.EPS).any()
     assert (v_list[-1].grad.abs() < gs.EPS).all()
+
+
+@pytest.mark.required
+@pytest.mark.precision("32")
+@pytest.mark.parametrize("backend", [gs.cpu])
+def test_differentiable_rigid(show_viewer):
+
```
duburcqa: Do not skip space here.

tests/test_grad.py
```diff
@@ -223,3 +223,93 @@ def test_differentiable_push(precision, show_viewer):
     for v_i in v_list[:-1]:
         assert (v_i.grad.abs() > gs.EPS).any()
     assert (v_list[-1].grad.abs() < gs.EPS).all()
+
+
+@pytest.mark.required
+@pytest.mark.precision("32")
+@pytest.mark.parametrize("backend", [gs.cpu])
```
duburcqa: Why are you enforcing both precision and backend? I guess it should work just fine without enforcing any of them? It is strictly harder to pass on CPU (because it would be implicitly checking more things, related to some optional features being automatically disabled or something)?

tests/test_grad.py
```diff
@@ -223,3 +223,93 @@ def test_differentiable_push(precision, show_viewer):
...
+            size=(0.1, 0.1, 0.2),
+        ),
+        surface=gs.surfaces.Default(
+            color=(0.9, 0.0, 0.0, 1.0),
+        ),
+    )
+    if show_viewer:
+        target = scene.add_entity(
+            gs.morphs.Box(
+                pos=goal_pos.cpu().tolist(),
```
duburcqa: It should be possible to pass `goal_pos` directly (same for `goal_quat`). If not, it is rather `gs.morphs.Box` that must be fixed to avoid forcing casting to list.

genesis/engine/solvers/rigid/constraint_noslip.py
duburcqa, line 1: Is noslip feature supported when enabling gradient computation? If not, you should raise an exception at init.

genesis/engine/solvers/rigid/constraint_noslip.py
duburcqa: After checking, it seems that another variable is specified when `is_backward=True`, so here if I understand correctly you are setting it to anything because it will not be used in practice. This does work in practice but I don't like it much... What about defining some extra free variable `PLACEHOLDER` in `array_class` (0D taichi tensor of type `array_class.V`) that we could use everywhere an argument is not used? This would clarify the intend and avoid any mistake because you cannot do much with such tensor.

genesis/utils/array_class.py
```diff
@@ -2133,6 +2171,37 @@ def __init__(self):
         return ClassEntitiesState()
 
 
+# =========================================== RigidAdjointCache ===========================================
+@dataclasses.dataclass
+class StructRigidAdjointCache:
```
duburcqa: Could you document in which context this is used and what this quantities are storing?

I know it was not documented for the other data structure, but it was a bad decision. Never to late to raise the bar of our standards.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -10,6 +10,7 @@
 import genesis.utils.geom as gu
 from genesis.engine.entities import AvatarEntity, DroneEntity, RigidEntity
 from genesis.engine.entities.base_entity import Entity
+from genesis.engine.states.cache import QueriedStates
 from genesis.engine.states.solvers import RigidSolverState
```
duburcqa: What about exposing all `*State(s)` class on the top-level of `states` submodule? That would be more convenient than having to go through the hierarchy for each State to import. To do this, just add this kind of lines to `genesis/engine/states/__init__.py`:
```python
from .solvers import *
```

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -131,6 +132,13 @@ def __init__(self, scene: "Scene", sim: "Simulator", options: RigidOptions) -> N
 
         self.qpos: ti.Template | ti.types.NDArray | None = None
 
+        self._queried_states = QueriedStates()
+
+        self.init_ckpt()
+
+    def init_ckpt(self):
+        self._ckpt = dict()
+
```
duburcqa: This is a bad pattern. All the attributes of a class should be unconditionally defined (set to `None` by default for optional attributes is acceptable), in the `__init__` function ONLY. We are violating these basic rules everywhere unfortunately.

If `init_ckpt` may be called outside init, it should rather be called something like `clear_*` or `reset_*`. If it is only called at init, then it should be part of `__init__`.

genesis/utils/geom.py
SonSang: Yes, here is the reproduction code.
```
import gstaichi as ti
import genesis as gs
import numpy as np

gs.init(precision="32", debug=True, backend=gs.cpu)

a = ti.field(dtype=gs.ti_vec3, shape=(), needs_grad=True)
b = ti.field(dtype=gs.ti_float, shape=(), needs_grad=True)

@ti.kernel
def foo(use_norm_sqr: ti.template()) -> None:
    if ti.static(use_norm_sqr):
        a_norm_sqr = a[None].norm_sqr()
        if a_norm_sqr < gs.EPS ** 2:
            b[None] = 0.0
        else:
            b[None] = ti.sqrt(a_norm_sqr)
    else:
        a_norm = a[None].norm()
        # When a_norm is 0, it does not affect [b] and thus the gradient of [a] should be 0. However, it turns out that
        # the gradient of [a] becomes NaN when a_norm is 0.
        if a_norm < gs.EPS:
            b[None] = 0.0
        else:
            b[None] = a_norm

for use_norm_sqr in [True, False]:
    foo(use_norm_sqr)
    b.grad[None] = 1.0
    foo.grad(use_norm_sqr)
    
    a_grad_isnan = np.isnan(a.grad.to_numpy()).any()
    if a_grad_isnan and not use_norm_sqr:
        print("a.grad is nan when use_norm_sqr is False, maybe because the gradient of the norm operation is NaN when the norm is 0.")
        print("Even though a_norm does not affect [b] when a_norm is 0, the gradient of the norm operation is seemingly still computed and used.")
```

As written here, I assume the gradient becomes `nan` because of the `norm` operation. Even when the result of the `norm` operation does not affect the final value, I assume the gradient is computed and reflected in the final gradient (maybe they simply multiply 0 to the `nan` gradient, which is still `nan`).

genesis/engine/states/entities.py
SonSang: I followed the naming convention of the other entity states (e.g. `FEMEntityState`, [here](https://github.com/Genesis-Embodied-AI/Genesis/blob/87a8e0b03ca2794c0d8725f639ede1a0bc0e3a91/genesis/engine/states/entities.py#L148)). Would it be better to rename the variable names for these other states as well?

Satvik1701:
Hey! I was just wondering if this PR is close to being merged soon, since the Heterogeneous Simulation PR is dependent on this. Thanks! PR: https://github.com/Genesis-Embodied-AI/Genesis/pull/1589

SonSang:
> Hey! I was just wondering if this PR is close to being merged soon, since the Heterogeneous Simulation PR is dependent on this. Thanks! PR: #1589

Hey, sorry for being late. This PR needs some polishing to pass some benchmark tests, and I'm working on it. Sorry for being late, I'll try to wrap up as soon as possible.

Satvik1701:
> > Hey! I was just wondering if this PR is close to being merged soon, since the Heterogeneous Simulation PR is dependent on this. Thanks! PR: #1589
> 
> Hey, sorry for being late. This PR needs some polishing to pass some benchmark tests, and I'm working on it. Sorry for being late, I'll try to wrap up as soon as possible.

Thanks so much! Yeah sorry for being pushy but do you have a rough timeline for this because I'm looking to train with the heterogeneous simulation and might reprioritize somethings.

SonSang:
> > > Hey! I was just wondering if this PR is close to being merged soon, since the Heterogeneous Simulation PR is dependent on this. Thanks! PR: #1589
> > 
> > 
> > Hey, sorry for being late. This PR needs some polishing to pass some benchmark tests, and I'm working on it. Sorry for being late, I'll try to wrap up as soon as possible.
> 
> Thanks so much! Yeah sorry for being pushy but do you have a rough timeline for this because I'm looking to train with the heterogeneous simulation and might reprioritize somethings.

It's hard to say, but I think at least a week is needed (because code review is needed again for merging). I'll let you know if I have a better estimate.

SamratSahoo:
following up on this - any updates on the eta? also looking to use heterogeneous simulations (#1589) which depends on this

SonSang:
> following up on this - any updates on the eta? also looking to use heterogeneous simulations (#1589) which depends on this

Hi, this PR will be merged tomorrow if code review goes well. Based on the review, it could be delayed a little, but I'll try to finish up until this weekend.

genesis/engine/states/entities.py
duburcqa: Ok let's keep it for now.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1989,6 +2131,18 @@ def set_quat(self, quat, envs_idx=None, *, relative=False, zero_velocity=True, u
         envs_idx : None | array_like, optional
             The indices of the environments. If None, all environments will be considered. Defaults to None.
         """
+        # Save in [tgt] for backward pass
+        if update_tgt:
```
duburcqa: `update_tgt` seems an internal feature. I would recommend calling `self.update_tgt` manually right before `set_quat` to avoid leaking this option to the user.

duburcqa: After checking twice, keep `self.update_tgt` call inside this function but rather use a global state variable `self._update_tgt` to track whether target must be updated or not, without exposing this option?

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -2005,6 +2159,18 @@ def set_quat(self, quat, envs_idx=None, *, relative=False, zero_velocity=True, u
         if zero_velocity:
             self.zero_all_dofs_velocity(envs_idx, unsafe=unsafe)
 
+    @gs.assert_built
+    def set_quat_grad(self, envs_idx, relative, unsafe, quat_grad):
+        tmp_quat_grad = quat_grad.unsqueeze(-2)
```
duburcqa: I don't get it. Why don't you pass `quat_grad.data` directly?

genesis/engine/solvers/rigid/constraint_solver_decomp_island.py
```diff
@@ -990,7 +990,7 @@ def _func_update_gradient(self, island, i_b):
                 i_e_ = self.contact_island.island_entity[island, i_b].start + i_island_entity
                 i_e = self.contact_island.entity_id[i_e_, i_b]
                 self._solver.mass_mat_mask[i_e_, i_b] = True
-            self._solver._func_solve_mass_batched(self.grad, self.Mgrad, i_b)
+            self._solver._func_solve_mass_batched(self.grad, self.Mgrad, None, i_b)
```
duburcqa: I don't think passing none is supported.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -250,10 +264,60 @@ def build(self):
             enable_joint_limit=getattr(self, "_enable_joint_limit", False),
             box_box_detection=getattr(self, "_box_box_detection", True),
             sparse_solve=getattr(self._options, "sparse_solve", False),
-            integrator=getattr(self, "_integrator", gs.integrator.implicitfast),
+            integrator=getattr(self, "_integrator", gs.integrator.approximate_implicitfast),
```
duburcqa: Weird to include this "bug fix" in this PR but OK.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -250,10 +264,60 @@ def build(self):
...
+                gs.raise_exception(
+                    "Only approximate_implicitfast integrator is supported yet when requires_grad is True."
+                )
+            from genesis.engine.couplers import SAPCoupler, IPCCoupler
+
+            if isinstance(self.sim.coupler, SAPCoupler):
+                gs.raise_exception("SAPCoupler is not supported yet when requires_grad is True.")
+
+            if isinstance(self.sim.coupler, IPCCoupler):
+                gs.raise_exception("IPCCoupler is not supported yet when requires_grad is True.")
```
duburcqa:
```python
if isinstance(self.sim.coupler, (SAPCoupler, IPCCoupler)):
    gs.raise_exception(f"{type(self.sim.coupler).__name__} is not supported yet when requires_grad is True.")

File: genesis/engine/solvers/rigid/rigid_solver_decomp.py [hunk 33]
```diff
@@ -250,10 +264,60 @@ def build(self):
...
+        )
+        self._static_rigid_sim_config.max_n_awake_links = (
+            getattr(self, "_n_links", 0) if self._static_rigid_sim_config.requires_grad else 0
+        )
+        self._static_rigid_sim_config.max_n_awake_entities = (
+            getattr(self, "_n_entities", 0) if self._static_rigid_sim_config.requires_grad else 0
+        )
+        self._static_rigid_sim_config.max_n_awake_dofs = (
+            getattr(self, "_n_dofs", 0) if self._static_rigid_sim_config.requires_grad else 0
+        )
```
duburcqa, line 328: Stop using `getattr`, it is a bad practice.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -291,7 +361,7 @@ def build(self):
             self._init_constraint_solver()
 
             self._init_invweight_and_meaninertia(force_update=False)
-            self._func_update_geoms(self._scene._envs_idx, force_update_fixed_geoms=True)
+            self._func_update_geoms(self._scene._envs_idx, force_update_fixed_geoms=True, is_backward=False)
```
duburcqa: We should just make `is_backward=False` the default value. It would be easier.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -1124,10 +1220,156 @@ def substep_pre_coupling(self, f):
...
 
     def substep_pre_coupling_grad(self, f):
-        pass
+        # Load the current state from adjoint cache
+        curr_qpos = self._rigid_adjoint_cache.qpos.to_numpy()[f]
+        curr_dofs_vel = self._rigid_adjoint_cache.dofs_vel.to_numpy()[f]
+        curr_dofs_acc = self._rigid_adjoint_cache.dofs_acc.to_numpy()[f]
+        self._rigid_global_info.qpos.from_numpy(curr_qpos)
+        self.dofs_state.vel.from_numpy(curr_dofs_vel)
+        self.dofs_state.acc.from_numpy(curr_dofs_acc)
```
duburcqa: This is too inefficient. We need to find a better way.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -1124,10 +1220,156 @@ def substep_pre_coupling(self, f):
...
+                rigid_global_info=self._rigid_global_info,
+                static_rigid_sim_config=self._static_rigid_sim_config,
+                force_update_fixed_geoms=False,
+                is_backward=True,
+            )
+
+        qpos_grad = self._rigid_global_info.qpos.grad.to_numpy()
+        dofs_vel_grad = self.dofs_state.vel.grad.to_numpy()
+        if np.isnan(qpos_grad).sum() > 0 or np.isnan(dofs_vel_grad).sum() > 0:
+            gs.raise_exception(f"Nan grad in qpos or dofs_vel found at step {self._sim.cur_step_global}")
```
duburcqa: Same. very inefficient. We cannot afford it I think.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -1124,10 +1220,156 @@ def substep_pre_coupling(self, f):
...
+
+        kernel_copy_next_to_curr.grad(
+            dofs_state=self.dofs_state,
+            rigid_global_info=self._rigid_global_info,
+            static_rigid_sim_config=self._static_rigid_sim_config,
+        )
```
duburcqa: All these small kernels are dominated by the overhead. You should rather make all of this a gigantic kernel.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -1156,33 +1399,76 @@ def substep_post_coupling(self, f):
...
+                qpos_grad = state.qpos.grad
+
+            if state.dofs_vel.grad is not None:
+                dofs_vel_grad = state.dofs_vel.grad
+
+            if state.links_pos.grad is not None:
+                links_pos_grad = state.links_pos.grad
+
+            if state.links_quat.grad is not None:
+                links_quat_grad = state.links_quat.grad
```
duburcqa, line 1395: Avoid skipping all these lines, you are diluting information.

genesis/utils/geom.py
duburcqa: Oh damn! Good find! Tracking this issue on compiler side.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -1280,16 +1574,51 @@ def set_state(self, f, state, envs_idx=None):
...
     def save_ckpt(self, ckpt_name):
-        pass
+        # Save ckpt only if we need gradients, because this operation is costly
+        if self._requires_grad:
+            if ckpt_name not in self._ckpt:
+                self._ckpt[ckpt_name] = dict()
+
+            self._ckpt[ckpt_name]["qpos"] = self._rigid_adjoint_cache.qpos.to_numpy()
+            self._ckpt[ckpt_name]["dofs_vel"] = self._rigid_adjoint_cache.dofs_vel.to_numpy()
+            self._ckpt[ckpt_name]["dofs_acc"] = self._rigid_adjoint_cache.dofs_acc.to_numpy()
```
duburcqa: If this is part of the hot path, avoid using `to_numpy` and prefer `ti_to_numpy`

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -1468,6 +1806,15 @@ def _sanitize_2D_io_variables(
...
 
+    def _sanitize_2D_io_variables_grad(
+        self,
+        grad_after_sanitization,
+        grad_before_sanitization,
+    ):
+        if grad_after_sanitization.shape != grad_before_sanitization.shape:
+            gs.raise_exception("Shape of grad_after_sanitization and grad_before_sanitization do not match.")
+        return grad_after_sanitization
+
```
duburcqa: This does not "sanitize" anything. It "validates".

duburcqa: Besides, these are static functions. So I would just move them outside this class (right before it, as free function).

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -1844,8 +2257,22 @@ def set_dofs_velocity(self, velocity, dofs_idx=None, envs_idx=None, *, skip_forw
...
+            velocity_grad, dofs_idx, self.n_dofs, envs_idx, skip_allocation=True, unsafe=unsafe
+        )
+        if self.n_envs == 0:
+            velocity_grad_ = velocity_grad_.unsqueeze(0)
+        kernel_set_dofs_velocity_grad(
+            velocity_grad_, dofs_idx, envs_idx, self.dofs_state, self._static_rigid_sim_config
+        )
+        if self.n_envs == 0:
+            velocity_grad_ = velocity_grad_.squeeze(0)
+        velocity_grad.data = self._sanitize_1D_io_variables_grad(velocity_grad_, velocity_grad)
```
duburcqa: Why don't you just pass `velocity_grad.data` directly to the kernel?

```python
        velocity_grad.data, dofs_idx, envs_idx = self._sanitize_1D_io_variables(
            velocity_grad, dofs_idx, self.n_dofs, envs_idx, skip_allocation=True, unsafe=unsafe
        )
        velocity_grad_ = velocity_grad.data
        if self.n_envs == 0:
             velocity_grad_ = velocity_grad_.unsqueeze(0)
        kernel_set_dofs_velocity_grad(
            velocity_grad_, dofs_idx, envs_idx, self.dofs_state, self._static_rigid_sim_config
        )
```

duburcqa: Why is this validation really needed?

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -6600,6 +7904,20 @@ def kernel_set_dofs_velocity(
...
+    velocity_grad: ti.types.ndarray(),
+    dofs_idx: ti.types.ndarray(),
+    envs_idx: ti.types.ndarray(),
+    dofs_state: array_class.DofsState,
+    static_rigid_sim_config: ti.template(),
+):
+    ti.loop_config(serialize=static_rigid_sim_config.para_level < gs.PARA_LEVEL.PARTIAL)
+    for i_d_, i_b_ in ti.ndrange(dofs_idx.shape[0], envs_idx.shape[0]):
+        velocity_grad[i_b_, i_d_] = dofs_state.vel.grad[dofs_idx[i_d_], envs_idx[i_b_]]
+        dofs_state.vel.grad[dofs_idx[i_d_], envs_idx[i_b_]] = 0.0
```
duburcqa, line 7638: Why do you need a specific kernel for this? You cannot use autodiff directly?

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -6312,6 +7587,35 @@ def kernel_set_links_quat(
...
+
+        if links_info.parent_idx[I_l] == -1 and links_info.is_fixed[I_l]:
+            for j in ti.static(range(4)):
+                quat_grad[i_b_, i_l_, j] = links_state.quat.grad[i_l, i_b][j]
+                links_state.quat.grad[i_l, i_b][j] = 0.0
+        else:
+            q_start = links_info.q_start[I_l]
+            for j in ti.static(range(4)):
+                quat_grad[i_b_, i_l_, j] = rigid_global_info.qpos.grad[q_start + j + 3, i_b]
+                rigid_global_info.qpos.grad[q_start + j + 3, i_b] = 0.0
```
duburcqa, line 7336: Why do you need a specific kernel for this? You cannot use autodiff directly?

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -6312,6 +7587,35 @@ def kernel_set_links_quat(
...
+    relative: ti.i32,
+    quat_grad: ti.types.ndarray(),
+    links_idx: ti.types.ndarray(),
+    envs_idx: ti.types.ndarray(),
+    links_info: array_class.LinksInfo,
+    links_state: array_class.LinksState,
+    rigid_global_info: array_class.RigidGlobalInfo,
+    static_rigid_sim_config: ti.template(),
+):
+
```
duburcqa: No need to skip line here.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -6206,6 +7418,7 @@ def kernel_set_state(
     ti.loop_config(serialize=static_rigid_sim_config.para_level < gs.PARA_LEVEL.ALL)
     for i_d, i_b_ in ti.ndrange(n_dofs, envs_idx.shape[0]):
         dofs_state.vel[i_d, envs_idx[i_b_]] = dofs_vel[envs_idx[i_b_], i_d]
+        dofs_state.acc[i_d, envs_idx[i_b_]] = dofs_acc[envs_idx[i_b_], i_d]
         dofs_state.ctrl_force[i_d, envs_idx[i_b_]] = gs.ti_float(0.0)
         dofs_state.ctrl_mode[i_d, envs_idx[i_b_]] = gs.CTRL_MODE.FORCE
```
duburcqa, line 7146: I think we should move adding acc to rigid state in another PR.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -5889,118 +7058,158 @@ def func_integrate(
...
+        # xanchor, xaxis
+        dst_joints_state.xanchor[i_j, i_b] = src_joints_state.xanchor[i_j, i_b]
+        dst_joints_state.xaxis[i_j, i_b] = src_joints_state.xaxis[i_j, i_b]
+
+    # geoms state
+    for i_g, i_b in ti.ndrange(src_geoms_state.pos.shape[0], src_geoms_state.pos.shape[1]):
+        # pos, quat, verts_updated
+        dst_geoms_state.pos[i_g, i_b] = src_geoms_state.pos[i_g, i_b]
+        dst_geoms_state.quat[i_g, i_b] = src_geoms_state.quat[i_g, i_b]
+        dst_geoms_state.verts_updated[i_g, i_b] = src_geoms_state.verts_updated[i_g, i_b]
```
duburcqa: Ideally we should avoid this kind of tiny kernels doing nothing, because overhead will dominates by a very large margin. I will try to see if we can leverage zero-copy for this. By ideally, it should be moved as part of a larger kernel.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -5889,118 +7058,158 @@ def func_integrate(
...
+    _B = dofs_state.vel.shape[1]
+
+    ti.loop_config(serialize=static_rigid_sim_config.para_level < gs.PARA_LEVEL.ALL)
+    for i_d, i_b in ti.ndrange(n_dofs, _B):
+        rigid_adjoint_cache.dofs_vel[f, i_d, i_b] = dofs_state.vel[i_d, i_b]
+        rigid_adjoint_cache.dofs_acc[f, i_d, i_b] = dofs_state.acc[i_d, i_b]
+
+    ti.loop_config(serialize=static_rigid_sim_config.para_level < gs.PARA_LEVEL.ALL)
+    for i_q, i_b in ti.ndrange(n_qs, _B):
+        rigid_adjoint_cache.qpos[f, i_q, i_b] = rigid_global_info.qpos[i_q, i_b]
```
duburcqa, line 6704: Same. This kernel is too small. Ideally it should be a ti.func that is part of a bigger kernel.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -5889,118 +7058,158 @@ def func_integrate(
...
+def kernel_copy_next_to_curr(
+    dofs_state: array_class.DofsState,
+    rigid_global_info: array_class.RigidGlobalInfo,
+    static_rigid_sim_config: ti.template(),
+):
+    func_copy_next_to_curr(
+        dofs_state=dofs_state,
+        rigid_global_info=rigid_global_info,
+        static_rigid_sim_config=static_rigid_sim_config,
+    )
```
duburcqa: Same. This kernel is too small. Ideally it should be a ti.func that is part of a bigger kernel.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -5647,68 +6745,63 @@ def func_update_force(
...
+                        if i_p != -1:
+                            links_state.cfrc_vel[i_p, i_b] += links_state.cfrc_vel[i_l, i_b]
+                            links_state.cfrc_ang[i_p, i_b] += links_state.cfrc_ang[i_l, i_b]
 
     # Clear coupling forces after use
     ti.loop_config(serialize=static_rigid_sim_config.para_level < gs.PARA_LEVEL.ALL)
-    for i_l, i_b in ti.ndrange(n_links, _B):
+    for i_l, i_b in ti.ndrange(links_info.root_idx.shape[0], links_state.pos.shape[1]):
         links_state.cfrc_coupling_ang[i_l, i_b] = ti.Vector.zero(gs.ti_float, 3)
         links_state.cfrc_coupling_vel[i_l, i_b] = ti.Vector.zero(gs.ti_float, 3)
```
duburcqa:
```python
    for I in ti.grouped(ti.ndrange(*links_state.cfrc_coupling_ang.shape)):
        links_state.cfrc_coupling_ang[I] = ti.Vector.zero(gs.ti_float, 3)
        links_state.cfrc_coupling_vel[I] = ti.Vector.zero(gs.ti_float, 3)
```

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -4313,32 +5269,95 @@ def func_COM_links(
...
+        ):
+            i_l = (
+                rigid_global_info.awake_links[i_l_, i_b] if ti.static(static_rigid_sim_config.use_hibernation) else i_l_
+            )
             I_l = [i_l, i_b] if ti.static(static_rigid_sim_config.batch_links_info) else i_l
 
             i_r = links_info.root_idx[I_l]
-            if i_l == i_r:
-                links_state.root_COM[i_l, i_b] = links_state.root_COM[i_l, i_b] / links_state.mass_sum[i_l, i_b]
+            if i_l == i_r and links_state.mass_sum[i_l, i_b] > 0.0:
```
duburcqa: This condition is already enforced by design. Why are you adding it? Did you observed any issue in practice?

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -2482,12 +2928,36 @@ def n_qs(self):
             return self._n_qs
         return sum([entity.n_qs for entity in self._entities])
 
+    @property
+    def max_n_qs_per_link(self):
+        if self.is_built:
+            return self._max_n_qs_per_link
+        return max([link.n_qs for link in self.links]) if len(self.links) > 0 else 0
```
duburcqa: `max(link.n_qs for link in self.links) if self.links else 0`

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -2482,12 +2928,36 @@ def n_qs(self):
...
     def n_dofs(self):
         if self.is_built:
             return self._n_dofs
         return sum(entity.n_dofs for entity in self._entities)
 
+    @property
+    def max_n_dofs_per_entity(self):
+        if self.is_built:
+            return self._max_n_dofs_per_entity
+        return max([entity.n_dofs for entity in self._entities]) if len(self._entities) > 0 else 0
```
duburcqa: `max(entity.n_dofs for entity in self._entities) if self._entities else 0`

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -2482,12 +2928,36 @@ def n_qs(self):
...
+    def max_n_dofs_per_link(self):
+        if self.is_built:
+            return self._max_n_dofs_per_link
+        return max([link.n_dofs for link in self.links]) if len(self.links) > 0 else 0
+
+    @property
+    def max_n_dofs_per_joint(self):
+        if self.is_built:
+            return self._max_n_dofs_per_joint
+        return max([joint.n_dofs for joint in self.joints]) if len(self.joints) > 0 else 0
```
duburcqa: Same.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -2410,18 +2838,36 @@ def n_links(self):
...
     def n_geoms(self):
         if self.is_built:
             return self._n_geoms
         return len(self.geoms)
 
+    @property
+    def max_n_geoms_per_entity(self):
+        if self.is_built:
+            return self._max_n_geoms_per_entity
+        return max([entity.n_geoms for entity in self._entities]) if len(self._entities) > 0 else 0
```
duburcqa: `max(len(link.joints) for link in self.links) if self.links else 0`

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -2410,18 +2838,36 @@ def n_links(self):
             return self._n_links
         return len(self.links)
 
+    @property
+    def max_n_links_per_entity(self):
+        if self.is_built:
+            return self._max_n_links_per_entity
+        return max([len(entity.links) for entity in self._entities]) if len(self._entities) > 0 else 0
```
duburcqa: `max(len(entity.links) for entity in self._entities) if self._entities else 0`

genesis/engine/solvers/rigid/rigid_solver_decomp.py
duburcqa: It seems more complex than necessary. Why cannot you assign `velocity_grad.data` directly and why do you need to validate the shape?

genesis/engine/solvers/rigid/rigid_solver_decomp.py
SonSang: I don't think we can use autodiff here. We need to copy the gradients in the taichi field to the tensor. Other part of our code is also doing the same thing ([here](https://github.com/Genesis-Embodied-AI/Genesis/blob/ea4c7ddb204fea910f98bb3e591e8b52e602e11d/genesis/engine/entities/mpm_entity.py#L352) calls [here](https://github.com/Genesis-Embodied-AI/Genesis/blob/ea4c7ddb204fea910f98bb3e591e8b52e602e11d/genesis/engine/solvers/mpm_solver.py#L828)).

genesis/engine/solvers/rigid/rigid_solver_decomp.py
SonSang: I don't think we can use autodiff here. We need to copy the gradients in the taichi field to the tensor. Other part of our code is also doing the same thing ([here](https://github.com/Genesis-Embodied-AI/Genesis/blob/ea4c7ddb204fea910f98bb3e591e8b52e602e11d/genesis/engine/entities/mpm_entity.py#L352) calls [here](https://github.com/Genesis-Embodied-AI/Genesis/blob/ea4c7ddb204fea910f98bb3e591e8b52e602e11d/genesis/engine/solvers/mpm_solver.py#L828)).

genesis/engine/solvers/rigid/rigid_solver_decomp.py
SonSang: The new unit test in this PR fails if we don't include it here. Can you reconsider it?

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1627,6 +1645,106 @@ def plan_path(
...
+            if "zero_velocity" in data_kwargs:
+                data_kwargs["zero_velocity"] = False
+            # Do not update [tgt], as input information is finalized at this point
+            self._update_tgt_while_set = False
+
+            match key:
+                case "pos":
+                    self.set_pos(**data_kwargs)
+                case "quat":
+                    self.set_quat(**data_kwargs)
```
duburcqa: "pos" and "quat" are redundant with "qpos" and should be removed.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1627,6 +1645,106 @@ def plan_path(
...
+                    quat = data_kwargs.pop("quat")
+                    if quat.requires_grad:
+                        quat._backward_from_ti(
+                            self.set_quat_grad, data_kwargs["envs_idx"], data_kwargs["relative"], data_kwargs["unsafe"]
+                        )
+
+                case "qpos":
+                    qpos = data_kwargs.pop("qpos")
+                    if qpos.requires_grad:
+                        raise NotImplementedError("Backward pass for set_qpos_grad is not implemented yet.")
```
duburcqa: Why is this case not supported? It is arguably more important than `set_pos` and `set_quat`.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1627,6 +1645,106 @@ def plan_path(
...
+        state = RigidEntityState(self, self._sim.cur_step_global)
+
+        solver_state = self._solver.get_state()
+        pos = solver_state.links_pos[:, self.base_link_idx]
+        quat = solver_state.links_quat[:, self.base_link_idx]
+
+        assert state._pos.shape == pos.shape
+        assert state._quat.shape == quat.shape
+        state._pos = pos
+        state._quat = quat
```
duburcqa, line 1720: You don't need any check. This is our own private internal machinery. It cannot fail.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -97,6 +98,23 @@ def __init__(
 
         self._load_model()
 
+        # Initialize target variables and checkpoint
+        self._tgt_keys = ["pos", "quat", "qpos", "dofs_velocity"]
```
duburcqa: Use tuple.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1627,6 +1645,106 @@ def plan_path(
...
+            # Do not update [tgt], as input information is finalized at this point
+            self._update_tgt_while_set = False
+
+            match key:
+                case "pos":
+                    self.set_pos(**data_kwargs)
+                case "quat":
+                    self.set_quat(**data_kwargs)
+                case "qpos":
+                    self.set_qpos(**data_kwargs)
```
duburcqa: Remove `qpos` if it is not supported.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -2015,6 +2133,18 @@ def set_pos(self, pos, envs_idx=None, *, relative=False, unsafe=False):
...
+        if self._update_tgt_while_set:
+            self._update_tgt(
+                "pos",
+                {
+                    "pos": pos,
+                    "envs_idx": envs_idx,
+                    "relative": relative,
+                    "unsafe": unsafe,
+                },
+            )
```
duburcqa: We should add a wrapper doing this job automatically. This is error-prone and clustering the code.

duburcqa:
```python
from functools import wraps

def tracked(fun):
    sig = inspect.signature(func)

    @wraps(fun)
    def wrapper(self, *args, **kwargs):
        bound = sig.bind(self, *args, **kwargs)
        bound.apply_defaults()
        args_dict = dict(tuple(bound.arguments.items())[1:])
        self._update_tgt(fun.__name__, args_dict)
        return func(*args, **kwargs)

    return wrapper

[...]

@gs.assert_built
@tracked
def set_pos(self, pos, envs_idx=None, *, relative=False, unsafe=False):
    [...]
```

genesis/engine/simulator.py
```diff
@@ -266,9 +266,9 @@ def f_global_to_s_global(self, f_global):
     # ------------------------------------------------------------------------------------
 
     def step(self, in_backward=False):
-        if self._rigid_only:  # "Only Advance!" --Thomas Wade :P
+        if self._rigid_only and (not self._requires_grad):  # "Only Advance!" --Thomas Wade :P
```
duburcqa: Remove parenthesis.

tests/test_rigid_physics.py
```diff
@@ -1293,9 +1293,11 @@ def test_set_root_pose(batch_fixed_verts, relative, show_viewer, tol):
         ):
             pos_zero = torch.tensor(pos_zero, device=gs.device, dtype=gs.tc_float)
             euler_zero = torch.deg2rad(torch.tensor(euler_zero, dtype=gs.tc_float))
+            quat_zero = gu.xyz_to_quat(euler_zero, rpy=True)
             assert_allclose(entity.get_pos(), pos_zero, tol=tol)
-            euler = gu.quat_to_xyz(entity.get_quat(), rpy=True)
-            assert_allclose(euler, euler_zero, tol=5e-4)
+            # Use quaternion for comparison to avoid gymbal lock issue in euler angles
+            quat = entity.get_quat()
+            assert_allclose(quat, quat_zero, tol=tol)
```
duburcqa, line 1300: Quaternion are not uniquely defined. The best approach would be:
```python
assert np.linalg.norm(gu.quat_to_rotvec(gu.transform_quat_by_quat(gu.inv_quat(quat), quat_zero))) < tol
```
or even better:
```python
assert_allclose(gu.quat_to_rotvec(gu.transform_quat_by_quat(gu.inv_quat(quat), quat_zero)), 0.0, tol=tol)
```

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -2447,12 +2865,36 @@ def n_qs(self):
...
+        if self.is_built:
+            return self._max_n_dofs_per_link
+        return max(link.n_dofs for link in self.links) if self.links else 0
+
+    @property
+    def max_n_dofs_per_joint(self):
+        if self.is_built:
+            return self._max_n_dofs_per_joint
+        return max(joint.n_dofs for joint in self.joints) if self.joints else 0
+
```
duburcqa: Do not expose additional function in the API if they are only used once at init, and I meaningless for users.

genesis/engine/entities/rigid_entity/rigid_entity.py
SonSang: There are many inputs to rigid entity, and in this PR I did not implement backward pass for all of them because this PR is more about backward pass of rigid "solver". I will make another PR for backward pass of the other inputs later.

PR #1810, opened 2025-10-08
[FEATURE] Avoid new COACD run when scale changes

SonSang:
## Description

When the mesh scale changes and the mesh requires convex decomposition, COACD runs again because it uses the raw vertex positions as key for cached directory.

```
cvx_path = get_cvx_path(mesh.vertices, mesh.faces, coacd_options)
```

To avoid this re-run, we first normalize the mesh by using bbox diagonal of the mesh as the mesh scale. Then, quantize the mesh vertices to get same hash key for slightly different floating point values. If there is a cache for the normalized mesh, re-use the result. Otherwise, run COACD again.

## Related Issue

## Motivation and Context

Avoid COACD re-run when the input mesh scale changes.

## Checklist:

### DIFF
--- a/genesis/utils/mesh.py
+++ b/genesis/utils/mesh.py
@@ -259,6 +259,7 @@ def surface_uvs_to_trimesh_visual(surface, uvs=None, n_verts=None):
 
     return visual
 
+
 def get_mesh_scale(verts):
     # compute mesh scale using bbox diagonal
     v = np.asarray(verts, dtype=np.float64, order="C")
@@ -266,20 +267,21 @@ def get_mesh_scale(verts):
     bmax = v.max(axis=0)
     ext = bmax - bmin
     scale = float(np.linalg.norm(ext))
-    
+
     return scale
+
 
 def convex_decompose(mesh, coacd_options):
     # compute mesh scale
     mesh_scale = get_mesh_scale(mesh.vertices)
-    
+
     # rescale mesh vertices to remove scale factor, and quantize to int
     if not (np.isinf(mesh_scale) or np.isnan(mesh_scale) or mesh_scale <= 0.0):
         _vertices = mesh.vertices / mesh_scale
         _vertices = np.round(_vertices / CVX_PATH_QUANTIZE_FACTOR).astype(np.int64, order="C")
     else:
         _vertices = mesh.vertices
-    
+
     # compute file name via hashing for caching
     cvx_path = get_cvx_path(_vertices, mesh.faces, coacd_options)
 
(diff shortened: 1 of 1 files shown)


YilingQiao:
Thank you very much! Could you add a test for it? Or some instructions for testing it?

duburcqa:
@YilingQiao Apparently it is already tested to some extent since the CI is now failing x)

genesis/utils/mesh.py
```diff
@@ -300,6 +333,7 @@ def convex_decompose(mesh, coacd_options):
             mesh_parts = []
             for vs, fs in result:
                 mesh_parts.append(trimesh.Trimesh(vs, fs))
+            mesh_parts.append(mesh_scale)
```
duburcqa: I don't know what you are trying to do here, but this is clearly invalid code.

duburcqa:
But yes, ideally a unit test should be added:
* monkeypatch `get_cvx_path` to add a hook that allows you to keep track of which path is being returned
* Create a scene with a sphere geometry (or ideally any non-convex mesh requiring decompoisition) with scale 2.0, then build it
* Create a new scene with a sphere with geometry with scale 3.0 then built it
* Make sure that the path returned by `get_cvx_path` was the same for the second run
* Check the vertices and make sure that the one is the upscaled version of the other

tests/test_mesh.py
```diff
@@ -510,3 +511,79 @@ def test_splashsurf_surface_reconstruction(show_viewer):
...
+    # check if the scaled parts match the scaled version of the original parts
+    cached_parts = seen_results[0]
+    scaled_parts = seen_results[-1]
+    assert len(scaled_parts) == len(cached_parts), "Expected %d scaled parts, got %d" % (
+        len(cached_parts),
+        len(scaled_parts),
+    )
+    for scaled_part, cached_part in zip(scaled_parts, cached_parts):
+        assert_allclose(scaled_part.vertices, cached_part.vertices * (second_scale / first_scale), rtol=1e-6)
+        assert_allclose(scaled_part.faces, cached_part.faces, atol=0)
```
duburcqa: You have `assert_array_equal` for this but no big deal.

tests/test_mesh.py
```diff
@@ -510,3 +511,79 @@ def test_splashsurf_surface_reconstruction(show_viewer):
...
+    scaled_path = seen_paths[-1]
+    assert cached_path == scaled_path, "Cached path and scaled path should be the same"
+
+    # check if the scaled parts match the scaled version of the original parts
+    cached_parts = seen_results[0]
+    scaled_parts = seen_results[-1]
+    assert len(scaled_parts) == len(cached_parts), "Expected %d scaled parts, got %d" % (
+        len(cached_parts),
+        len(scaled_parts),
+    )
```
duburcqa: I don't think it is necessary to add this assert comments. It is self-explanatory here, and right and left side values of equality checks while be displayed automatically by pytest.

duburcqa: Here is an example:
```python
    def test_dummy(show_viewer):
        a = 1
        b = 2
>       assert a == b
E       assert 1 == 2

```

PR #1838, opened 2025-10-10
[BUG FIX] Fix loading MJCF file with includes.

YilingQiao:


genesis/utils/mjcf.py
```diff
@@ -138,6 +145,26 @@ def build_model(xml, discard_visual, default_armature=None, merge_fixed_links=Fa
...
+                        mj.jnt_solreflimit[:, 0] = MIN_TIMECONST
+                    else:
+                        mj.jnt_solref[:, 0] = MIN_TIMECONST
+                if ("joint", "solreffriction") in _applied_defaults and mj.njnt > 0:
+                    if hasattr(mj, "jnt_solreffriction"):
+                        mj.jnt_solreffriction[:, 0] = MIN_TIMECONST
+                    else:
+                        mj.jnt_solref[:, 0] = MIN_TIMECONST
+                if ("joint", "armature") in _applied_defaults and default_armature is not None:
+                    mj.dof_armature[:] = default_armature
```
duburcqa: I don't like this entire logic duplication. I hope there should be a better way. We put a lot of effort into unifying URDF and MJCF loading, it would be a shame to do a step back.

Issue #1820, opened 2025-10-09
[Bug]: Issues loading Allegro hand URDF model


ebianchi:
### Bug Description

The Allegro robotic hand URDF [from this repo](https://github.com/RobotLocomotion/models/tree/master/allegro_hand_description) does not load as expected in Genesis.  

Here's how the URDF looks when loaded into Genesis (broken):
<img width="506" height="679" alt="Image" src="https://github.com/user-attachments/assets/86fab167-6f47-491c-8bd2-dd9d94fe5ee1" />

Here's how the URDF looks when loaded into Drake's model visualizer (correct):
<img width="833" height="802" alt="Image" src="https://github.com/user-attachments/assets/e9ddc619-74d2-445f-a898-13bba715a250" />

It appears the `axis` and/or `origin` tags under `joint` definitions are not respected when loaded into Genesis.

### Steps to Reproduce

Download and extract this folder, containing a slightly edited Allegro portion of the RobotLocomotion `models` repo.

[allegro_hand_description.zip](https://github.com/user-attachments/files/22800051/allegro_hand_description.zip)

Load in the model and visualize:

```python
# Code snippet showing the incorrectly reconstructed Allegro
import genesis as gs
gs.init()
scene = gs.Scene(show_viewer=True)
broken = scene.add_entity(
    gs.morphs.URDF(file='allegro_hand_description/urdf/allegro_hand_description_right.urdf'),
)
scene.build()
breakpoint()  # Observe the model in the visualizer window
```

### Expected Behavior

The expected behavior is the Allegro hand assembles properly, such as in the meshcat screenshot.

### Relevant log output

```shell

```

### Environment

- OS:  Ubuntu 22.04
- GPU:  NVIDIA GeForce RTX 4090
- GPU-driver version:  560.35.05
- CUDA / CUDA-toolkit version:  12.6

### Release version or Commit ID

5a9008d1fd9820bf125fcad33c91e45a29721440

ebianchi:
The issue may have something to do with how Genesis loads the gltf visual geometries.  If I use the collision geometries (made from URDF box and sphere primitives) also as the visual geometries, the layout appears correct.  See below for a 2-finger test.

<img width="1282" height="1058" alt="Image" src="https://github.com/user-attachments/assets/02fa1ace-4f75-4882-8494-bb4da50f49a1" />

duburcqa:
@ACMLCZH Could you have a look to this issue?

PR #1856, opened 2025-10-13
[MISC] Fix gltf Loading for URDF

ACMLCZH:
- For Issue #1820 , when URDF loads gltf meshes, it uses "y-up" by default. It is currently loaded as "z-up" now.
franka_1


YilingQiao:
I saw you are updating the hugging face commit hash but did not see any test. Did you forget to add your unit test?

PR #1857, opened 2025-10-13
[MISC] Fix gltf Loading for URDF

### LINKED ISSUE #1820: [Bug]: Issues loading Allegro hand URDF model

ACMLCZH:
When URDF loads gltf meshes, it uses "y-up" by default, but it currently loaded as "z-up" now.
franka_1

Resolves #1820

### DIFF
--- a/genesis/engine/mesh.py
+++ b/genesis/engine/mesh.py
@@ -343,7 +343,7 @@ class Mesh(RBC):
             if morph.is_format(gs.options.morphs.MESH_FORMATS):
                 meshes = mu.parse_mesh_trimesh(morph.file, morph.group_by_material, morph.scale, surface)
             elif morph.is_format(gs.options.morphs.GLTF_FORMATS):
-                if not morph.parse_glb_with_trimesh:
+                if not morph.parse_glb_with_zup:
                     gs.logger.warning(
                         "GLTF is using y-up while Genesis uses z-up. Please set parse_glb_with_zup=True"
                         " in morph options if you find the mesh is 90-degree rotated. We will set parse_glb_with_zup=True"


YilingQiao:
can you add a unit test? Just for this scene and compare the image pixel?

duburcqa:
> can you add a unit test? Just for this scene and compare the image pixel? Pixel machine may not be the best, because we don't really look at them :/ We about computing the principal axes and check if it matches the one for the collision geometry?

YilingQiao:
> how about computing the principal axes and check if it matches the one for the collision geometry?

how to define the principal axes? Maybe we can find the longest dimension in an object. I don't have a strong opinion though.

duburcqa:
> how to define the principal axes? 

You ask trimesh to do it for you :) 

`_components, vectors = trimesh.inertia.principal_axis(tmesh.moment_inertia)`

tests/test_mesh.py
```diff
@@ -370,11 +371,18 @@ def test_urdf_with_existing_glb(tmp_path, show_viewer):
...
-    robot = scene.add_entity(
+    robot_urdf = scene.add_entity(
         gs.morphs.URDF(
             file=urdf_path,
         ),
     )
+    robot_mesh = scene.add_entity(
+        gs.morphs.Mesh(
+            file=glb_path,
+            parse_glb_with_zup=True,
```
YilingQiao, line 382: Is `parse_glb_with_zup` option purely for this test?

YilingQiao: if it is, maybe we don't need to add this option in the main code. You can manually do this transform in the test. What's your preference @duburcqa ?

ACMLCZH: Not really. It is also discussed two months ago that we want to add an option for glb to convert it from y_up to z_up. Since glb is always stored as y_up but we now load it in Genesis as z_up. It is expected to gradually add this option to users and finally change its behavior.

duburcqa: I’m not 100% sure to follow. It looks like this option should be removed. I don’t think users care about this kind of considérations but I may be wrong.

YilingQiao: After giving it a second thought, I think this option could be useful. Many background scenes are stored in GLTF, and when they’re loaded, they appear rotated by 90 degrees — it often takes a moment to figure out why and how to fix it. With this option, users can easily adjust the orientation.

duburcqa: But is it really supposed to be set manually? It looks like something that should just work out of the box without user intervention ?

YilingQiao: If it “just works out of the box” (i.e., the flag is set to true by default), it will change the behavior for all existing GLTF meshes. This flag provides a way for users to set it manually while preserving the current default loading behavior.

duburcqa: Changing a default behavior that is broken for something that just work should not be a blocker. I must have missed something here. Are you suggesting there is no universal way to interpret GLTF files? This sounds fishy.

YilingQiao (approved):

genesis/ext/urdfpy/utils.py
Milotrince, line 1: This change causes [allegro_hand_right_glb.urdf](https://huggingface.co/datasets/Genesis-Intelligence/assets/blob/main/allegro_hand/allegro_hand_right_glb.urdf) to be loaded with flipped axes. I see that there is the `parse_glb_with_zup` option in `gs.morphs.Mesh` but no such option seems to be exposed for `gs.morphs.URDF`

YilingQiao: I see. Is this PR to fix Allegro hand? There are two versions of this hand?
https://github.com/Genesis-Embodied-AI/Genesis/issues/1820

ACMLCZH: Add an option for URDF: #1938

Milotrince: yeah seems that there are different versions floating around. There is no "fixing" the hand, just need to make it clear know what the default convention is and provide an option to load it with different axes.
Thanks @ACMLCZH for the quick PR

PR #1871, opened 2025-10-17
[MISC] Add get_particles_pos to sph

YilingQiao:
## Description

Add an get_particles_pos API for SPH entities.

## Related Issue

## Checklist:

### DIFF
--- a/tests/test_deformable_physics.py
+++ b/tests/test_deformable_physics.py
@@ -186,28 +186,27 @@ def test_deformable_parallel(show_viewer):
     )
     scene.build(n_envs=2)
 
-    mpm_cube_pos = mpm_cube.get_particles_pos()
-    cloth_pos = cloth.get_particles_pos()
-    water_pos = water.get_particles_pos()
-
-    mpm_cube_active = mpm_cube.get_particles_active()
-    cloth_active = cloth.get_particles_active()
-    water_active = water.get_particles_active()
-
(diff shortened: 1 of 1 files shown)


tests/test_deformable_physics.py
```diff
@@ -191,6 +191,25 @@ def test_deformable_parallel(show_viewer):
...
+    mpm_cube_vel = mpm_cube.get_particles_vel()
+    cloth_vel = cloth.get_particles_vel()
+    water_vel = water.get_particles_vel()
+
+    assert mpm_cube_pos.shape == mpm_cube_vel.shape
+    assert mpm_cube_pos.shape == (mpm_cube_active.shape[0], mpm_cube.n_particles, 3)
+    assert cloth_pos.shape == cloth_vel.shape
+    assert cloth_pos.shape == (cloth_active.shape[0], cloth.n_particles, 3)
+    assert water_pos.shape == water_vel.shape
+    assert water_pos.shape == (water_active.shape[0], water.n_particles, 3)
```
duburcqa: It is not checking much, just that the API is available. Not really useful. The optional arguments are not tested and the returned value is completely ignored. It would be nice to check that the value is correct, and if too complicated, at least checking that it makes sense (i.e. non-zero and in valid range). You should also check that the value is changing over time. I think you should know the initial and final position and velocity, it could be possible to assert that.

YilingQiao: updated

PR #1874, opened 2025-10-17
[WIP] [FEATURE] Add support of GsTaichi dynamic array to SAPCoupler.

duburcqa:


genesis/utils/array_class.py
```diff
@@ -2135,28 +2134,26 @@ def __init__(self):
...
-cache_value = partial(dataclasses.field, metadata={FIELD_METADATA_CACHE_VALUE: True})
 
-
-@dataclasses.dataclass
+@dataclass
 class StaticRigidSimCacheKey:
-    para_level: int = cache_value()
-    use_gjk_collision: bool = cache_value()
-    use_hibernation: bool = cache_value()
-    batch_links_info: bool = cache_value()
```
YilingQiao, line 2146: why are you doing this? This is to remind compiler that those values need to be considered when computing cache key

genesis/utils/array_class.py
```diff
@@ -2260,3 +2256,59 @@ def __init__(self, solver):
 ContactIslandState = ti.template() if not gs.use_ndarray else StructContactIslandState
 DiffContactInput = ti.template() if not gs.use_ndarray else StructDiffContactInput
 AABBState = ti.template()
+
+
+# ===================================== SAP Coupler Hydroelastic =====================================
+
+
+@dataclass
+class StructHydroelasticRigidData:
```
YilingQiao, line 2265: as a first step, we don't necessarily need to take those fields out of the coupler class. Fields can still be accessed through `self`. We just need to ensure the rigid body related properties are passed in kernels as arguments

genesis/utils/array_class.py
duburcqa: From what I understood, it is no longer necessary after merging migration to gstaichi 3.0?
https://github.com/Genesis-Embodied-AI/Genesis/pull/1870

genesis/utils/array_class.py
duburcqa: Ok I see. Much less invasive indeed.

PR #1925, opened 2025-10-29
[MISC] Added warning message about stable time step for SPH solver

SonSang:
## Description

When using SPH solver, the timestep size should be adjusted based on particle viscosity (`mu`) and particle size to get stable results. Specifically, in the current implementation, particle acceleration is computed as ([here](https://github.com/Genesis-Embodied-AI/Genesis/blob/ebedb969ccee4bfac23d9c49f761a8b1cc4becd9/genesis/engine/solvers/sph_solver.py#L262)):
```
f_v = d * mu_i * (mass_j / rho_j) * v_ij / (dist**2 + 0.01 * self._support_radius**2) * self.cubic_kernel_derivative(d_ij))
```

Therefore, acceleration magnitude is proportional to the viscosity (`mu_i`) and particle size (`1.0 / (0.01 * self._support_radius**2)`). Note that `support_radius = 2.0 * particle_size` as shown [here](https://github.com/Genesis-Embodied-AI/Genesis/blob/ebedb969ccee4bfac23d9c49f761a8b1cc4becd9/genesis/options/solvers.py#L484). We do not care properties of `j`-th particle because these are neighboring particles of `i`-th particle and thus dynamically change. 

We compute the stable timestep as follows:
```
stable_dt = self._stable_dt_coef * (0.01 * self._support_radius**2) / self.particles_info.mu.to_numpy().max()
```

Where `_stable_dt_coef` is a tunable coefficient to determine the stable time step size, which we can adjust based on experiments. For now, it is set as `1.0`. When the current substep dt size exceeds this value, print out a warning, so that user could adjust the settings for stable simulation.

For instance, in this example:
```
scene_args = {'show_viewer': False, 'vis_options': {'show_world_frame': False}, 'profiling_options': {'show_FPS': True}, 'sph_options': {'particle_size': 0.01, 'lower_bound': [-1, -1, 0], 'upper_bound': [1, 1, 1]}, 'sim_options': {'dt': 0.0015, 'substeps': 8, 'gravity': [0, 0, -9.81]}}
...
chocolate_emitter = scene.add_emitter(
    material=gs.materials.SPH.Liquid(
        rho=1200.0,
        stiffness=50000.0,
        exponent=7.0,
        mu=2.0,
        gamma=0.02,
        sampler='regular'
    ),
    max_particles=60000,
    surface=gs.surfaces.Smooth(color=(0.25, 0.12, 0.07), double_sided=True, vis_mode='particle')
)
```

The `mu` is too large (default = `0.005`) and `particle_size` is also small (default = `0.02`). In this case, a warning is printed as:
```
[Genesis] [18:57:44] [WARNING] Current timestep (0.0001875) is larger than the stable timestep (2e-06) for SPH solver. Please consider reducing the timestep, liquid viscosity (mu), or increasing the particle size.
```

The simulation is unstable and penetration happens under this setting:

https://github.com/user-attachments/assets/b5326350-75cd-4f67-aee3-ba89464e7677

Based on the warning message, we can increase the number of substep from `8` to `256` and reduce `mu` from `2.0` to `0.5`. Then, the stable timestep increases from `2e-6` to `8e-6`, and the substep dt decreases from `0.0001875` to `5e-6`. Therefore, it satisfies the stability condition and penetration does not happen in the simulation:

https://github.com/user-attachments/assets/31ec99e2-d560-4f07-9d4d-11d95e95184e

## Related Issue

## Checklist:


YilingQiao (approved):

YilingQiao:
Usually a bug fix should come with a test. But here it's not "fixing" anything but throwing a warning (no behavior change). I don't have strong opinion regarding the unit test.

duburcqa:
@SonSang Why just a warning? I'm in favour of errors instead of allowing broken stuffs to happen.

SonSang:
@duburcqa Right now I set the `_stable_dt_coef` to 1 based on this single example. So I was thinking of adjusting this value over time based on the other examples. I thought it would be better to raise error after we get the reliable estimate of `_stable_dt_coef` later, but I do not have strong opinion about it. Maybe we can compute the `stable_dt` conservatively, and raise error instead of warning.

PR #1941, opened 2025-11-03
[CHANGING][BUG FIX] Fix external coupling forces from other solvers not affecting rigid bodies.

SonSang:
## Description

In the current implementation, every external force exerted on rigid bodies are stored in `cfrc_applied_vel` and `cfrc_applied_ang` for each link. These external forces include forces from user input and those from other solvers (e.g. force from MPM particles). However, these external forces require different handling:

* External forces from user input: They should stay the same over the entire simulation step (across substeps), and cleared to zero at the end of the simulation step.
* External forces from other solvers: They should be applied to the next substep, and be cleared to zero after being used.

Right now, the logic for `cfrc_applied_vel` and `cfrc_applied_ang` only follows the one for the forces from user input. Therefore, the external forces from other solvers are cleared at the end of simulation step without being used in the next step. Because of this, when we use `substeps=1`, the external forces from other solvers do not affect the rigid bodies at all:

https://github.com/user-attachments/assets/2c251460-27b0-4dbe-a808-95b2fafa4f08

If we use `substeps=10`, the external forces are applied to the rigid body and the simulation result is very different:

https://github.com/user-attachments/assets/a5011ed4-eeb1-4ca8-bda9-3bfbdce1d46b

Also, even when we use `substeps=10`, the simulation result is not correct because the external forces from the other solvers are not cleared after each substep -- the external forces are accumulated over the substeps. This makes the simulation results from using different `substeps` quite inconsistent to each other. For example, when we use `substeps=30`, the result is:

https://github.com/user-attachments/assets/9ae8e8b4-4588-40b8-9154-df4352c86f23

To fix this issue, this PR adds other variables, `cfrc_coupling_vel` and `cfrc_coupling_ang` to store the external forces from the other solvers. As described above, they are applied to the next substep, and cleared to zero as soon as being used. With this fix, the external forces are applied to rigid bodies even when using `substeps=1`:

https://github.com/user-attachments/assets/c38200bf-dd3c-4027-b992-328c8887eefb

This result is also consistent with the result when using `substeps=10` or `substeps=30`:

https://github.com/user-attachments/assets/f69b013a-7d24-4328-8a15-b6997c2759fe

https://github.com/user-attachments/assets/e17b0ad7-6b44-4482-a245-b3435be8cd5a

Because this PR changes the simulation result, tagged as [CHANGING], but it could be also seen as bug fix.

The code to reproduce the videos:

```
import genesis as gs

########################## init ##########################
gs.init()

########################## create a scene ##########################

dt_per_substep = 4e-4
horizon_substeps = 2000
substeps = 10
fps = int(60.0 / (substeps / 10.0))
scene = gs.Scene(
        dt       = dt_per_substep * substeps,
        substeps = substeps,
    ),
        lower_bound   = (-0.5, -1.0, 0.0),
        upper_bound   = (0.5, 1.0, 1),
    ),
    vis_options=gs.options.VisOptions(
        visualize_mpm_boundary = True,
    ),
        camera_fov=30,
    ),
    show_viewer = True,
)

########################## entities ##########################
plane = scene.add_entity(
    morph=gs.morphs.Plane(),
)

obj_rigid = scene.add_entity(
    material=gs.materials.Rigid(
        rho=200.0,
    ),
    morph=gs.morphs.Box(
        pos  = (0.0, -0.5, 0.25),
        size = (0.2, 0.2, 0.2),
    ),
    surface=gs.surfaces.Default(
        color    = (1.0, 0.4, 0.4),
        vis_mode = 'visual',
    ),
)

obj_sand = scene.add_entity(
    material=gs.materials.MPM.Liquid(),
    morph=gs.morphs.Box(
        pos  = (0.0, 0.0, 0.25),
        size = (0.3, 0.3, 0.3),
    ),
    surface=gs.surfaces.Default(
        color    = (0.3, 0.3, 1.0),
        vis_mode = 'particle',
    ),
)

camera = scene.add_camera(
    res=(800, 600),
    pos=(3.5, 0.5, 2.5),
    lookat=(0.0, 0.0, 0.5),
    up=(0.0, 0.0, 1.0),
    fov=40
)

########################## build ##########################
scene.build()

camera.start_recording()
horizon = horizon_substeps // substeps
for i in range(horizon):
    scene.step()
    camera.render()
camera.stop_recording("test.mp4", fps=fps)
```

## Related Issue

## Checklist:

### DIFF
--- a/tests/test_hybrid.py
+++ b/tests/test_hybrid.py
@@ -3,6 +3,7 @@ import pytest
 import torch
 
 import genesis as gs
+from genesis.utils.misc import tensor_to_array
 
 from .utils import assert_allclose, get_hf_dataset
 
@@ -344,5 +345,5 @@ def test_rigid_mpm_legacy_coupling(substeps, show_viewer):
         scene.step()
 
     # Check that the sand moved the box along the negative Y direction
-    pos = obj_rigid.get_pos().cpu().numpy()
+    pos = tensor_to_array(obj_rigid.get_pos())
     assert pos[1] + 0.25 < 0.0


YilingQiao:
Thank you very much! Is it easy to add a test? Like in the previous implementation, if substep is 1, the rigid body will not be impacted by coupling force (which is wrong), and the current implementation can fix it.

SonSang:
Sure, I'll add a test!

tests/test_hybrid.py
```diff
@@ -278,3 +278,74 @@ def test_sap_fem_vs_robot(show_viewer):
...
+
+
+@pytest.mark.required
+@pytest.mark.parametrize("substeps", [1, 10])
+def test_rigid_mpm_legacy_coupling(substeps, show_viewer):
+    # This test is aimining at two things:
+    # 1) When substep = 1, the rigid object should be affected by the MPM particles
+    # 2) Regardless of substeps, as far as the total step dt is the same, they should give consistent results
+    dt_per_substep = 4e-4
+    horizon_substeps = 2000
```
YilingQiao: let's see how long will this test run.

tests/test_hybrid.py
```diff
@@ -278,3 +278,74 @@ def test_sap_fem_vs_robot(show_viewer):
...
+
+    scene.build()
+
+    horizon = horizon_substeps // substeps
+    for i in range(horizon):
+        scene.step()
+
+    # Check that the sand moved the box to the correct position
+    pos = obj_rigid.get_pos()
+    target_pos = np.array([0.0, -0.55, 0.1])
```
YilingQiao: why is this target_pos the expected pos? Does this mean the sand push the objects in the -y direction?

YilingQiao (approved):

tests/test_hybrid.py
SonSang: It's based on experiments, there is no closed form formulation to get this expected pos. Would it be better to only check if the box has been moved or not, rather than using this expected pos?

YilingQiao: I think it's okay for now and don't expect the exact closed form formulation. "the sand push the objects in the -y direction" is enough. If the sand doesn't have an impact, the box would be at (0, -0.5, 0.1).

YilingQiao:
looks like we have many network issues in CI recently @duburcqa

SonSang:
Hmm the benchmark regression also seems to be an outlier, these are the results for the failed benchmark (`batch_size=2048-constraint_solver=Newton-enable_mujoco_compatibility=False-env=box_pyramid#5-gjk_collision=False-use_contact_island=False`) on my machine:

[After fix]
wandb: Run summary:
wandb:    compile_time 17.56327
wandb: realtime_factor 3307.42994
wandb:     runtime_fps 330742.99375

[Original]
wandb: Run summary:
wandb:    compile_time 17.34628
wandb: realtime_factor 3304.67197
wandb:     runtime_fps 330467.19712

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -3810,6 +3810,11 @@ def func_forward_dynamics(
         rigid_global_info=rigid_global_info,
         static_rigid_sim_config=static_rigid_sim_config,
     )
+    func_clear_coupling_force(
+        links_state=links_state,
+        rigid_global_info=rigid_global_info,
+        static_rigid_sim_config=static_rigid_sim_config,
+    )
```
duburcqa: Do not add dedicated clear ingredients kernel. This will significantly affect simulation performance. Just add the clearing code in `func_update_force` kernel, at the very end.

tests/test_hybrid.py
duburcqa: No. This is not a good approach. This is the reason why running the unit tests now takes 5h30 on windows without cache. You MUST design your test to be as short as possible. Otherwise it will seriously hinder our capability to add more unit tests in the future.

tests/test_hybrid.py
```diff
@@ -278,3 +278,74 @@ def test_sap_fem_vs_robot(show_viewer):
...
+
+@pytest.mark.required
+@pytest.mark.parametrize("substeps", [1, 10])
+def test_rigid_mpm_legacy_coupling(substeps, show_viewer):
+    # This test is aimining at two things:
+    # 1) When substep = 1, the rigid object should be affected by the MPM particles
+    # 2) Regardless of substeps, as far as the total step dt is the same, they should give consistent results
+    dt_per_substep = 4e-4
+    horizon_substeps = 2000
+    atol = 1e-2
```
duburcqa: If it is only used once, inline the value on the assert itself. This will improve readability.

duburcqa:
@SonSang The title of the PR is not correct. You should not highly the solution but what you are trying to achieve from the end-user perspective. Keep in mind that this will be used to generate the release note, do you think users care about "additional variable to store external forces applied to rigid bodies"?

I updated the title accordingly.

duburcqa:
> Hmm the benchmark regression also seems to be an outlier, these are the results for the failed benchmark

Ignore the regression check. We merged too many big changes recently, it is completely messed up.

SonSang:
> @SonSang The title of the PR is not correct. You should not highly the solution but what you are trying to achieve from the end-user perspective. Keep in mind that this will be used to generate the release note, do you think users care about "additional variable to store external forces applied to rigid bodies"?
> 
> I updated the title accordingly.

Thanks! I'll do that for the future PRs.

duburcqa (approved):

PR #1947, opened 2025-11-05
[BUG FIX] Fixed bug in benchmark ID definition and regression detection.

SonSang:
## Description

There was a mismatch in defining benchmark IDs between W&B and artifact:
```
# W&B
self.hparams = {**hparams, "dtype": "ndarray" if gs.use_ndarray else "field"}
self.benchmark_id = "-".join((BENCHMARK_NAME, pprint_oneline(hparams, delimiter="-")))
self.wandb_run = wandb.init(
    project="genesis-benchmarks",
    name="-".join((self.benchmark_id, revision)), ...

# Artifact
msg = (
    pprint_oneline(self.hparams, delimiter=" \t| ")
    + " \t| "
    + pprint_oneline(items, delimiter=" \t| ", digits=1)
)
```
Note that W&B uses `hparams`, and artifact uses `self.hparams`.

Therefore, during detecting regression, we could not search W&B results that have the same benchmark IDs from artifacts. To fix this, changed W&B to use the same `self.hparams` to define benchmark ID.

Also, in the previous regression detection logic, the search succeeded only when all of the previous commits had the same benchmark ID:

```
def mean_of(metric, bid):
    tot = 0.0
    for record in records_by_rev.values():
        stats = record.get(bid)
        if stats is None:
            break
```

In this PR, relaxed this condition:

```
def mean_of(metric, bid):
    tot = 0.0
    count = 0
    for record in records_by_rev.values():
        stats = record.get(bid)
        if stats is None:
            continue
        tot += record[bid][metric]
        count += 1
    
    if count == 0:
        return float("nan")
    return tot / count
```

Also, changed to print out `Not Found` instead of `nan` when there is no matching previous benchmark result.

## Related Issue

## Motivation and Context

The benchmark regression results showed `nan` values.

## Checklist:

### DIFF
--- a/.github/workflows/alarm.yml
+++ b/.github/workflows/alarm.yml
@@ -109,9 +109,15 @@ jobs:
               return out
 
           def fmt_num(v):
-              return f"{int(v):,}" if isinstance(v, int) or v.is_integer() else f"{v:.2f}"
+              if math.isnan(v):
+                  return "---"
+              if isinstance(v, int) or v.is_integer():
+                  return f"{int(v):,}"
+              return f"{v:.2f}"
 
           def fmt_pct(v, *, highlight=False):
+              if math.isnan(v):
+                  return "---"
               s = f"{v:+.2f}%"
               return f"**{s}**" if highlight else s
 
@@ -211,16 +217,16 @@ jobs:
               }
 
           def mean_of(metric, bid):
-              tot = 0.0
-              count = 0
+              tot, count = 0.0, 0
               for record in records_by_rev.values():
                   stats = record.get(bid)
                   if stats is None:
                       continue
                   tot += record[bid][metric]
                   count += 1
-              
-              mean = float("nan") if count == 0 else tot / count
+              if count == 0:
+                  return float("nan"), 0
+              mean = tot / count
               return mean, count
 
           # ----- build TWO tables -----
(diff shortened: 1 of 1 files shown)


YilingQiao (approved):

SonSang:
Changed to print `undefined` when the number of previous samples is less than the given number of samples (now 5). When `undefined`, we do not raise warning even if there is regression, because it could have high variance. Only when there are enough number of samples, we raise regression warning. The example output is like:

```
| ℹ️ | `batch_size=30000-constraint_solver=Newton-dtype=field-env=batched_franka-gjk_collision=False-use_contact_island=False` | 41.20 | Undefined (nan) | +nan% |
| ℹ️ | `batch_size=30000-constraint_solver=Newton-dtype=field-env=batched_franka-gjk_collision=True-use_contact_island=False` | 40 | Undefined (nan) | +nan% |
| ℹ️ | `batch_size=30000-dtype=field-env=random-use_contact_island=False` | 39.90 | Undefined (nan) | +nan% |
| ℹ️ | `batch_size=8192-constraint_solver=CG-dtype=field-env=cube#10-gjk_collision=False-use_contact_island=False` | 42.40 | Undefined (37) | +14.59% |
| ℹ️ | `batch_size=8192-constraint_solver=CG-dtype=field-env=cube#10-gjk_collision=True-use_contact_island=False` | 42.30 | Undefined (40) | +5.75% |
```

Note that `Undefined (nan)` is printed when there is no data at all. When there is at least one data, we still report the statistics like `Undefined (xxx)`.

PR #1948, opened 2025-11-05
[FEATURE] Implement position-velocity controller.

matthieuvigne:
In robotics, a PD controller write:

```math 
\tau = - K_p (pos - pos_{target})  - K_d (vel - vel_{target}) 
```
where vel_target is the derivative of the position target.

Howerver, the controller implemented in `control_dofs_position` is 
```math 
\tau = - K_p (pos - pos_{target})  - K_d vel
```
The lack of the velocity feedforward term leads to a 'drag' tracking error when following dynamic trajectory.

This PR adds a new function `control_dofs_position_velocity` which implements the above-mentioned PD. This leads to improved joint tracking - illustrated here on a single robotic joint tracking a sinusoid - with the orange curve almost perfectly overlapping the target:

image

TODO: 
 - [x] Add unit test
 - [x] Add comparative example

### DIFF
--- a/examples/rigid/control_mesh.py
+++ b/examples/rigid/control_mesh.py
@@ -1,7 +1,5 @@
 import argparse
 
-import numpy as np
-
 import genesis as gs
 
 
@@ -45,17 +43,13 @@ def main():
     scene.build()
 
     dofs_idx = duck.base_joint.dofs_idx
+    duck.set_dofs_kp((0.3,) * 6, dofs_idx)
+    duck.set_dofs_kv((1.0,) * 6, dofs_idx)
 
-    duck.set_dofs_kv(
-        np.array([1, 1, 1, 1, 1, 1]) * 50.0,
-        dofs_idx,
-    )
     pos = duck.get_dofs_position()
     pos[-1] = 1.0  # rotate around intrinsic z axis
-    duck.control_dofs_position(
-        pos,
-        dofs_idx,
-    )
+    duck.control_dofs_position(pos, dofs_idx)
+
     for i in range(1000):
         scene.step()
 


genesis/constants.py
```diff
@@ -47,6 +47,7 @@ class CTRL_MODE(IntEnum):
     FORCE = 0
     VELOCITY = 1
     POSITION = 2
+    POSITION_VELOCITY = 3
```
duburcqa: I don't think internally you need this extra mode. I would just remove `POSITION` and use `POSITION_VELOCITY` everywhere instead.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -1824,7 +1824,7 @@ def control_dofs_velocity(self, velocity, dofs_idx=None, envs_idx=None, *, unsaf
         )
         if not unsafe and not has_gains:
```
duburcqa, line 1878: `kernel_control_dofs_velocity` should be modified to only consider `kv` and ignore `kp`

genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -1824,7 +1824,7 @@ def control_dofs_velocity(self, velocity, dofs_idx=None, envs_idx=None, *, unsaf
...
             raise gs.raise_exception(
-                "Please set control gains kp,kv using `set_dofs_kp`,`set_dofs_kv` prior to calling this method."
+                "Please set control gains kv using `set_dofs_kp`,`set_dofs_kv` prior to calling this method."
```
duburcqa: Remove "using `set_dofs_kp`"

duburcqa:
It would be nice to add a unit test showing good tracking performance on a simple pendulum.

genesis/constants.py
matthieuvigne: Hum I had understood for our discussion IRL that you explicitly wanted a different mode, to clearly outline the difference between a Mujoco-compatible POSITION mode and this one...

So you propose to get rid of this mode and use `kernel_control_dofs_position` to explicitly set the velocity target to 0, while using the same formula for both modes ?

genesis/engine/solvers/rigid/rigid_solver_decomp.py
matthieuvigne: I don't understand what you are referring to: this was already the case, only the documentation was wrong.

genesis/constants.py
duburcqa: > Hum I had understood for our discussion IRL that you explicitly wanted a different mode, to clearly outline the difference between a Mujoco-compatible POSITION mode and this one...

Only for the public API. For the private API, body is supposed to know what is happening under the hood, these modes are implementation details. I just initially thought it was necessary, but just setting the target velocity to zero is a nice trick.

genesis/engine/solvers/rigid/rigid_solver_decomp.py
duburcqa: No. The kernel itself is wrong. it is checking that kp is non zeros, which should not be required for velocity control.

matthieuvigne: Ah ok, I was confused as I was looking at my local code - you are right it is not the case on master, but this is already implemented in this MR

```diff
@@ -6589,7 +6619,7 @@ def kernel_control_dofs_velocity(
        I_d = [i_d, i_b] if ti.static(static_rigid_sim_config.batch_dofs_info) else i_d
        dofs_state.ctrl_mode[i_d, i_b] = gs.CTRL_MODE.VELOCITY
        dofs_state.ctrl_vel[i_d, i_b] = velocity[i_b_, i_d_]
-        if (dofs_info.kp[I_d] > gs.EPS) | (dofs_info.kv[I_d] > gs.EPS):
+        if (dofs_info.kv[I_d] > gs.EPS):
            has_gains = True
    return has_gains


```

genesis/engine/solvers/rigid/rigid_solver_decomp.py
YilingQiao:
```
Please set control gains kv using set_dofs_kp,set_dofs_kv prior to calling this method.
```
It still has `set_dofs_kp`

duburcqa (dismissed):

PR #1949, opened 2025-11-06
[BUG FIX] Fix silent process killing issue in MPM simulation by raising an exception

SonSang:
## Description

When running MPM simulation, unexpected `nan` value leads to wrong rendering (see below) and silent process failure. Even when there is no logical bug, these `nan` values could arise when the simulation time step is too small or material properties are bad. 

To fix this process-killing bug and tell user about the reason of failure, added a `nan` detection logic for MPM simulation. Specifically, it detects if a grid cell's `new_vel` becomes `nan` in `g2p` function, as all MPM states depend on `new_vel`:
```
class RETURN_CODE(IntEnum):
    """
    Return codes for detecting potential issues during MPM simulation.
    """
    SUCCESS = 0
    NAN = 1

...
def g2p():
...
    if ti.math.isnan(new_vel).any():
        self.kernel_flag[None] = ti.atomic_max(self.kernel_flag[None], RETURN_CODE.NAN)
...
```

## Motivation and Context

(Silent) Process killing bug of MPM simulation, caused by `nan` value.

## Checklist:

### DIFF
--- a/genesis/engine/entities/emitter.py
+++ b/genesis/engine/entities/emitter.py
@@ -163,7 +163,7 @@ class Emitter(RBC):
             vels = speed * direction
 
             if n_particles > self._entity.n_particles:
-                gs.logger.warning(
+                gs.raise_exception(
                     f"Number of particles to emit ({n_particles}) at the current step is larger than the maximum "
                     f"number of particles ({self._entity.n_particles})."
                 )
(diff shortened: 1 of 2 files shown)


duburcqa:
I don’t think you need to store this kernel flag anywhere, and accessing it in python would be slow. Just return a standard boolean. In this case I don’t think you need atomic to protect for simulatenous write since it is a single bit and every thread would be writing exactly the same value anyway.

duburcqa:
A simple “is_success” Boolean flag without enumerating would be sufficient for now. Let’s complexity only if needed ;)

duburcqa (approved):

duburcqa (approved):

PR #1959, opened 2025-11-07
[BUG FIX] Fix wrong computation of accelerometer signal.

pschoc:
## Related Issue

## Checklist:

### DIFF
--- a/genesis/engine/sensors/imu.py
+++ b/genesis/engine/sensors/imu.py
@@ -10,7 +10,12 @@ from genesis.options.sensors import (
     MaybeMatrix3x3Type,
     IMU as IMUOptions,
 )
-from genesis.utils.geom import inv_transform_by_trans_quat, transform_by_quat, transform_quat_by_quat
+from genesis.utils.geom import (
+    inv_transform_by_trans_quat,
+    transform_by_quat,
+    transform_quat_by_quat,
+    inv_transform_by_quat,
+)
 from genesis.utils.misc import concat_with_tensor, make_tensor_field, tensor_to_array
 
 from .base_sensor import (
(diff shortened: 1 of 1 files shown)


pschoc:
also I have trouble understanding the skew:

how is the "skew" meant to be used or what is it expressing? From what I see in imu.py/_get_skew_to_alignment_matrix() it generates a rotation matrix of the form:
[ 1   v1  v2 ]
[ v0  1   v2 ]
[ v0  v1  1  ]
given a skew vector [v0, v1, v2]. That makes no sense to me, I was expecting sth like eye(3) + hat([v0,v1,v2]) = 
[ 1   -v2 v1 ]
[ v2  1   -v0 ]
[ -v1  v0  1  ]
or handing over the vector of angular misalignments which then get mapped from euler angles to so3

duburcqa:
@Milotrince Why the unit test is passing ?

Milotrince:
@pschoc Thanks for the catch.

@duburcqa I will update the unit test to cover this, i think the current test is too simple (local orientation = global orientation) so it didn't catch it.

> also I have trouble understanding the skew:

The axis skew is supposed to represent sensor axis misalignment / cross-axis coupling.

matlab also uses this definition, for reference: https://www.mathworks.com/help/nav/ref/imu.html
Screenshot 2025-11-07 at 10 10 22

Milotrince:
maybe we could rename `axes_skew` to `axes_misalignment` or `axes_coupling` to avoid confusion ?

genesis/engine/sensors/imu.py
```diff
@@ -211,7 +216,9 @@ def _update_shared_ground_truth_cache(
         local_ang = inv_transform_by_trans_quat(ang, shared_metadata.offsets_pos, offset_quats)
 
         *batch_size, n_imus, _ = local_acc.shape
-        local_acc = local_acc - gravity.unsqueeze(-2).expand((*batch_size, n_imus, -1))
+        local_acc = local_acc - inv_transform_by_quat(
+            gravity.unsqueeze(-2).expand((*batch_size, n_imus, -1)), offset_quats
+        )
```
duburcqa, line 221: Just subtract gravity BEFORE moving to local frame.

duburcqa:
Superseded by https://github.com/Genesis-Embodied-AI/Genesis/pull/1962

PR #1965, opened 2025-11-10
[MISC] Format benchmark params as table and support download as csv.

SonSang:
## Description

This PR changes `alarm` workflow in two ways:

1) Parse benchmark name and reframe comparison table based on it:
image

2) Enable downloading of `csv` and `jsonl` file for the comparison table

## Related Issue

## Checklist:

### DIFF
--- a/.github/workflows/alarm.yml
+++ b/.github/workflows/alarm.yml
@@ -234,7 +234,13 @@ jobs:
           for metric, alias in (("runtime_fps", "FPS"), ("compile_time", "compile")):
               rows_md = []
 
-              header_cells = ["status"] + params_name + [f"current {alias}", f"baseline {alias} [last (mean ± std)]", f"Δ {alias}"]
+              header_cells = (
+                  "status",
+                  *params_name,
+                  f"current {alias}",
+                  f"baseline {alias} [last (mean ± std)] (*1)",
+                  f"Δ {alias} (*2)"
+              )
               header = "| " + " | ".join(header_cells) + " |"
               align  = "|:------:|" + "|".join([":---" for _ in params_name]) + "|---:|---:|---:|"
 
@@ -319,6 +325,9 @@ jobs:
                   "",
                   "### Compile Time",
                   *tables["compile_time"],
+                  "",
+                  f"- (*1) last: last commit on main, mean/std: stats over revs {MAX_VALID_REVISIONS} commits if available.",
+                  f"- (*2) Δ: relative difference between PR and last commit on main, i.e. (PR - main) / main * 100%.",
               ]
           )
 


.github/workflows/alarm.yml
```diff
@@ -204,11 +212,36 @@ jobs:
...
+                      if token and "=" in token:
+                          k, v = token.split("=", 1)
+                          kv[k.strip()] = v.strip()
+              return kv
+
+          id2kv = {bid: parse_norm_id(bid) for bid in current_bm.keys()}
+          all_keys = set()
+          for kv in id2kv.values():
+              all_keys.update(kv.keys())
+          ordered_keys = sorted(all_keys)
```
duburcqa: We already of `normalize_benchmark_id` and `_normalize_kv_id`. Maybe you should remove them and type benchmark id as a sorted tuple of pairs ? (not a dict, to make sure it is "frozen" and == is well defined).

.github/workflows/alarm.yml
```diff
@@ -228,27 +262,38 @@ jobs:
...
+                  rows_md.append("| " + " | ".join([picto] + key_cells + [value_repr, stats_repr, delta_repr]) + " |")
+
+                  # CSV/JSONL
+                  rows_for_csv[metric].append({
+                      "metric": metric,
+                      "benchmark_id": bid,
+                      "status": "regression" if is_reg else ("not_enough_baselines" if (len(values_prev) < MAX_VALID_REVISIONS) else "ok"),
+                      **{k: kv.get(k, "-") for k in ordered_keys},
+                      "current": float(value_cur) if isinstance(value_cur,(int,float)) else None,
+                      "baseline_mean": float(value_ref) if isinstance(value_ref,(int,float)) else None,
```
duburcqa: You should export `mean`, `last`, `min`, `max`.

.github/workflows/alarm.yml
```diff
@@ -274,15 +319,41 @@ jobs:
...
+              w.writeheader()
+              for rec in rows_for_csv["runtime_fps"]:
+                  w.writerow(rec)
+
+          # compile_time.csv
+          with csv_compile.open("w", newline="", encoding="utf-8") as f:
+              w = csv.DictWriter(f, fieldnames=csv_header)
+              w.writeheader()
+              for rec in rows_for_csv["compile_time"]:
+                  w.writerow(rec)
```
duburcqa: Avoid duplicated code. Use a for-loop.

.github/workflows/alarm.yml
```diff
@@ -274,15 +319,41 @@ jobs:
...
 
-          # Write files
-          check_body_path.write_text(check_body + "\n", encoding="utf-8")
-          pr_comment_path.write_text(comment_body + "\n", encoding="utf-8")
+          # common CSV header
+          csv_header = (
+              ["metric", "benchmark_id", "status"]
+              + ordered_keys
+              + ["current", "baseline_last", "baseline_mean", "baseline_std", "delta_pct"]
+          )
```
duburcqa:
```python
csv_header = (
    "metric", "benchmark_id", "status", *ordered_keys, "current", "baseline_last", "baseline_mean", "baseline_std", "delta_pct"
)
```

.github/workflows/alarm.yml
duburcqa: `ordered_keys` is not a good name. `params_name` of something like this would be more appropriate.

duburcqa:
@SonSang Could you add a legend to explain what `last, mean, min, max` stand for? What "baseline" and "current" stand for. What "-" in params table means (i.e. using default value). And explain that only `last` will be reported if there not enough commits are available for this specific benchmark id.

.github/workflows/alarm.yml
SonSang: You mean we can remove `normalize_benchmark_id` and `_normalize_kv_id`, and typing benchmark id (`bid` here) as a sorted tuple of pairs? It is currently a string, not a dict, so just to confirm.

SonSang:
> @SonSang Could you add a legend to explain what `last, mean, min, max` stand for? What "baseline" and "current" stand for. What "-" in params table means (i.e. using default value). And explain that only `last` will be reported if there not enough commits are available for this specific benchmark id.

Currently we export `last, min, max` only to `csv` files, not in the markdown report. Do you think we should also include them in the markdown report? And when there is not enough commits, we just use average of what we have, not the last value. Would it be better to change to using only the last value?

duburcqa:
Why not just to do the same as we are already doing for the existing markdown report? It already display last, mean, std.

SonSang:
> Why not just to do the same as we are already doing for the existing markdown report? It already display last, mean, std.

image

Right now we only report mean and std, not the last value. I can make another columns for last, max, and min.

+) Oh sorry, I misunderstood, you're right. I will add columns for max and min (the last is also reported now)

duburcqa:
> +) Oh sorry, I misunderstood, you're right. I will add columns for max and min (the last is also reported now)

I don't think it is necessary in this markdown report. Too much information would be harder to read at first glance.

SonSang:
> > +) Oh sorry, I misunderstood, you're right. I will add columns for max and min (the last is also reported now)
> 
> I don't think it is necessary in this markdown report. Too much information would be harder to read at first glance.

Oh you mean we don't have to display max / min in the markdown report? Or do you mean we do not have to add separate columns for them (maybe displaying in the same column as last, mean, std?)

.github/workflows/alarm.yml
duburcqa: yes

PR #1968, opened 2025-11-11
[MISC] Cleanup PARA_LEVEL. Add 'GS_PARA_LEVEL' env var.

duburcqa:
* Cleanup PARA_LEVEL. 
* Add 'GS_PARA_LEVEL' env var.
* Add global EPS to Data manager to please GsTaichi fastcache.


YilingQiao:
Is this error because of synchronization?

duburcqa:
Good question, it is not clear for now. I will investigate.

PR #1970, opened 2025-11-11
[FEATURE] IPC Coupling with ExternalArticulationConstraint & TwoWaySoftConstraint

Roushelfy:
This PR introduces two major improvements to the IPC coupling system:

1. Simplified rigid body mesh handling
   - Use uipc.geometry.trimesh directly instead of converting to tetmesh
   - ABD works with surface meshes, no need for volumetric tetmesh
   - Improves performance and simplifies mesh processing pipeline

2. IPC-only rigid bodies (one-way coupling)
   - Add set_link_only_in_ipc(entity) API for entities that only exist in IPC
   - These entities use full density (no mass splitting with Genesis)
   - Skip SoftTransformConstraint (no coupling forces)
   - Directly set Genesis transforms from IPC results
   - Fix quaternion conversion: scipy [x,y,z,w] -> Genesis [w,x,y,z]
   - Zero velocities after transform updates to avoid spurious forces
   - Useful for passive rigid objects like screws, nuts that only need accurate contact

Usage:
    scene.sim.coupler.set_link_only_in_ipc(entity=screw)


YilingQiao (dismissed):

genesis/engine/couplers/ipc_coupler.py
```diff
@@ -889,11 +964,128 @@ def _retrieve_rigid_states(self, f):
...
 
+        # Handle IPC-only links: directly set Genesis transform to IPC result (one-way coupling)
+        self._set_genesis_transforms_from_ipc(abd_data_by_link)
+
         # Apply coupling forces from IPC ABD to Genesis rigid bodies (two-way coupling)
         # Based on soft_transform_constraint.cu gradient computation
         if self.options.two_way_coupling:
             self._apply_abd_coupling_forces(abd_data_by_link)
 
+    def _set_genesis_transforms_from_ipc(self, abd_data_by_link):
```
duburcqa: This code is very inefficient and should be improved before merging. Please implement a dedicated taichi kernel for this.

duburcqa (changes requested):
Must improve performance

duburcqa:
Moreover, unit tests should be implemented, with skip if module is not available.

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -0,0 +1,3458 @@
...
+
+        # ExternalArticulationConstraint instance (created in _init_ipc if needed)
+        self._ipc_eac = None
+
+        # Mapping from link_idx to ABD geometry for articulation constraint
+        # Structure: {(env_idx, link_idx): abd_geometry}
+        self._link_to_abd_geo = {}
+        # Mapping from link_idx to ABD geometry slot for articulation constraint
+        # Structure: {(env_idx, link_idx): abd_geometry_slot}
+        self._link_to_abd_slot = {}
```
alanray-tech, line 987: `_link_to_abd_slot` is able to get `abd_geo` by `geo:SimplicialComplex=slot.geometry()` it seems we don't need the `self._link_to_abd_geo` ?

PR #1977, opened 2025-11-12
[MISC] Enable back GsTaichi dynamic array mode by default except for MacOS.

duburcqa:
* Get around GsTaichi bug causing dof frictionloss to break on Apple Metal.
* Enable back GsTaichi dynamic array mode by default except for MacOS.

### DIFF
--- a/genesis/__init__.py
+++ b/genesis/__init__.py
@@ -118,8 +118,7 @@ def init(
 
     # Configure GsTaichi fast cache and array type
     global use_ndarray, use_fastcache
-    # is_ndarray_disabled = (os.environ.get("GS_ENABLE_NDARRAY") or ("0" if sys.platform == "darwin" else "1")) == "0"
-    is_ndarray_disabled = os.environ.get("GS_ENABLE_NDARRAY", "0") == "0"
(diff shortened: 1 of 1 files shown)


genesis/__init__.py
```diff
@@ -118,8 +118,7 @@ def init(
 
     # Configure GsTaichi fast cache and array type
     global use_ndarray, use_fastcache
-    # is_ndarray_disabled = (os.environ.get("GS_ENABLE_NDARRAY") or ("0" if sys.platform == "darwin" else "1")) == "0"
-    is_ndarray_disabled = os.environ.get("GS_ENABLE_NDARRAY", "0") == "0"
+    is_ndarray_disabled = (os.environ.get("GS_ENABLE_NDARRAY") or ("0" if sys.platform == "darwin" else "1")) == "0"
```
YilingQiao, line 121: if env var has GS_ENABLE_NDARRAY, `is_ndarray_disabled=True`?

duburcqa: Why do you think so? It will be disabled if the user set this env var to “0”. If empty then the platform-specific default value will be used. This means that if the user does `GS_ENABLE_NDARRAY=` , then it will be interpreted as `GS_ENABLE_NDARRAY=1` except on MacOS. I can change this logic if you want.

YilingQiao: oh, I get it. I misunderstood the parenthesis

PR #1984, opened 2025-11-15
[MISC] Only enable GJK by default if gradient computation is required.

duburcqa:
* Increase default number of compile threads.
* Do not forcibly disable self-collision in benchmarks anymore.
* Only enable GJK by default if gradient computation is required.

### DIFF
--- a/genesis/options/solvers.py
+++ b/genesis/options/solvers.py
@@ -349,7 +349,8 @@ class RigidOptions(Options):
     max_dynamic_constraints : int, optional
         Maximum number of dynamic constraints (like suction cup). Defaults to 8.
     use_gjk_collision: bool, optional
-        Whether to use GJK for collision detection instead of MPR. Defaults to True.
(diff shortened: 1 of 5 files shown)


genesis/options/solvers.py
```diff
@@ -400,7 +401,7 @@ class RigidOptions(Options):
     enable_mujoco_compatibility: bool = False
 
     # GJK collision detection
-    use_gjk_collision: bool = True
```
YilingQiao, line 403: can we still use gjk by default?

duburcqa: GJK is much slower than MPR+SDF without any proven benefit at the moment. It is clearly a more future-proof algorithm than MPR+SDF but I think it requires more polishing before being pushed to the user by default. Still we do not have any recorded failure cases for the latest version of MPR+SDF in practice. In my opinion, its dramatic impact on runtime speed makes a worst default choice than MPR+SDF for most users that do not need extreme robustness.

YilingQiao: sure!

Issue #1556, opened 2025-08-13
[Feature]: Text display or support more display method(class)?


prexhu:
### What feature or enhancement are you proposing?

Some user input commands (such as linear/angular velocity) are defined in the example script `locomoiton/go2_train.py`. They are working fine in genesis.

But it should be better for showing the result if it could display the user commands dynamically in the render window.

I have read the source code, and I find some text-show methods like `render_text/render_texts` in `ext/pyrender/renderer.py`.

I also tried to make some modifications to source-code for my ideas. However, limited to my beginner-coding ability, these modifications cause simulation crash:(

Besides, adding more indicators seems feasible, like attaching an arrow that indicates an external force applied to the legged robot.

<img width="330" height="172" alt="image" src="https://github.com/user-attachments/assets/a1f4d3cc-e285-4b51-bc54-a89ee40bfb12" />

> Picture from Learning Quadrupedal Locomotion over Challenging Terrain

### Motivation

Better observe the robot's response to input commands.

### Potential Benefit

For a better interactive/presentation experience.

### What is the expected outcome of the implementation work?

- [ ] Show some information dynamically in the render window

duburcqa:
This issue has been fixed on main!

duburcqa:
Fixed by #2004

Issue #713, opened 2025-02-12
[Feature]: Capture keyboard commands


tesslerc:
### What feature or enhancement are you proposing?

Enabling function callbacks on custom keyboard commands sent to the viewer.

### Motivation

As a researcher working on robotic simulation
I want to dynamically interact with the simulation. For example, a button for screen capture, resetting the scene, changing camera focus to another agent, etc...
Because this enables improved debugging and visualization.

### Potential Benefit

Aligning with the feature set of other simulators, such as IsaacGym and IsaacLab.

### What is the expected outcome of the implementation work?

- Programmer can define keyboard commands and what should happen when the user clicks them.

PR #2004, opened 2025-11-17
[FEATURE] Introduce interactive viewer plugins.

Milotrince:
## Description
nice-to-have feature, allows users to integrate interactive viewer plugins.

- Remove dependency on `pynput`
- `gs.vis.keybindings.Keybind` callback registration
- in `gs.options.ViewerOptions`:
  - `disable_keyboard_shortcuts` replaced with `disable_default_keybinds`
  - add `disable_help_text`

## Related Issue

Resolves https://github.com/Genesis-Embodied-AI/Genesis/issues/713
Resolves https://github.com/Genesis-Embodied-AI/Genesis/issues/1556

## Motivation and Context
- `ViewerPlugin` enables directly interfacing with pyviewer mouse/keyboard controls.
- Replaces `pynput`
- Simple keybind callback registration

added `tests/test_viewer.py`

Screenshot 2026-01-30 at 16 08 08
registered keys automatically show in help text :)

## Checklist:

### DIFF
--- a/examples/IPC_Solver/ipc_arm_cloth.py
+++ b/examples/IPC_Solver/ipc_arm_cloth.py
@@ -310,6 +310,9 @@ def run_sim(scene, entities, add_keybinds, mode="interactive", trajectory_file=N
 
             scene.step()
             step_count += 1
+
+            if "PYTEST_VERSION" in os.environ:
+                break
     except KeyboardInterrupt:
         gs.logger.info("Simulation interrupted, exiting.")
 
--- a/examples/drone/interactive_drone.py
+++ b/examples/drone/interactive_drone.py
@@ -73,19 +73,19 @@ def main():
     scene.build()
 
     # Register keybindings
-    def direction_keybinds(key: Key, name: str, direction: tuple[float, float, float, float]):
+    def direction_keybinds(name: str, key: Key, direction: tuple[float, float, float, float]):
         """Helper to create press/release keybinds for a direction"""
         dir_arr = np.array(direction)
         return [
             Keybind(
-                key_name=f"{name}_press",
+                name=f"{name}_press",
                 key=key,
                 key_action=KeyAction.PRESS,
                 callback=controller.add_direction,
                 args=(dir_arr,),
             ),
             Keybind(
-                key_name=f"{name}_release",
+                name=f"{name}_release",
                 key=key,
                 key_action=KeyAction.RELEASE,
                 callback=controller.add_direction,
@@ -93,13 +93,20 @@ def main():
             ),
         ]
 
+    is_running = True
+
+    def stop():
+        nonlocal is_running
+        is_running = False
+
     scene.viewer.register_keybinds(
-        *direction_keybinds(Key.UP, "move_forward", (1.0, 1.0, -1.0, -1.0)),
-        *direction_keybinds(Key.DOWN, "move_backward", (-1.0, -1.0, 1.0, 1.0)),
-        *direction_keybinds(Key.LEFT, "move_left", (-1.0, 1.0, -1.0, 1.0)),
-        *direction_keybinds(Key.RIGHT, "move_right", (1.0, -1.0, 1.0, -1.0)),
+        *direction_keybinds("move_forward", Key.UP, (1.0, 1.0, -1.0, -1.0)),
+        *direction_keybinds("move_backward", Key.DOWN, (-1.0, -1.0, 1.0, 1.0)),
+        *direction_keybinds("move_left", Key.LEFT, (-1.0, 1.0, -1.0, 1.0)),
+        *direction_keybinds("move_right", Key.RIGHT, (1.0, -1.0, 1.0, -1.0)),
         Keybind("accelerate", Key.SPACE, KeyAction.HOLD, callback=controller.accelerate),
         Keybind("decelerate", Key.LSHIFT, KeyAction.HOLD, callback=controller.decelerate),
+        Keybind("quit", Key.ESCAPE, KeyAction.PRESS, callback=stop),
     )
 
     # Print control instructions
@@ -110,15 +117,12 @@ def main():
     print("→ - Move Right (East)")
     print("space - Increase RPM")
     print("shift - Decrease RPM")
-    print("\nPlus all default viewer controls (press 'i' to see them)\n")
-    print("Initial hover RPM:", controller.thrust)
 
     # Run simulation
     try:
-        while True:
+        while is_running:
             # Update and apply RPMs based on current direction
             rpms = controller.update_rpms()
-            print("Current RPMs:", rpms)
             drone.set_propellels_rpm(rpms)
 
             # Step simulation
--- a/examples/keyboard_teleop.py
+++ b/examples/keyboard_teleop.py
@@ -15,6 +15,7 @@ esc	- Quit
 Plus all default viewer controls (press 'i' to see them)
 """
 
+import os
 import random
 
 import numpy as np
@@ -164,6 +165,9 @@ if __name__ == "__main__":
             robot.control_dofs_position(q[:-2], motors_dof)
 
             scene.step()
+
+            if "PYTEST_VERSION" in os.environ:
+                break
     except KeyboardInterrupt:
         gs.logger.info("Simulation interrupted, exiting.")
     finally:


duburcqa:
This feature looks very convenient!

Milotrince:
- Should we keep under `genesis/ext/pyrender/interaction/plugins/` or `genesis/vis/plugins` ?
- Do we want to use Vec3, Quat, Color, Ray, etc. classes in the codebase?
- now that pyrender viewer mouse/keyboard interaction is exposed, we can get rid of `pynput` in examples

duburcqa:
> Should we keep under genesis/ext/pyrender/interaction/plugins/ or genesis/vis/plugins ?

Do you consider this as a "universal" plugin or something pyrender-specific? This would give you the answer :)

> Do we want to use Vec3, Quat, Color, Ray, etc. classes in the codebase?

I'm not a fan of these objects. Overly complicated and not very pythonic in my view.

> now that pyrender viewer mouse/keyboard interaction is exposed, we can get rid of pynput in examples

That would be great! But we need a way to expose this in a clean way.

duburcqa:
@Milotrince If you could finish this PR I would be happy to review it and merge it :)

Milotrince:
@duburcqa Could I get your opinion on the state of the PR? there are some design decisions I'd like your input on!
- removed pynput, see examples/* changes for new usage
- I kept the mouse spring interaction implementation and moved it as a plugin
- there existed a register key system in pyrender viewer but was not exposed in genesis viewer. I made a Keybind/Keybindings class and exposed register_keybind to viewer

I still need to add tests.

duburcqa:
> Could I get your opinion on the state of the PR? there are some design decisions I'd like your input on!

Sure ! I will do it this week ! Could you fix the conflicts in the meantime? Thank you :)

duburcqa:
Small comment on snapshots PR: https://huggingface.co/datasets/Genesis-Intelligence/snapshots/discussions/6

examples/drone/interactive_drone.py
```diff
@@ -145,14 +75,41 @@ def main():
...
-    # Start keyboard listener.
-    # Note that instantiating the listener after building the scene causes segfault on MacOS.
-    listener = keyboard.Listener(on_press=controller.on_press, on_release=controller.on_release)
-    listener.start()
-
     # Build scene
     scene.build()
 
+    # Register keybindings
+    from pyglet.window import key
```
duburcqa: Could you add `from pyglet.window import key` in your new `keybindings` module, so that users can do:
```python
from genesis.ext.pyrender.interaction.keybindings import key
```

This would avoid breaking encapsulation of `pyglet`, that should be an implementation details for users.


Similarly, it would be great to expose your keybindings module outside `ext`, that is also supposed to be an implementation details for the end user. This would be relaying not the same mechanism I just mentioned. Ideally the user should be able to replace `genesis.ext.pyrender.interaction.keybindings` by `genesis.vis.keybindings`

duburcqa:
Could you split the introduction of `ViewerPlugin` feature, with the ability to interactively add markers on geometries from the viewer? Because this PR is getting quite bug already and both features are quite orthogonal. It would make my life easier for the review.

You can open a (stacked) PR for the second part right away, merging on top of this branch instead of main (with draft status).

examples/drone/interactive_drone.py
```diff
@@ -1,112 +1,41 @@
...
-        # Update physics
-        scene.step(refresh_visualizer=False)
-
-        # Limit simulation rate
-        time.sleep(1.0 / scene.viewer.max_FPS)
+    def accelerate(self):
+        """All rotors spin faster"""
+        self.thrust += self.thrust_delta
+        self.rpms = [self.thrust] * 4
+        self.rpms = np.clip(self.rpms, 0, 25000)
```
duburcqa: Are you sure about your formula?

This is a bit dirty. It would be more clean to allocate rpm once (ndarray) and update its value by reference. By the way, there is no need to clip `self.thrust`? Here you keep incrementing something that is out of range, which is weird, because the user won't be able to decelerate for a while if accelerating too much beyond the limit.

duburcqa: Assuming the formula is correct:
```python
self.thrust = min(max(self.thrust + self.thrust_delta, 0), 25000)
self.rpms[:] = self.thrust
```

duburcqa: Similarly, it is really relevant to store both thrust and rpms if they are both the same but expended? Am i missing something?

examples/drone/interactive_drone.py
```diff
@@ -145,14 +75,41 @@ def main():
...
+        ]
+
+    scene.viewer.register_keybinds(
+        *direction_keybinds(key.UP, "move_forward", (1.0, 1.0, -1.0, -1.0)),
+        *direction_keybinds(key.DOWN, "move_backward", (-1.0, -1.0, 1.0, 1.0)),
+        *direction_keybinds(key.LEFT, "move_left", (-1.0, 1.0, -1.0, 1.0)),
+        *direction_keybinds(key.RIGHT, "move_right", (1.0, -1.0, 1.0, -1.0)),
+        Keybind(key.SPACE, KeyAction.HOLD, name="accelerate", callback=controller.accelerate),
+        Keybind(key.LSHIFT, KeyAction.HOLD, name="decelerate", callback=controller.decelerate),
+    )
```
duburcqa, line 110: it would be nice to have a way to detect conflicting key bindings with the default viewer controls, then raise an exception if any, or just a warning eventually.

examples/IPC_Solver/ipc_arm_cloth.py
```diff
@@ -171,14 +147,14 @@ def build_scene(use_ipc=False, show_viewer=False, enable_ipc_gui=False):
...
+def run_sim(scene, entities, mode="interactive", trajectory_file=None):
     robot = entities["robot"]
     target_entity = entities["target"]
 
     robot_init_pos = np.array([0.5, 0, 0.55])
     robot_init_R = R.from_euler("y", np.pi)
-    target_pos = robot_init_pos.copy()
-    target_R = robot_init_R
+    target_pos = [robot_init_pos.copy()]  # Use list for mutability in closures
+    target_R = [robot_init_R]  # Use list for mutability in closures
```
duburcqa: I don't understand. numpy arrays are already mutable. You can modify then in place. So doing this should not be necessary.

examples/IPC_Solver/ipc_arm_cloth.py
```diff
@@ -189,18 +165,56 @@ def run_sim(scene, entities, clients, mode="interactive", trajectory_file=None):
...
+
     def reset_scene():
-        nonlocal target_pos, target_R
-        target_pos = robot_init_pos.copy()
-        target_R = robot_init_R
-        target_quat = target_R.as_quat(scalar_first=True)
-        target_entity.set_qpos(np.concatenate([target_pos, target_quat]))
-        q = robot.inverse_kinematics(link=ee_link, pos=target_pos, quat=target_quat)
+        target_pos[0] = robot_init_pos.copy()
+        target_R[0] = robot_init_R
```
duburcqa: Just update them in place: `target_pos[:] = robot_init_pos`, `target_R[:] = robot_init_R`

examples/IPC_Solver/ipc_arm_cloth.py
```diff
@@ -14,45 +14,21 @@
...
-import numpy as np
 import csv
 import os
 from datetime import datetime
-from pynput import keyboard
-from scipy.spatial.transform import Rotation as R
+
+import numpy as np
 from huggingface_hub import snapshot_download
+from scipy.spatial.transform import Rotation as R
```
duburcqa: The use of scipy should be avoided if possible, but it was already there before your PR.

examples/IPC_Solver/ipc_arm_cloth.py
```diff
@@ -248,99 +262,57 @@ def reset_scene():
...
+        print("\nPlus all default viewer controls (press 'i' to see them)")
 
     # reset scene before starting teleoperation
     reset_scene()
 
     # start teleoperation or playback
-    stop = False
     step_count = 0
 
-    while not stop:
```
duburcqa, line 260: Why the stop mechanism had to be removed in favor of try-except?

examples/sensors/lidar_teleop.py
```diff
@@ -193,48 +136,79 @@ def apply_pose_to_all_envs(pos_np: np.ndarray, quat_np: np.ndarray):
...
+        target_euler[1] += KEY_DANGLE
+
+    def pitch_down():
+        target_euler[1] -= KEY_DANGLE
+
+    def yaw_ccw():
+        target_euler[2] += KEY_DANGLE
+
+    def yaw_cw():
+        target_euler[2] -= KEY_DANGLE
```
duburcqa: This is very verbose for little benefit. What about only exposing `translate`, `rotate`, with some "axis: int, is_negative: bool" argument or just `axis: np.ndarray`, and use `partial` to keybind them? That would be more flexible and simpler I think.

examples/viewer_plugin/mesh_point_selector.py
duburcqa, line 1: Could we move this feature in a separate PR?

examples/viewer_plugin/mouse_spring.py
duburcqa, line 1: Same.

examples/keyboard_teleop.py
```diff
@@ -113,115 +84,83 @@ def build_scene():
...
 
-    def reset_scene():
-        nonlocal target_pos, target_R
-        target_pos = robot_init_pos.copy()
-        target_R = robot_init_R
-        target_quat = target_R.as_quat(scalar_first=True)
-        target_entity.set_qpos(np.concatenate([target_pos, target_quat]))
+    # Initialize target pose
+    target_pos = robot_init_pos.copy()
+    target_R = [robot_init_R]  # Use list to make it mutable in closures
```
duburcqa: Same as before. No need to use list encapsulation. Just modify in-place the pre-allocated array.

examples/keyboard_teleop.py
```diff
@@ -113,115 +84,83 @@ def build_scene():
...
+        # Randomize cube position
+        cube.set_pos((random.uniform(0.2, 0.4), random.uniform(-0.2, 0.2), 0.05))
+        cube.set_quat(R.from_euler("z", random.uniform(0, np.pi * 2)).as_quat(scalar_first=True))
+
+    # Initialize robot pose
+    reset_robot()
+
+    # Robot teleoperation callback functions
+    def move(dpos: tuple[float, float, float]):
+        global target_pos
```
duburcqa: Remove this global line entirely. It is not needed and probably confusing for users.

genesis/engine/sensors/raycaster.py
```diff
@@ -195,8 +85,7 @@ def kernel_cast_rays(
         ray_dir_local = ti.math.vec3(ray_directions[i_p, 0], ray_directions[i_p, 1], ray_directions[i_p, 2])
         ray_direction_world = ti_normalize(ti_transform_by_quat(ray_dir_local, link_quat), gs.EPS)
 
-        # --- 2. BVH Traversal ---
-        # FIXME: this duplicates the logic in LBVH.query() which also does traversal
```
duburcqa, line 199: Why FIXME disappeared? You actually implemented the fix?!

examples/drone/interactive_drone.py
Milotrince: leaving a note here: on mac (untested with other machines), when i `import pyglet` or `pyglet.window.key` at the top level before viewer does, I get
```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[NSApplication macOSVersion]: unrecognized selector sent to instance 0x1618b7df0'

*** First throw call stack:
(
        0   CoreFoundation                      0x0000000198f498ec __exceptionPreprocess + 176
        1   libobjc.A.dylib                     0x0000000198a22418 objc_exception_throw + 88
        2   CoreFoundation                      0x000000019902b7e8 +[NSObject(NSObject) instanceMethodSignatureForSelector:] + 0
...
        87  python3.12                          0x000000010436e1f0 builtin_exec + 404
        88  python3.12                          0x00000001042bb45c cfunction_vectorcall_FASTCALL_KEYWORDS + 92
        89  python3.12                     
libc++abi: terminating due to uncaught exception of type NSException
```
By claude's suggestion I added it as a lazily loading module to safely expose  `key` at `genesis.vis.keybindings`.

examples/drone/interactive_drone.py
Milotrince: yes, this is in `Keybindings.register()`

examples/drone/interactive_drone.py
Milotrince: the use of thrust and rpms is how it was before. I'll update it based on your suggestions!

examples/IPC_Solver/ipc_arm_cloth.py
Milotrince: replaced with `genesis.geom.utils as gu`!

examples/IPC_Solver/ipc_arm_cloth.py
Milotrince: my bad, leftover codingagent slop

genesis/engine/sensors/raycaster.py
Milotrince: I removed it because I think the FIXME note isn't quite accurate, `raycaster.py` does ray-AABB tests while `LBVH.query()` checks AABB-AABB. There might be a bit of duplicated logic but i think it'd be more complicated to split it up into smaller `ti.func`s so may as well leave as is.

Milotrince:
> Could you split the introduction of `ViewerPlugin` feature, with the ability to interactively add markers on geometries from the viewer? Because this PR is getting quite bug already and both features are quite orthogonal. It would make my life easier for the review.

Ok, I removed the `mouse_spring` and `mesh_point_selector` plugins. I am keeping the base `ViewerPlugin` since I interpreted the default viewer controls as a kind of `ViewerPlugin`. To disable the default viewer key interactions, one would specify `gs.options.ViewerOptions(viewer_plugin=None)` (as opposed to the default `gs.options.ViewerOptions(viewer_plugin=DefaultControlsPlugin)`). If this seems strange we can put the default controls back onto `Viewer`.

> You can open a (stacked) PR for the second part right away, merging on top of this branch instead of main (with draft status).

I can't directly open a stacked PR on the Genesis repo since the parent branch lives on my forked repo. I opened one on my fork for sake of preview https://github.com/Milotrince/Genesis/pull/3/changes

examples/drone/interactive_drone.py
```diff
@@ -1,112 +1,36 @@
...
-        time.sleep(1.0 / scene.viewer.max_FPS)
-
-        if "PYTEST_VERSION" in os.environ:
-            break
+        self.cur_dir = np.array([0.0, 0.0, 0.0, 0.0])  # rotor directions
+
+    def update_rpms(self):
+        """Compute RPMs based on current direction and thrust"""
+        clipped_dir = np.clip(self.cur_dir, -1.0, 1.0)
+        print(clipped_dir)
```
YilingQiao: delete this

genesis/vis/viewer.py
```diff
@@ -40,16 +40,12 @@ def __init__(self, options: "ViewerOptions", context):
         self._camera_init_lookat = np.asarray(options.camera_lookat, dtype=gs.np_float)
         self._camera_up = np.asarray(options.camera_up, dtype=gs.np_float)
         self._camera_fov = options.camera_fov
-        self._enable_interaction = options.enable_interaction
```
YilingQiao, line 43: how should we disable interaction with the UI? Some users need to customize they keyboard binding outside genesis

genesis/utils/raycast.py
```diff
@@ -0,0 +1,121 @@
...
+    # Check if ray intersects AABB
+    if t_near <= t_far:
+        result = t_near
+
+    return result
+
+
+@ti.kernel
+def kernel_update_aabbs(
+    free_verts_state: ti.template(),
```
YilingQiao: This kernel won't work with NDArray on my side.
Try `python examples/sensors/lidar_teleop.py`
Can you confirm on your side the above example still works?
If it doesn't, we should probably raise an exception in any function that calls this kernel. We can also check whether it is using ti.field (by setting performance_mode = True or setting the environment variable GS_ENABLE_NDARRAY=0).

YilingQiao:
Thank you for the PR! After addressing those comments, I think I'm good with this PR.

genesis/vis/viewer.py
Milotrince: > To disable the default viewer key interactions, one would specify `gs.options.ViewerOptions(viewer_plugin=None)` (as opposed to the default `gs.options.ViewerOptions(viewer_plugin=DefaultControlsPlugin)`).

genesis/utils/raycast.py
Milotrince: i updated this to use `array_class`!

examples/drone/interactive_drone.py
```diff
@@ -1,112 +1,35 @@
...
 import genesis as gs
+from genesis.vis.keybindings import Key, KeyAction, Keybind
 
 
 class DroneController:
     def __init__(self):
-        self.thrust = 14475.8  # Base hover RPM - constant hover
-        self.rotation_delta = 200.0  # Differential RPM for rotation
+        self.thrust = 14475.8  # Base RPM for constant hover
+        self.rotation_delta = 100.0  # Differential RPM for rotation
```
duburcqa, line 12: Are you sure about this change? If you tested it and it works fine then I'm ok with it.

examples/IPC_Solver/ipc_arm_cloth.py
duburcqa, line 1: Did you validate that it works fine for all possible user arguments that may be impacted by your changes?

```python
    parser.add_argument("--ipc", action="store_true", default=False, help="Enable IPC coupling")
    parser.add_argument(
        "--mode",
        type=str,
        default="playback",
        choices=["interactive", "record", "playback"],
        help="Mode: interactive, record (save trajectory), or playback (replay trajectory, default)",
    )
    parser.add_argument(
        "--trajectory",
        type=str,
        default="grasp_cloth1.csv",
        help="Trajectory file to load (for playback mode, default: grasp_cloth1.csv)",
    )
    parser.add_argument("--list", action="store_true", help="List available trajectories and exit")
    parser.add_argument("-v", "--vis", action="store_true", default=False, help="Show Genesis viewer")
    parser.add_argument("--vis_ipc", action="store_true", default=False, help="Show IPC GUI")
```

genesis/ext/pyrender/interaction/plugins/default_controls.py
```diff
@@ -0,0 +1,159 @@
...
+
+if TYPE_CHECKING:
+    from genesis.engine.scene import Scene
+    from genesis.ext.pyrender.node import Node
+
+INSTR_KEYBIND_NAME = "toggle_instructions"
+
+
+@register_viewer_plugin(DefaultControlsOptions)
+class DefaultControls(HelpTextPlugin):
```
duburcqa: We need a registration mechanism of this? It is specific to pyrender now? Why would someone need to implement its own plugin? Moreover, why do you need HelpTextPlugin + DefaultControls with some inheritance pattern here? It seems this options are only ever going to be used for text help anyway?

I think it is great to have some options to customise help text (and ideally it would be nice to have a mode to disable text overlay because it is surprising slow to render and kind of annoying when doing screen recording)

genesis/ext/pyrender/interaction/plugins/help_text.py
duburcqa, line 1: As mentioned in previous comment, I'm not sure we need both default controls and help text. It seems they could be merged as one class. And the use case for registration is not very clear. If it is possible to disable default controls completely, then combined with register_keybindings method we have all we need?

genesis/ext/pyrender/interaction/raycaster.py
```diff
@@ -0,0 +1,282 @@
...
+class RayHit(NamedTuple):
+    """Ray hit result with distance, position, normal, and optional geometry."""
+
+    distance: float
+    position: np.ndarray  # (3,)
+    normal: np.ndarray  # (3,)
+    geom: "RigidGeom | None"
+
+
+@ti.kernel
```
duburcqa: It is compatible with fast cache I think? (ie it is stateless)
`@ti.kernel(fastcache=gs.use_fastcache)`

genesis/ext/pyrender/interaction/raycaster.py
duburcqa, line 1: Could you clarify in comment on top of this file about this differs from all the raycasting kernels that have been implemented for raycast sensors and why both are needed without more code sharing? And why it is so specific to the viewer so that this module is stored under `genesis/ext/pyrender/interaction` instead of `genesis.utils.raycast`?

genesis/ext/pyrender/interaction/raycaster.py
```diff
@@ -0,0 +1,282 @@
...
+    Returns:
+        result[0]: distance to hit point (NO_HIT_DISTANCE if no hit)
+        result[1]: geom_idx of hit geometry
+        result[2]: hit_point x coordinate
+        result[3]: hit_point y coordinate
+        result[4]: hit_point z coordinate
+        result[5]: normal x coordinate
+        result[6]: normal y coordinate
+        result[7]: normal z coordinate
+        result[8]: env_idx of hit environment
```
duburcqa: It is much slower to compute when using a dataclass (see `array_class.py`)? I think that just retrieving the result (I mean converting taichi tensor(s) to numpy / torch) should not be slower with dataclass than a single vector now that efficient zero-copy memory sharing has been introduced.

duburcqa: Because I'm afraid that even if you rigorously documented what `result` is, it still adds complexity and should be avoided if it does not provide (necessary) performance benefits.

genesis/ext/pyrender/interaction/plugins/help_text.py
duburcqa: Note that I'm not against it. Just trying to avoid unnecessary complexity. If all of this is necessary then I'm ok of course.

duburcqa:
> Could I get your opinion on the state of the PR? there are some design decisions I'd like your input on!

Overall, I really like what you are doing with this PR. I think it is going in the right direction and is more future proof than what we currently have.

examples/drone/interactive_drone.py
duburcqa: This error is very strange... I think the lazy loading is a perfectly acceptable solution if necessary. Just make sure to add a comment to remember why we did this.

duburcqa:
> To disable the default viewer key interactions, one would specify gs.options.ViewerOptions(viewer_plugin=None) (as opposed to the default gs.options.ViewerOptions(viewer_plugin=DefaultControlsPlugin)). 

Overall, this is a good idea I think. But it could be nice to provide a list of plugins, and make it empty to disable them all, no? Moreover, I would recommend to keep the options `disable_keyboard_shortcuts` and `enable_interaction` for convenience, but make them append/remove plugins from the list under the hood.

genesis/ext/pyrender/interaction/plugins/help_text.py
duburcqa: Ok all of this was not very fresh in my mind. It may actually be the right design from the start. What is your opinion @Milotrince ? Do you think there is any level of abstraction that would be removed? I think the plugin mechanism is interesting after all, but what about merging DefaultControls and HelpTextPlugin as one class without inheritance?

examples/drone/interactive_drone.py
Milotrince: Yeah, I found that controlling the drone felt unstable with 200 delta but more reasonable at 100

genesis/ext/pyrender/interaction/plugins/help_text.py
Milotrince: Use case 1: someone doesnt want the default controls, but registers their own keybinds. Setting `viewer_plugin=HelpTextPlugin()` means they can still see the `[i] toggle instructions` for their registered keys.
Use case 2: a `CustomViewerPlugin(HelpTextPlugin)` 

but this can be achievable with a different design as well.
e.g. `DefaultControls(include_default_keybinds=False)`

genesis/ext/pyrender/interaction/plugins/default_controls.py
Milotrince: I thought the registration mechanism might be most consistent with other usage, but we could also do something like
`scene.add_viewer_plugin(ViewerPlugin())` without options/registration

genesis/ext/pyrender/interaction/raycaster.py
Milotrince: my bad, I should have left the raycaster changes for the next PR. It is for screen point to scene raycasting (adrian's MouseSpring had a slow version of this).

Milotrince:
> Overall, this is a good idea I think. But it could be nice to provide a list of plugins, and make it empty to disable them all, no? Moreover, I would recommend to keep the options disable_keyboard_shortcuts and enable_interaction for convenience, but make them append/remove plugins from the list under the hood.

ah yeah i think a list of ViewerPlugin objects would make more sense. Which do we prefer for the usage, the options/registration pattern or `scene.add_viewer_plugin(ViewerPlugin())`, @duburcqa @YilingQiao  ?

duburcqa:
What does `add_viewer_plugin` exactly? Only register or enable a given plugin?

Milotrince:
> What does add_viewer_plugin exactly? Only register or enable a given plugin?

how about this: Viewer stores a list of plugins, e.g. `[DefaultControlsPlugin(), HelpTextPlugin()]`, calling `add_viewer_plugin(CustomPlugin)` enables new plugin by adding it to the list `[DefaultControlsPlugin(), HelpTextPlugin(), CustomPlugin()]`. The default ones can be disabled with option flags `disable_default_keybinds` and `disable_help_text`

duburcqa:
> how about this: Viewer stores a list of plugins, e.g [...]

I think this mechanism is more natural and less disruptive compared to what we have now. What do you think @Milotrince ?

duburcqa:
@Milotrince CI is failing!

Milotrince:
updated snapshot pr: https://huggingface.co/datasets/Genesis-Intelligence/snapshots/discussions/7
posting the "stacked pr" here again (raycaster changes moved here) https://github.com/Milotrince/Genesis/pull/3

genesis/vis/keybindings.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def __dir__(self):
+        if type(self)._module is None:
+            from pyglet.window import key as pyglet_key
+
+            type(self)._module = pyglet_key
+        return dir(type(self)._module)
+
+
+Key = _LazyPygletKeyModule()
```
duburcqa: I would recommend to just `from pyglet.window import key as pyglet_key` on top of every single method that is calling `pyglet_key`. Another option would be to make `pyglet_key` an attribute of KeyBindings class at __init__ (via `self.pyglet_key = pyglet_key`).

duburcqa: Ok I see why you need this. Because this part is user facing... What about exposing our own enum class for keys, that we will internally convert to pyglet. Breaking pyglet encapsulation by asking users to pass pyglet enum is problematic. First, we do not want to get struck with pyglet backend, in the future we may extend this to more backends. Second, the users may try to import pyglet themselves, which would cause segfault.

genesis/vis/viewer.py
```diff
@@ -40,16 +41,18 @@ def __init__(self, options: "ViewerOptions", context):
...
-        self._disable_keyboard_shortcuts = options.disable_keyboard_shortcuts
+
+        self._disable_help_text = options.disable_help_text
+        self._viewer_plugins: list["ViewerPlugin"] = []
+        if not options.disable_default_keybinds:
+            from genesis.vis.viewer_plugins import DefaultControlsPlugin
+
+            self._viewer_plugins.append(DefaultControlsPlugin())
```
duburcqa: Why is it necessary to delay import?

tests/test_viewer.py
duburcqa, line 1: I just merge your snapshot PR (probably a bit too fast but it is not a big deal). Did you remove old now relevant screenshots if any?

Disabling shadows is not a good unit test, because it is not enabled in the first place if no GPU is detected. Your wireframe and disable shadow systematically (see other unit tests doing this).

genesis/vis/viewer.py
Milotrince: it's not necessary, moved it up!

genesis/vis/keybindings.py
Milotrince: how about defining Key as our own enum, with int values compatible with pyglet's? (see [commit](https://github.com/Milotrince/Genesis/blob/864e36b9bde50fa64fdfbe2ebfe9947f6a82b4af/genesis/vis/keybindings.py))

tests/test_viewer.py
Milotrince: i updated this test to not use snapshot and just check if the internal boolean is toggled is that ok?

duburcqa: I guess so! Even better. We need to revert snapshot PR I guess?

YilingQiao:
Look forward to landing this PR. I created a GUI on top of that

https://github.com/user-attachments/assets/ba4459a5-788c-4bda-8457-ed1ef5703b16

Milotrince:
@duburcqa some tests are failing from `importlib.metadata.PackageNotFoundError: No package metadata was found for gstaichi` do i need to worry about this?

genesis/vis/keybindings.py
duburcqa: > how about defining Key as our own enum, with int values compatible with pyglet's?

Yes! this is exactly what I meant by: 

> What about exposing our own enum class for keys, that we will internally convert to pyglet.

duburcqa:
> @duburcqa some tests are failing from importlib.metadata.PackageNotFoundError: No package metadata was found for gstaichi do i need to worry about this?

This is concerning. We should fix this before merging. Weird.

EDIT: the only failure I could find on the CI is this one:

image

EDIT2: Ok I found the bug!

tests/test_viewer.py
duburcqa: Let's do this in a follow up PR!

.github/workflows/examples.yml
```diff
@@ -54,7 +54,6 @@ jobs:
         run: |
           python -m pip install --upgrade pip setuptools wheel
           pip install torch --index-url https://download.pytorch.org/whl/cpu
-          pip install -e '.[dev,usd]' pynput
```
duburcqa, line 57: `pip install -e '.[dev,usd]'`

realmtai:
I'm trying to `from genesis.vis.keybindings import Key, KeyAction, Keybind`

and getting `ModuleNotFoundError: No module named 'genesis.vis.keybindings'`

- This is due to `gensis-world 0.3.13` does not have the keybindings yet? and will most likely make it into a later release

duburcqa:
> This is due to gensis-world 0.3.13 does not have the keybindings yet?

Yes, you need to install Genesis main branch. It will be part of 0.3.14 release.

PR #2005, opened 2025-11-18
[MISC] Fix car emoji in Genesis welcome message.

zhouxian:


duburcqa:
I don't have this issue locally:
image

zhouxian:
Interesting. It's an issue on both yiling and my linux desktop, as well as my M1 mac.
Let's change the emoji to some more standard one?
image

Issue #1978, opened 2025-11-13
[Bug]: pdoc error in genesis import


mstoisa:
### Bug Description

# pdoc error #
Using [pdoc](https://pdoc.dev/docs/pdoc.html) to generate documentation of a package that imports genesis causes an error.
### cause ###
pdoc imports packages to read docs/annotations/typing and produce documentation. The error occurs because `sys.stderr` is accessed in [genesis.utils.misc.redirect_libc_stderr](https://github.com/Genesis-Embodied-AI/Genesis/blob/main/genesis/utils/misc.py#L49), but it is not a valid file descriptor under pdoc execution.

### Steps to Reproduce

`uv run pdoc my_package_that_imports_genesis/`

### Expected Behavior

During import of genesis package, the access to `sys.stderr`should not be assumed, or should be protected by try/except mechanism.

### Relevant log output

```shell
> uv run pdoc ./test_genesis_import/

[...]

Warn: Couldn't import test_genesis_import.test_genesis_import:
Traceback (most recent call last):
  File "/home/me/my_package/.venv/lib/python3.12/site-packages/pdoc/extract.py", line 220, in load_module
    return importlib.import_module(module)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/me/.local/share/uv/python/cpython-3.12.10-linux-x86_64-gnu/lib/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 999, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "/home/me/my_package/test_genesis_import/test_genesis_import.py", line 1, in <module>
    import genesis
  File "/home/me/my_package/.venv/lib/python3.12/site-packages/genesis/__init__.py", line 389, in <module>
    with open(os.devnull, "w") as stderr, redirect_libc_stderr(stderr):
[...]

PR #2011, opened 2025-11-19
[FEATURE] Leverage GsTaichi zero-copy in data accessors.

duburcqa:
Resolves https://github.com/Genesis-Embodied-AI/Genesis/issues/1978
Resolves https://github.com/Genesis-Embodied-AI/Genesis/discussions/2016

### DIFF
--- a/genesis/__init__.py
+++ b/genesis/__init__.py
@@ -41,6 +41,7 @@ device: torch.device | None = None
 backend: gs_backend | None = None
 use_ndarray: bool | None = None
 use_fastcache: bool | None = None
+use_zerocopy: bool | None = None
 EPS: float | None = None
 
 
@@ -117,7 +118,7 @@ def init(
         backend = gs_backend.cpu
 
     # Configure GsTaichi fast cache and array type
-    global use_ndarray, use_fastcache
+    global use_ndarray, use_fastcache, use_zerocopy
     is_ndarray_disabled = (os.environ.get("GS_ENABLE_NDARRAY") or ("0" if backend == gs_backend.metal else "1")) == "0"
     if use_ndarray is None:
         _use_ndarray = not (is_ndarray_disabled or performance_mode)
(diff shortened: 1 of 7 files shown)


YilingQiao:
Any methods to test the performance impact of this PR?

genesis/__init__.py
```diff
@@ -136,6 +137,15 @@ def init(
             raise_exception("Genesis previous initialized. GsTaichi fast cache mode cannot be disabled anymore.")
     use_ndarray, use_fastcache = _use_ndarray, _use_fastcache
 
+    # Unlike dynamic vs static array mode, and fastcache, zero-copy can be toggle on/off between init without issue
+    _use_zerocopy = int(os.environ["GS_ENABLE_ZEROCOPY"]) if "GS_ENABLE_ZEROCOPY" in os.environ else None
+    if use_ndarray and backend in (gs_backend.cpu, gs_backend.cuda):
+        if _use_zerocopy is None:
+            _use_zerocopy = True
+    elif _use_zerocopy:
+        raise_exception(f"Zero-copy only support by GsTaichi dynamic array mode on CPU and CUDA backend.")
```
YilingQiao: ti.field won't support zero copy?

YilingQiao (approved):

genesis/__init__.py
duburcqa: Yes ! It will, but not right now, due to a bug in torch itself. It has been fixed by @hughperkins but it is in review, then if it gets merged we may be able to use it with the next torch release in 3 months x)

YilingQiao: I thought that's bug only for Mac? If it is, can we turn it on for other platform?

duburcqa: No it is not a bug related to Mac OS in particular.

hughperkins: I'm not sure I follow. Field works except on Metal.

hughperkins: The reason it doesnt work on metal is that torch expects a MTLBuffer object, and you can't do pointer arithmetic on that.

On CPU, we get a raw memory pointer, and can do pointer arithmetic.

On CUDA, we get a pointer that behaves like a raw memory pointer, and can do pointer arithmetic.

(of course, in reality, these addresses are likely all virtual memory addresses, but either way, they allow pointer arithmetic).

duburcqa: Ok I see. Still, I will disable zero-copy for field for now, because it seems to be buggy. Better safe than sorry. I will investigate.

PR #2025, opened 2025-11-24
[FEATURE] Add support of zero-copy to 'set_qpos' and 'set_dofs_velocity'.

duburcqa:
* Improve numerical stability of tower.
* Do not clear error code in 'set_qpos' and 'set_dofs_position'.
* Add support of zero-copy to 'set_dofs_velocity'.
* Skip velocity update if not necessary.
* Add support of zero-copy to 'set_qpos'.
* Do not expose 'skip_forward' option for setters that are discouraged in hot path.
* Add 'set_qpos', 'set_dofs_velocity' to accessor benchmark.

### DIFF
--- a/tests/test_rigid_benchmarks.py
+++ b/tests/test_rigid_benchmarks.py
@@ -420,7 +420,18 @@ def _batched_franka(solver, n_envs, gjk, is_collision_free, accessors):
     if n_envs > 0:
         ctrl = torch.tile(ctrl, (n_envs, 1))
     if is_collision_free:
+        franka.set_qpos(ctrl)
         franka.control_dofs_position(ctrl)
+
+    vel0 = torch.zeros((franka.n_qs,), dtype=gs.tc_float, device=gs.device)
+    if n_envs > 0:
+        n_reset_envs = int(0.02 * n_envs)
+        reset_envs_idx = torch.randperm(n_envs)[:n_reset_envs]
+        vel0 = torch.tile(vel0, (n_reset_envs, 1))
+        qpos0 = ctrl[reset_envs_idx]
+    else:
+        reset_envs_idx = None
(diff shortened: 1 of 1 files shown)


genesis/engine/solvers/rigid/rigid_solver_decomp.py
```diff
@@ -1657,21 +1657,29 @@ def set_geoms_friction_ratio(self, friction_ratio, geoms_idx=None, envs_idx=None
...
-        kernel_set_qpos(qpos, qs_idx, envs_idx, self._rigid_global_info, self._static_rigid_sim_config)
-
-        self._errno[None] = 0
-        self.collider.reset(envs_idx)
-        self.collider.clear(envs_idx)
-        if self.constraint_solver is not None:
-            self.constraint_solver.reset(envs_idx)
-            self.constraint_solver.clear(envs_idx)
+    def set_qpos(
+        self, qpos, qs_idx=None, envs_idx=None, *, skip_forward=False, skip_clear_collisions=False, unsafe=False
```
YilingQiao: when should we set skip_clear_collisions True?

duburcqa: When you know that you will be doing a simulation step right after without accessing contact information in the meantime. At least that was my intention. We may need to double check that it works properly.

duburcqa: Actually I don't think it is a good idea. I will remove this option and rather implement a more efficient version based on zero-copy.

Kashu7100: sounds good to me

YilingQiao (approved):

PR #2087, opened 2025-12-08
[FEATURE] Add support of boolean masking as index.

duburcqa:
* Further optimize geometry utils.
* Check for errno systematically at the beginning of the step.
* Support boolean mask as index.

### DIFF
--- a/genesis/utils/misc.py
+++ b/genesis/utils/misc.py
@@ -646,13 +646,15 @@ def ti_to_python(
 
 
 def indices_to_mask(
-    *indices: Any, keepdim: bool = True, to_torch: bool = True, raise_if_fancy: bool = False
+    *indices: Any, keepdim: bool = True, to_torch: bool = True, boolean_mask: bool = False, raise_if_fancy: bool = False
 ) -> tuple[slice | int | torch.Tensor, ...]:
     """Converts a sequence of slice-like objects into a multi-dimensional mask corresponding to their cross-product.
 
     Args:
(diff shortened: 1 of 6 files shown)


genesis/utils/misc.py
```diff
@@ -646,13 +646,15 @@ def ti_to_python(
...
-    *indices: Any, keepdim: bool = True, to_torch: bool = True, raise_if_fancy: bool = False
+    *indices: Any, keepdim: bool = True, to_torch: bool = True, boolean_mask: bool = False, raise_if_fancy: bool = False
 ) -> tuple[slice | int | torch.Tensor, ...]:
     """Converts a sequence of slice-like objects into a multi-dimensional mask corresponding to their cross-product.
 
     Args:
         keepdim (bool): Whether to keep all dimensions even if masks are integers. Defaults to True.
         to_torch (bool): Whether to force casting collections to torch.Tensor.
+        boolean_mask (bool): Whether boolean mask are supported more must be converted to indices via `torch.nonzero`.
+        raise_if_fancy (bool): Whether fancy indexing is supported for should raise an exception.
```
YilingQiao, line 657: sorry maybe I missed several PRs, what is fancy indexing btw?

duburcqa: https://numpy.org/doc/stable/glossary.html#term-advanced-indexing

duburcqa: Advanced indexing, so-called fancy indexing, consist in passing tensors as index, instead of slice. This involves random memory access, which forces returning a copy when extracting data, and slows down assignment operator.

PR #2100, opened 2025-12-10
[CHANGING][FEATURE] Improve inertial estimation if undefined.

YilingQiao:
### **Problem**
The original implementation computed inertial properties by composing all geometries into a single mesh first, then calculating mass/COM/inertia. This was inefficient and potentially inaccurate for complex geometries with multiple components.

### **Solution**
Rewrote the mass property computation to process each geometry individually, then compose their properties using proper inertial mechanics.

### **Key Changes**

1. **Added `compose_inertial_properties()` helper function** in `genesis/utils/urdf.py`:
   - Composes inertial properties of two bodies using parallel axis theorem
   - Handles center-of-mass calculation and inertia tensor translation

2. **Refactored `rigid_link.py` mass computation**:
   - Process each `RigidGeom`/`RigidVisGeom` individually 
   - Compute per-geometry mass, COM, and inertia in local frame
   - Transform to link frame using `init_pos`/`init_quat`
   - Compose using the helper function
   - Apply material density per-geometry
   - Convexify only when needed for inertia computation

3. **Added comprehensive unit tests** in `tests/test_utils.py`:
   - Tests inertial composition for various scenarios
   - Validates physics correctness (parallel axis theorem)

4. **Fixed mesh construction** to properly create `trimesh.Trimesh` objects from geometry data

### **Fallback Logic**
- **Geoms available**: Use `RigidGeom` list (collision geoms)
- **No geoms, vgeoms available**: Use `RigidVisGeom` list (visual geoms)  
- **Neither available**: Default to mass=1e-3, COM=zeros, inertia=zeros

### **Test Command**
```bash
cd /home/qq/Desktop/github/Genesis && conda activate gs312 && python -m pytest tests/test_utils.py::test_compose_inertial_properties -v
```

### **Validation Command**
```bash
python debug/251209_core_teleop/teleop.py
```

## Checklist:

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_link.py
+++ b/genesis/engine/entities/rigid_entity/rigid_link.py
@@ -163,7 +163,9 @@ class RigidLink(RBC):
                     if not inertia_mesh.is_watertight:
                         inertia_mesh = trimesh.convex.convex_hull(inertia_mesh)
 
-                    # If the volume is negative, it means the mesh is inverted.
+                    # TODO: without this check, some geom will have negative volume
+                    # even after the above convex hull operation.
+                    # tests/test_examples.py::test_example[rigid/terrain_from_mesh.py-None]
                     if inertia_mesh.volume < -gs.EPS:
                         inertia_mesh.invert()
 


tests/test_utils.py
```diff
@@ -435,3 +436,78 @@ def test_fps_tracker():
...
+
+    mass2, com2 = 0.0, np.array([1.0, 0.0, 0.0])
+    inertia2 = np.zeros((3, 3))
+
+    combined_mass, combined_com, combined_inertia = compose_inertial_properties(
+        mass1, com1, inertia1, mass2, com2, inertia2
+    )
+
+    assert_allclose(combined_mass, 0.0, tol=TOL)
+    assert_allclose(combined_com, np.array([0.0, 0.0, 0.0]), tol=TOL)  # Should use com1 when total_mass == 0
```
duburcqa: Just implement a single test case that combines all the effects at once.

genesis/utils/urdf.py
```diff
@@ -419,6 +419,39 @@ def rotate_inertia(I, R):
...
+    Returns:
+        combined_mass: Combined mass
+        combined_com: Combined center of mass (3,) array
+        combined_inertia: Combined inertia tensor (3,3) array
+    """
+    combined_mass = mass1 + mass2
+    if combined_mass > 0:
+        combined_com = (mass1 * com1 + mass2 * com2) / combined_mass
+    else:
+        combined_com = com1.copy()
```
duburcqa: Just raise an exception is both masses are not `> gs.EPS`

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -121,82 +122,88 @@ def _build(self):
...
-                    vert_offset += geom.init_verts.shape[0]
-            elif len(self._vgeoms) > 0:  # use vgeom if there's no geom
-                for vgeom in self._vgeoms:
-                    init_verts.append(gu.transform_by_trans_quat(vgeom.init_vverts, vgeom.init_pos, vgeom.init_quat))
-                    init_faces.append(vgeom.init_vfaces + vert_offset)
-                    vert_offset += vgeom.init_vverts.shape[0]
-            init_verts = np.concatenate(init_verts)
-            init_faces = np.concatenate(init_faces)
-            return trimesh.Trimesh(init_verts, init_faces)
+        print(self.name, self._inertial_mass)
```
duburcqa: Remove print.

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -121,82 +122,88 @@ def _build(self):
...
+            if len(self._geoms) > 0:
+                geom_list = self._geoms
+                get_mesh = lambda geom: trimesh.Trimesh(geom.init_verts, geom.init_faces)
+                get_pos = lambda geom: geom._init_pos
+                get_quat = lambda geom: geom._init_quat
+            elif len(self._vgeoms) > 0:
+                geom_list = self._vgeoms
+                get_mesh = lambda geom: trimesh.Trimesh(geom.init_vverts, geom.init_vfaces)
+                get_pos = lambda geom: geom._init_pos
+                get_quat = lambda geom: geom._init_quat
```
duburcqa: Avoid doing this. Just inline these function calls directly where they are used.

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -121,82 +122,88 @@ def _build(self):
...
             else:
-                if self._init_mesh.is_watertight:
-                    self._inertial_mass = self._init_mesh.volume * self.entity.material.rho
-                else:  # TODO: handle non-watertight mesh
-                    self._inertial_mass = 1.0
+                # Fallback: use default values
+                self._inertial_mass = 1e-3
```
duburcqa: Use 0.0. There is no reason to add some random mass in the kinematic tree.

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -121,82 +122,88 @@ def _build(self):
...
+                geom_list = self._vgeoms
+                get_mesh = lambda geom: trimesh.Trimesh(geom.init_vverts, geom.init_vfaces)
+                get_pos = lambda geom: geom._init_pos
+                get_quat = lambda geom: geom._init_quat
             else:
-                if self._init_mesh.is_watertight:
-                    self._inertial_mass = self._init_mesh.volume * self.entity.material.rho
-                else:  # TODO: handle non-watertight mesh
-                    self._inertial_mass = 1.0
+                # Fallback: use default values
```
duburcqa: This is not a fallback. It is the right solution where there is no geometry attached to a link.

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -121,82 +122,88 @@ def _build(self):
...
-            if len(self._geoms) == 0 and len(self._vgeoms) == 0:
-                self._inertial_mass = 0.0
+        if self._inertial_mass is None or self._inertial_pos is None or self._inertial_i is None:
+            # Determine which geom list to use: geoms first, then vgeoms, then fallback
+            if len(self._geoms) > 0:
+                geom_list = self._geoms
+                get_mesh = lambda geom: trimesh.Trimesh(geom.init_verts, geom.init_faces)
+                get_pos = lambda geom: geom._init_pos
+                get_quat = lambda geom: geom._init_quat
+            elif len(self._vgeoms) > 0:
```
duburcqa, line 130: It is not clear if collision geometry of visual geometry should be used to compute such property. I have the feeling that collision geometries may be very off. There are often rough approximation using primitives, unless they are the output of convex decomposition. You can easily check in metadata of meshes whether they are the output of convex decomposition and prefer them in such a case (or more generally if they are true meshes instead of primitive geometries), otherwise ressort on visual. Another option would be to add some flag in morph to control whether collision or visual should be preferred in such a case, but I would rather avoid that.

I would recommend adding some info message to explain that spatial inertia informations are being deduced from the geometry.

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -121,82 +122,88 @@ def _build(self):
...
+
+                for geom in geom_list:
+                    inertia_mesh = get_mesh(geom)
+                    geom_pos = get_pos(geom)
+                    geom_quat = get_quat(geom)
+
+                    if inertia_mesh.is_watertight:
+                        geom_mass = inertia_mesh.volume * self.entity.material.rho
+                    else:
+                        geom_mass = 1e-3
```
duburcqa: Same. I don't like these random number. Why are you not used the convex hull in such a case?

You should also add a warning saying that inertia information may be wrongly estimated.

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -121,82 +122,88 @@ def _build(self):
...
+
+                    geom_com_local = np.array(inertia_mesh.center_mass, dtype=gs.np_float)
+
+                    if not inertia_mesh.is_watertight:
+                        inertia_mesh = trimesh.convex.convex_hull(inertia_mesh)
+
+                    if inertia_mesh.is_watertight and inertia_mesh.mass > 0:
```
duburcqa: You should just raise an exception is `inertia_mesh.mass < gs.EPS`. No point to support this, it will just cause more issues later on.

duburcqa: This condition is completely irrelevant. A convex hull cannot be non-watertight.

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -121,82 +122,88 @@ def _build(self):
...
+                        geom_inertia_local = inertia_mesh.moment_inertia / inertia_mesh.mass * geom_mass
+
+                    else:
+                        # Approximate with sphere
+                        radius = (
+                            max(inertia_mesh.bounds[1] - inertia_mesh.bounds[0]) / 2.0
+                            if inertia_mesh.bounds is not None
+                            else 0.1
+                        )
+                        geom_inertia_local = 0.4 * geom_mass * radius**2 * np.eye(3)
```
duburcqa: Remove this branch.

genesis/engine/entities/rigid_entity/rigid_link.py
YilingQiao: This code is mainly for mesh entities. Usually the visual and collision geoms differ a lot in carefully robot description like URDF, and the mass for each link are specified.

YilingQiao: Also, can we fix this if we find a real use case or bug. For this PR, I prefer to keep the logic same as we have been using.

duburcqa: I'm not convinced. The old behaviour was garbage in the first place. There is no good reason to keep it. If t is only for mesh, then detect it and raise an exception.

tests/test_utils.py
```diff
@@ -435,3 +436,51 @@ def test_fps_tracker():
...
+    total_mass, total_com, total_inertia = compose_inertial_properties(
+        combined_mass_12, combined_com_12, combined_inertia_12, mass3, com3, inertia3
+    )
+
+    # Total mass = 2.0 + 1.0 + 1.5 = 4.5
+    expected_mass = 4.5
+
+    # Center of mass calculation:
+    # COM_x = (2.0*0 + 1.0*3 + 1.5*0) / 4.5 = 3/4.5 = 0.667
+    # COM_y = (2.0*0 + 1.0*0 + 1.5*2) / 4.5 = 3/4.5 = 0.667
```
duburcqa: Ideally it would be nice to have a different value for x, y, and z.

tests/test_utils.py
```diff
@@ -435,3 +436,51 @@ def test_fps_tracker():
...
+    assert_allclose(total_com, expected_com, tol=TOL)
+
+    # Verify inertia tensor is symmetric and positive definite
+    assert_allclose(total_inertia, total_inertia.T, tol=TOL)  # Symmetric
+    eigenvals = np.linalg.eigvals(total_inertia)
+    assert np.all(eigenvals >= -TOL)  # Positive semi-definite (within tolerance)
+
+    # Verify physical consistency: trace of inertia relates to total mass distribution
+    # (This is a rough sanity check - the exact value depends on the specific configuration)
+    assert total_inertia.trace() > 0  # Should be positive for real physical systems
```
duburcqa: Compute the exact value and check that the entire matrix matches expectation, ideally with some number that can be easily checked by hand but not necessarily.

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -121,83 +122,73 @@ def _build(self):
...
+                    # Transform geom properties to link frame
+                    geom_com_link = gu.transform_by_quat(geom_com_local, geom_quat) + geom_pos
+                    geom_inertia_link = rotate_inertia(geom_inertia_local, gu.quat_to_R(geom_quat))
+
+                    # Compose with accumulated properties
+                    if total_mass < gs.EPS:
+                        # First geom
+                        total_mass = geom_mass
+                        total_com = geom_com_link
+                        total_inertia = geom_inertia_link
```
duburcqa: I'm not convinced it makes sense to have such branching.

genesis/engine/entities/rigid_entity/rigid_link.py
YilingQiao: added log info

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -121,83 +122,77 @@ def _build(self):
...
+                    else:
+                        inertia_mesh = trimesh.Trimesh(geom.init_verts, geom.init_faces)
+
+                    geom_pos = geom._init_pos
+                    geom_quat = geom._init_quat
+
+                    if not inertia_mesh.is_watertight:
+                        inertia_mesh = trimesh.convex.convex_hull(inertia_mesh)
+
+                    # If the volume is negative, it means the mesh is inverted.
```
duburcqa: I don't think this is true.

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -121,83 +122,77 @@ def _build(self):
...
+                    f"link '{self.name}'. Inertial and mass are set to zero."
+                )
+                self._inertial_mass = gs.EPS
+                self._inertial_pos = np.zeros(3, dtype=gs.np_float)
+                self._inertial_i = np.eye(3, dtype=gs.np_float) * gs.EPS
+                geom_list = []
+
+            # Process each geom individually and compose their properties
+            if len(geom_list) > 0:
+                total_mass = gs.EPS
```
duburcqa: Can you add a comment to explain what using EPS mass is necessary? It does not make sense physically so it must be strongly motivated.

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -121,83 +122,77 @@ def _build(self):
...
+                total_com = np.zeros(3, dtype=gs.np_float)
+                total_inertia = np.eye(3, dtype=gs.np_float) * gs.EPS
```
duburcqa, line 151: Same.

genesis/engine/entities/rigid_entity/rigid_link.py
YilingQiao: I'm afraid inv_mass would be nan for a link without geom

YilingQiao: added a comment

tests/test_rigid_physics.py
```diff
@@ -1841,8 +1841,12 @@ def test_frictionloss_advanced(show_viewer, tol):
     assert_allclose(robot.get_AABB()[0, 2], 0.0, tol=2e-4)
     box_pos = box.get_pos()
     assert box_pos[0] > 0.6
-    assert_allclose(box_pos[1:], 0.0, tol=0.02)
-    assert_allclose(box.get_dofs_velocity(), 0.0, tol=5 * tol)
+    # This is to check collision detection is working correctly on metal
+    # The box will collide with the robot and rolling on the ground,
+    # We check whether it's rolling within a reasonable range and not blowing up.
+    # Behavior on mdetial is different from other platforms
```
duburcqa, line 1847: This is due to buggy compiler on Metal.

duburcqa (dismissed):

PR #2104, opened 2025-12-10
[MISC] Further optimize constraint solver GPU block dim.

duburcqa:


YilingQiao (approved):

duburcqa:
It makes contact rich heterogeneous simulation faster, but it is not a viable solution I think, because it is slower in all other cases.

PR #2116, opened 2025-12-15
[BUG FIX] Update Camera Up vector to reduce discontinuities in smooth animations.

Rush2k:
## Description
This PR updates 'up' direction of the camera to prevent discontinuity at singularity points; for example when the camera is pointing down and we need to pick a camera frame.

## Screenshots
Motion of the camera before the fix: Camera is looking down at the same target, traveling along the x-axis
https://github.com/user-attachments/assets/e6cf3e12-d57d-494b-885a-561a6eb25433
And along the y-axis
https://github.com/user-attachments/assets/4476a8e5-e990-473d-ba57-34700d809d7d

Motion of the camera afterwards, without the discontinuity twist in the middle
https://github.com/user-attachments/assets/0b3ec77b-0b8a-43c2-9f2d-21c5fc46c5f1
https://github.com/user-attachments/assets/38179e29-e28e-4934-87aa-dc98d1ac49e3


YilingQiao:
Thank you! Could you please provide a unit test example so it's easier to see what's going wrong in that specific case. And also prevent similar errors in the future?

duburcqa (approved):

duburcqa:
@YilingQiao Let's merge this. I think writing a unit test for it would be quite tricky and it is not really a "feature", more like a convenient property.

PR #2148, opened 2025-12-26
[MISC] Reduce warnings spam on rigid_joint.py

hughperkins:
## Description

Reduce warnings spam on rigid_joint.py, and make more specific

Before:
Screenshot 2025-12-26 at 10 38 37

After:
Screenshot 2025-12-26 at 10 42 50

## Checklist:

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_joint.py
+++ b/genesis/engine/entities/rigid_entity/rigid_joint.py
@@ -270,7 +270,9 @@ class RigidJoint(RBC):
         This property either returns a list, an integer, or None depending on whether the joint has multiple DoFs, a
         single one, or none, respectively.
         """
-        warn_once("This property is deprecated and will be removed in future release. Please use 'dofs_idx' instead.")
+        warn_once(
+            f"dof_idx is deprecated on {__file__} and will be removed in future release. Please use 'dofs_idx' instead."
+        )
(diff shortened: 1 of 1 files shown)


hughperkins:
For the benchmark, some compile times faster, some slower. Seems like noise?

duburcqa:
I’m not ok with this PR because it does not go through our logger facility. I would rather fix the code that triggers these warnings, or just deprecate them completely and throw an error instead.

hughperkins:
> I’m not ok with this PR because it does not go through our logger facility. I would rather fix the code that triggers these warnings, or just deprecate them completely and throw an error instead.

genesis.utils.warn_once does use gs.logger.warning: https://github.com/Genesis-Embodied-AI/Genesis/blob/2d98bbb786e94b3f6c4e7171c87b4ff31ff3ccdf/genesis/utils/warnings.py#L12

duburcqa:
> genesis.utils.warn_once does use gs.logger.warning:

Oh, that's great! I just noticed that I looked different on your screenshot. Still, I'm more in favour or throwing an exception. I don't think hiding legitimate warnings behind `warn_once` that avoid spamming the user is really helping.

duburcqa:
Superseded by https://github.com/Genesis-Embodied-AI/Genesis/pull/2148

PR #2166, opened 2025-12-30
[MISC] Split benchmarks into multiple steps, migrate to uv, mount uv cache.

hughperkins:
## Related Issue

## Checklist:

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -109,41 +109,51 @@ jobs:
           # Checkout full history is required to shallow cloning while mark HEAD as "grafted". This breaks remote
           # tracking thereby making it impossible to detect whether a commit is contained in upstream main.
           fetch-depth: 0
-      - name: Run benchmarks
+      - name: Start container
         run: |
           SLURM_JOB_NAME="$(uuidgen)_$(date +%Y%m%d_%H%M%S)"
-          echo "SLURM_JOB_NAME=${SLURM_JOB_NAME}" >> $GITHUB_ENV
-
+          CONTAINER_NAME="${SLURM_JOB_NAME}"
+          SRUN_CONTAINER_OPTS="--container-name=${CONTAINER_NAME} \
+               --container-mounts=/mnt/data/artifacts:/mnt/data/artifacts,${{ github.workspace }}:/root/workspace,${HOME}/.cache/uv:/root/.cache/uv \
+               --no-container-mount-home \
+               --container-workdir=/root/workspace"
           SLURM_ENV_VARS="NVIDIA_DRIVER_CAPABILITIES=all,BASH_ENV=/root/.bashrc,HF_TOKEN,GS_ENABLE_NDARRAY=${GS_ENABLE_NDARRAY}"
           if [[ "${{ github.repository }}" == 'Genesis-Embodied-AI/Genesis' && "${{ github.ref }}" == 'refs/heads/main' ]] ; then
             SLURM_ENV_VARS="${SLURM_ENV_VARS},WANDB_API_KEY"
           fi
 
-          srun \
-            --container-image="/mnt/data/images/genesis-v${GENESIS_IMAGE_VER}.sqsh" \
-            --container-mounts=\
-          "${HOME}/.venv":/root/.venv,\
-          /mnt/data/artifacts:/mnt/data/artifacts,\
-          "${{ github.workspace }}":/root/workspace \
-            --no-container-mount-home --container-workdir=/root/workspace \
-            --export=${SLURM_ENV_VARS} \
-            --partition=hpc-mid --nodes=1 --gpus=8 --exclusive --time="${TIMEOUT_MINUTES}" \
-            --job-name=${SLURM_JOB_NAME} \
-            bash -e -s << 'EOF'
-              # sudo apt update
-              # sudo apt install -y tmate
-              # tmate -S /tmp/tmate.sock new-session -d
-              # tmate -S /tmp/tmate.sock wait tmate-ready
-              # tmate -S /tmp/tmate.sock display -p '#{tmate_ssh}'
+          JOBID_FIFO="${{ github.workspace }}/.slurm_job_id_fifo"
+          [[ -e "$JOBID_FIFO" ]] && rm -f "$JOBID_FIFO"
+          mkfifo "$JOBID_FIFO"
+          salloc --job-name="${SLURM_JOB_NAME}" \
+                 --partition=hpc-mid --nodes=1 --gpus=8 --exclusive \
+                 --time="${TIMEOUT_MINUTES}" \
+                 bash -c "echo \$SLURM_JOB_ID > $JOBID_FIFO; sleep ${TIMEOUT_MINUTES}m" &
+          SLURM_JOB_ID=$(cat "$JOBID_FIFO")
+          rm -f "$JOBID_FIFO"
+          SRUN_COMMON="--overlap --jobid=${SLURM_JOB_ID} ${SRUN_CONTAINER_OPTS} --export=${SLURM_ENV_VARS}"
+          srun --jobid=${SLURM_JOB_ID} \
+               --container-image=/mnt/data/images/genesis-v${GENESIS_IMAGE_VER}.sqsh \
+               ${SRUN_CONTAINER_OPTS} \
+               --export=${SLURM_ENV_VARS} \
+               echo "Container ready"
 
-              source /root/.venv/bin/activate
-              pip install --no-input ".[dev,render]"
+          echo "SLURM_JOB_NAME=${SLURM_JOB_NAME}" >> $GITHUB_ENV
+          echo "SRUN_COMMON=${SRUN_COMMON}" >> "$GITHUB_ENV"
 
-              pytest --print -x -m "benchmarks" ./tests
-              cat speed_test*.txt > "/mnt/data/artifacts/speed_test_${SLURM_JOB_NAME}.txt"
+      - name: Build
+        run: |
+          srun ${SRUN_COMMON} \
+               bash .github/workflows/scripts/production_build.sh
 
-              # tmate -S /tmp/tmate.sock wait tmate-exit
+      - name: Run benchmarks
+        run: |
+          srun ${SRUN_COMMON} bash -e -s <<'EOF'
+            source /venv/bin/activate
+            pytest --print -x -m "benchmarks" ./tests
+            cat speed_test*.txt > "/mnt/data/artifacts/speed_test_${SLURM_JOB_NAME}.txt"
           EOF
+
       - name: Kill srun job systematically
         if: always()
         run: |
--- a/.github/workflows/scripts/production_build.sh
+++ b/.github/workflows/scripts/production_build.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+set -ex
+
+# sudo apt update
+# sudo apt install -y tmate
+# tmate -S /tmp/tmate.sock new-session -d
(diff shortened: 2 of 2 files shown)


.github/workflows/scripts/production_build.sh
duburcqa, line 1: I would much prefer to keep the workflows self-contained instead of calling external scripts. Much easier for me the get the full picture, and also easier to much sure nobody is tempering with security.

.github/workflows/examples.yml
duburcqa, line 1: You should restore with workflow.

.github/workflows/generic.yml
duburcqa, line 1: You should restore this workflow

.github/workflows/production.yml
```diff
@@ -109,47 +109,115 @@ jobs:
...
+      - name: Test
+        run: |
+          srun ${SRUN_COMMON} bash -e -s <<'EOF'
+            source /venv/bin/activate
+            pytest --print -x -m "benchmarks" ./tests
+            cat speed_test*.txt > "/mnt/data/artifacts/speed_test_${SLURM_JOB_NAME}.txt"
           EOF
-      - name: Kill srun job systematically
+
+      - name: Cleanup allocation
```
duburcqa: Why are you changing this? It was working fine so far.

.github/workflows/production.yml
```diff
@@ -109,47 +109,115 @@ jobs:
...
+          SRUN_INIT="--container-image=/mnt/data/images/genesis-v${GENESIS_IMAGE_VER}.sqsh ${SRUN_CONTAINER_OPTS} --export=${SLURM_ENV_VARS}"
+          
+          # For subsequent commands: use container name
+          SRUN_COMMON_BARE="--jobid=${SLURM_JOB_ID} ${SRUN_CONTAINER_OPTS} --export=${SLURM_ENV_VARS}"
+          echo "SRUN_COMMON_BARE=${SRUN_COMMON_BARE}" >> "$GITHUB_ENV"
+
+          SRUN_COMMON="--overlap ${SRUN_COMMON_BARE}"
+          echo "SRUN_COMMON=${SRUN_COMMON}" >> "$GITHUB_ENV"
+
+          # Start the container and keep it alive
```
duburcqa: Why the logic is so complicated compared to before?

.github/workflows/production.yml
```diff
@@ -109,47 +109,115 @@ jobs:
...
+              echo "ERROR: Container failed to become ready after 60 attempts" >&2
+              exit 1
+            fi
+            echo "Waiting for container to be ready (attempt $i/60)..."
+            sleep 2
+          done
+
+      - name: Build
+        run: |
+          srun ${SRUN_COMMON} \
```
duburcqa, line 146: I’m not a fan of this. Much harder to me to copy paste on the cluster to check if it works.

.github/workflows/production.yml
```diff
@@ -109,47 +109,115 @@ jobs:
...
+                 --partition=hpc-mid --nodes=1 --gpus=8 --exclusive \
+                 --time="${TIMEOUT_MINUTES}" \
+                 bash -c "sleep ${TIMEOUT_MINUTES}m" &
+          echo "SLURM_ALLOC_PID=$!" >> $GITHUB_ENV
+
+          for i in $(seq 1 30); do
+            SLURM_JOB_ID="$(squeue --noheader -o '%A' --name "${SLURM_JOB_NAME}" | head -n1)"
+            test -n "$SLURM_JOB_ID" && break
+            sleep 3
+          done
```
duburcqa: I'm against sleep-based approaches. I would much rather prefer any other approach that would avoid this.

.github/workflows/production.yml
```diff
@@ -109,47 +109,115 @@ jobs:
...
+            exit 1
+          fi
+          echo "SLURM_JOB_ID=${SLURM_JOB_ID}" >> $GITHUB_ENV
+
+          SLURM_NODE="$(squeue --noheader -o '%N' -j "${SLURM_JOB_ID}")"
+          echo "Allocated node: ${SLURM_NODE}"
 
+          # Base container configuration (mounts and workdir)
+          SRUN_CONTAINER_OPTS="--container-name=${CONTAINER_NAME}"
+          SRUN_CONTAINER_OPTS="${SRUN_CONTAINER_OPTS} --container-mounts=/mnt/data/artifacts:/mnt/data/artifacts,${{ github.workspace }}:/root/workspace,${HOME}/.cache/uv:/root/.cache/uv"
```
duburcqa: `/mnt/data/artifacts:/mnt/data/artifacts` should not be added for unit tests. I would rather only add it for benchmarks to reduce the surface of attack.

.github/workflows/production.yml
```diff
@@ -109,47 +109,115 @@ jobs:
...
+          
+          # For initial container start: specify image to create/extract the container
+          SRUN_INIT="--container-image=/mnt/data/images/genesis-v${GENESIS_IMAGE_VER}.sqsh ${SRUN_CONTAINER_OPTS} --export=${SLURM_ENV_VARS}"
+          
+          # For subsequent commands: use container name
+          SRUN_COMMON_BARE="--jobid=${SLURM_JOB_ID} ${SRUN_CONTAINER_OPTS} --export=${SLURM_ENV_VARS}"
+          echo "SRUN_COMMON_BARE=${SRUN_COMMON_BARE}" >> "$GITHUB_ENV"
+
+          SRUN_COMMON="--overlap ${SRUN_COMMON_BARE}"
+          echo "SRUN_COMMON=${SRUN_COMMON}" >> "$GITHUB_ENV"
```
duburcqa: I don't like this approach of splitting every options in 20 envs vars. Much harder to read and copy paste for  me.

.github/workflows/production.yml
```diff
@@ -109,47 +109,115 @@ jobs:
...
+
+          # Start the container and keep it alive
+          srun --jobid=${SLURM_JOB_ID} ${SRUN_INIT} sleep "${TIMEOUT_MINUTES}m" &
+          echo "BASE_SRUN_PID=$!" >> $GITHUB_ENV
+          BASE_SRUN_PID=$!
+
+          # Wait for container extraction/initialization to complete
+          # The srun process will only continue once the container is fully extracted
+          echo "Waiting for container extraction to complete..."
+          sleep 10  # Give it a moment to start
```
duburcqa: Avoid sleep.

.github/workflows/production.yml
```diff
@@ -109,47 +109,115 @@ jobs:
...
+              break
+            fi
+
+            if [ $i -eq 60 ]; then
+              echo "ERROR: Container failed to become ready after 60 attempts" >&2
+              exit 1
+            fi
+            echo "Waiting for container to be ready (attempt $i/60)..."
+            sleep 2
+          done
```
duburcqa: Avoid waiting. Find a way to block on job completion.

.github/workflows/production.yml
```diff
@@ -109,47 +109,115 @@ jobs:
...
+          sleep 2  # Give it a moment to die
+          
+          # Then remove the container
+          if [ -n "${SLURM_JOB_ID}" ] && [ -n "${CONTAINER_NAME}" ]; then
+            srun --jobid=${SLURM_JOB_ID} --overlap enroot remove -f "${CONTAINER_NAME}" 2>/dev/null || true
           fi
+          
+          # Finally cancel the job and kill salloc
+          scancel "${SLURM_JOB_ID}" 2>/dev/null || true
+          kill "${SLURM_ALLOC_PID}" 2>/dev/null || true
```
duburcqa: * Do not do `2>/dev/null`, I want to see the output.

* Do not hide errors via `|| true`. This may explode at our faces.

.github/workflows/production.yml
```diff
@@ -109,47 +109,115 @@ jobs:
...
         run: |
-          if [ -n "${SLURM_JOB_NAME}" ] ; then
-            scancel --user=${USER} --name="${SLURM_JOB_NAME}"
+          # Kill the background sleep process first
+          kill "${BASE_SRUN_PID}" 2>/dev/null || true
+          sleep 2  # Give it a moment to die
+          
+          # Then remove the container
+          if [ -n "${SLURM_JOB_ID}" ] && [ -n "${CONTAINER_NAME}" ]; then
+            srun --jobid=${SLURM_JOB_ID} --overlap enroot remove -f "${CONTAINER_NAME}" 2>/dev/null || true
```
duburcqa: Same, avoid `2>/dev/null || true`

.github/workflows/production.yml
hughperkins: - necessary for benchmarks otherwise we get:
Screenshot 2026-01-02 at 14 57 17

- dont think its added to unit tests?

.github/workflows/production.yml
hughperkins: The 'job' in this case is simply waiting for 60 minutes (the timeout time), to keep the container alive.

What we are waiting on here is for the container to spin up, which we detect when the sleep command is visible in `ps` output.

.github/workflows/production.yml
hughperkins: fused a bunch.

.github/workflows/production.yml
hughperkins: reverted

.github/workflows/production.yml
hughperkins: Updated to use a FIFO. Thoughts?

.github/workflows/production.yml
hughperkins: Updated to run a simple synchronous echo to wait for expansion.

.github/workflows/production.yml
```diff
@@ -109,41 +109,48 @@ jobs:
...
+          [[ -e "$JOBID_FIFO" ]] && rm -f "$JOBID_FIFO"
+          mkfifo "$JOBID_FIFO"
+          salloc --job-name="${SLURM_JOB_NAME}" \
+                 --partition=hpc-mid --nodes=1 --gpus=8 --exclusive \
+                 --time="${TIMEOUT_MINUTES}" \
+                 bash -c "echo \$SLURM_JOB_ID > $JOBID_FIFO; sleep ${TIMEOUT_MINUTES}m" &
+          SLURM_JOB_ID=$(cat "$JOBID_FIFO")
+          rm -f "$JOBID_FIFO"
+          SRUN_COMMON="--overlap --jobid=${SLURM_JOB_ID} ${SRUN_CONTAINER_OPTS} --export=${SLURM_ENV_VARS}"
+          srun --jobid=${SLURM_JOB_ID} --container-image=/mnt/data/images/genesis-v${GENESIS_IMAGE_VER}.sqsh ${SRUN_CONTAINER_OPTS} --export=${SLURM_ENV_VARS} \
```
duburcqa: Could you skip line for each arg?

duburcqa (approved):

.github/workflows/production.yml
duburcqa: Looks good!

.github/workflows/production.yml
duburcqa: > necessary for benchmarks

Yes it is!

> dont think its added to unit tests?

I think it was at some point during your refactoring.

.github/workflows/production.yml
hughperkins: done

.github/workflows/generic.yml
hughperkins: done

.github/workflows/examples.yml
hughperkins: done

PR #2171, opened 2025-12-31
reduce memory leak

YilingQiao:


duburcqa:
Rebooted here https://github.com/Genesis-Embodied-AI/Genesis/pull/2173

PR #2173, opened 2025-12-31
[BUG FIX] Fix 'pysplashsurf' memory leak causing OOM error.

duburcqa:

```python
import genesis as gs

gs.init(backend=gs.gpu, logging_level='debug')

options_mapping = {
    'sim_options': gs.options.SimOptions,
    'vis_options': gs.options.VisOptions,
    'coupler_options': gs.options.BaseCouplerOptions,
    'rigid_options': gs.options.RigidOptions,
    'mpm_options': gs.options.MPMOptions,
    'sph_options': gs.options.SPHOptions,
    'fem_options': gs.options.FEMOptions,
    'pbd_options': gs.options.PBDOptions,
    'profiling_options': gs.options.ProfilingOptions,
    'viewer_options': gs.options.ViewerOptions,
}
scene_args = {'show_viewer': False, 'vis_options': {'link_frame_size': 0.01, 'world_frame_size': 0.1, 'show_world_frame': False, 'segmentation_level': 'link', 'shadow': True, 'background_color': [0.05, 0.05, 0.05], 'plane_reflection': True, 'visualize_sph_boundary': False, 'visualize_mpm_boundary': False, 'visualize_pbd_boundary': False, 'render_particle_as': 'sphere', 'show_link_frame': False, 'show_cameras': False}, 'profiling_options': {'show_FPS': True}, 'sim_options': {'dt': 0.0002, 'substeps': 20, 'gravity': [0, 0, 0]}, 'mpm_options': {'upper_bound': [1.5, 0.5, 0.5], 'lower_bound': [-0.5, -0.5, -0.5], 'grid_density': 200, 'enable_CPIC': True, 'particle_size': 0.0025}}
camera_args = {'res': [1280, 720], 'pos': [-0.3, -0.15, 0.1], 'lookat': [0.2, 0, 0], 'up': [0, 0, 1], 'fov': 50, 'model': 'pinhole', 'aperture': 2.8, 'spp': 256}

scene_args['renderer'] = gs.options.renderers.Rasterizer()

for key, value in scene_args.items():
    if isinstance(value, dict):
        scene_args[key] = options_mapping[key](**value)
    else:
        scene_args[key] = value
scene = gs.Scene(**scene_args)
camera = scene.add_camera(**camera_args)

water_sphere = scene.add_entity(
    morph=gs.morphs.Sphere(pos=(0.5, 0, 0), radius=0.2, fixed=False),
    material=gs.materials.MPM.Liquid(rho=1000.0, viscous=False, sampler='pbs'),
    surface=gs.surfaces.Default(color=(0.2, 0.5, 1.0), vis_mode='recon')
)
bullet = scene.add_entity(
    morph=gs.morphs.Cylinder(pos=(-0.2, 0, 0), euler=(0, 90, 0), radius=0.025, height=0.05, fixed=False),
    material=gs.materials.Rigid(rho=11340.0),
    surface=gs.surfaces.Metal(metal_type='iron')
)
scene.build()

horizon = 150
bullet.set_dofs_velocity([100.0, 0.0, 0.0, 0.0, 0.0, 0.0])
for _ in range(horizon):
    gs.logger.info(f"{_}/{horizon} steps advanced")
    camera.render()
    scene.step()
```

### DIFF
--- a/genesis/utils/particle.py
+++ b/genesis/utils/particle.py
@@ -1,3 +1,4 @@
+import ctypes
 import os
 import pickle as pkl
 import platform
@@ -347,7 +348,6 @@ def particles_to_mesh(positions, radius, backend):
         faces = mesh.triangles.reshape([-1, 3])
 
         return trimesh.Trimesh(vertices, faces, process=False)
-
     elif "splashsurf" in backend:
         # Suggested value is 1.4-1.6, but 1.0 seems more detailed
         mesh_with_data, _ = pysplashsurf.reconstruction_pipeline(
@@ -365,10 +365,11 @@ def particles_to_mesh(positions, radius, backend):
         )
         normals = mesh_with_data.get_point_attribute("normals")
         vertices, triangles = mesh_with_data.take_mesh().take_vertices_and_triangles()
-        mesh = trimesh.Trimesh(vertices=vertices, faces=triangles, face_normals=normals)
+        mesh = trimesh.Trimesh(vertices=vertices, faces=triangles, face_normals=normals, process=False)
+        # FIXME: Reclaiming free-ed head memory manually is necessary to avoid unbounded growth
+        ctypes.CDLL(ctypes.util.find_library("c")).malloc_trim(0)
         gs.logger.debug(f"[splashsurf]: reconstruct vertices: {mesh.vertices.shape}, {mesh.faces.shape}")
         return mesh
-
     else:
         gs.raise_exception(f"Unsupported backend: {backend}.")
 


duburcqa:
@w1th0utnam3 Hi Fabian! Any idea about what is happening here? It would be great if this issue could be fixed on pysplashsurf side. Happy new year by the way!

w1th0utnam3:
> @w1th0utnam3 Hi Fabian! Any idea about what is happening here? It would be great if this issue could be fixed on pysplashsurf side. Happy new year by the way!

Happy new year. Can you try upgrading to v0.14.0? Then you can access the data directly using e.g. `mesh_with_data.mesh.vertices/triangles` which returns a Numpy view. I can imagine that the leak is related to the old `take...` methods. If this doesn't fix the issue I have to check this more in-depth.

duburcqa:
@w1th0utnam3 Thank you for your feedback! I'm trying that right now.

duburcqa:
@w1th0utnam3 This is not working. I still have to do this: https://github.com/Genesis-Embodied-AI/Genesis/pull/2176

If you are running on linux, it is easy to observe the memory leak by commenting `malloc_trim(0)` and running my reproduction script (>100gb RAM after only 60 iterations).

w1th0utnam3:
While I cannot completely rule out a memory leak in the Python bindings at the moment, I focused on the Rust side first. I exported the particle positions from your script until time step 59 and ran them with the same settings through the `splashsurf` CLI. I noticed that starting from time step 44, memory consumption during each reconstruction increases rapidly to about 25GB for step 59. I found out that this huge allocation of memory is caused by a large amount of particles positioned at the same coordinate. In step 44 there are 11k particles at coordinate `[1.485, 0.485, 0.485]`, this increases to more than 33k particles at this coordinate in step 59. These particles are all neighbors of each other and the corresponding neighborhood list of the surrounding subdomain then ends up with 33kˆ2 entries which consumes several gigabytes of memory (this list is needed due to the definition of the implicit surface via SPH interpolation). So this is already a less than ideal starting point for the reconstruction.

If these particles are supposed to be "deactivated" or something like that I would recommend to move them to a specific coordinate and to define a bounding box for the active particles to exclude this coordinate. Inside of the library there are also methods that support excluding particles with a mask/flag array.

I verified that there is no leak on the Rust side - between reconstructions, allocated memory drops to an almost constant amount again. So if there is a leak, it is in the bindings. Considering this, I'm a bit confused why it blows up so fast on the Python side without the trim operation. The only thing stored between reconstructions are the inputs and outputs but those are only take up a fraction of the memory needed during reconstruction. 

To debug the Python side I wrote a small script calling `pySplashsurf` on the exported particle files with the same arguments you used in Genisis and indeed, on Linux it looked like a memory leak. Meanwhile on MacOS (where I did most of the prior debugging) it allocated a lot of memory but it didn't accumulate over time. Maybe on Linux the allocator doesn't "hand back" memory as quickly between reconstructions, accumulating over time, which is "fixed" by the trim. I also cannot reproduce any leak-like behavior on Linux if I keep repeating reconstructions of e.g. the first 10 time steps of your simulation (which do not have this pathological particle configuration) or any of our typical SPH simulation outputs, even if they have millions of particles.

Is there any way you can prevent the input from having so many particles at the same coordinate? I'm not sure I can do much about that on the splashsurf side because it's very hard to efficiently find and filter out such configurations beforehand in a generic way (except if you restrict it to perfectly identical coordinates but this still requires sorting all particles). Such (expensive) functionality also has questionable benefit because this does not occur naturally in SPH simulations due to the pressure solver.

PR #2185, opened 2026-01-04
[FEATURE] Add method to compute axis-aligned bounding boxes of visual geometries.

duburcqa:
* Add method to compute axis-aligned bounding boxes of visual geometries.
* Cleanup logics to estimate spatial inertia from geometries if unknown.
* Cleanup rigid geom, link, and entity data accessors.
* Fix support of torch>2.9.1.

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_geom.py
+++ b/genesis/engine/entities/rigid_entity/rigid_geom.py
@@ -10,11 +10,10 @@ import torch
 import trimesh
 
 import genesis as gs
-import genesis.utils.array_class as array_class
 import genesis.utils.geom as gu
 import genesis.utils.mesh as mu
 from genesis.repr_base import RBC
-from genesis.utils.misc import tensor_to_array, DeprecationError
+from genesis.utils.misc import tensor_to_array, ti_to_torch, DeprecationError
 
 if TYPE_CHECKING:
     from genesis.engine.materials.rigid import Rigid as RigidMaterial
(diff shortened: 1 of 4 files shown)


genesis/engine/entities/rigid_entity/rigid_geom.py
```diff
@@ -839,6 +842,14 @@ def __init__(
         self._solver = link.entity.solver
         self._vmesh = vmesh
 
+        # Note that ggressiveness has been tuned to give sub-millimeter accuracy most of the time
+        aabb_mesh = vmesh.copy()
+        aabb_mesh.convexify()
+        aabb_mesh.decimate(decimate_face_num=NUM_VERTS_VISUAL_GEOM_AABB, decimate_aggressiveness=5, convexify=False)
```
YilingQiao: Will this take a long time to preprocess every time if vmesh is very big?

duburcqa: No I don't think so. These steps are very similar to what we are doing for convex decomposition.

duburcqa: Note that this is done only once, before building the scene.

YilingQiao: yes, like what we are doing for collision mesh, this can be slow (that's why we had the cache mechanism for collision mesh.) Right now we will do it no matter whether we need the visual mesh, and there is no cache mechanism

duburcqa: It is the convex decomposition algorithm that is slow, not mesh preprocessing. This is not cached I think. I will double check and add some caching mechanism if necessary.

duburcqa: Note that we only perform decimation on the convex hull. This should greatly reduce the complexity of the mesh that is being decimated.

duburcqa: I can also compute this low-res mesh only the first time it is needed.

YilingQiao: I see, maybe I can try some complex gltf mesh (and also some common robots) and see? If it's not slow, we can go with the current solution.

duburcqa: It takes 1ms for a glb with 50000 vertices (a tree). I think it should be fine.

duburcqa: I added lazy initialization, just in case.

PR #2193, opened 2026-01-06
[MISC] Prevent defining non-fixed plane primitive.

duburcqa:


genesis/options/morphs.py
```diff
@@ -433,8 +433,6 @@ class Plane(Primitive):
     collision : bool, optional
         Whether the entity needs to be considered for collision checking. Defaults to True.
         `visualization` and `collision` cannot both be False. **This is only used for RigidEntity.**
-    fixed : bool, optional
```
YilingQiao, line 436: can we raise an exception when fixed=False? There are (cautious) users who set fixed=True, we don't need to break their code

duburcqa: Yes we can!

PR #2195, opened 2026-01-06
[MISC] Migrate from pip to uv for unit tests venv in production CI.

hughperkins:


.github/workflows/production.yml
```diff
@@ -45,49 +45,59 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
-      - name: Run unit tests
+      - name: Start container
         if: github.event_name == 'pull_request'
         run: |
           SLURM_JOB_NAME="$(uuidgen)_$(date +%Y%m%d_%H%M%S)"
+          CONTAINER_NAME="${SLURM_JOB_NAME}"
```
hughperkins, line 52: Personally, I would like this bit to be offloaded to a common script, but I understand you would prefer not to do that.

.github/workflows/scripts/production_build.sh
duburcqa, line 1: Why not to use the same script in both cases? I see no reason to use different venv config in both cases.

.github/workflows/scripts/production_build.sh
```diff
@@ -0,0 +1,16 @@
...
+
+curl -LsSf https://astral.sh/uv/install.sh | sh
+which uv
+uv --version
+
+uv venv --python '3.10' /venv
+source /venv/bin/activate
+# TODO: USD baking does not currently support Python 3.11 since
+# NVIDIA does not currently release `omniverse-kit==107.3` on PyPI.
+# See: https://github.com/Genesis-Embodied-AI/Genesis/pull/1300
```
duburcqa, line 10: This comment should probably go above, when the desired python version is specified.

.github/workflows/scripts/production_unittests_build.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -ex
+
+curl -LsSf https://astral.sh/uv/install.sh | sh
+which uv
```
duburcqa: Not sure if this one is really useful.

.github/workflows/scripts/production_build.sh
```diff
@@ -0,0 +1,16 @@
...
+uv pip install omniverse-kit --index-url https://pypi.nvidia.com/
+uv pip install torch --index-url https://download.pytorch.org/whl/cu129
```
duburcqa, line 16: Maybe add a comment to specify that the version of cuda must tightly align with what is being installed in the Slurm container image, otherwise poorly packaged libraries such as libuipc may fail to import.

.github/workflows/production.yml
duburcqa: Yes, I would prefer not to do this indeed, for security concerns as you already know.

.github/workflows/production.yml
```diff
@@ -45,49 +45,59 @@ jobs:
...
-              # tmate -S /tmp/tmate.sock wait tmate-ready
-              # tmate -S /tmp/tmate.sock display -p '#{tmate_ssh}'
-
-              pytest -v -ra --backend gpu --dev --forked ./tests
-
-              # tmate -S /tmp/tmate.sock wait tmate-exit
+      - name: Run benchmarks
+        if: github.event_name == 'pull_request'
+        run: |
+          srun ${SRUN_COMMON} bash -e -s <<'EOF'
```
duburcqa, line 87: I'm wondering if it would be possible to using the directive:
```yaml
shell: srun ${SRUN_COMMON} bash -e {0}
run: |
    source /venv/bin/activate

    # sudo apt update
    # sudo apt install -y tmate
    # tmate -S /tmp/tmate.sock new-session -d
    # tmate -S /tmp/tmate.sock wait tmate-ready
    # tmate -S /tmp/tmate.sock display -p '#{tmate_ssh}'

    pytest -v -ra --backend gpu --dev --forked ./tests

   # tmate -S /tmp/tmate.sock wait tmate-exit
```

That would be awesome isn't it?!

.github/workflows/production.yml
```diff
@@ -45,49 +45,59 @@ jobs:
...
-            --no-container-mount-home --container-workdir=/root/workspace \
-            --export=NVIDIA_DRIVER_CAPABILITIES=all,BASH_ENV=/root/.bashrc,HF_TOKEN,GS_ENABLE_NDARRAY=${GS_ENABLE_NDARRAY} \
-            --partition=hpc-mid --nodes=1 --gpus=8 --exclusive --time="${TIMEOUT_MINUTES}" \
-            --job-name=${SLURM_JOB_NAME} \
-            bash -e -s << 'EOF'
-              if test -n "$(find /root/.venv -maxdepth 0 -empty)"; then
-                python3 -m venv --system-site-packages /root/.venv
-                source /root/.venv/bin/activate
-                pip install --no-input --upgrade pip pkg-info wheel
-                pip install --no-input --ignore-installed --upgrade blinker pyparsing setuptools
```
duburcqa, line 74: This commands are missing in the setup script. They should be added back.

.github/workflows/alarm.yml
duburcqa, line 1: For the record, there is no need to comment out the alarm workflow. It is not going to change anything because changes to this file are only taken into account AFTER merging for security concerns since it can be triggered by forks.

hughperkins:
Not needed with uv

hughperkins:
I found them necessary in my pip workflow, but not in the uv workflow.

The pip version:
https://github.com/hughperkins/Genesis/blob/a6f8559fe3aaf8b7fbe78dd21c7580e774c5291d/.github/workflows/scripts/production_build.sh#L30

hughperkins:
The difference is that for uv, we dont have conflicts with system installed
packages.

duburcqa:
> The difference is that for uv, we dont have conflicts with system installed
packages.

Yes, we do not want to rely on system dependencies.

duburcqa:
Still, it is important to make sure the latest versions of `pip wheel setuptools` are installed on the system to avoid build failure when building packages from source. It is just being extra cautious but I don't want to hit hard-to-debug pip install failures.

.github/workflows/production.yml
hughperkins: Doesn't seem to work?

Screenshot 2026-01-07 at 13 02 57

hughperkins: https://github.com/Genesis-Embodied-AI/Genesis/actions/runs/20791370477/job/59714073874?pr=2195

duburcqa: Could you try `shell: srun ${{ env.SRUN_COMMON }} bash -e {0}`  ?

hughperkins: VS linter puts squiggly red line:

Screenshot 2026-01-08 at 08 41 08

duburcqa: I think it is worth trying anyway. You are doing this at other places.

.github/workflows/scripts/production_build.sh
```diff
@@ -3,10 +3,15 @@
...
 curl -LsSf https://astral.sh/uv/install.sh | sh
-which uv
 uv --version
 
+# TODO: USD baking does not currently support Python 3.11 since
+# NVIDIA does not currently release `omniverse-kit==107.3` on PyPI.
+# See: https://github.com/Genesis-Embodied-AI/Genesis/pull/1300
 uv venv --python '3.10' /venv
 source /venv/bin/activate
+# Note: the version of cuda must tightly align with what is being installed
```
duburcqa, line 13: Could you add `uv pip install pip wheel setuptools` ?

hughperkins: No because it will slow down benchmarks.

They are installed for unit tests though. https://github.com/Genesis-Embodied-AI/Genesis/pull/2195/files#diff-d5171f9b8c42863b1e934fda1fefedf968913c65aa5d715555a696fb3cbd4c7eR84

hughperkins: alternative approach: https://github.com/Genesis-Embodied-AI/Genesis/blob/8c494052aa113eeaa669cfdb1dd249ebfea0f974/.github/workflows/scripts/production_build.sh

duburcqa: > No because it will slow down benchmarks.

Do you have numbers?

duburcqa:
```bash
uv pip install ".[dev,render]"
if [[ "${ADD_USD}" == "1" ]]; then
    uv pip install ".[usd]"
fi
```

I do not recommend doing this because it must resolves dependencies twice, which is usually costly. Just do:
```bash
if [[ "${ADD_USD}" == "1" ]]; then
    uv pip install ".[dev,usd,render]"
else
    uv pip install ".[dev,render]"
fi
```
Or
```bash
GENESIS_OPTIONAL_DEPS="dev,render"
if [[ "${ADD_USD}" == "1" ]]; then
    GENESIS_OPTIONAL_DEPS="$GENESIS_OPTIONAL_DEPS,usd"
fi
uv pip install ".[$GENESIS_OPTIONAL_DEPS]"
```

hughperkins: > > No because it will slow down benchmarks.
> 
> Do you have numbers?

Yes, the three changes cost 30 seconds overall.

hughperkins: well, you can compare the time fot he 'Build' step for unit tests vs benchmarks:

Screenshot 2026-01-08 at 10 57 31
Screenshot 2026-01-08 at 10 57 36

hughperkins: actually, more like 45 seconds.

duburcqa: I see. It depends how much more complex it is to maintain both. I tend to think that it is important to use the same venv configuration for unit tests and benchmarks, to make it easier / less error prone to reproduce failures / issues.

.github/workflows/production.yml
duburcqa: Ok I'm sad.
image

.github/workflows/scripts/production_build.sh
hughperkins: Fair enough. Updated in 057c1ca5

duburcqa (dismissed):

duburcqa:
Blocked by https://github.com/Genesis-Embodied-AI/Genesis/pull/2208

hughperkins:
Thanks! 🙌

PR #2200, opened 2026-01-07
[MISC] Add memory benchmarks to alarm.yml

hughperkins:
## Description

- add upload of memroy benchmarks to wandb
- add alarm.yml script to process the mem.csv, and the historical data in wandb
- create test harness to run alarm.yml
- create unit tests using the test harness

Example output:
Screenshot 2026-01-07 at 11 12 15

## Related Issue

## Checklist:


duburcqa:
I'm not in favour of adding unit test for your CI scripts. It just makes things more complicated for something that has nothing to do with Genesis itself. You have 1200 lines of unit tests for a script of 100 lines originally. This is hardcore.

.github/workflows/alarm.yml
duburcqa, line 1: I don't understand. I though to wanted to move the python code outside in its own script? I'm ok with it if it makes your life easier.

hughperkins:
> I'm not in favour of adding unit test for your CI scripts. It just makes things more complicated for something that has nothing to do with Genesis itself. You have 1200 lines of unit tests for a script of 100 lines originally. This is hardcore.

It makes testing and development much easier.

I'm not a fan of "testing on github" on the whole. Especially when we cannot test until we merge to main.

.github/workflows/alarm.yml
hughperkins: Needs a test harness anyway. Might as well keep it in alarm.yml therefore.

duburcqa:
Wait, there may be some confusion. I suggest to keep your test harness, but not all the other tests that come with it. Just keep all you need for "making testing and development much easier."

> I'm not a fan of "testing on github" on the whole. Especially when we cannot test until we merge to main.

Yes, we agree on this point. But I don't think `tests/test_upload_mem_to_wandb.py‎` and `tests/test_alarm_workflow.py‎` after necessary for this.

.github/workflows/alarm.yml
```diff
@@ -240,14 +294,85 @@ jobs:
...
+                      if not isinstance(max_mem_mb, (float, int)) or math.isnan(max_mem_mb):
+                          continue
+                  except (KeyError, TypeError):
+                      continue
+                  
+                  # Store memory record
+                  nbid = normalize_benchmark_id(bid)
+                  mem_records_by_rev.setdefault(rev, {})[nbid] = {
+                      "max_mem_mb": max_mem_mb
+                  }
```
duburcqa: There is not way to share the logics with what we are already doing for runtime and compile time? We could probably define some helper function that would support both cases?

.github/workflows/alarm.yml
```diff
@@ -325,6 +450,99 @@ jobs:
...
+                      mem_info["status"] = "n/a"
+                      picto = "ℹ️"
+                      stats_repr = "---"
+                      delta_repr = "---"
+
+                  rows_md_mem.append("| " + " | ".join((picto, *params_repr, value_repr, stats_repr, delta_repr)) + " |")
+                  rows_for_csv["mem"].append(mem_info)
+
+              tables["mem"] = [header_mem, align_mem] + rows_md_mem
+
```
duburcqa: Why are you duplicating the entire logics? We manage the share the same code for runtime and compile time. It should be possible to do the same for memory I think.

tests/upload_mem_to_wandb.py
```diff
@@ -0,0 +1,172 @@
...
+
+        # Parse GitHub URL to get org/repo
+        if "github.com" in remote:
+            # Handle both SSH (git@github.com:org/repo.git) and HTTPS (https://github.com/org/repo.git)
+            org_repo = remote.split("github.com")[-1].strip("/:").replace(".git", "")
+            return f"{rev}@{org_repo}"
+
+        return f"{rev}@unknown"
+    except subprocess.CalledProcessError:
+        return "unknown@unknown"
```
duburcqa: This function looks like a duplicate of what we already have elsewhere. Why not just call the existing method?

hughperkins:
done some refactorization. (note: heavily Cursored)

hughperkins:
Done a bunch of refactorization in this.

PR #2203, opened 2026-01-07
Texture Path Preservation

Rush2k:
## Description
Preserve texture file paths alongside image data so exporters can reference original texture files instead of re-encoding pixel data.

## Changes

1. genesis/options/textures.py - Allow ImageTexture to have both image_path and image_array
Changed validation from XOR (exactly one) to "require at least one"
Only load image from disk if image_array is not already provided
Preserve image_path even when image_array is present
2. genesis/engine/mesh.py - Extract path from trimesh materials
Extract image_path from material.kwargs.get("image_path") when loading SimpleMaterial
Pass extracted path to create_texture()
3. genesis/utils/mesh.py - Plumbing for path preservation
create_texture(): Added image_path parameter, pass to ImageTexture constructor
create_plane(): Store image_path="textures/checker.png" in plane material
surface_uvs_to_trimesh_visual(): Store texture.image_path in material kwargs when converting to trimesh
4. tests/test_mesh.py - Test for path preservation
Added test_plane_texture_path_preservation() to verify planes preserve texture paths

### DIFF
--- a/genesis/engine/sensors/camera.py
+++ b/genesis/engine/sensors/camera.py
@@ -96,6 +96,9 @@ class BatchRendererCameraWrapper(BaseCameraWrapper):
         self.idx = len(sensor._shared_metadata.sensors)  # Camera index in batch
         self.debug = False
 
+        # Camera model attribute for BatchRenderer.build() compatibility
+        self.model = getattr(sensor._options, "model", "pinhole")
+
         # Initial pose
         pos = torch.tensor(sensor._options.pos, dtype=gs.tc_float, device=gs.device)
         lookat = torch.tensor(sensor._options.lookat, dtype=gs.tc_float, device=gs.device)


YilingQiao:
Please change the code format. Other than that, it looks good to me

tests/test_mesh.py
```diff
@@ -510,6 +510,28 @@ def test_2_channels_luminance_alpha_textures(show_viewer):
...
+    assert vgeom.vmesh is not None
+    assert vgeom.vmesh.surface is not None
+
+    texture = vgeom.vmesh.surface.diffuse_texture
+    assert texture is not None
+    assert isinstance(texture, gs.textures.ImageTexture)
+
+    # The texture should have both image_path and image_array
+    assert texture.image_path is not None
+    assert texture.image_array is not None
```
YilingQiao, line 532: those asserts are not helpful

tests/test_mesh.py
```diff
@@ -510,6 +510,28 @@ def test_2_channels_luminance_alpha_textures(show_viewer):
...
+
+    # Check that the plane's vgeom has a texture with a path
+    assert len(plane.vgeoms) > 0
+    vgeom = plane.vgeoms[0]
+    assert vgeom.vmesh is not None
+    assert vgeom.vmesh.surface is not None
+
+    texture = vgeom.vmesh.surface.diffuse_texture
+    assert texture is not None
+    assert isinstance(texture, gs.textures.ImageTexture)
```
duburcqa, line 528: I don't understand. These checks seem to be completely unrelated to what this test is pretending to check. Either broaden the scope of this unit test (test name, docstring, comments...) or remove these checks.

tests/test_mesh.py
duburcqa: I agree. Either redundant or out of scope.

genesis/engine/mesh.py
```diff
@@ -232,6 +232,7 @@ def from_trimesh(
         must_update_surface = True
         roughness_factor = None
         color_image = None
+        color_image_path = None
```
duburcqa, line 235: I would recommend not supporting specifying both and raise an exception if it happens. Then load image files directly inside this method if provided.

genesis/utils/mesh.py
```diff
@@ -499,9 +500,14 @@ def tonemapped(image):
...
-def create_texture(image, factor, encoding):
+def create_texture(image, factor, encoding, image_path=None):
     if image is not None:
-        return gs.textures.ImageTexture(image_array=image, image_color=factor, encoding=encoding)
+        return gs.textures.ImageTexture(
+            image_array=image,
+            image_path=image_path,
+            image_color=factor,
+            encoding=encoding,
+        )
```
duburcqa, line 510: Same. This API is a bit broken I think. What about adding other method called `create_texture_from_path` or something, that would do the internal plumbing? Because allowing to specify both `image` and `image_path` is weird.

genesis/utils/mesh.py
```diff
@@ -928,6 +934,7 @@ def create_plane(normal=(0.0, 0.0, 1.0), plane_size=(1e3, 1e3), tile_size=(1, 1)
             ),
             material=trimesh.visual.material.SimpleMaterial(
                 image=Image.open(os.path.join(get_assets_dir(), "textures/checker.png")),
+                image_path="textures/checker.png",
```
duburcqa, line 937: This looks like dark magic. This is not officially supported by Trimesh from what I know, but maybe I'm wrong.

genesis/options/textures.py
```diff
@@ -112,7 +112,8 @@ class Config:
     def __init__(self, **data):
         super().__init__(**data)
 
-        if not (self.image_path is None) ^ (self.image_array is None):
+        # Allow both image_path and image_array, but require at least one
+        if self.image_path is None and self.image_array is None:
```
duburcqa, line 116: I don't think it makes sense.

genesis/engine/sensors/camera.py
```diff
@@ -96,6 +96,9 @@ def __init__(self, sensor: "BatchRendererCameraSensor"):
         self.idx = len(sensor._shared_metadata.sensors)  # Camera index in batch
         self.debug = False
 
+        # Camera model attribute for BatchRenderer.build() compatibility
+        self.model = getattr(sensor._options, "model", "pinhole")
+
```
YilingQiao, line 101: https://github.com/Genesis-Embodied-AI/Genesis/commit/d3198385879374207a142a54538e3e9ab19e6b15

can we change to use this fix instead?

duburcqa: That would be much better indeed.

duburcqa: (and update comment accordingly of course)

genesis/engine/mesh.py
```diff
@@ -340,6 +343,9 @@ def from_morph_surface(cls, morph, surface=None):
         if isinstance(morph, gs.options.morphs.Mesh):
             if morph.is_format(gs.options.morphs.MESH_FORMATS):
                 meshes = mu.parse_mesh_trimesh(morph.file, morph.group_by_material, morph.scale, surface)
+                if morph.parse_glb_with_zup:
+                    for mesh in meshes:
+                        mesh.convert_to_zup()
```
duburcqa, line 348: This seems completely unrelated to this PR? We need a unit test for this.

duburcqa:
This PR is mixing at least 3 different completely unrelated "features":
* GLB z-up
* Camera type
* Texture path preservation

Since texture path preservation is not ready, I would suggest to split this PR in at least 2 and merge can be, to avoid holding all these features for longer.

PR #2207, opened 2026-01-08
[FEATURE] Enabling rasterizer camera sensor to work with n_envs > 1.

Narsil:
## Description

This reallows RasterizerCameraSensor to work with n_envs > 1.

The PyRenderer requires `env_separate_rigid=True` to correctly separate the rendering.
This requires OpenGL 4.2 which is not available on Metal.

I moved the errorring out to the `.build()` since I think it's easier on users to have errors earlier rather than later.

I chose not to change the batched dimension on the Rasterizer (it can return a non batched variant when use_separate_rigid=False) to minimize the surface of this change.

Diff unfortunately includes some `ruff format` modifications, not reverted automatically by `black`. 
I can manually revert if necessary.

## Motivation and Context

It's important to have rendering when using `n_envs>1` in the context of leveraging the batched simulation.
This version works, even if not optimal compared to a purely batched rendering.

I decided to not change other renderers as they belong in other PRs IMHO.

https://huggingface.co/datasets/Genesis-Intelligence/snapshots/commit/1029f0b5ec874b93dc5170179f7f5a3e770b10ea

## Checklist:

### DIFF
--- a/tests/test_sensor_camera.py
+++ b/tests/test_sensor_camera.py
@@ -1,8 +1,8 @@
+import sys
+
 import numpy as np
-from numpy.__config__ import show
 import pytest
 import torch
-import sys
 
 import genesis as gs
 from genesis.utils.misc import tensor_to_array


duburcqa:
> Diff unfortunately includes some ruff format modifications, not reverted automatically by black.
I can manually revert if necessary.

Yes, that would be great indeed.

tests/test_sensor_camera.py
```diff
@@ -130,3 +131,66 @@ def _get_camera_world_pos(sensor):
...
     assert cam_move_dist_offset_T > 1e-2
     assert_allclose(cam_move_dist_offset_T, cam_move_dist, atol=1e-2)
+
+
+# ========================== Multi-environment tests ==========================
+
+
+@pytest.fixture
+def rasterizer_camera_scene():
+    """Create a simple scene with rasterizer camera sensor for multi-env tests."""
```
duburcqa: Not sure it is worth adding this fixture at this point since it is used only once and it seems quite specific.

tests/test_sensor_camera.py
```diff
@@ -130,3 +131,66 @@ def _get_camera_world_pos(sensor):
...
+
+    return scene, sphere, camera
+
+
+def test_rasterizer_camera_sensor_n_envs(rasterizer_camera_scene, png_snapshot):
+    scene, sphere, camera = rasterizer_camera_scene
+    n_envs = 2
+
+    scene.build(n_envs=n_envs)
+    sphere.set_pos(torch.tensor([[0.0, 0.0, 1.0], [0.2, 0.0, 0.5]]))
```
duburcqa: `sphere.set_pos([[0.0, 0.0, 1.0], [0.2, 0.0, 0.5]])` is good enough.

tests/test_sensor_camera.py
```diff
@@ -130,3 +131,66 @@ def _get_camera_world_pos(sensor):
...
+
+    scene.step()
+
+    data = camera.read()
+
+    assert data.rgb.shape == (2, 64, 64, 3), f"Expected shape (2, 64, 64, 3), got {data.rgb.shape}"
+    assert data.rgb.dtype == torch.uint8, f"Expected dtype torch.uint8, got {data.rgb.dtype}"
+    assert (data.rgb[0] - data.rgb[1]).abs().sum() != 0, "We should have different frames"
```
duburcqa: I'm not convinced these assert descriptions are bringing any value.

duburcqa: Why not just `assert (data.rgb[0] != data.rgb[1]).any()` ?

tests/test_sensor_camera.py
```diff
@@ -130,3 +131,66 @@ def _get_camera_world_pos(sensor):
...
+        pos=(3.0, 0.0, 1.5),
+        lookat=(0.0, 0.0, 0.5),
+        fov=60.0,
+    )
+    camera = scene.add_sensor(options)
+
+    return scene, sphere, camera
+
+
+def test_rasterizer_camera_sensor_n_envs(rasterizer_camera_scene, png_snapshot):
```
duburcqa: Missing `@pytest.mark.required` decorator.

tests/test_sensor_camera.py
Narsil: I took it from the other PR.

What do you prefer:
- Inline in the test ?
- Simple function call (instead of fixture).
- Other existing figure/scene regularly used for tests ?

tests/test_sensor_camera.py
duburcqa: You should probably add:
```python
if sys.platform == "darwin":
    pytest.xskip("Batched rendering with Rasterizer backend is not supported on this machine because it requires OpenGL 4.2.")
```

tests/test_sensor_camera.py
Narsil: No particular reason, I'm used to accepting floating point errors, `.abs().sum()` allows showing the tolerance very easily, doesn't really applies here.

Narsil: Regularly I'm hitting errors and not understanding what's going on.

`frames` is the keyword here (imho it adds from `data.rgb` which is not super clear if you're not familiar with the test).

The other 2 are definitely redundant, I'll remove them.

tests/test_sensor_camera.py
Narsil: Done both.

tests/test_sensor_camera.py
Narsil: Done.

tests/test_sensor_camera.py
Narsil: I went with `.any()`.

tests/test_sensor_camera.py
duburcqa: Inline in the test

genesis/engine/sensors/camera.py
```diff
@@ -423,11 +428,22 @@ def build(self):
 
     def _create_standalone_context(self, scene):
         """Create a simplified RasterizerContext for camera sensors."""
+        if not scene.sim._rigid_only and scene.n_envs > 1:
+            gs.raise_exception("Rasterizer with n_envs > 1, does not work when using non rigid simulation")
+        if sys.platform == "darwin":
+            if scene.n_envs > 1:
+                gs.raise_exception(
+                    "Rasterizer with n_envs > 1, does not work on Metal because it doesn't support OpenGL 4.3"
```
duburcqa: `OpenGL 4.2`

duburcqa:
You should probably also apply this patch:
```diff
diff --git a/genesis/ext/pyrender/jit_render.py b/genesis/ext/pyrender/jit_render.py
index e163af05..beacb87c 100644
--- a/genesis/ext/pyrender/jit_render.py
+++ b/genesis/ext/pyrender/jit_render.py
@@ -374,7 +374,7 @@ class JITRenderer:
         self.gl = GLWrapper()

         IS_OPENGL_42_AVAILABLE = hasattr(self.gl.wrapper_instance, "glDrawElementsInstancedBaseInstance")
-        OPENGL_42_ERROR_MSG = "Seperated env rendering not supported because OpenGL 4.2 not available on this machine."
+        OPENGL_42_ERROR_MSG = "Separated env rendering not supported because OpenGL 4.2 not available on this machine."

         @nb.jit(
             nb.none(
diff --git a/genesis/ext/pyrender/renderer.py b/genesis/ext/pyrender/renderer.py
index 4f72e509..a3d7c2ed 100644
--- a/genesis/ext/pyrender/renderer.py
+++ b/genesis/ext/pyrender/renderer.py
@@ -152,7 +152,7 @@ class Renderer(object):

         if flags & RenderFlags.ENV_SEPARATE and flags & RenderFlags.OFFSCREEN:
             n_envs = scene.n_envs
-            use_env_idx = True
+            use_env_idx = True and scene.n_envs > 1
         else:
             n_envs = 1
             use_env_idx = False
diff --git a/genesis/vis/camera.py b/genesis/vis/camera.py
index c1fb8bfb..32cedefa 100644
--- a/genesis/vis/camera.py
+++ b/genesis/vis/camera.py
@@ -156,7 +156,7 @@ class Camera(RBC):
                 self._raytracer.add_camera(self)
             else:
                 self._is_batched = False
-                if self._visualizer.scene.n_envs > 0 and self._visualizer._context.env_separate_rigid:
+                if self._visualizer.scene.n_envs > 1 and self._visualizer._context.env_separate_rigid:
                     gs.logger.warning(
                         "Batched rendering via 'VisOptions.env_separate_rigid=True' is only partially supported by "
                         "Rasterizer for now. The same camera transform will be used for all the environments."
@@ -165,7 +165,7 @@ class Camera(RBC):
                 if self._env_idx is None:
                     if not self._is_batched:
                         self._env_idx = int(self._visualizer._context.rendered_envs_idx[0])
-                        if self._visualizer.scene.n_envs > 0:
+                        if self._visualizer.scene.n_envs > 1:
                             gs.logger.info(
                                 "Raytracer and Rasterizer requires binding to the camera with a specific environment "
                                 "index. Defaulting to 'rendered_envs_idx[0]'. Please specify 'env_idx' if necessary."
```

genesis/engine/sensors/camera.py
Narsil: Oops. Thanks.

genesis/engine/sensors/camera.py
```diff
@@ -423,11 +428,22 @@ def build(self):
 
     def _create_standalone_context(self, scene):
         """Create a simplified RasterizerContext for camera sensors."""
+        if not scene.sim._rigid_only and scene.n_envs > 1:
+            gs.raise_exception("Rasterizer with n_envs > 1, does not work when using non rigid simulation")
```
duburcqa, line 443: You should probably add a warning about bad performance using Batched rendering with rasterizer, plus mention that camera pose is not batched.

Narsil: I added an exception for attached camera, it's not doing the proper rendering right ?

I'll add the warning too.

Narsil:
What does `use_env_idx` mean. I've seen it in the code, but I dont' really understand its purpose...

tests/test_sensor_camera.py
Narsil: Done.

genesis/engine/sensors/camera.py
duburcqa: > I added an exception for attached camera, it's not doing the proper rendering right ?

Yes I think it is a good idea.

duburcqa:
> What does use_env_idx mean. I've seen it in the code, but I dont' really understand its purpose...

`use_env_idx` is an internal flag of pyrender to determine whether it should filter "geometries" based on env idx (which requires OpenGL>4.2) and just consider all of them no matter what.

genesis/engine/sensors/camera.py
duburcqa: Yeah, my bad it is me that mentioned 4.3, but I double checked and it was 4.2.

tests/test_sensor_camera.py
```diff
@@ -130,3 +132,106 @@ def _get_camera_world_pos(sensor):
...
+        res=(64, 64),
+        pos=(3.0, 0.0, 1.5),
+        lookat=(0.0, 0.0, 0.5),
+        fov=60.0,
+        entity_idx=sphere.idx,
+    )
+    scene.add_sensor(options)
+    n_envs = 2
+
+    with pytest.raises(gs.GenesisException):
```
duburcqa: Could you add a comment to explain what you are trying to catch here?

tests/test_sensor_camera.py
```diff
@@ -130,3 +132,106 @@ def _get_camera_world_pos(sensor):
...
+        show_viewer=False,
+    )
+
+    # Add a plane
+    scene.add_entity(
+        morph=gs.morphs.Plane(),
+        surface=gs.surfaces.Rough(color=(0.4, 0.4, 0.4)),
+    )
+
+    # Add a sphere
```
duburcqa: You could remove the plane. Not sure that adding a surface is relevant here but as you want.

tests/test_sensor_camera.py
```diff
@@ -130,3 +132,106 @@ def _get_camera_world_pos(sensor):
...
+    if sys.platform == "darwin":
+        pytest.skip(
+            "Batched rendering with Rasterizer backend is not supported on this machine because it requires OpenGL 4.2."
+        )
+
+    scene = gs.Scene(
+        rigid_options=gs.options.RigidOptions(
+            enable_collision=True,
+            gravity=(0, 0, -9.8),
+        ),
```
duburcqa: Remove these options, they are irrelevant in this context.

tests/test_sensor_camera.py
```diff
@@ -130,3 +132,106 @@ def _get_camera_world_pos(sensor):
     cam_move_dist_offset_T = np.linalg.norm(cam_pos_final_offset_T - cam_pos_initial_offset_T)
     assert cam_move_dist_offset_T > 1e-2
     assert_allclose(cam_move_dist_offset_T, cam_move_dist, atol=1e-2)
+
+
+# ========================== Multi-environment tests ==========================
+
+
+@pytest.mark.required
+def test_rasterizer_camera_sensor_n_envs(png_snapshot):
```
duburcqa: Consider supporting `show_viewer`.  This "global" flag is used to enable interactive viewer within unit tests. This is meant to use debugging in case of failure. It is enabled via the comment line argument `--vis`. Not sure if it is useful in this context. In this case, it should enable `show_viewer=True` at scene level, but also `draw_debug=True` in `gs.sensors.RasterizerCameraOptions`.

tests/test_sensor_camera.py
```diff
@@ -130,3 +132,106 @@ def _get_camera_world_pos(sensor):
...
+
+    scene.step()
+
+    data = camera.read()
+
+    assert data.rgb.shape == (2, 64, 64, 3)
+    assert data.rgb.dtype == torch.uint8
+    assert (data.rgb[0] != data.rgb[1]).any(), "We should have different frames"
+    from PIL import Image
+    import io
```
duburcqa: You can move these imports only top level of this file.

tests/test_sensor_camera.py
```diff
@@ -130,3 +132,106 @@ def _get_camera_world_pos(sensor):
...
+
+    for i in range(n_envs):
+        buffer = io.BytesIO()
+        img0 = Image.fromarray(data.rgb[i].numpy())
+        img0.save(buffer, format="PNG")
+        png = buffer.getvalue()
+        assert png == png_snapshot
```
duburcqa: Just do `rgb_array_to_png_bytes(data.rgb[i]) == png_snapshot`

duburcqa: `from .utils import rgb_array_to_png_bytes`

tests/test_sensor_camera.py
```diff
@@ -1,5 +1,7 @@
 import numpy as np
 import pytest
+import torch
+import sys
```
duburcqa: We are (manually sadly...) organising imports. Here is the rule: You should first import standards libraries, skip line, then ubiquitous third-party libraries (torch, numpy, tqdm...etc), skip line, then random third-party libraries, skip line, then our own custom libraries, skip line, then relative genesis imports. For each group, make sure it is sorted alphabetically.

Narsil: > (manually sadly...) 

`ruff check` should do that on its own, doesn't it ? (ruff format + ruff check work great in general, that's my default setup).

tests/test_sensor_camera.py
Narsil: Great, much better !

tests/test_sensor_camera.py
Narsil: Done (well they are not needed anymore)

tests/test_sensor_camera.py
Narsil: Done. Hopefully correctly (it's a fixture right?)

tests/test_sensor_camera.py
Narsil: Removed.

tests/test_sensor_camera.py
Narsil: I removed the plane. I find the snapshots less easy to "See" (no reference plane). But the tests are indeed simpler.

tests/test_sensor_camera.py
Narsil: I added a match statement for the exception we're supposed to catch. That should act the same as the comment and be checked.

tests/test_sensor_camera.py
duburcqa: I guess. I should have a look at it, but I'm stuck at extinguishing fires forever...

tests/test_sensor_camera.py
duburcqa: > it's a fixture right?

Technically it is a fixture, but it is just a flag in practice:
```python
@pytest.fixture(scope="session")
def show_viewer(pytestconfig):
    return pytestconfig.getoption("--vis") and IS_INTERACTIVE_VIEWER_AVAILABLE
```

tests/test_sensor_camera.py
duburcqa: > I removed the plane. I find the snapshots less easy to "See" (no reference plane). But the tests are indeed simpler.

I was only talking about the second test that is supposed to fail at build x) Not the first one, for which I agree the plane makes things more visual. It would be better to add it back for this test if you removed it. Sorry for the confusion....

Narsil: Done ! Makes sense.

tests/test_sensor_camera.py
```diff
@@ -1,9 +1,12 @@
 import numpy as np
+from numpy.__config__ import show
```
duburcqa: You added this line by mistake.

tests/test_sensor_camera.py
```diff
@@ -1,9 +1,12 @@
...
 import pytest
+import torch
+import sys
```
duburcqa: `import sys` should be moved on top, separated from the rest.

PR #2212, opened 2026-01-09
[FEATURE] Support attaching camera sensors with Rasterizer batched rendering.

Narsil:
## Description

This PR re-enables having `n_envs > 1` with `add_Sensors` Rasterizer camera where the camera is LINKED to an entity (so the renderer needs to use the proper env camera pose).

## Related Issue

## Checklist:

### DIFF
--- a/genesis/ext/pyrender/scene.py
+++ b/genesis/ext/pyrender/scene.py
@@ -602,12 +602,8 @@ class Scene(object):
 
     def sorted_mesh_nodes(self):
         cam_pos = self.get_pose(self.main_camera_node)
-        if len(cam_pos.shape) == 3:
-            batched_pos = True
-            cam_loc = cam_pos[:, :3, 3]
-        else:
-            batched_pos = False
-            cam_loc = cam_pos[:3, 3]
+        cam_loc = cam_pos[..., :3, 3]
+        batched_pos = len(cam_pos.shape) == 3
         solid_nodes = []
         trans_nodes = []
         for node in self.mesh_nodes:


genesis/engine/sensors/camera.py
```diff
@@ -448,9 +443,6 @@ def _create_standalone_context(self, scene):
                 )
             env_separate_rigid = False
         else:
```
Narsil, line 445:
```suggestion
```

pyproject.toml
```diff
@@ -139,6 +139,10 @@ force-exclude = '''
 )/
 '''
 
+
+[tool.ruff]
+line-length = 120
```
Narsil, line 144: Reduces the differences between black and ruff to a bare minimum (I could split it up, but it's important so I don't spend my time reverting my autoformatting (when I forgot disabling it).

genesis/ext/pyrender/scene.py
```diff
@@ -612,7 +618,11 @@ def sorted_mesh_nodes(self):
                 solid_nodes.append(node)
 
         # TODO BETTER SORTING METHOD
-        trans_nodes.sort(key=lambda n: -np.linalg.norm(self.get_pose(n)[:3, 3] - cam_loc))
-        solid_nodes.sort(key=lambda n: -np.linalg.norm(self.get_pose(n)[:3, 3] - cam_loc))
+        if batched_pos:
+            trans_nodes.sort(key=lambda n: -np.linalg.norm(self.get_pose(n)[:3, 3] - cam_loc[0]))
+            solid_nodes.sort(key=lambda n: -np.linalg.norm(self.get_pose(n)[:3, 3] - cam_loc[0]))
```
Narsil, line 620: This part is happening before the rendering pass so we don't really have `env_idx` here.

Before (with `cam_loc` unbatched) we were already using this behavior. So this is not modifying behavior even if it looks wrong/odd.

Given the TODO I decided to leave it as-is as it might warrant a bigger change to "fix" this (if it's needed at all, I'm not sure how strict the ordering should be).

genesis/ext/pyrender/scene.py
```diff
@@ -601,7 +601,13 @@ def from_trimesh_scene(trimesh_scene, bg_color=None, ambient_light=None):
...
 
     def sorted_mesh_nodes(self):
-        cam_loc = self.get_pose(self.main_camera_node)[:3, 3]
+        cam_pos = self.get_pose(self.main_camera_node)
+        if len(cam_pos.shape) == 3:
+            batched_pos = True
+            cam_loc = cam_pos[:, :3, 3]
+        else:
+            batched_pos = False
+            cam_loc = cam_pos[:3, 3]
```
duburcqa: What about

```python
cam_pos = self.get_pose(self.main_camera_node)[..., :3, 3]
batched_pos = len(cam_pos.shape) == 2
```

?

genesis/ext/pyrender/scene.py
duburcqa: This part is not important. Maybe add a fixme to mention that it should be camera-dependent in theory. But we are not using transparency a lot in practice, so it is not a big deal.

tests/test_sensor_camera.py
```diff
@@ -206,11 +211,28 @@ def test_rasterizer_camera_sensor_n_envs_attached_camera():
...
-        gs.sensors.RasterizerCameraOptions(
-            entity_idx=sphere.idx,
-        )
+    options = gs.sensors.RasterizerCameraOptions(
+        res=(64, 64),
+        pos=(-0.4, 0.1, 2.0),
+        lookat=(-0.6, 0.4, 1.0),
+        fov=60.0,
+        entity_idx=sphere.idx,
     )
```
duburcqa, line 221: Maybe enable debug_draw with show_viewer ?

genesis/ext/pyrender/scene.py
Narsil: It works, I tend to dislike `...` (and working with unknown rank tensors in general).

If you feel strongly about it I can change it. I tend to prefer slightly my version because the ranks are obviously either 3 or 2 (base on the slicing calls).

genesis/ext/pyrender/scene.py
Narsil: Done.

tests/test_sensor_camera.py
Narsil: Done.

genesis/ext/pyrender/scene.py
duburcqa: We are using `...` everywhere because the batch dimension may or may not be there (for now...). I tend to prefer following the usual pattern. 

> working with unknown rank tensors in general

Well, the issue is that we are exactly doing this here. If the expected number of dimension is unique, then I never use `...`, but if it is variable, I strongly advise using `...` to clarify that it is changing dynamically so one must be careful when manipulating this tensor.

Narsil: Noted, I'll update !

Narsil: Done.

PR #2220, opened 2026-01-12
minor fix for batch renderer

YilingQiao:
## Related Issue

## Checklist:

### DIFF
--- a/genesis/engine/sensors/camera.py
+++ b/genesis/engine/sensors/camera.py
@@ -95,6 +95,8 @@ class BatchRendererCameraWrapper(BaseCameraWrapper):
         super().__init__(sensor)
         self.idx = len(sensor._shared_metadata.sensors)  # Camera index in batch
         self.debug = False
+        # Camera model attribute for BatchRenderer.build() compatibility
+        self.model = getattr(sensor._options, "model", "pinhole")
 
         # Initial pose
         pos = torch.tensor(sensor._options.pos, dtype=gs.tc_float, device=gs.device)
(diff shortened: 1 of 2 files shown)


genesis/engine/sensors/camera.py
```diff
@@ -95,6 +95,8 @@ def __init__(self, sensor: "BatchRendererCameraSensor"):
         super().__init__(sensor)
         self.idx = len(sensor._shared_metadata.sensors)  # Camera index in batch
         self.debug = False
+        # Camera model attribute for BatchRenderer.build() compatibility
+        self.model = getattr(sensor._options, "model", "pinhole")
```
Narsil, line 99: This is a wrong fix imho, because there is no `model` defined in the option, so this would always be pinhole, whereas there is some check for `fisheye` variants in the batchRenderer. The `model` in Raytracer is defined as `model: str = "pinhole"`  in the Options which seems like we should copy from. This should make this line become:

```python
self.model = sensor._options.model
```

Actually I would suggest using :
```python
model: Literal["pinhole", "fisheye"] = "pinhole"
```

So the typecheckers can know statically which variants are acceptable, but that's. a refinement.

YilingQiao: sounds good, will do this

duburcqa: Yes, Let's stop using `getattr`.

YilingQiao: btw, this PR is not into main. I will create a simple fix into the main

duburcqa: Yes I know :)

PR #2227, opened 2026-01-13
[MISC] Store texture path for primitive morphs as metadata.

Rush2k:
Store "default" texture path for the plane primitive in Mesh.metadata, similar to other paths, so it can be retrieved later by other tools, exporters, etc.
Changes summary:
1- genesis/utils/mesh.py - create_plane(...)
The function that sets the default texture on the plane, also adds it to a metadata dictionary, and returns the meta data to its caller.
2- genesis/engine/entities/rigid_entity/rigid_entity.py - _load_primitive(...)
Here we add the metadata from create_plane to the gs.Mesh.from_trimesh call

The unit test in test_mesh.py::test_plane_texture_path_preservation just makes sure that texture path is preserved in the vgeom.vmeh.metadata

### DIFF
--- a/genesis/utils/mesh.py
+++ b/genesis/utils/mesh.py
@@ -37,7 +37,7 @@ CVX_PATH_QUANTIZE_FACTOR = 1e-6
 Y_UP_TRANSFORM = np.asarray(  # translation on the bottom row
     [[1.0, 0.0, 0.0, 0.0], [0.0, 0.0, 1.0, 0.0], [0.0, -1.0, 0.0, 0.0], [0.0, 0.0, 0.0, 1.0]], dtype=np.float32
 )
-DEFAULT_PLANE_TEXTURE_PATH = "textures/checker.png"
+DEFAULT_PLANE_TEXTURE_PATH = "textures/checker.png"  # use checkerboard texture by default
 
 
 class MeshInfo:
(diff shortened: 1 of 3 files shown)


tests/test_mesh.py
```diff
@@ -510,6 +510,22 @@ def test_2_channels_luminance_alpha_textures(show_viewer):
...
+    scene.build()
+
+    # Check that the plane's vgeom has metadata with texture_path
+    assert len(plane.vgeoms) > 0
+    vgeom = plane.vgeoms[0]
+    assert vgeom.vmesh is not None
+
+    # The texture path should be stored in metadata
+    assert "texture_path" in vgeom.vmesh.metadata
+    assert vgeom.vmesh.metadata["texture_path"] == "textures/checker.png"
```
duburcqa: * No need to build the scene
  * You can remove Scene options if you don't build, they won't be used anyway
* assert "texture_path" in vgeom.vmesh.metadata is still not necessary. pytest will throw a clear exception anyway (plain asserts are being post-processed to ease debugging).

```python
    scene = gs.Scene()
    plane = scene.add_entity(gs.morphs.Plane())

    # The texture path should be stored in metadata
    assert plane.vgeoms[0].vmesh.metadata["texture_path"] == "textures/checker.png"
```

genesis/utils/mesh.py
```diff
@@ -927,15 +930,17 @@ def create_plane(normal=(0.0, 0.0, 1.0), plane_size=(1e3, 1e3), tile_size=(1, 1)
...
         )
+        # Store texture path in metadata since we applied a texture
+        metadata["texture_path"] = DEFAULT_PLANE_TEXTURE_PATH
     else:
         vmesh.visual = trimesh.visual.ColorVisuals(
             vertex_colors=np.tile(np.asarray(color, dtype=np.float32), (len(vmesh.vertices), 1))
         )
 
-    return vmesh, mesh
+    return vmesh, mesh, metadata
```
duburcqa: I'm not a huge fan of returning metadata this way, because this helpers are not aware of what `metadata` is. 

Maybe create plane should take some `color_to_texture_path` input argument, that default to some global variable `DEFAULT_PLANE_TEXTURE_PATH` that would be defined on top of this module. Then when you create the plane, you can explicitly specify `color_or_texture`, and fill your metadata accordingly.

Rush2k: Fixed

tests/test_mesh.py
Rush2k: Fixed

PR #2237, opened 2026-01-15
[BREAKING][FEATURE] Improve import of y-up vs z-up file meshes.

AnisB:


genesis/engine/mesh.py
```diff
@@ -340,20 +343,19 @@ def from_morph_surface(cls, morph, surface=None):
...
-                if morph.parse_glb_with_zup:
-                    for mesh in meshes:
-                        mesh.convert_to_zup()
-                else:
-                    gs.logger.warning(
-                        "GLTF is using y-up while Genesis uses z-up. Please set parse_glb_with_zup=True"
-                        " in morph options if you find the mesh is 90-degree rotated. We will set parse_glb_with_zup=True"
-                        " and rotate glb mesh by default later and gradually enforce this option."
-                    )
+                # The GLF spec defines that all meshes are Yup, ignoring the FileMorph.file_meshes_are_zup option.
```
Kashu7100, line 355: I think it is still worth to make configurable. I found some assets that has been made z up already.
Adding this conversion by default makes sense though.

duburcqa: If you can prove that they are properly supported by other simulator, we may consider this.

duburcqa: https://github.com/Genesis-Embodied-AI/genesis-doc/blob/main/source/user_guide/getting_started/conventions.md

Kashu7100: The asset was from Maniskill: https://github.com/haosulab/ManiSkill/blob/3bc2381e7a3f99b77c2f7de80f057a62a2421cfd/mani_skill/utils/scene_builder/table/assets/table.glb
But z-up for glb is rare.

PR #2258, opened 2026-01-19
[MISC] Split Rigid engine related files into different submodules.

YilingQiao:
## Description

Refactor the file structures in rigid solver to reduce the length of a single file.

## Related Issue

## Checklist:


duburcqa:
What about submodules instead of prefixed names? It would be much more standard.

YilingQiao:
updated @duburcqa

genesis/engine/entities/rigid_entity/impl/entity_loader.py
```diff
@@ -0,0 +1,665 @@
...
+        geom_inertia = rotate_inertia(geom_inertia, gu.quat_to_R(geom_quat))
+
+        total_mass, total_com, total_inertia = compose_inertial_properties(
+            total_mass, total_com, total_inertia, geom_mass, geom_com_world, geom_inertia
+        )
+
+    return total_mass, total_com, total_inertia
+
+
+class RigidEntityLoaderMixin:
```
duburcqa: No class needed here. Just adding useless complexity. Remove this and define flat methods.

genesis/engine/entities/rigid_entity/impl/entity_loader.py
```diff
@@ -0,0 +1,665 @@
...
+            self._load_scene(morph, self._surface)
+        elif isinstance(morph, gs.morphs.Primitive):
+            self._first_g_infos = self._load_primitive(morph, self._surface)
+        elif isinstance(morph, gs.morphs.Terrain):
+            self._load_terrain(morph, self._surface)
+        else:
+            gs.raise_exception(f"Unsupported morph: {morph}.")
+
+        # Load heterogeneous variants (if any)
+        self._load_heterogeneous_morphs()
```
duburcqa: I don't think it make sense to move this specific method outside rigid_entity.py. It is just confusing and harder to maintain.

genesis/engine/entities/rigid_entity/impl/entity_loader.py
```diff
@@ -0,0 +1,665 @@
...
+
+    def _load_heterogeneous_morphs(self):
```
duburcqa: Same. Keep this in rigid_entity. Only move read-only methods that can be purely functional (take input, return output, without updating the internal state).

genesis/engine/entities/rigid_entity/impl/entity_loader.py
```diff
@@ -0,0 +1,665 @@
...
+                    n_dofs=0,
+                    type=gs.JOINT_TYPE.FIXED,
+                )
+            ],
+            g_infos=g_infos,
+            morph=morph,
+            surface=surface,
+        )
+
+    def _load_scene(self, morph, surface):
```
duburcqa: Same. I don't think it makes sense to move this outside rigid_entity.

genesis/engine/entities/rigid_entity/impl/entity_loader.py
```diff
@@ -0,0 +1,665 @@
...
+        self._links = gs.List()
+        self._joints = gs.List()
+        self._equalities = gs.List()
+
+        self._load_morph(self._morph)
+
+        self._requires_jac_and_IK = self._morph.requires_jac_and_IK
+        self._is_local_collision_mask = isinstance(self._morph, gs.morphs.MJCF)
+
+    def _load_primitive(self, morph, surface, load_geom_only_for_heterogeneous=False):
```
duburcqa: I don't think it makes sense to move this outside rigid_entity.

genesis/engine/entities/rigid_entity/impl/entity_kinematics.py
```diff
@@ -0,0 +1,1067 @@
...
+import torch
+
+import genesis as gs
+from genesis.utils import array_class
+from genesis.utils import geom as gu
+from genesis.utils import linalg as lu
+from genesis.utils.misc import broadcast_tensor, ti_to_torch
+
+
+class RigidEntityKinematicsMixin:
```
duburcqa: Do not use Mixin pattern. It is usually an anti-pattern unless in very specific cases.

genesis/engine/entities/rigid_entity/impl/entity_kinematics.py
```diff
@@ -0,0 +1,1067 @@
...
+    """Mixin class providing kinematics functionality for RigidEntity."""
+
+    def get_jacobian(self, link, local_point=None):
```
duburcqa: Refactor all these methods to convert them as input-output function instead of class methods.

genesis/engine/entities/rigid_entity/rigid_entity.py
duburcqa, line 1: I think all these modification should be reverted. Using mixin design is not correct.

If you want to reduce the complexity, using inheritance, move complexity outside by defining input-output functions that will be called inside the class. If not possible, then just keep the implementation where it is.

genesis/engine/solvers/rigid/abd/misc.py
duburcqa, line 1: Move init kernels in their own file.

Create a 'misc.py' for things that are not utils.py. A utils is a generic helper function, which `func_wakeup_entity_and_its_temp_island` is definitely not. `func_check_index_range` is a utils though.

genesis/engine/solvers/rigid/abd/accessor.py
duburcqa, line 1: Why is this file called "control"? This is weird. I would call this "accessor". It is a bit loose but much clearer I think.

genesis/engine/solvers/rigid/articulation/__init__.py
duburcqa, line 1: I don't think 'articulation' is a good name. I would recommend either 'rbd_algorithms' (rigid body dynamics algorithm), because it is exactly what it is, or just 'forward', which is more specific and therefore maybe better.

genesis/engine/solvers/rigid/abd/diff.py
duburcqa, line 1: This should not be called 'backward'. This is confusing because backward in Rigid Body Dynamics is a reserved terms. Call this 'diff', or 'differentiability', or 'autodiff', or 'grad'...etc

genesis/engine/solvers/rigid/abd/forward_kinematics.py
duburcqa, line 1: I would be more specific and call this forward_kinematics.

genesis/engine/solvers/rigid/abd/forward_dynamics.py
duburcqa, line 1: I would be more specific and call this forward_dynamics.

genesis/engine/solvers/rigid/abd/forward_kinematics.py
duburcqa: By the way, it would be nice to add 'inverse_kinematics' and store its implementation here.

genesis/engine/entities/rigid_entity/rigid_entity.py
duburcqa: By the way, motion planning should be moved in its own dedicated module.

YilingQiao: into solver or entity?

YilingQiao:
1. Reverted rigid_entity mixin pattern
  - Restored rigid_entity.py from main branch
  2. Renamed articulation/ → abd
  3. File renames within abd/
  4. Move inverse_kinematics to abd

hughperkins:
Nice! 🙌

genesis/engine/entities/rigid_entity/rigid_entity.py
duburcqa: As you want. It is not very clear actually. Could even be in utils.

PR #2261, opened 2026-01-19
[MISC] Improve code factorization of constraint solver.

duburcqa:


genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -1256,25 +1331,148 @@ def func_nt_hessian_direct(
...
+
+@ti.func
+def func_hessian_direct_tiled(
+    constraint_state: array_class.ConstraintState,
+    rigid_global_info: array_class.RigidGlobalInfo,
+):
+    """Compute the Hessian matrix `H = M + J.T @ D @ J of the optimization problem for all environment at once.
+
+    This implementation is specialized for GPU backend and highly optimized for it using shared memory and cooperative
+    threading. Under the hood, it implements the blocked matrix multiplication algorithm to support arbritrary matrix
```
hughperkins: "a blocked"
"arbitrary matrices"
"kB" (means "kilobytes", cf kb, I interpret as "kilobits")
Note that the shared memory size is architecture specific. Also, I think the base limit is 48kB, not 47kB?

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -1127,108 +1194,116 @@ def add_frictionloss_constraints(
...
+def func_hessian_direct_batch(
     i_b,
     entities_info: array_class.EntitiesInfo,
     constraint_state: array_class.ConstraintState,
     rigid_global_info: array_class.RigidGlobalInfo,
     static_rigid_sim_config: ti.template(),
 ):
+    """Compute the Hessian matrix `H = M + J.T @ D @ J of the optimization problem for a given environment `i_b`.
+
+    Note that only the upper triangular part will be updated for efficiency, because the Hessian matrix is symmetric.
```
hughperkins: Please comment on what we do with the lower triangle
- does the client need to copy from upper to lower triangle?
- does this function copy from upper to lower at the end?
- do our solver functions all assume the Hessian is symmetric, and only ever use values from the upper triangle, so no copy is needed?
- somethign else?

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -1256,25 +1331,148 @@ def func_nt_hessian_direct(
...
+
+    # Performance is optimal for BLOCK_DIM = MAX_DOFS_PER_BLOCK = 64.
+    BLOCK_DIM = ti.static(64)
+    MAX_DOFS_PER_BLOCK = ti.static(64)
+    MAX_CONSTRAINTS_PER_BLOCK = ti.static(32)
+
+    n_lower_tri = n_dofs * (n_dofs + 1) // 2
+
+    # FIXME: Adding `serialize=False` is causing sync failing for some reason...
+    # TODO: Consider moving `H += M` in a dedicated CUDA kernel. It should be bother simpler and faster.
```
hughperkins: "better"

hughperkins: oh, actually "both"

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -1256,25 +1331,148 @@ def func_nt_hessian_direct(
...
+        tid = i % BLOCK_DIM
+        i_b = i // BLOCK_DIM
+        if i_b >= _B:
+            continue
+
+        jac_row = ti.simt.block.SharedArray((MAX_CONSTRAINTS_PER_BLOCK, MAX_DOFS_PER_BLOCK), gs.ti_float)
+        jac_col = ti.simt.block.SharedArray((MAX_CONSTRAINTS_PER_BLOCK, MAX_DOFS_PER_BLOCK), gs.ti_float)
+        efc_D = ti.simt.block.SharedArray((MAX_CONSTRAINTS_PER_BLOCK,), gs.ti_float)
+
+        # Loop over all the constraints and accumulate their respective contribution to the hessian matrix
```
hughperkins: "contributions"

hughperkins: "Hessian"

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -1256,25 +1331,148 @@ def func_nt_hessian_direct(
...
+                    # being considered is a diagonal block.
+                    if not is_diag_tile:
+                        i_c_ = tid
+                        while i_c_ < n_conts_tile:
+                            for i_d_ in range(n_dofs_tile_col):
+                                jac_col[i_c_, i_d_] = constraint_state.jac[i_c_start + i_c_, i_d2_start + i_d_, i_b]
+                            i_c_ = i_c_ + BLOCK_DIM
+                        ti.simt.block.sync()
+
+                    # Compute `H += J.T @ D @ J` for the a single hessian block
```
hughperkins: "for a"

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -1256,25 +1331,148 @@ def func_nt_hessian_direct(
...
 @ti.func
-def func_nt_chol_factor(
+def func_cholesky_factor_direct_batch(
     i_b,
     constraint_state: array_class.ConstraintState,
     rigid_global_info: array_class.RigidGlobalInfo,
 ):
+    """Compute the Cholesky factorization L of the Hessian matrix H = L @ L.T for all environments at once.
+
+    Beware the Hessian matrix is re-purposed to stored its Cholesky factorization to sparse memory resources.
```
hughperkins: "to store"
"to spare"

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: > "a blocked"

Why? I think there is only one. I'm not talking about blocked GPU threading or stuff, just blocked matrix multiplication: 
https://math.stackexchange.com/questions/787909/block-matrix-multiplication
https://en.wikipedia.org/wiki/Block_matrix#Multiplication

Well, actually it is square block matrix multiplication to be precise.

> Also, I think the base limit is 48kB, not 47kB?

Yes I think you are right. 

> Note that the shared memory size is architecture specific. 

The amount of shared memory is architecture specific but the 48kB default limit without custom flag is architecture agnostic from what I know.

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -1290,121 +1488,320 @@ def func_nt_chol_factor(
...
+                i_b,
+                entities_info=entities_info,
+                rigid_global_info=rigid_global_info,
+                constraint_state=constraint_state,
+                static_rigid_sim_config=static_rigid_sim_config,
+            )
+    else:
+        _B = constraint_state.jac.shape[2]
+
+        # Compute the hessian matrix
```
hughperkins: let's remove this comment, since this information is contained in the function name.

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -1290,121 +1488,320 @@ def func_nt_chol_factor(
...
+        func_hessian_direct_tiled(constraint_state, rigid_global_info)
+
+        if ti.static(static_rigid_sim_config.enable_tiled_cholesky_hessian):
+            # Compute Cholesky decomposition of the Hessian matrix
```
hughperkins: let's remove this comment, since this information is contained in the function name.

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -1290,121 +1488,320 @@ def func_nt_chol_factor(
...
-                    i_b,
-                    entities_info=entities_info,
-                    dofs_state=dofs_state,
-                    rigid_global_info=rigid_global_info,
-                    constraint_state=constraint_state,
-                    static_rigid_sim_config=static_rigid_sim_config,
-                )
-                if not constraint_state.improved[i_b]:
-                    break
+    if ti.static(static_rigid_sim_config.sparse_solve or static_rigid_sim_config.backend == gs.cpu):
```
hughperkins: I would be fairly partial for adding a comment
```
# CPU
```
on the following line. Not sure your thoughts on this? I think it might make ti easier for me to scan the code.

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -1290,121 +1488,320 @@ def func_nt_chol_factor(
...
+        ti.loop_config(serialize=static_rigid_sim_config.para_level < gs.PARA_LEVEL.ALL, block_dim=32)
+        for i_b in range(_B):
+            func_hessian_and_cholesky_factor_direct_batch(
+                i_b,
+                entities_info=entities_info,
+                rigid_global_info=rigid_global_info,
+                constraint_state=constraint_state,
+                static_rigid_sim_config=static_rigid_sim_config,
+            )
+    else:
```
hughperkins, line 1619: and then here we would have a comment 
```
# GPU
```

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -1290,121 +1488,320 @@ def func_nt_chol_factor(
...
+        _B = constraint_state.jac.shape[2]
+
+        # Compute the hessian matrix
+        func_hessian_direct_tiled(constraint_state, rigid_global_info)
+
+        if ti.static(static_rigid_sim_config.enable_tiled_cholesky_hessian):
```
hughperkins, line 1623: why is Hessian always tiled, but Cholesky is configurable? Does this relate to Cholesky being more complicated to parallelize, and hence higher risk etc?

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -1290,121 +1488,320 @@ def func_nt_chol_factor(
...
+    is_degenerated = False
+    for i_c in range(constraint_state.n_constraints[i_b]):
+        is_active = constraint_state.active[i_c, i_b]
+        is_active_prev = constraint_state.prev_active[i_c, i_b]
+        if is_active ^ is_active_prev:
+            sign = 1.0 if is_active else -1.0
+
+            efc_D_sqrt = ti.sqrt(constraint_state.efc_D[i_c, i_b])
+            if ti.static(static_rigid_sim_config.sparse_solve):
+                for i_d_ in range(constraint_state.jac_n_relevant_dofs[i_c, i_b]):
```
hughperkins: I have a gentle preference to split out the sparse implementation and the dense implementation into separate inline functions. But I can undersatnd that you might prefer both implementations side by side like this, so I'm just stating a gentle preference, not requesting this.

If we do want to keep both implementations in the same function, preference to label each one with
```
# Sparse
```
and 
```
# Dense
```
please

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -2444,133 +2531,213 @@ def func_init_solver(
...
+            entities_info=entities_info,
+            rigid_global_info=rigid_global_info,
+            constraint_state=constraint_state,
+            static_rigid_sim_config=static_rigid_sim_config,
+        )
 
-            equalities_info.sol_params[i_e, i_b] = ti.Vector(
-                [2 * rigid_global_info.substep_dt[None], 1.0, 0.9, 0.95, 0.001, 0.5, 2.0]
-            )
+        tol_scaled = (rigid_global_info.meaninertia[i_b] * ti.max(1, n_dofs)) * rigid_global_info.tolerance[None]
```
hughperkins: Gentle preference to factorize this block of code out into a separate inline function, so that this containing function contains relatively high level concepts, like `func_hessian_and_cholesky_factor_direct_batch`, and whatever this block is doing (updating search direction?)

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -2444,133 +2531,213 @@ def func_init_solver(
...
-                if i_e < constraint_state.ti_n_equalities[i_b] - 1:
-                    equalities_info.eq_type[i_e, i_b] = equalities_info.eq_type[
-                        constraint_state.ti_n_equalities[i_b] - 1, i_b
-                    ]
-                constraint_state.ti_n_equalities[i_b] = constraint_state.ti_n_equalities[i_b] - 1
+    _B = constraint_state.grad.shape[1]
+
+    ti.loop_config(serialize=static_rigid_sim_config.para_level < gs.PARA_LEVEL.ALL, block_dim=32)
+    for i_b in range(_B):
+        # t0_start = ti.clock_counter()
```
hughperkins: this seems like dead code?

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -2444,133 +2531,213 @@ def func_init_solver(
...
+                    dofs_state=dofs_state,
+                    rigid_global_info=rigid_global_info,
+                    constraint_state=constraint_state,
+                    static_rigid_sim_config=static_rigid_sim_config,
+                )
+                if not constraint_state.improved[i_b]:
+                    break
+        else:
+            constraint_state.improved[i_b] = False
+        # t0_end = ti.clock_counter()
```
hughperkins: dead code

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: from your link:
Screenshot 2026-01-19 at 20 48 55

hughperkins: > The amount of shared memory is architecture specific but the 48kB default limit without custom flag is architecture agnostic from what I know.

Yes, that's my understanding, from searching briefly.

hughperkins:
Looks nice! Thank you 🙌

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: Yes, it is on purpose. I want to remember how to use taichi clock counter. Are you ok with me keeping these comments?

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: Ok done.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: I adding more details. Let me know if you are happy with it.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: I think it makes sense. There is almost nothing shared because these 2 implementations...

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: Make sense. I will do it.

duburcqa: done.

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: Not a huge fan of dead code. Let's create a notion page for performance measurement, e.g. this one https://www.notion.so/Measure-performance-in-Genesis-and-gstaichi-2ee0c06194e180ae8ebdce4da609daab

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: Done ! Let me know if it is fine now.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: Added a global static flag to toggle on/off in-kernel timers.

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -2444,133 +2643,185 @@ def func_init_solver(
...
+                    rigid_global_info=rigid_global_info,
+                    constraint_state=constraint_state,
+                    static_rigid_sim_config=static_rigid_sim_config,
+                )
+                if not constraint_state.improved[i_b]:
+                    break
+        else:
+            constraint_state.improved[i_b] = False
+        t0_end = ti.clock_counter() if ti.static(static_rigid_sim_config.enable_kernel_timers) else 0
+        constraint_state.timers[0, i_b] = t0_end - t0_start
```
hughperkins: You are measuring the time of a top level kernel. Couldn't you just measure the time using pytorch profiler?

duburcqa: No because it is per-batch. Anyway I removed this entirely in the end.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: Removed this entirely.

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -1290,121 +1492,372 @@ def func_nt_chol_factor(
...
+        constraint_state.Mgrad[i_d, i_b] = curr_out / constraint_state.nt_H[i_b, i_d, i_d]
+
+
+@ti.func
+def func_cholesky_solve_tiled(
+    constraint_state: array_class.ConstraintState,
+    static_rigid_sim_config: ti.template(),
+):
+    # Performance is optimal for BLOCK_DIM = MAX_DOFS_PER_BLOCK = 64.
+    # Note that the current Warp reduction is CUDA-specific for now. Although Warp reduction is supposed to be supported
```
hughperkins: Do you have some more details on what you mean by "GsTaichi does not provide a unified API for it yet"? Do you mean:
- gstaichi doesn't provide access to shared memory across all platforms?
- gstaichi doesn't provide access to warp synchronization primitives across all platforms?
- whilst gstaichi does provide access to shared memory and warp synchronization primtives across all platforms, there is no unified api for calling these, that is uniform across all platforms?
- something else?

( No need to add as an actual comment in the code, but I'm curious for you to provide details into this PR thread please).

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: Interesting. I feel it would be good to specify this in some central location. Since nearly half the values of nt_H are undefined therefore. Perhaps we could add a comment about this into array_class.py, in the definition of StructConstraintState, just above declaration of nt_H?

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -1256,25 +1333,150 @@ def func_nt_hessian_direct(
...
 @ti.func
-def func_nt_chol_factor(
+def func_cholesky_factor_direct_batch(
     i_b,
     constraint_state: array_class.ConstraintState,
     rigid_global_info: array_class.RigidGlobalInfo,
 ):
+    """Compute the Cholesky factorization L of the Hessian matrix H = L @ L.T for all environments at once.
+
+    Beware the Hessian matrix is re-purposed to stored its Cholesky factorization to spare memory resources.
```
hughperkins: Since the Cholesky is lower-triangular I think? Does this mean that the repurposed Hessian is now lower triangular rather than upper triangular?
- could we add also a comment into array_class.py, next to the declaration of nt_H, stating that nt_H is also used to store the Cholesky factorization, which in this case is lower-triangular, please?

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -1966,8 +2338,138 @@ def func_update_gradient(
...
+):
+    """
+    Unified implementation of gradient updated optimized for both CPU and GPU backends.
+
+    The tiled optimization is only supported on GPU backend and specifically optimized for it, falling back to the
+    classical batched implementation when running on CPU backend.
+
+    Note that the tiled cholesky factorization and solving is not systematically enabled because it is not always
+    superior in terms of performance and does not support arbitrary matrix sizes. More specifically, tiling gets
+    more beneficial as n_dofs increases, but n_dofs>=96 is not supported for now.
```
hughperkins: what happens for n_dofs >= 96? Does it crash? Give incorrect results? Fall back to another algorithm?

hughperkins: Also, why is it limited to n_dofs < 96? Limitations of shared memory?

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -1290,121 +1492,372 @@ def func_nt_chol_factor(
...
+            continue
+
+        H = ti.simt.block.SharedArray((MAX_DOFS, MAX_DOFS + 1), gs.ti_float)
+        v = ti.simt.block.SharedArray((MAX_DOFS,), gs.ti_float)
+        partial = ti.simt.block.SharedArray(
+            (NUM_WARPS if ti.static(ENABLE_WARP_REDUCTION) else BLOCK_DIM,), gs.ti_float
+        )
+
+        i_flat = tid
+        while i_flat < n_dofs_2:
```
hughperkins, line 1813: could we have a short few-word comment for each of these child for/while loops stating what they are doing please?

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -1290,121 +1492,372 @@ def func_nt_chol_factor(
...
+@ti.func
+def func_hessian_and_cholesky_factor_incremental_batch(
+    i_b,
+    constraint_state: array_class.ConstraintState,
+    rigid_global_info: array_class.RigidGlobalInfo,
+    static_rigid_sim_config: ti.template(),
+) -> bool:
+    is_degenerated = False
+    if ti.static(static_rigid_sim_config.sparse_solve):
+        # Sparse
```
hughperkins: since you have factorized out the implementation (thank you!), lets remove the comment, which contains the same information as the name of the sub function, pelase.

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -1290,121 +1492,372 @@ def func_nt_chol_factor(
...
+        is_degenerated = func_hessian_and_cholesky_factor_incremental_sparse_batch(
+            i_b, constraint_state, rigid_global_info
+        )
+    else:
+        # Dense
```
hughperkins: since you have factorized out the implementation (thank you!), lets remove the comment, which contains the same information as the name of the sub function, pelase.

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -1290,121 +1492,372 @@ def func_nt_chol_factor(
...
+
+
+@ti.func
+def func_hessian_and_cholesky_factor_incremental_batch(
+    i_b,
+    constraint_state: array_class.ConstraintState,
+    rigid_global_info: array_class.RigidGlobalInfo,
+    static_rigid_sim_config: ti.template(),
+) -> bool:
+    is_degenerated = False
```
hughperkins, line 1731: does gstaichi require we declare is_degenerated here? If not, could we remove it please, since it's defined in both if branches? (If gstaichi requires this declaration here, then obviously we should keep it :) )

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -1290,121 +1492,372 @@ def func_nt_chol_factor(
...
-            for i_dir in range(4):
-                d = (2 * (i_dir % 2) - 1) * (d1 if i_dir < 2 else d2)
-                n = d * contact_data_friction - contact_data_normal
-                force = force + n * constraint_state.efc_force[i_c * 4 + i_dir + const_start, i_b]
+        # Padding +1 to avoid memory bank conflicts that would cause access serialization
+        H = ti.simt.block.SharedArray((MAX_DOFS, MAX_DOFS + 1), gs.ti_float)
 
-            collider_state.contact_data.force[i_c, i_b] = force
+        i_pair = tid
+        while i_pair < n_lower_tri:
```
hughperkins, line 1537: could we add a few-word comment to each of these child while/for loops, stating what the code in the loop is doing, please?

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: Yes if you want!

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: * gstaichi doesn't provide access to warp synchronization primitives across all platforms?

AND

* there is no unified api for calling these, that is uniform across all platforms

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: Good point... let me check.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: I can add more details about these points.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: I'm 99% sure gstaichi does require this.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: sure.

duburcqa: sure.

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: Great!

Question: what is the layout of nt_H? Is it a 1-d tensor, with each column laid out one by one, like:

```
r0c0

r0c1
r1c1

r0c2
r1c2
r2c2

[...]
```

Or is it a 2d tensor, with just under half the values being invalid?

If it is the latter, is this because there is some fundamental reason why the former layout will be slower?

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: Interesting, thanks!

genesis/engine/solvers/rigid/constraint_solver.py
```diff
@@ -1966,8 +2338,138 @@ def func_update_gradient(
...
+    classical batched implementation when running on CPU backend.
+
+    Note that the tiled cholesky factorization and solving is not systematically enabled because it is not always
+    superior in terms of performance and does not support arbitrary matrix sizes. More specifically, tiling gets
+    more beneficial as n_dofs increases, but n_dofs>=96 is not supported for now.
+    """
+    _B = constraint_state.jac.shape[2]
+
+    if ti.static(
+        not static_rigid_sim_config.enable_tiled_cholesky_hessian or static_rigid_sim_config.backend == gs.cpu
```
hughperkins, line 2429: It looks like we are switching on the tiled backend inside RigidSolver.build ? It looks like that function uses the following heuristic? :
- backend != cpu
- 16 <= dofs <= max_n_threads
- n_envs <= 16384

Per other comments on this PR, it looks like tiled doesn't work on non-CUDA GPUs? Do we need to update RigidSolver.build to check also that we are on CUDA?

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: Yup, just checked. It does.

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: created a linear ticket and project for exploring adding warp sync primtiaves on other platforms https://linear.app/genesis-ai-company/issue/CMP-235/explore-adding-warp-synchronization-primitives-to-non-cuda-platforms

hughperkins: and added an additional ticket to create cross-platform unified api https://linear.app/genesis-ai-company/issue/CMP-236/create-unified-warp-synchronization-cross-platform-api

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: Add will add more details in comment.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: Actually it is the lower part that is being computed.... updated documentation.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: It does work on Apple Metal.

hughperkins: Oh ok. Does it also work on AMD? What does it not work on? Vulkan?

hughperkins: (how does the sharedmemory compare across the various platforms it does work on?)

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: done.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: > Does it also work on AMD? What does it not work on? Vulkan? (how does the sharedmemory compare across the various platforms it does work on?)

I have no idea. It was only tested on CUDA and Metal. All I know is that `ti.simt.block.SharedArray` is supposed to be backend-agnostic API. Therefore I did not add any backend-specific branching.

hughperkins: You stated elsewhere in this PR that gstaichi does not make warp synchroniztaion primitives available across all platforms. I feel that contradicst your assertion that this kernel is backend-agnostic?

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: done.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: done.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: done.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: done.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: done.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: > You stated elsewhere in this PR that gstaichi does not make warp synchroniztaion primitives available across all platforms.

No. Warp synchronisation (I mentioned reduction) is something else.

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: "to store" not addressed yet

hughperkins: Screenshot 2026-01-20 at 08 17 01

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: `ti.simt.warp.shfl_down_f32` is a warp-level operation we are relying on. This is only supported on CUDA for now.

duburcqa: By the way, `warp` terminology is kind of proprietary from Nvidia, or at least strongly implies that it is Nvidia specific. For instance, Apple calls it `SIMD-group`, which I think is much more generic. I would advocate the rename `warp` in `group` if there was a plan to provide a backend-agnostic variant.

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: - I think I'd prefer to make ti clear somehow that we are assuming the most conservative shared memory on CUDA. As we both agree, there seem to be options to use more shared memory
     - maybe `base shared memory` `default shared memory`, soemthign like thta?
- what is the shared memory size on other supported platforms, such as Metal and AMD?

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: Thanks!

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: Not seeing this question answered?
Screenshot 2026-01-20 at 08 24 03

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: Thanks!

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: Thank you!

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: Thanks!

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: Oh, shfl is a fairly cuda specific thing yeah. I ended up making a shim for this in coriander https://github.com/hughperkins/coriander/blob/5f2ca1e56214e96acabe5f7841ad5a4cac941765/src/shims.cpp#L24-L38 Might have become more generalized across other platofrms now though? 🤔

hughperkins: We could of course add this shim into gstaichi though :) (slower, but at least runs)

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: Per my understanding the exact specific reason this si cuda specific si because of `__shfl_down` ?Please could you state that concretely, in the comment?

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: > `It is the responsibility of the developper
    to manually configure the static global flag `enable_tiled_cholesky_hessian` accordingly`

Aren't we the developer? Aren't we setting this in RigidSolver.build?

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: Thanks!

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: Thanks!

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: what do you mean by 'done'? You state,d and I confirmed, taht we cannot remove this declaration in gstaichi?

hughperkins: Oh, is this because you blanket wrote 'done' in every resolved comment?

Yes, confirmed the decalaratoin is still there, which we both agree is correct. Thanks!

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: > By the way, warp terminology is kind of proprietary from Nvidia, or at least strongly implies that it is Nvidia specific. For instance, Apple calls it SIMD-group, which I think is much more generic. I would advocate the rename warp in group if there was a plan to provide a backend-agnostic variant.

Good observation. Thanks!

hughperkins: Back to the top question in this thread:
- it looks like tiled cholesky only works on cuda, because of __shfl_down?
- so shouldnt RigidConfig.build only enable tiled cholesky on cuda platforms?

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: Damn, indeed!

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: (confirmed offline; this question has been addressed. Cholesky uses more cuda things (specificaly shlf_down), so is less cross-platofmr than the hessian)

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: done.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: done.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: done.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: removed.

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: seems not to have been addressed? (function `func_cholesky_factor_direct_tiled` I think?)

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: > it looks like tiled cholesky only works on cuda, because of __shfl_down?

No, this is guarded and there are different code path inside the tiled implementation depending on whether this feature is available or not.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: > I think I'd prefer to make ti clear somehow that we are assuming the most conservative shared memory on CUDA.

I already made it clear: "Note that the amount of shared memory available is hardware-specific, but the 48kB default limit without enabling dedicated GPU context flag is hardware agnostic on modern GPUs."

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: done.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: I don't understand. I already wrote: "Note that the current Warp reduction is CUDA-specific for now. Although Warp reduction is supposed to be supported by all major GPUs if not all (incl. Apple Silicon chips under naming 'SIMD-group'), GsTaichi does not provide a unified API for it yet."

duburcqa: Added more details.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: > Aren't we the developer? 

We are.

> Aren't we setting this in RigidSolver.build?

We do.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: > seems not to have been addressed? (function func_cholesky_factor_direct_tiled I think?)

Indeed, but this is a complete different function x). On it.

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: I think the failure is specifcally because of __shfl_down right? Nothing else is cuda-only? If so, let's call that operation out specifically please. This makes it easier to understand when this function will or won't work.

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: Ok. I'm not sure I'm convinced that we want to frame things as though our own calls from our own code are from external clients, beyond our control. Are we expecting extenral users to call this kernel directly?

Maybe your goal is to make this comment not break if the calling code changes?

In that case, maybe somethign like "It is the responsibilty of calling code ..."

And potentially add "(at the time of writing, this was handled by RigidSolver.build)", though I can accept not putting this because of maintenance issues.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: done.

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa:
```
    # Note that the current "Warp" reduction is CUDA-specific for now (to be clear, it concerns all methods in
    # `ti.simt.warp` sub-module, but only `shfl_down_f32` is being used here).
    # Although "Warp" reduction is supposed to be supported by all major GPUs if not all (incl. Apple Silicon chips
    # under naming 'SIMD-group'), GsTaichi does not provide a unified API for it yet.
```

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: > Ok. I'm not sure I'm convinced that we want to frame things as though our own calls from our own code are from external clients, beyond our control.

Developper for me are only internal. I don't call users this way. I call them... users.

duburcqa: > And potentially add "(at the time of writing, this was handled by RigidSolver.build)", though I can accept not putting this because of maintenance issues.


Yes,  I don't want to do this.

duburcqa: I will clarify what is a dev.

duburcqa: Calling code is fine I think.

duburcqa: Note that calling anything that is not part of the public API (which excludes anything that is part of engine submodules) is unsupported and I considered that these use-cases do not exist at all.

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: Thanks! Are you pasting that becuase that's what it already said, but somehow I missed that it mentioned shfl down?

Either way, thank you :)

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: Thanks! 🙌

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: ok, so, working through this...
- searching for "if ti.static(
        not static_rigid_sim_config.enable_tiled_cholesky_hessian or static", I find `func_update_gradient` function
- `func_update_gradient` calls `func_update_gradient_tiled`
-  `func_update_gradient_tiled` calls `func_cholesky_solve_tiled`
- `func_cholesky_solve_tiled` says only supported on warp

Ergo func_update_gradient should only call func_update_gradient on a CUDA device? What am I missing here?

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: This is the updated version, that I wrote before your reply (at least if the page was properly refreshed...)

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: > func_cholesky_solve_tiled says only supported on warp
 
 No. It has features that are only supported on warp but disabled if not available. I will be more specific in documentation.

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: Thanks! 🙌

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: also at line 1510, in func_cholesky_factor_direct_tiled
Screenshot 2026-01-20 at 10 18 42

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: (I'm actually working off a local clone. I found the github diff for this PR quite unreadable...)

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: oh I see, using `ENABLE_WARP_REDUCTION`. Yes, please clarify this point in the comment. It was not obvious to me.

duburcqa: done.

duburcqa:
```
    # Performance is optimal for BLOCK_DIM = 64.
    BLOCK_DIM = ti.static(64)
    MAX_DOFS = ti.static(static_rigid_sim_config.tiled_n_dofs)

    # Note that the current "Warp" reduction is CUDA-specific for now (to be clear, it concerns all warp-level
    # intrinsics in `ti.simt.warp` sub-module, of which only `shfl_down_f32` is being used here).
    # Although "Warp" reduction is supposed to be supported by all major GPUs if not all (incl. Apple Silicon chips
    # under naming 'SIMD-group'), GsTaichi does not provide a unified API for it yet.
    # As a result, warp-level intrinsics are currently disabled of not running on CUDA backend. On top of that, most
    # if not all, Warp-level intrinsics are only supporting 32bits precision. A less efficient fallback has been
    # implemented if this feature is not available.
    ENABLE_WARP_REDUCTION = ti.static(static_rigid_sim_config.backend == gs.cuda and gs.ti_float == ti.f32)
    WARP_SIZE = ti.static(32)
    NUM_WARPS = ti.static(BLOCK_DIM // WARP_SIZE)
```

hughperkins: Thanks! 🙌

genesis/engine/solvers/rigid/constraint_solver.py
hughperkins: Thanks! 🙌

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: More cleanup:
```
    """Compute the solution of H @ grad = Mgrad st H = L @ L.T for all environments at once.

    This implementation is specialized for GPU backend and highly optimized for it using shared memory and cooperative
    threading. The current implementation only supports n_dofs <= 64 for 64bits precision and n_dofs <= 92 for 32bits
    precision. See `func_cholesky_factor_direct_tiled` documentation for details.

    Note that this implementation leverages warp-level reduction whenever supported, a generic fallback otherwise. At
    the time of writing, all warp-level intrinsics in `ti.simt.warp` sub-module are CUDA-specific, of which only
    `shfl_down_f32` is being used here. Although some of these warp-level instrinsics are supposed to be supported by
    all major GPUs if not all (incl. Apple Silicon chips under naming 'SIMD-group'), GsTaichi does not provide a unified
    API for it yet. As a result, warp-level intrinsics are currently disabled if not running on CUDA backend. On top of
    that, most if not all, Warp-level intrinsics are only supporting 32bits precision.
    """
 ```

genesis/engine/solvers/rigid/constraint_solver.py
duburcqa: Damn copy-paste x)

duburcqa: done.

hughperkins: haha :) Thanks ! :)

duburcqa:
> Awesome 🔥 Thank you

Amazing! Thank you for tackling this hard quest together.

Issue #2267, opened 2026-01-20
[Bug]: Scenes with more than 2 raytracers return no data for the sensors


JackLowry:
### Bug Description

In scenes with more than 2 raytracer sensors, all additional raytracers' sensor values are all 0s. I've traced this to being an issue with sensor_cache_offsets in [https://github.com/Genesis-Embodied-AI/Genesis/blob/main/genesis/engine/sensors/raycaster.py](url). The kernel_cast_rays states it should be a cumulative offset but this isn't implemented. 

### Steps to Reproduce

If possible, provide a script triggering the bug, e.g.
```python
import genesis as gs
import numpy as np
import time
import torch

########################## init ##########################
gs.init()

########################## create a scene ##########################

scene = gs.Scene(
    sim_options=gs.options.SimOptions(
        dt       = 4e-3,
        substeps = 10,
    ),
    mpm_options=gs.options.MPMOptions(
        # lower_bound   = (-0.5, -1.0, 0.0),
        lower_bound   = (-0.2, -.40, 0.0),
        upper_bound   = (0.2, .40, .25),
    ),
    vis_options=gs.options.VisOptions(
        visualize_mpm_boundary = True,
    ),
    viewer_options=gs.options.ViewerOptions(
        camera_fov=30,
    ),
    show_viewer = False,
)

########################## entities ##########################
plane = scene.add_entity(
    morph=gs.morphs.Plane(),
)

cube = scene.add_entity(
    # material=gs.materials.MPM.Elastic(),
    morph=gs.morphs.Box(
        pos  = (0.0, -1.0, 0.5),
        size = (0.1, 0.1, 1.),
    ),
)

########################## robot ##########################

sensor1 = scene.add_sensor(
        gs.sensors.Raycaster(
            pattern=gs.sensors.raycaster.SphericalPattern(
                n_points=(6, 6),
                fov=(60.0, (-120.0, -60.0)),
            ),
            entity_idx=cube.idx,
            pos_offset=(0.0, 0.5, 0.0),
            return_world_frame=False,
            draw_debug=True,
            debug_sphere_radius=0.01,
            debug_ray_start_color=(1.0, 1.0, 0.0, 1.0),
            debug_ray_hit_color=(0.5, 1.0, 1.0, 1.0),
        )
    )

sensor2 = scene.add_sensor(
        gs.sensors.Raycaster(
            pattern=gs.sensors.raycaster.SphericalPattern(
                n_points=(6, 6),
[...]

PR #2268, opened 2026-01-20
[BUG FIX] Sensor cache offset fix for scenes with multiple raycast sensors

JackLowry:
## Description
Minor change to sensor_cache_offsets, which correctly implements it as a cumulative offset per the comments in kernel_cast_rays

## Related Issue

Resolves #2267 .

## Motivation and Context
Bug Fix

## How Has This Been / Can This Be Tested?
Tested using the script from #2267. Can be verified by testing scenes with more than 2 raytrace sensors

## Checklist:

- [x] I read the **CONTRIBUTING** document.
- [x] I followed the `Submitting Code Changes` section of **CONTRIBUTING** document.
- [x] I tagged the title correctly (including BUG FIX/FEATURE/MISC/BREAKING)
- [x] I updated the [documentation](https://github.com/Genesis-Embodied-AI/genesis-doc) accordingly or no change is needed.
- [x] I tested my changes and added instructions on how to test it for reviewers.

- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #2269, opened 2026-01-21
[BUG FIX] Fix BVH morton code indexing for multi-env raycasting.

d-corsi:
## Description
Fixes incorrect Raycaster outputs when running multiple environments in parallel.

In `kernel_cast_rays`, BVH traversal correctly indexes nodes per environment (`bvh_nodes[i_b, ...]`), but the leaf-to-triangle mapping incorrectly used `bvh_morton_codes[0, ...]` (always env 0). This causes env>0 to map leaf nodes to the wrong triangle indices, producing incorrect / zeroed lidar values for non-zero environments.

This PR changes the morton-code lookup to index with the current batch/env index (`i_b`).

## Related Issue
Resolves Genesis-Embodied-AI/Genesis#2267

## Motivation and Context
The BVH data structures are batched per environment. Using a hard-coded batch index for `bvh_morton_codes` breaks correctness when multiple envs are simulated in parallel, since the leaf ordering / mapping is environment-specific.

1. Run a scene with `n_envs > 1` and a Raycaster sensor enabled.
2. Compare Raycaster outputs across envs.
   - Before this change: only env 0 produces correct lidar values; env>0 is incorrect/zeroed.
   - After this change: lidar values are correct for all envs.

(Verified locally with multiple parallel environments.)

### DIFF
--- a/tests/test_sensors.py
+++ b/tests/test_sensors.py
@@ -476,6 +476,7 @@ def test_raycaster_hits(show_viewer, n_envs):
     assert_allclose(grid_distances, grid_distances_ref, tol=1e-3)
 
 
+@pytest.mark.required
 def test_lidar_bvh_parallel_env(show_viewer, tol):
     """Verify each environment receives a different lidar distance when geometries differ."""
     scene = gs.Scene(
@@ -546,7 +547,8 @@ def test_lidar_bvh_parallel_env(show_viewer, tol):
     expected_distances = front_positions - sensor_positions[:, 0]
     assert_allclose(lidar_distances, expected_distances, tol=tol)
 
-    
+
+@pytest.mark.required
 def test_lidar_cache_offset_parallel_env(show_viewer, tol):
     scene = gs.Scene(
         show_viewer=show_viewer,


genesis/engine/sensors/raycaster.py
```diff
@@ -329,7 +329,7 @@ def __init__(
     @classmethod
     def _update_bvh(cls, shared_metadata: RaycasterSharedMetadata):
         """Rebuild BVH from current geometry in the scene."""
-        from genesis.engine.solvers.rigid.rigid_solver import kernel_update_all_verts
+        from genesis.engine.solvers.rigid.rigid_solver_decomp import kernel_update_all_verts
```
YilingQiao: revert this change

YilingQiao:
could you please add a unit test.

genesis/engine/sensors/raycaster.py
```diff
@@ -385,7 +385,7 @@ def build(self):
 
         # These fields are used to properly index into the big cache tensor in kernel_cast_rays
         self._shared_metadata.sensor_cache_offsets = concat_with_tensor(
-            self._shared_metadata.sensor_cache_offsets, self._cache_size
+            self._shared_metadata.sensor_cache_offsets, self._cache_size * (self._idx + 1)
```
YilingQiao, line 388: this is also mentioned by #2268

d-corsi:
Thanks for the review!

- [x] Reverted the unrelated kernel_update_all_verts import change as requested.
- [x] Added a unit test covering the multi-environment Raycaster behavior: tests/test_raycaster_multi_env.py. The test builds n_envs=2 with different geometry per env and verifies that the returned lidar distances differ accordingly.
- [x] Verified locally that the test fails with the buggy behavior and passes with this fix.

Please let me know if you’d like the test adjusted further.

genesis/engine/sensors/raycaster.py
d-corsi: Reverted the import change as requested.

tests/test_raycaster_multi_env.py
duburcqa, line 1: Could you move this test in our `test_sensors.py` module?

genesis/engine/sensors/raycaster.py
duburcqa, line 388: This fix seems unrelated no? Please remove it if it is not covered by your unit test.

PR #2279, opened 2026-01-22
[BREAKING][BUG FIX] Do not officially support importing GLTF morph Mesh as Z-UP.

duburcqa:


genesis/options/morphs.py
```diff
@@ -727,12 +731,33 @@ class Mesh(FileMorph, TetGenMixin):
...
+                    "Specifying 'file_meshes_are_zup' for GLTF/GLB files is not supported. A rotation will be applied "
+                    "explicitly on the morph instead. Please consider fixing your asset to use Y-UP convention."
+                )
+                quat = (0.707, -0.707, 0.0, 0.0)
+                if self.quat is None:
+                    self.quat = quat
+                else:
+                    self.quat = gu.transform_quat_by_quat(
+                        np.array(quat, dtype=gs.np_float), np.array(self.quat, dtype=gs.np_float)
+                    )
```
duburcqa, line 756: @YilingQiao I could rather just raise an exception asking to user to manually specify a rotation if you think it is better.

PR #2272, opened 2026-01-21
[MISC] Unify mem and perf file formats; upload mem results for test run to single wandb run

hughperkins:
## Description

Example of resulting summary data in wandb overview:

Screenshot 2026-01-21 at 08 52 39

## Related Issue

Resolves Genesis-Embodied-AI/Genesis#<your-issue-number>

## Checklist:

- [x] I read the **CONTRIBUTING** document.
- [x] I followed the `Submitting Code Changes` section of **CONTRIBUTING** document.
- [x] I tagged the title correctly (including BUG FIX/FEATURE/MISC/BREAKING)
- [ ] I updated the [documentation](https://github.com/Genesis-Embodied-AI/genesis-doc) accordingly or no change is needed.
- [ ] I tested my changes and added instructions on how to test it for reviewers.

- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #2281, opened 2026-01-22
[MISC] Add memory to CI performance monitoring report.

hughperkins:
## Description

Note that this stacks on top of https://github.com/Genesis-Embodied-AI/Genesis/pull/2272

## Related Issue

Tested by runnig the following script:

```bash
mkdir -p /tmp/out/artifacts/mem-test-results
mkdir -p /tmp/out/artifacts/speed-test-results
mkdir -p /tmp/out/output

cat <<EOF > /tmp/out/artifacts/mem-test-results/mem_test.txt
env=franka 	| constraint_solver=None 	| gjk_collision=True 	| batch_size=30000 	| backend=cuda 	| dtype=field 	| max_mem_mb=123
env=box_pyramid_3 | batch_size=4096 | backend=gpu | dtype=field | max_mem_mb=150
env=go2 	| batch_size=4096 	| backend=gpu 	| dtype=field | gjk_collision=True | max_mem_mb=2700
env=anymal 	| constraint_solver=Newton 	| gjk_collision=False 	| batch_size=0 	| backend=cpu 	| dtype=ndarray 	| max_mem_mb=234
EOF

cat <<EOF > /tmp/out/artifacts/speed-test-results/speed_test.txt
solver=PBD | backend=cpu | n_envs=128 | compile_time=2.52 | runtime_fps=990.0 | realtime_factor=49.5
env=box_pyramid_6 | batch_size=4096 | use_contact_island=False | gjk_collision=False | dtype=ndarray | backend=cuda | compile_time=1.50 | runtime_fps=1200.0 | realtime_factor=60.0
solver=PBD | backend=gpu | n_envs=1024 | compile_time=2.54 | runtime_fps=985.0 | realtime_factor=49.3
env=franka | batch_size=0 | use_contact_island=False | dtype=ndarray | backend=cpu | compile_time=1.20 | runtime_fps=1500.0 | realtime_factor=75.0
solver=MPM | backend=cpu | n_envs=64 | compile_time=2.53 | runtime_fps=988.0 | realtime_factor=49.4
EOF

export SPEED_ARTIFACTS_DIR=/tmp/out/artifacts/speed-test-results
export MEM_ARTIFACTS_DIR=/tmp/out/artifacts/mem-test-results
export MAX_VALID_REVISIONS=2
export MAX_FETCH_REVISIONS=2
export RUNTIME_FPS_REGRESSION_TOLERANCE_PCT=10
export COMPILE_TIME_REGRESSION_TOLERANCE_PCT=10
export MEM_REGRESSION_TOLERANCE_PCT=10
export CSV_RUNTIME_FPS_PATH=/tmp/out/output/runtime_fps.csv
export CSV_COMPILE_TIME_PATH=/tmp/out/output/compile_time.csv
export CSV_MEM_PATH=/tmp/out/output/mem.csv
export CHECK_BODY_PATH=/tmp/out/output/output.md
export EXIT_CODE_REGRESSION=42
export EXIT_CODE_ALERT=43

export WANDB_ENTITY=genesis-ai-company
export WANDB_PROJECT_OLD_FORMAT=genesis-benchmarks
export WANDB_PROJECT_NEW_FORMAT=genesis-benchmarks-2
export WANDB_SILENT=true

if [[ -f ${CHECK_BODY_PATH} ]]; then {
  rm ${CHECK_BODY_PATH}
} fi

python .github/workflows/scripts/alarm.py \
  --speed-artifacts-dir "${SPEED_ARTIFACTS_DIR}" \
  --mem-artifacts-dir "${MEM_ARTIFACTS_DIR}" \
  --max-valid-revisions "${MAX_VALID_REVISIONS}" \
  --max-fetch-revisions "${MAX_FETCH_REVISIONS}" \
  --runtime-fps-regression-tolerance-pct "${RUNTIME_FPS_REGRESSION_TOLERANCE_PCT}" \
  --compile-time-regression-tolerance-pct "${COMPILE_TIME_REGRESSION_TOLERANCE_PCT}" \
  --mem-regression-tolerance-pct "${MEM_REGRESSION_TOLERANCE_PCT}" \
  --csv-runtime-fps-path "${CSV_RUNTIME_FPS_PATH}" \
  --csv-compile-time-path "${CSV_COMPILE_TIME_PATH}" \
  --csv-mem-path "${CSV_MEM_PATH}" \
  --check-body-path "${CHECK_BODY_PATH}" \
  --exit-code-regression "${EXIT_CODE_REGRESSION}" \
  --exit-code-alert "${EXIT_CODE_ALERT}" \
  --dev-allow-all-branches
  # --dev-skip-speed \

echo exit code $?
cat ${CHECK_BODY_PATH}
```

which gave output like:
```
Thresholds: runtime ± 10%, compile ± 10%, mem ± 10%

### Runtime FPS
| status | solver | backend | n_envs | env | batch_size | use_contact_island | gjk_collision | dtype | current FPS | baseline FPS [last (mean ± std)] (*1) | Δ FPS (*2) |
|:------:|:---|:---|:---|:---|:---|:---|:---|:---|---:|---:|---:|
| ℹ️ | MPM | cpu | 64 | - | - | - | - | - | 988 | --- | --- |
| ℹ️ | PBD | cpu | 128 | - | - | - | - | - | 990 | --- | --- |
| ℹ️ | PBD | gpu | 1024 | - | - | - | - | - | 985 | --- | --- |
| ℹ️ | - | cpu | - | franka | 0 | False | - | ndarray | 1,500 | 3,279 | -54.3% |
| ℹ️ | - | cuda | - | box_pyramid_6 | 4096 | False | False | ndarray | 1,200 | 5,137 | -76.6% |
**Baselines considered:** **2** commits
- Commit 1: e5639c507b17efa8cdb5801945a047ba1d22e9a2
- Commit 2: e52e73877a6ef9e338c06df05f59b53239c9f9b1

### Compile Time
| status | solver | backend | n_envs | env | batch_size | use_contact_island | gjk_collision | dtype | current compile | baseline compile [last (mean ± std)] (*1) | Δ compile (*2) |
|:------:|:---|:---|:---|:---|:---|:---|:---|:---|---:|---:|---:|
| ℹ️ | MPM | cpu | 64 | - | - | - | - | - | 2.53 | --- | --- |
| ℹ️ | PBD | cpu | 128 | - | - | - | - | - | 2.52 | --- | --- |
| ℹ️ | PBD | gpu | 1024 | - | - | - | - | - | 2.54 | --- | --- |
| ℹ️ | - | cpu | - | franka | 0 | False | - | ndarray | 1.20 | 37.76 | -96.8% |
| ℹ️ | - | cuda | - | box_pyramid_6 | 4096 | False | False | ndarray | 1.50 | 46.22 | -96.8% |
**Baselines considered:** **2** commits
- Commit 1: e5639c507b17efa8cdb5801945a047ba1d22e9a2
- Commit 2: e52e73877a6ef9e338c06df05f59b53239c9f9b1

### Memory usage
| status | env | constraint_solver | gjk_collision | batch_size | backend | dtype | current memory | baseline memory [last (mean ± std)] (*1) | Δ memory (*2) |
|:------:|:---|:---|:---|:---|:---|:---|---:|---:|---:|
| ℹ️ | anymal | Newton | False | 0 | cpu | ndarray | 234 | --- | --- |
| ℹ️ | box_pyramid_3 | - | - | 4096 | gpu | field | 150 | 1,547 | -90.3% |
| ℹ️ | franka | None | True | 30000 | cuda | field | 123 | --- | --- |
| ℹ️ | go2 | - | True | 4096 | gpu | field | 2,700 | 3,221 | -16.2% |
**Baselines considered:** **1** commits
- Commit 1: bc74cac1fa0dc6037cc87de0e38663274abf6777

- (*1) last: last commit on main, mean/std: stats over commit hashes 2 commits if available.
- (*2) Δ: relative difference between PR and last commit on main, i.e. (PR - main) / main * 100%.
```

## Checklist:

### DIFF
--- a/tests/upload_benchmarks_table_to_wandb.py
+++ b/tests/upload_benchmarks_table_to_wandb.py
@@ -109,7 +109,7 @@ def upload_results_to_wandb(
     print(f"\n✅ Upload complete: {uploaded_count} results processed, {skipped_count} skipped")
 
 
-def main():
+if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="Upload benchmark results to W&B")
     parser.add_argument("--in-file", required=True, help="Path to results file")
     parser.add_argument(
@@ -123,10 +123,6 @@ def main():
         help="Metric field names to upload (e.g., max_mem_mb compile_time runtime_fps). If not specified, all numeric fields are uploaded.",
     )
     args = parser.parse_args()
-    return upload_results_to_wandb(
+    upload_results_to_wandb(
         run_prefix=args.run_prefix, results_file_path=args.in_file, project_name=args.project, metric_names=args.metrics
     )
-
-
-if __name__ == "__main__":
-    main()


.github/workflows/scripts/alarm.py
```diff
@@ -0,0 +1,630 @@
...
+    return results
+
+
+def fmt_num(v, is_int: bool):
+    """
+    converts number to string where:
+    - ints => displays as int
+    - floats => displays to 2 decimal places
+    """
+    if v is (math.inf):
```
YilingQiao: if v == math.inf:

.github/workflows/scripts/alarm.py
```diff
@@ -0,0 +1,630 @@
...
+            print("len(commit_hashes)", len(commit_hashes))
+            if len(commit_hashes) == self.MAX_FETCH_REVISIONS:
+                break
+
+            # Early return if enough complete records have been collected
+            complete_records = [
+                benchmark_under_test.all_config_param_fdicts.issubset(record.keys())
+                for record in records_by_commit_hash.values()
+            ]
+            print("sum complete reocrds", sum(complete_records))
```
YilingQiao: reocrds -> records

.github/workflows/scripts/alarm.py
```diff
@@ -0,0 +1,630 @@
...
+        self.MEM_METRIC_KEYS = (self.metric_max_mem_mb,)  # note: make sure is a tuple
+
+        self.dev_skip_speed = args.dev_skip_speed
+        self.dev_allow_all_branches = args.dev_allow_all_branches
+
+        assert "WANDB_API_KEY" in os.environ
+
+        self.ENTITY = os.environ["WANDB_ENTITY"]
+        self.PROJECT_OLD = os.environ["WANDB_PROJECT_OLD_FORMAT"]
+        self.PROJECT_NEW = os.environ["WANDB_PROJECT_NEW_FORMAT"]
```
YilingQiao: what are them?

.github/workflows/scripts/alarm.py
```diff
@@ -0,0 +1,630 @@
...
+                row_data["baseline_last"] = int(value_last) if is_int else float(value_last)
+
+                stats_repr = f"{fmt_num(value_last, is_int)}"
+                delta_repr = f"{delta:+.1f}%"
+                if len(values_prev) >= self.MAX_VALID_REVISIONS:
+                    row_data["baseline_mean"] = int(value_ref) if is_int else float(value_ref)
+                    row_data["baseline_min"] = int(min(values_prev)) if is_int else float(min(values_prev))
+                    row_data["baseline_max"] = int(max(values_prev)) if is_int else float(max(values_prev))
+
+                    value_ci95 = statistics.stdev(values_prev) / len(values_prev) * 1.96 if len(values_prev) > 1 else math.inf
```
YilingQiao: what is this?

hughperkins: CI95. "confidence interval 95". http://www.stat.yale.edu/Courses/1997-98/101/confint.htm

hughperkins: whoops I forgot the square root . will add

.github/workflows/scripts/alarm.py
hughperkins: - old format: one metric per wandb run => very slow to query, because wandb is optimized for running very long running experiments each having many metrics and data
- new format: one wandb run for all speed metrics for a single run through all benchmarks, and another wandb run for all memory metrics for a single run through all benchmarks. Faster to query.

.github/workflows/scripts/alarm.py
hughperkins: addressed in d8069f24

.github/workflows/scripts/alarm.py
hughperkins: addressed in 0ab19574

.github/workflows/scripts/alarm.py
hughperkins: addressed in ee026061

.github/workflows/scripts/alarm.py
```diff
@@ -0,0 +1,633 @@
...
+        key_list = []
+        for col in config_param_names:
+            if col in d:
+                val = d[col]
+                key_list.append((0, val))
+            else:
+                key_list.append((1, None))
+        return key_list
+
+    return sort_key
```
duburcqa: Wait? Why are you doing this?

You should rather use:
```python
from functools import partial

def sort_key(d: dict[str, Any], config_param_names: Iterable[str]):
    nonlocal config_param_names
    key_list = []
    for col in config_param_names:
         if col in d:
            val = d[col]
            key_list.append((0, val))
        else:
            key_list.append((1, None))
    return key_list

key = partial(config_param_names=config_param_names)
```


Moreover, it is not important here but I would recommend this pattern:
```python
         val = g.get(col)
         key_list.append((int(val is None), val))
```
Unless you want to distinguish `col in d` from None value. In which case I would add a comment in the code to state this.

.github/workflows/scripts/alarm.py
```diff
@@ -0,0 +1,633 @@
...
+      (in the context of this script, this lets us ignore the values of
+      metrics during sorting)
+    - when a param_name is present in the dictionary, the tuple
+      contains (0, value), otherwise (1, None)
+
+    Since the resulting tuples will be used for sorting, the result
+    is that we will first sort the incoming dictionaries by the first param_name,
+    then the second, etc
+    - for a particular param_name, the dicts without that param_name will
+      be placed after the dicts with that param name, since 1 is after 0.
```
duburcqa: I would rather use true / false instead of 1 / 0 but not a strong opinion.

.github/workflows/scripts/alarm.py
```diff
@@ -0,0 +1,633 @@
...
+        config_param_dict: dict[str, str] = dict(  # type: ignore
+            map(str.strip, p.split("=", 1))
+            for p in line.split("|")
+            if "=" in p  # type: ignore
+        )
+        metrics: dict[str, float | int] = {}
+        for k in metric_keys:
+            try:
+                # removes metric keys from the config param dict, and adds
+                # to the metric kv dict
```
duburcqa: why 2 lines if it fits on 1?

.github/workflows/scripts/alarm.py
```diff
@@ -0,0 +1,633 @@
...
+        print("records_by_commit_hash before update:")
+        dev_dump_records_by_commit_hash(records_by_commit_hash)
+        for k, v in summary.items():
+            if k.startswith("_"):
+                continue
+            metric_name, _, kv_pairs_str = k.partition("-")
+            kv_pairs_fdict = config_params_str_to_fdict(kv_pairs_str)
+            records_by_commit_hash[commit_hash][kv_pairs_fdict][metric_name] = v
+        print("records_by_commit_hash after update:")
+        dev_dump_records_by_commit_hash(records_by_commit_hash)
```
duburcqa: I'm not a fan of all these prints but ok.

.github/workflows/scripts/alarm.py
```diff
@@ -0,0 +1,633 @@
...
+        self.all_config_param_fdicts = frozenset(self.results.keys())
+        assert self.all_config_param_fdicts
+
+        # ordered list of the config parameter names
+        self.config_param_names = merge_string_tuples(tuple((tuple(kv.keys())) for kv in self.results.keys()))
+
+
+@dataclasses.dataclass
+class Table:
+    markdown_rows: list[str]
```
duburcqa: Why are you using a dataclass of this? It seems overkill.

.github/workflows/scripts/alarm.py
```diff
@@ -0,0 +1,633 @@
...
+
+
+class Alarm:
+    def __init__(self, args: argparse.Namespace) -> None:
```
duburcqa, line 283: I don't think it is relevant to use a class of this. I would rather just go for a `main`. Object Oriented design does not bring any benefit here and is just making my life harder for debugging.

.github/workflows/scripts/alarm.py
```diff
@@ -0,0 +1,633 @@
...
+            self.metric_compile_time: Path(args.csv_compile_time_path).expanduser().resolve(),
+            self.metric_max_mem_mb: Path(args.csv_mem_path).expanduser().resolve(),
+        }
+
+        self.SPEED_METRIC_KEYS = (
+            self.metric_compile_time,
+            self.metric_runtime_fps,
+            self.metric_realtime_factor,
+        )
+        self.MEM_METRIC_KEYS = (self.metric_max_mem_mb,)  # note: make sure is a tuple
```
duburcqa: Only use UPPER_CASE formatting for global constants on top of the module if any.

.github/workflows/alarm.yml
```diff
@@ -28,379 +28,76 @@ jobs:
...
+          SPEED_ARTIFACTS_DIR: ${{ steps.dl_speed.outputs.download-path }}
+          MEM_ARTIFACTS_DIR: ${{ steps.dl_mem.outputs.download-path }}
           CHECK_BODY_PATH: check_output.md
-          CSV_RUNTIME_PATH: runtime_fps.csv
-          CSV_COMPILE_PATH: compile_time.csv
+          CSV_RUNTIME_FPS_PATH: runtime_fps.csv
+          CSV_COMPILE_TIME_PATH: compile_time.csv
+          CSV_MEM_PATH: mem.csv
           EXIT_CODE_REGRESSION: 42
           EXIT_CODE_ALERT: 43
```
duburcqa: Remove these environment variables and inline them directly in `python .github/workflows/scripts/alarm.py` command.

.github/workflows/scripts/alarm.py
hughperkins: partial is not type safe. I'm not a huge fan.

hughperkins: (partial also does not enable IDE auto-complete, which I dislike)

hughperkins: autocomplete doesnt work:
Screenshot 2026-01-23 at 02 05 33
typing doesnt work
Screenshot 2026-01-23 at 02 05 46

hughperkins: addressed key_list append in ef87fb96

hughperkins: Moved sort_key to non-nested function ,and wrapped in type-safe functor, in 9f33c96a

.github/workflows/scripts/alarm.py
```diff
@@ -0,0 +1,633 @@
...
+        key_list.append((int(val is None), val))
+    return key_list
+
+
+class SortKey:
+    def __init__(self, config_param_names: Iterable[str]) -> None:
+        self.config_param_name = config_param_names
+
+    def __call__(self, d: frozendict[str, Any]) -> list[tuple[int, int | float | None]]:
+        return sort_key(config_param_names=self.config_param_name, d=d)
```
hughperkins: I guess I could fuse `__call__` with sort_key, and just paste the contents of sort_key function into `__call__`?

.github/workflows/scripts/alarm.py
hughperkins: It makes it easier to share variables around. It is much easier/cleaner than having global variables I feel. I much prefer. This is a pretty strong opinion. What issues do you see for debugging?

.github/workflows/scripts/alarm.py
hughperkins: yes, historical, will remove

hughperkins: removed in 5277d198

.github/workflows/scripts/alarm.py
hughperkins: Ok. I can remove them.

hughperkins: prints removed in 973575e

.github/workflows/scripts/alarm.py
hughperkins: joined in 28ce3ec5

.github/workflows/scripts/alarm.py
hughperkins: addessed in bae25b04

.github/workflows/scripts/alarm.py
duburcqa: > partial is not type safe. I'm not a huge fan.

Ok for this time, but I think you need to get along with it at some point. Partial if the correct way to bind arguments to a function that you want to pass as callback or something to some other black-box function. It is designed to be efficient and implements many trick under the hood, eg nested partial elision). Moreover, it is detected by many tools and unfolded accordingly to extract the raw function pointer by JIT libraries and the likes.

.github/workflows/scripts/alarm.py
hughperkins: addressed i 041f95d9

.github/workflows/scripts/alarm.py
hughperkins: If we need it for perf sometime we can revisit at that point.

In general, I feel partial is fairly 'hacky' because of the lack of type safety etc, and I would prefer it not to be used, unless there is some strong reason, that cannot be addressed by any other means.

.github/workflows/scripts/alarm.py
```diff
@@ -0,0 +1,612 @@
...
+    return results
+
+
+def fmt_num(v, is_int: bool):
+    """
+    converts number to string where:
+    - ints => displays as int
+    - floats => displays to 2 decimal places
+    """
+    if v == math.inf:
```
duburcqa: `if not math.isfinite(v):`

.github/workflows/scripts/alarm.py
```diff
@@ -0,0 +1,612 @@
...
+        return "∞"
```
duburcqa: I'm not convinced we should print inf in this case. When is this supposed to happen? I don't think we should add branch for stuff that never occurs. Just raise an exception if you have no proof that it is necessary.

hughperkins: It happened when I was testing with --max-valid-revisions 1

hughperkins: (which I was doing because I only had a single run in wandb)

.github/workflows/scripts/alarm.py
duburcqa: > In general, I feel partial is fairly 'hacky' because of the lack of type safety etc

partial is not hacky. It is how it should be done. nonlocal and global ARE hacky. 

> because of the lack of type safety etc

Absolutely no relation between something being hacky and type safety. partial is there because typing ever existed in Python.

> If we need it for perf sometime we can revisit at that point. [...] I would prefer it not to be used, unless there is some strong reason, that cannot be addressed by any other means.

I disagree. In my view using a class with magic dunder __call__ is way more hacky than partial just to store a single parameter in a callable that would be otherwise stateless. But in this case I don't think it is worth fighting.

.github/workflows/scripts/alarm.py
duburcqa: Yes. Please do.

hughperkins: fused in f4ca5d6a

.github/workflows/scripts/alarm.py
hughperkins: addressed in fc20acd9

.github/workflows/scripts/alarm.py
duburcqa: I have a strong opinion about restricting object-oriented programming to the use cases it is really necessary. I like the clear and expressive interfaces that come with functional programming: you give inputs, you get outputs, as simple as that. Of course there are many cases where functional programming actually makes things worst, or increases complexity. Especially when it comes to user-facing interfaces. Typically I think your object oriented design for the solvers is the right approach. But here it is not the case. You instantiate a class that is instantiated once and destroyed right away in the scope of a single function. This is typically for me an anti-pattern to use class in this scenario.

When it comes to debugging. I cannot just execute main in ipython in inspect the value of the variables, I have to use actual breakpoints instead. Not a huge pain but definitely annoying.

DEPRECATION_SHIMS_README.md
duburcqa, line 1: I'm in favour of removing this file. Importing rigid solver is NOT supported. It is part of private API. I see no reason to explain users with provide backward compatibility shims for something that is an implementation details they should not even consider it exists.

.github/workflows/alarm.yml
```diff
@@ -28,379 +28,59 @@ jobs:
...
+            --max-fetch-revisions "40" \
+            --runtime-fps-regression-tolerance-pct "8" \
+            --compile-time-regression-tolerance-pct "16" \
+            --mem-regression-tolerance-pct "10" \
+            --csv-runtime-fps-path "runtime_fps.csv" \
+            --csv-compile-time-path "compile_time.csv" \
+            --csv-mem-path "mem.csv" \
+            --check-body-path "check_output.md" \
+            --exit-code-regression "42" \
+            --exit-code-alert "43"
```
duburcqa: You don't need quote around numbers. It is just weird and confusing I think.

DEPRECATION_SHIMS_README.md
hughperkins: Not sure what this file is. Prboably from Cursor, from previous version. Not intentionally included. Will remove.

hughperkins: removed in ba546cdf

.github/workflows/alarm.yml
hughperkins: removed in 62acc4b0

.github/workflows/scripts/alarm.py
duburcqa: Do you think printing inf makes sense in this case. It sounds more like a nan ("-") no?

hughperkins: yeah, I was pondering that.

hughperkins: addressed in d6fa78de

PR #2282, opened 2026-01-22
[BUG FIX] Clean up offscreen renderer when scene is destroyed.

nimrod-gileadi:
## Description

This PR ensures that `Scene.destroy()` properly cleans up camera sensor resources, including the `OffscreenRenderer` and `RasterizerContext`. Previously, destroying a scene with camera sensors could leave OpenGL resources unreleased, leading to a context leak and eventual crash.

Changes:
- Added `destroy()` method to `Simulator` that delegates to `SensorManager.destroy()`.
- Added `destroy()` method to `SensorManager` that iterates through all sensors and calls their `destroy()` methods if they exist.
- Added `destroy()` method to `RasterizerCameraSensor` that removes cameras from the renderer and cleans up shared metadata (renderer, context) when the last camera is destroyed.
- Fixed an existing typo in `Rasterizer.remove_camera()`: `removenode` → `remove_node`.
- Added null check in `RasterizerContext.remove_node()` to handle cases where visualizer doesn't exist.

Run the new tests:
```
pytest -v --forked -m required tests/test_scene_destroy.py tests/test_sensor_camera.py::test_cameras_share_metadata tests/test_sensor_camera.py::test_camera_destroy_cleans_shared_metadata tests/test_sensor_camera.py::test_multiple_cameras_destroy_cleans_shared_renderer
```

Checklist:
[x] I read the CONTRIBUTING document.
[x] I followed the Submitting Code Changes section of CONTRIBUTING document.
[x] I tagged the title correctly (including BUG FIX/FEATURE/MISC/BREAKING)
[x] I updated the documentation accordingly or no change is needed.
[x] I tested my changes and added instructions on how to test it for reviewers.
[x] I have added tests to cover my changes.
[x] All new and existing tests passed.

### DIFF
--- a/genesis/engine/sensors/camera.py
+++ b/genesis/engine/sensors/camera.py
@@ -155,6 +155,19 @@ class RasterizerCameraSharedMetadata(RigidSensorMetadataMixin, SharedSensorMetad
     # Track when rasterizer cameras were last updated
     last_render_timestep: int = -1
 
+    def destroy(self):
+        super().destroy()
+
+        if self.renderer is not None:
+            self.renderer.destroy()
+            self.renderer = None
+        if self.context is not None:
+            self.context.destroy()
+            self.context = None
+        self.lights = None
+        self.image_cache = None
+        self.sensors = None
+
 
 @dataclass
 class RaytracerCameraSharedMetadata(RigidSensorMetadataMixin, SharedSensorMetadata):
@@ -170,6 +183,13 @@ class RaytracerCameraSharedMetadata(RigidSensorMetadataMixin, SharedSensorMetada
     image_cache: Optional[Dict[int, np.ndarray]] = None
     # Track when raytracer cameras were last updated
     last_render_timestep: int = -1
+
+    def destroy(self):
+        super().destroy()
+
+        self.renderer = None
+        self.sensors = None
+        self.image_cache = None
 
 
 @dataclass
@@ -188,6 +208,14 @@ class BatchRendererCameraSharedMetadata(RigidSensorMetadataMixin, SharedSensorMe
     last_render_timestep: int = -1
     # MinimalVisualizerWrapper instance
     visualizer_wrapper: Optional["MinimalVisualizerWrapper"] = None
+
+    def destroy(self):
+        super().destroy()
+
+        self.renderer = None
+        self.sensors = None
+        self.image_cache = None
+        self.visualizer_wrapper = None
 
 
 # ========================== Base Camera Sensor ==========================
@@ -427,19 +455,6 @@ class RasterizerCameraSensor(BaseCameraSensor):
         w, h = self._options.res
         self._shared_metadata.image_cache[self._idx] = torch.zeros((_B, h, w, 3), dtype=torch.uint8, device=gs.device)
 
-    @classmethod
-    def destroy(cls, shared_metadata: RasterizerCameraSharedMetadata):
-        if not shared_metadata.sensors:
-            return
-
-        shared_metadata.renderer.destroy()
-        shared_metadata.renderer = None
-        shared_metadata.context.destroy()
-        shared_metadata.context = None
-        shared_metadata.lights = None
-        shared_metadata.image_cache = None
-        shared_metadata.sensors = None
-
     def _create_standalone_context(self, scene):
         """Create a simplified RasterizerContext for camera sensors."""
         if not scene.sim._rigid_only and scene.n_envs > 1:
@@ -644,15 +659,6 @@ class RaytracerCameraSensor(BaseCameraSensor):
         w, h = self._options.res
         self._shared_metadata.image_cache[self._idx] = torch.zeros((_B, h, w, 3), dtype=torch.uint8, device=gs.device)
 
-    @classmethod
-    def destroy(cls, shared_metadata: RaytracerCameraSharedMetadata):
-        if not shared_metadata.sensors:
-            return
-
-        shared_metadata.renderer = None
-        shared_metadata.sensors = None
-        shared_metadata.image_cache = None
-
     @gs.assert_built
     def move_to_attach(self):
         # Bypass original implementation since it will be handled by visualizer
@@ -774,16 +780,6 @@ class BatchRendererCameraSensor(BaseCameraSensor):
         _B = max(self._manager._sim.n_envs, 1)
         w, h = self._options.res
         self._shared_metadata.image_cache[self._idx] = torch.zeros((_B, h, w, 3), dtype=torch.uint8, device=gs.device)
-
-    @classmethod
-    def destroy(cls, shared_metadata: BatchRendererCameraSharedMetadata):
-        if not shared_metadata.sensors:
-            return
-
-        shared_metadata.renderer = None
-        shared_metadata.sensors = None
-        shared_metadata.image_cache = None
-        shared_metadata.visualizer_wrapper = None
 
     def _render_current_state(self):
         """Perform the actual render for the current state."""
--- a/genesis/engine/sensors/sensor_manager.py
+++ b/genesis/engine/sensors/sensor_manager.py
@@ -75,9 +75,10 @@ class SensorManager:
                 sensor._is_built = True
 
     def destroy(self):
-        for sensor_cls, _sensors in self._sensors_by_type.items():
-            sensor_cls.destroy(self._sensors_metadata[sensor_cls])
-            del self._sensors_metadata[sensor_cls]
+        for sensors_metadata in self._sensors_metadata.values():
+            if sensors_metadata is not None:
+                sensors_metadata.destroy()
+        self._sensors_metadata.clear()
         self._sensors_by_type.clear()
 
     def reset(self, envs_idx=None):
--- a/tests/test_sensor_camera.py
+++ b/tests/test_sensor_camera.py
@@ -1,6 +1,5 @@
-import gc
-import weakref
 import sys
+import weakref
 
 import numpy as np
 import pytest
@@ -359,7 +358,6 @@ def test_rasterizer_destroy():
     offscreen_renderer_ref = weakref.ref(cam1._shared_metadata.renderer._renderer)
 
     scene.destroy()
-    gc.collect()
     assert offscreen_renderer_ref() is None
 
 
--- a/genesis/engine/sensors/base_sensor.py
+++ b/genesis/engine/sensors/base_sensor.py
@@ -50,6 +50,20 @@ class SharedSensorMetadata:
     cache_sizes: list[int] = field(default_factory=list)
     delays_ts: torch.Tensor = make_tensor_field((0, 0), dtype_factory=lambda: gs.tc_int)
 
+    def __del__(self):
+        try:
+            self.destroy()
+        except Exception:
+            pass
+
+    def destroy(self):
+        """
+        Destroy shared metadata.
+
+        This method is called by SensorManager when the scene is destroyed. his should remove any references to the
+        sensors from the shared metadata, and clean up any resources associated with the sensors.
+        """
+
 
 SharedSensorMetadataT = TypeVar("SharedSensorMetadataT", bound=SharedSensorMetadata)
 
@@ -113,16 +127,6 @@ class Sensor(RBC, Generic[SharedSensorMetadataT]):
             dim=1,
         )
         self._shared_metadata.cache_sizes.append(self._cache_size)
-
-    @classmethod
-    def destroy(cls, shared_metadata: SharedSensorMetadataT):
-        """
-        Destroy the sensor.
-
-        This method is called by SensorManager when the scene is destroyed.
-        This should remove any references to the sensors from the shared metadata,
-        and clean up any resources associated with the sensors.
-        """
 
     @classmethod
     def reset(cls, shared_metadata: SharedSensorMetadataT, envs_idx):


genesis/engine/sensors/camera.py
```diff
@@ -427,6 +427,23 @@ def build(self):
         w, h = self._options.res
         self._shared_metadata.image_cache[self._idx] = torch.zeros((_B, h, w, 3), dtype=torch.uint8, device=gs.device)
 
+    def destroy(self):
+        if not self._shared_metadata.sensors:
+            return
+        self._shared_metadata.sensors.remove(self)
+        self._shared_metadata.renderer.remove_camera(self._camera_wrapper)
+        # TODO: should remove the camera from the renderer too.
```
YilingQiao: why is it TODO? Isn't it the line above already removing camera?

genesis/engine/sensors/camera.py
```diff
@@ -631,6 +648,12 @@ def build(self):
         w, h = self._options.res
         self._shared_metadata.image_cache[self._idx] = torch.zeros((_B, h, w, 3), dtype=torch.uint8, device=gs.device)
 
+    def destroy(self):
+        if not self._shared_metadata.sensors:
+            return
+        self._shared_metadata.sensors.remove(self)
+        # FIXME: there are probably other cleanups that should be done here, if this is the last sensor.
```
YilingQiao: what could be the other cleanups? Any known issue?

genesis/engine/sensors/sensor_manager.py
```diff
@@ -74,6 +74,13 @@ def build(self):
                 sensor.build()
                 sensor._is_built = True
 
+    def destroy(self):
+        for _sensor_cls, sensors in self._sensors_by_type.items():
+            for sensor in sensors:
+                if hasattr(sensor, "destroy"):
```
YilingQiao: Should the destroy() method be added to BaseSensor as an default no-op method to avoid the hasattr check?

genesis/engine/scene.py
```diff
@@ -303,6 +303,10 @@ def destroy(self):
             self._visualizer.destroy()
             self._visualizer = None
 
+        if getattr(self, "_sim", None) is not None:
```
YilingQiao, line 306: Is there a situation _sim might not exist as an attribute (requiring getattr)?

tests/test_sensor_camera.py
```diff
@@ -317,6 +317,162 @@ def test_batch_renderer(n_envs, png_snapshot):
...
 
 
+@pytest.mark.required
+def test_destroy_unbuilt_scene_with_camera():
+    """Test that destroy on an unbuilt scene with cameras doesn't crash."""
+    scene = gs.Scene(show_viewer=False)
+    scene.add_entity(morph=gs.morphs.Plane())
+    scene.add_sensor(gs.sensors.RasterizerCameraOptions(res=(64, 64)))
+
+    # Destroy without building - should not crash
```
duburcqa: I think it should "crash" actually. In a sense that only a scene that was build should be destroyed. Here it should raise an exception. Of course, no rendering backend should be started before build.

tests/test_sensor_camera.py
```diff
@@ -317,6 +317,162 @@ def test_batch_renderer(n_envs, png_snapshot):
...
+    """Test that calling destroy twice on a scene with cameras doesn't crash."""
+    scene = gs.Scene(show_viewer=False)
+    scene.add_entity(morph=gs.morphs.Plane())
+    camera = scene.add_sensor(gs.sensors.RasterizerCameraOptions(res=(64, 64)))
+
+    scene.build()
+    camera.read()
+
+    scene.destroy()
+    scene.destroy()  # Should not crash
```
duburcqa: Same. I think it should crash.

tests/test_sensor_camera.py
```diff
@@ -317,6 +317,162 @@ def test_batch_renderer(n_envs, png_snapshot):
...
+
+    cam1 = scene.add_sensor(gs.sensors.RasterizerCameraOptions(res=(64, 64)))
+    cam2 = scene.add_sensor(gs.sensors.RasterizerCameraOptions(res=(32, 32)))
+
+    scene.build()
+    cam1.read()
+    cam2.read()
+
+    assert cam1._shared_metadata is cam2._shared_metadata
+    assert len(cam1._shared_metadata.sensors) == 2
```
duburcqa: I don't think this test is relevant but I may be wrong. What are you trying to enforce here? It looks like implementation details.

tests/test_sensor_camera.py
```diff
@@ -317,6 +317,162 @@ def test_batch_renderer(n_envs, png_snapshot):
...
+
+
+@pytest.mark.required
+def test_camera_destroy_cleans_shared_metadata():
+    import gc
+    import weakref
+
+    scene = gs.Scene(show_viewer=False)
+    scene.add_entity(morph=gs.morphs.Plane())
```
duburcqa: You don't need to do this.

tests/test_sensor_camera.py
```diff
@@ -317,6 +317,162 @@ def test_batch_renderer(n_envs, png_snapshot):
...
+    cam2.read()
+
+    assert cam1._shared_metadata is cam2._shared_metadata
+    assert len(cam1._shared_metadata.sensors) == 2
+
+
+@pytest.mark.required
+def test_camera_destroy_cleans_shared_metadata():
+    import gc
+    import weakref
```
duburcqa: No local imports except absolutely necessary.

tests/test_sensor_camera.py
```diff
@@ -317,6 +317,162 @@ def test_batch_renderer(n_envs, png_snapshot):
...
+
+    camera = scene.add_sensor(gs.sensors.RasterizerCameraOptions(res=(64, 64)))
+
+    scene.build()
+    camera.read()
+
+    shared_metadata = camera._shared_metadata
+    renderer = shared_metadata.renderer
+    offscreen_renderer = renderer._renderer
+    offscreen_renderer_ref = weakref.ref(offscreen_renderer)
```
duburcqa: I would rather keep it short to avoid distraction and make it easier to understand what this test is doing at first glance:
```python
offscreen_renderer_ref = weakref.ref(camera._shared_metadata.renderer._renderer)
```

tests/test_sensor_camera.py
```diff
@@ -317,6 +317,162 @@ def test_batch_renderer(n_envs, png_snapshot):
...
+
+    scene.destroy()
+
+    assert shared_metadata.renderer is None
+    assert shared_metadata.context is None
+    assert shared_metadata.sensors is None
+    assert renderer._renderer is None
+
+    del offscreen_renderer
```
duburcqa: Following my previous recommendation, this del is no longer necessary.

tests/test_sensor_camera.py
```diff
@@ -317,6 +317,162 @@ def test_batch_renderer(n_envs, png_snapshot):
...
+    gc.collect()
```
duburcqa: In my view it is a requirement to release memory right away without waiting for garbage collection. If someone creates another scene right away, it may cause OpenGL related race condition or similar.

tests/test_sensor_camera.py
```diff
@@ -317,6 +317,162 @@ def test_batch_renderer(n_envs, png_snapshot):
...
+    assert offscreen_renderer_ref() is None
+
+
+@pytest.mark.required
+def test_multiple_cameras_destroy_cleans_shared_renderer():
+    import gc
+    import weakref
```
duburcqa: I would rather add a single destroy test that will combine this one and the previous one.

tests/test_sensor_camera.py
```diff
@@ -317,6 +317,162 @@ def test_batch_renderer(n_envs, png_snapshot):
...
+    assert offscreen_renderer_ref() is None
+
+
+@pytest.mark.required
+@pytest.mark.parametrize("backend", [gs.cuda])
+@pytest.mark.skipif(not ENABLE_MADRONA, reason="BatchRenderer is not supported because 'gs_madrona' is not available.")
+def test_batch_renderer_destroy():
+    scene = gs.Scene(show_viewer=False)
+    scene.add_entity(morph=gs.morphs.Plane())
+    scene.add_entity(morph=gs.morphs.Sphere(pos=(0.0, 0.0, 1.0)))
```
duburcqa: I don't think adding this is relevant.

tests/test_sensor_camera.py
```diff
@@ -317,6 +317,162 @@ def test_batch_renderer(n_envs, png_snapshot):
...
+        renderer=gs.renderers.RayTracer(
+            env_surface=gs.surfaces.Emission(
+                emissive_texture=gs.textures.ColorTexture(color=(0.2, 0.3, 0.5)),
+            ),
+            env_radius=20.0,
+        ),
+        show_viewer=False,
+    )
+    scene.add_entity(morph=gs.morphs.Plane())
+    scene.add_entity(morph=gs.morphs.Sphere(pos=(0.0, 0.0, 1.0)))
```
duburcqa: No need to do this.

genesis/engine/sensors/camera.py
```diff
@@ -427,6 +427,23 @@ def build(self):
...
+        # A sort of reference counting to destroy the renderer once there are no more cameras.
+        if not self._shared_metadata.sensors:
+            self._shared_metadata.renderer.destroy()
+            self._shared_metadata.renderer = None
+            self._shared_metadata.context.destroy()
+            self._shared_metadata.context = None
+            self._shared_metadata.lights = None
+            self._shared_metadata.image_cache = None
+            # Make sure a new one is created next time a camera is built.
+            self._shared_metadata.sensors = None
```
duburcqa: What does this means? I don't think it sound be possible to manually destroy a camera. This should rather be `__del__` and the sensor manager forcibly calling `del camera`. Otherwise the user may assume destroying camera is allowed and part of the public API. You may also add some `_destroy` prefix to make it "private" if you prefer.

genesis/engine/sensors/camera.py
```diff
@@ -631,6 +648,12 @@ def build(self):
         w, h = self._options.res
         self._shared_metadata.image_cache[self._idx] = torch.zeros((_B, h, w, 3), dtype=torch.uint8, device=gs.device)
 
+    def destroy(self):
+        if not self._shared_metadata.sensors:
+            return
+        self._shared_metadata.sensors.remove(self)
```
duburcqa: I don't think it should work this way. Either all sensors are destroyed or not. They should never be destroyed one by one. Destroy should be a classmethod acting only on shared data basically.

genesis/engine/sensors/camera.py
```diff
@@ -753,6 +776,12 @@ def build(self):
...
+        # FIXME: there are probably other cleanups that should be done here, if this is the last sensor.
+
```
duburcqa: Same as what I said above.

genesis/engine/sensors/sensor_manager.py
duburcqa: Yes I agree.

genesis/engine/scene.py
duburcqa: Yes, if initializing the scene fails for whatever reason. You need to be very careful in a destructor because it is impossible to anticipate every possible scenario and this method is going to be called systematically, no matter if the scene was properly initialised or not.

genesis/engine/simulator.py
```diff
@@ -206,6 +206,9 @@ def build(self):
 
         self._sensor_manager.build()
 
+    def destroy(self):
+        self._sensor_manager.destroy()
+
```
duburcqa, line 211: Same. This should probably not be public API. But it is much less of a problem because `Simulator` object is not supposed to be exposed to the user, so it is acceptable.

tests/test_sensor_camera.py
nimrod-gileadi: I added a comment to explain why I think it's necessary. Because `__del__` calls `destroy`, a crash would result in log space (uncaught exception), and garbage collection before build() or after destroy() is expected.

tests/test_sensor_camera.py
nimrod-gileadi: Added a comment.

tests/test_sensor_camera.py
nimrod-gileadi: Removed.

tests/test_sensor_camera.py
nimrod-gileadi: Done.

tests/test_sensor_camera.py
nimrod-gileadi: Actually add_sensor(RasterizerCameraOptions) was crashing if there's no entity in the scene. Now fixed with a change in genesis/options/sensors/options.py

tests/test_sensor_camera.py
nimrod-gileadi: Done.

tests/test_sensor_camera.py
nimrod-gileadi: The code does release the memory right away. I could check that by holding a reference to OffscreenRenderer and checking the private member _platform is cleared (to show that OffscreenRenderer.delete() was called). Would that be better?

tests/test_sensor_camera.py
nimrod-gileadi: Done.

tests/test_sensor_camera.py
nimrod-gileadi: Done.

tests/test_sensor_camera.py
nimrod-gileadi: Done.

genesis/engine/sensors/camera.py
nimrod-gileadi: Oops - leftover from development.

genesis/engine/sensors/camera.py
nimrod-gileadi: I understand your suggestion below to make this classmethod and will try now.

My proposed Sensor interface was that `build()` can be called at some point, and it will check for the existence of `shared_metadata.sensors` to decide whether it's the first sensor to be created and whether initialization can happen.
For symmetry, `destroy()` can be called at some point, and it will check whether it's the last sensor to be destroyed to decide whether to release the resources.

I don't think using `__del__` is right, because it's not guaranteed to be called immediately (though it normally is with CPython), and it might mysteriously stop working if there's code that holds a reference to a sensor outside the scene.

genesis/engine/sensors/camera.py
nimrod-gileadi: Updated to remove the circular reference in shared_metadata.

nimrod-gileadi:
I resolved the comments which didn't require a big restructuring of the code. Now I'll try to apply Alexis' suggestion of using a class method for destroying sensors with shared_metadata and will let you know how it goes.

genesis/engine/sensors/camera.py
nimrod-gileadi: Updated to make destroy a classmethod.

genesis/engine/sensors/camera.py
nimrod-gileadi: Done.

genesis/engine/simulator.py
nimrod-gileadi: Acknowledged. I'll leave it unless you insist.

genesis/engine/sensors/camera.py
duburcqa: > I don't think using __del__ is right, because it's not guaranteed to be called immediately

This is wrong. Combined with `del` it will be called right away as I mentioned.

tests/test_sensor_camera.py
duburcqa: `__del__` could call destroy only if built. We are tracking this already. But it may just add unnecessary complexity, make users' life harder, and after checking, we are already officially supporting doing this. So let's stick to what we are already doing.

tests/test_sensor_camera.py
duburcqa: > Actually add_sensor(RasterizerCameraOptions) was crashing if there's no entity in the scene. Now fixed with a change in genesis/options/sensors/options.py

WoW this is a bug that should be fixed. Well spotted.

tests/test_sensor_camera.py
duburcqa: If it releases memory right away, why would you need to change anything exception removing call to `gs.collect()` ?

duburcqa: If you mean because your check is a bit loose, yes it may be nice to make it more strict and check whether `_platform` is cleared.

genesis/engine/sensors/camera.py
duburcqa: Much like RigidEntity and the likes, sensors should be empty shell not owning "data" by just exposing them. The scene-global shared memory should be deleted by the sensor manager, without involving the sensors themselves.

PR #2067, opened 2025-12-03
[FEATURE] Add support of USD import for Rigid Body.

alanray-tech:
## Description

This PR adds comprehensive USD (Universal Scene Description) import functionality to Genesis, enabling users to directly import USD stages containing articulated structures and rigid bodies into Genesis simulation scenes.

The implementation includes:

- **USD Parser System**: A modular parser architecture with separate parsers for materials, articulations, and rigid bodies
- **Articulation Support**: Full parsing of USD articulations with support for multiple joint types (revolute, prismatic, spherical, fixed)
- **Rigid Body Support**: Parsing of standalone rigid bodies with collision and visual geometry
- **Material Parsing** (Placeholder, need further development): Automatic extraction and conversion of USD rendering materials 
- **Scene Integration**: New `Scene.add_stage()` method for easy USD file import
- **New Morph Types**: `USDArticulation` and `USDRigidBody` morph classes for USD-based entities

Key components:
- `UsdParser`: Main parser entrance with `import_from_stage()` method
- `UsdArticulationParser`: Extracts and parses articulation structures from USD stages
- `UsdRigidBodyParser`: Extracts and parses rigid body prims from USD stages
- `UsdRenderingMaterialParser`: Parses USD materials and shaders
- `UsdParserContext`: Context manager for tracking parsed entities and materials
- `UsdParserUtils`: Utility functions for transform computation, mesh conversion, and coordinate system transformations

## Related issues

Resolves https://github.com/Genesis-Embodied-AI/Genesis/issues/1419

## Motivation and Context

USD is a widely-used format in the robotics and simulation community, particularly in tools like NVIDIA Isaac Sim, Omniverse, and other industry-standard platforms. Adding USD import capability allows Genesis users to:

1. Import assets directly from USD files without manual conversion
2. Leverage existing USD asset libraries and pipelines
3. Work seamlessly with USD-based workflows and tools
4. Import complex articulated structures (robots, mechanisms) with proper joint definitions
5. Import rigid bodies with correct collision and visual geometry

This feature significantly improves Genesis's interoperability with the broader robotics and simulation ecosystem.

## How Has This Been / Can This Be Tested?

The feature can be tested using the provided example script:

```python
import genesis as gs
gs.init(backend=gs.cpu)

scene = gs.Scene(
    viewer_options=gs.options.ViewerOptions(
        camera_pos=(3.5, 0.0, 2.5),
        camera_lookat=(0.0, 0.0, 0.5),
        camera_fov=40,
[...]

PR #2286, opened 2026-01-22
[Feature] Support Rendering Textures for USD Articulation and Other Refactors. 

ACMLCZH:
Based on #2067 
- **Supported Features:**
  - Support Rendering Textures for USD Articulations:
   kitchen_cam1_step000

  - Support `scale` for USD stages and morphs.
  - Support `.usdz` file parsing.
  - Support `GeomSubset` rendering.
  - USD baking can be used in python 3.11 and 3.12 now.

- **API changes:**
  - `scene.add_stage` now accepts `gs.morphs.USD` as `stage`, `material` and `surface`. It parses all trees with `ArticulationRootAPI`, `RigidBodyAPI` and `CollisionAPI` roots as rigid entities and then add them into the scene. `material` and `surface` will be applied to all rigid entities parsed from USD stage. It returns a `list` of `gs.Entity` instead of a `dict` because the prim path can be retrieved from `entity.name`.
  - `scene.add_entity` also accepts `gs.morphs.USD` as `morph`. You can add a specified `prim` or the default `prim` in the USD file into the scene as a rigid entity. Note that the transforms of this prim inside the file will be parsed as the root link poses.
  - `gs.morphs.USD` updated paramters:
    - `prim_path`: the path to the prim you want to add in the USD stage. If set to None, use the default prim in the USD file.
    - `scale`: now the entity-level scale can be applied to the USD entities.

- **Refactors:**
  - Move shared data into `UsdContext`, including:
    - An `XformCache` for speeding up computing transforms.
    - Parsed material properties shared across different entities.
    - The stage's metadata (e.g. `up_axis`, `meters_per_unit`, etc.)
  - Rewrite visual and collision geoms detection logic, now it aligns with both articulation and mesh-only USD assets.
  - Rename most of the files and classes. 
  - Rewrite `UsdParserSpec.md`.

- **TODOs:**
  - Parse geom frictions (and other properties) with `physics` materials.
  - Parse link/joint physics properties in a more appropriate way.
  - Parse articulation structures more precisely.
  - Determine the default density of Articulations.
  - Some speeding up operations.


genesis/engine/scene.py
```diff
@@ -373,9 +373,14 @@ def add_entity(
...
 
-        if isinstance(morph_for_checks, (gs.morphs.URDF, gs.morphs.MJCF, gs.morphs.Terrain)):
+        if isinstance(morph_for_checks, (gs.morphs.URDF, gs.morphs.MJCF, gs.morphs.USD, gs.morphs.Terrain)):
             if not isinstance(material, (gs.materials.Rigid, gs.materials.Hybrid)):
                 gs.raise_exception(f"Unsupported material for morph: {material} and {morph_for_checks}.")
+            if isinstance(material, gs.materials.Hybrid):
+                # TODO: there should be a way to distinguish between custom overrided rho and default rho
+                material._material_rigid._rho = 1000.0
+            else:
+                material._rho = 1000.0  # Default density for URDF, MJCF and USD
```
YilingQiao: why do you overwrite this rho for URDF/MJCF/USD?

genesis/options/morphs.py
```diff
@@ -1045,15 +1048,15 @@ class Drone(FileMorph):
         sufficient to met the desired accuracy (see 'decompose_(robot|object)_error_threshold' documentation). The
         module 'coacd' is used for this decomposition process. If not given, it defaults to `True` for `RigidEntity`
         and `False` for other deformable entities.
-    decompose_nonconvex : bool, optional
+    decompose_nonconvex : float, optional
```
YilingQiao: why change to float?

genesis/engine/scene.py
```diff
@@ -464,13 +469,46 @@ def add_entity(
...
         visualize_contact: bool = False,
+        vis_mode: Literal["visual", "collision"] = "visual",
     ):
-        from genesis.utils.usd import import_from_stage
+        """
+        Add a stage to the scene.
+
+        Parameters
+        ----------
+        stage : gs.morphs.Stage
```
YilingQiao: should it be gs.morphs.USD according to the parameter type

genesis/engine/scene.py
ACMLCZH: Because MJCF uses 1000.0 as the default density. When inertia_mass is not specified in the mjcf file, mujoco will automatically use 1000.0 as density and the volume to compute the mass of the link. And in order to align with MJCF, I set the default rho of USD rigid entities also to 1000.0. The original default density 200.0 does not work.

genesis/utils/usd/usd_context.py
```diff
@@ -0,0 +1,324 @@
...
+
+from .usd_material import parse_material_preview_surface
+from .usd_parser_utils import extract_scale
+
+# Check for Omniverse Kit support (required for USD baking)
+# Note: CI workflows should set OMNI_KIT_ACCEPT_EULA=yes in their env section
+try:
+    import omni.kit_app
+
+    device, *_ = gs.utils.get_device(gs.cuda)
```
YilingQiao: why do you need to get device during module import time? Is this useful at all?

genesis/engine/scene.py
```diff
@@ -464,13 +469,46 @@ def add_entity(
...
+            morphs = parse_usd_stage(stage)
+        else:
+            gs.raise_exception(f"Unsupported stage: {stage}.")
+
+        entities = []
+        for morph in morphs:
+            entities.append(self.add_entity(morph, material, surface, visualize_contact, vis_mode))
 
-        return import_from_stage(self, morph.file, vis_mode, morph, visualize_contact)
+        return entities
```
YilingQiao, line 515: scene.add_stage now returns list[Entity] instead of dict[str, Entity]? Why do we change it this way? dictionary is more convenient for managing the entities

genesis/utils/usd/usd_context.py
ACMLCZH: Because USD baking force to use CUDA. Request CUDA here to check the availability of USD baking. Or I should choose a more appropriate position to check it?

genesis/engine/scene.py
ACMLCZH: First, the dict keys are prim paths, which are hard to understand for users. Additionally, it can be retrieved from entity.morph.prim_path if users understand it.

genesis/engine/scene.py
ACMLCZH: @duburcqa  MuJoCo's default geom density is 1000.0. If an MJCF asset doesn't explicitly author a density value, MuJoCo uses this default value together with the mesh volume to compute the link mass. As a result, when we import such an MJCF entity, its effective density/mass can differ from what the user specified via `gs.materials.Rigid.rho`. Do you consider this a bug/issue on our side, or expected behavior?

genesis/engine/scene.py
ACMLCZH: Using `dict` will also provide a well-defined list and be easy for some users for refer required entities.
@duburcqa What do you think about the API design here?

genesis/engine/scene.py
```diff
@@ -376,11 +376,6 @@ def add_entity(
         if isinstance(morph_for_checks, (gs.morphs.URDF, gs.morphs.MJCF, gs.morphs.USD, gs.morphs.Terrain)):
             if not isinstance(material, (gs.materials.Rigid, gs.materials.Hybrid)):
                 gs.raise_exception(f"Unsupported material for morph: {material} and {morph_for_checks}.")
-            if isinstance(material, gs.materials.Hybrid):
-                # TODO: there should be a way to distinguish between custom overrided rho and default rho
-                material._material_rigid._rho = 1000.0
-            else:
-                material._rho = 1000.0  # Default density for URDF, MJCF and USD
```
duburcqa: Ok I will take care of fixing this. Good spot.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -3,6 +3,7 @@
 from itertools import chain
 from typing import TYPE_CHECKING, Literal, Any
 from functools import wraps
+import os
```
duburcqa: We are (manually sadly...) organising imports. Here is the rule: You should first import standards libraries, skip line, then ubiquitous third-party libraries (torch, numpy, tqdm...etc), skip line, then random third-party libraries, skip line, then our own custom libraries, skip line, then relative genesis imports. For each group, make sure it is sorted alphabetically.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -607,10 +608,10 @@ def _load_scene(self, morph, surface):
                     + str(e).replace("\n", " - ")
                 )
         elif isinstance(morph, gs.morphs.USD):
-            from genesis.utils.usd import parse_usd_rigid_entity
+            import genesis.utils.usd as usu
 
             # Unified parser handles both articulations and rigid bodies
-            l_infos, links_j_infos, links_g_infos, eqs_info = parse_usd_rigid_entity(morph, surface)
+            l_infos, links_j_infos, links_g_infos, eqs_info = usu.parse_usd(morph, surface)
```
duburcqa: Remove this change. It does not bring any benefit.

genesis/engine/mesh.py
```diff
@@ -63,9 +63,9 @@ def __init__(
 
         if self._surface.requires_uv():  # check uvs here
             if self._uvs is None:
-                if "mesh_path" in metadata:
+                if "mesh_path" in self._metadata:
                     gs.logger.warning(
-                        f"Texture given but asset missing uv info (or failed to load): {metadata['mesh_path']}"
+                        f"Texture given but asset missing uv info (or failed to load): {self._metadata['mesh_path']}"
```
duburcqa, line 82: Why this change?

ACMLCZH: Because at this point `metadata` can still be `None`.

genesis/engine/scene.py
duburcqa: I tend to think returning a flat sequence makes sense because for now our entities are name-less.

examples/usd/import_stage.py
```diff
@@ -79,7 +79,7 @@ def main():
         gs.morphs.Plane(),
     )
     entities = scene.add_stage(
-        morph=gs.morphs.USD(
+        stage=gs.morphs.USD(
```
duburcqa: Revert this change. `gs.morphs.USD` is a morph not a stage.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -607,10 +608,10 @@ def _load_scene(self, morph, surface):
                     + str(e).replace("\n", " - ")
                 )
         elif isinstance(morph, gs.morphs.USD):
-            from genesis.utils.usd import parse_usd_rigid_entity
+            from genesis.utils.usd import parse_usd
 
             # Unified parser handles both articulations and rigid bodies
-            l_infos, links_j_infos, links_g_infos, eqs_info = parse_usd_rigid_entity(morph, surface)
+            l_infos, links_j_infos, links_g_infos, eqs_info = parse_usd(morph, surface)
```
duburcqa: Why are you changing the name? I don't understand. Can you load something different from a rigid entity? I don't think so for now. Once we will support rendering-only elements such as cameras it will be different, but based on the name of this PR it is not the case.

genesis/engine/mesh.py
duburcqa: You are right.

genesis/engine/mesh.py
```diff
@@ -350,10 +350,6 @@ def from_morph_surface(cls, morph, surface=None):
                     for mesh in meshes:
                         mesh.convert_to_zup()
                     gs.logger.debug(f"Converting the geometry of the '{morph.file}' file to zup.")
-            elif morph.is_format(gs.options.morphs.USD_FORMATS):
-                import genesis.utils.usd.usda as usda_utils
-
-                meshes = usda_utils.parse_mesh_usd(morph.file, morph.group_by_material, morph.scale, surface)
```
duburcqa, line 375: Why is this no longer useful? 

USD mesh is no longer supported? It is because it does not really exist and everything is basically just a stage. Considering them as morph Mesh is basically incorrect?

genesis/engine/scene.py
```diff
@@ -464,13 +464,46 @@ def add_entity(
...
+        stage : gs.morphs.USD
+            The stage to add to the scene.
+        material : gs.materials.Material | None, optional
+            The material of the stage. If None, use ``gs.materials.Rigid()`` for all morphs.
+        surface : gs.surfaces.Surface | None, optional
+            The surface of the stage. If None, use ``gs.surfaces.Default()`` for all morphs.
+        visualize_contact : bool
+            Whether to visualize contact forces applied to this stage as arrows in the viewer and rendered images. Note that this will not be displayed in images rendered by camera using the `RayTracer` renderer.
+        vis_mode : str | None, optional
+            The visualization mode of the stage. This is a handy shortcut for setting `surface.vis_mode` without explicitly creating a surface object.
```
duburcqa: Do not exceed 120 chars linewidth.

genesis/engine/entities/rigid_entity/rigid_entity.py
ACMLCZH: I renamed it just to align with `uu.parse_urdf` and `mju.parse_xml`. So should I add the detail `rigid_entity` here?

genesis/engine/scene.py
```diff
@@ -464,13 +464,46 @@ def add_entity(
...
+        entities : List[genesis.Entity]
+            The created entities.
+        """
+        morphs = []
+        if isinstance(stage, gs.morphs.USD):
+            from genesis.utils.usd import parse_usd_stage
+
+            morphs = parse_usd_stage(stage)
+        else:
+            gs.raise_exception(f"Unsupported stage: {stage}.")
```
duburcqa: Still, unsupported "morph" not "stage". USD morph describes a stage but it is still a morph. 

By the way, it would be nice to clarify that this method only support USD formats.

genesis/engine/scene.py
```diff
@@ -464,13 +464,46 @@ def add_entity(
...
+
+        entities = []
+        for morph in morphs:
+            entities.append(self.add_entity(morph, material, surface, visualize_contact, vis_mode))
```
duburcqa: I don't understand what is happening here, `parse_usd_stage` creates sub-morphs that you need to add manually afterward?

genesis/engine/mesh.py
ACMLCZH: `usda.py` is also a complete pipeline to parse usd file (but only the whole mesh without rigid properties), reasonably there should be only one pipeline to do USD parsing. In PR #2067 , Genesis uses `gs.morphs.Mesh()` to parse USD mesh-only and `gs.morphs.USD()` to parse USD articulations. While in this PR, we are able to handle both cases with only `gs.morphs.USD()`, so it is no longer used.

genesis/engine/scene.py
ACMLCZH: `parse_usd_stage` truly returns list of `morphs`. I think there is no need for `parse_usd_stage` to touch the `entity` level. It is better to put all `morphs` processing and `entity` creating inside `scene`.

genesis/options/morphs.py
```diff
@@ -612,7 +613,9 @@ def _repr_type(self):
         return f"<gs.morphs.{self.__class__.__name__}(file='{self.file}')>"
 
     def is_format(self, format):
-        return self.file.lower().endswith(format)
+        if not isinstance(self.file, (str, os.PathLike, Path)):
+            return False
+        return str(self.file).lower().endswith(format)
```
duburcqa, line 615: I think you should just raise an exception. You may re-raise it with some clearer error message, but accommodating clearly invalid and unsupported argument types is not a good approach.

genesis/options/morphs.py
```diff
@@ -1394,51 +1397,23 @@ class USD(FileMorph):
...
     visual_mesh_prim_patterns : List[str], optional
         List of regex patterns to match visual mesh prim names. Patterns are tried in order.
-        Defaults to [r"^([vV]isual).*", r"^.*"].
-
-    Geometry Decomposition Options
-    -------------------------------
-    convexify : bool, optional
-        Whether to convexify the entity. When convexify is True, all the meshes in the entity will each be converted
-        to a set of convex hulls. The mesh will be decomposed into multiple convex components if the convex hull is not
-        sufficient to meet the desired accuracy. The module 'coacd' is used for this decomposition process.
```
duburcqa, line 1405: Why are you remove this docstring?

genesis/options/morphs.py
```diff
@@ -1394,51 +1397,23 @@ class USD(FileMorph):
...
     ----------------
     prim_path : str, optional
         The parsing target prim path. Defaults to None.
-    parser_ctx : Any, optional
+    usd_ctx : Any, optional
         The parser context. Defaults to None.
+    geometry_only: bool, optional
+        Whether to skip rigid/articulation validation for the specified prim. Defaults to False.
+    use_bake_cache: bool, optional
+        Whether to use cached baked assets. Defaults to True.
```
duburcqa: What are these parameters? Why are they useful? Is it possible to remove them?

genesis/options/morphs.py
duburcqa: You should keep the doc even it is inherited from the based class.

genesis/utils/usd/__init__.py
```diff
@@ -1,24 +1,13 @@
...
-
-This package provides an extendable USD parser system with:
-- UsdParserUtils: Utility functions for transforms, mesh conversion, etc.
-- UsdParserContext: Context for tracking materials, articulations, and rigid bodies
-- parse_all_materials: Function for parsing rendering materials
-- parse_usd_rigid_entity: Unified parser for both articulations and rigid bodies
-- import_from_stage: Main parser function for importing from a USD stage
-- import_from_usd: Main parser function for importing from a USD file
-"""
-
```
duburcqa, line 14: Why are you removing this?

genesis/utils/usd/__init__.py
```diff
@@ -1,24 +1,13 @@
...
 # Check if USD support is available before importing modules that depend on it
 try:
     from pxr import Usd
-except ImportError as e:
-    gs.raise_exception_from("pxr module not found. Please install it with `pip install genesis-world[usd]`.", e)
-else:
```
duburcqa, line 20: Why are you adding more stuff in the try-except block? The less the better. if guarding pxr import is sufficient, you should not add anything else in exception handling. If there is another module that may cause issues, then add a dedicated try-exception block for it.

genesis/utils/usd/usd_bake.py
duburcqa, line 1: Which version of `omni.kit_app` is supported? you should update pyproject requirements accordingly if necessary. (I don't know if you did for now, you going through your PR linearly.)

genesis/utils/gltf.py
```diff
@@ -392,7 +392,8 @@ def parse_mesh_glb(path, group_by_material, scale, surface):
             mesh_info, first_created = mesh_infos.get(group_idx)
             if first_created:
                 mesh_info.set_property(
-                    surface=material, metadata={"path": path, "name": material_name if group_by_material else mesh_name}
+                    surface=material,
+                    metadata={"mesh_path": path, "name": material_name if group_by_material else mesh_name},
```
duburcqa, line 396: Why are you doing this? Remove unnecessary formatting changes.

tests/test_mesh.py
```diff
@@ -12,62 +12,40 @@
 
 from .utils import assert_allclose, assert_array_equal, get_hf_dataset
 
-# Check for USD support by testing if pxr module (from usd-core package) is available
-try:
-    import pxr.Usd
+VERTICES_TOL = 1e-05  # Transformation loses a little precision in vertices
+NORMALS_TOL = 1e-02  # Conversion from .usd to .glb loses a little precision in normals
```
duburcqa: These tolerances are very bad, especially for the normal. 1e-2 is not a precision loss. It is almost 2 degrees of error.

tests/test_rigid_physics.py
```diff
@@ -2072,7 +2072,6 @@ def test_mesh_repair(convexify, show_viewer, gjk_collision):
             file=f"{asset_path}/work_table.glb",
             pos=(0.4, 0.0, -0.54),
             fixed=True,
-            file_meshes_are_zup=True,
```
duburcqa, line 2075: Why are you doing this? Avoid unnecessary changes.

pyproject.toml
```diff
@@ -115,9 +115,7 @@ render = [
 ]
 usd = [
     # Used for parsing `.usd` mesh files
-    "usd-core<25.11",
-    # import stubs to static code analysis
-    "types-usd==24.5.2",
```
duburcqa, line 122: This is now useless?

duburcqa:
@ACMLCZH Why did you have to remove and add so many modules? Are you just renaming them somehow?

pyproject.toml
ACMLCZH: It is just a typing-stubs package for code completion for `usd-core`. However, their versions do not match (`usd-core==25.11` requires `type-usd==25.11` but it is not published), so it should be removed from here.

genesis/utils/gltf.py
ACMLCZH: `"path"` is changed to `mesh_path` (the metadata name aligning with that in mjcf, urdf... and unittests will use this metadata). And if I don't start a new line, the character number will exceed 120.

genesis/utils/usd/usd_bake.py
ACMLCZH: `omniverse-kit` cannot be installed in arm platform, and different python versions install different versions of omniverse-kit (e.g. python==3.10, 3.11 -> omniverse-kit==107.3 and python==3.12 -> omniverse-kit==109.2). so it cannot be generally written into `pyproject.toml`. In genesis doc installation there is a guide that the users should manually install it through `pip install --extra-index-url https://pypi.nvidia.com omniverse-kit`.

genesis/options/morphs.py
ACMLCZH: `geometry_only`: the selected prim is not required to have RigidBodyAPI etc.. It is useful for importing pure mesh USD assets (with no rigid properties).
`use_bake_cache`: If set to True, reuse the material baking results from the last run will not bake again. Otherwise, it will rebake the materials.

ACMLCZH:
It is truly renaming, but each renamed file would have huge differences in this PR, maybe more like rewriting.

genesis/utils/usd/__init__.py
ACMLCZH: I just see almost none of `__init__.py` in Genesis has such doc, so I removed it.

genesis/engine/entities/rigid_entity/rigid_entity.py
alanray-tech: we may support USD softbody in the future I think.  `parse_usd_rigid_entity` only deals with the articulation and rigid body

pyproject.toml
duburcqa: I see. Ok.

genesis/utils/gltf.py
duburcqa: Ok sorry I missed that. Good catch!

genesis/utils/usd/usd_bake.py
duburcqa: Ok I see. Damn it is painful to install. Could you point to Genesis documentation to install Nvidia omniverse in the exception?

genesis/utils/usd/__init__.py
duburcqa: I think it is good to have them. We are trying to raise the bar for new modules that are being added, including enforcing typing. It is too much to refactor the entire codebase at once, but we can at least make sure that new additions are up to standard. This kind of doctring contributes to this effort.

genesis/options/morphs.py
duburcqa: I don’t think it makes sense to have such option as `use_bake_cache=False`. What is supposed to be the use case?

Same for geometry_only. It should be handled automatically. No need to have a dedicated option for this.

genesis/engine/scene.py
duburcqa: What kind of morphs is being returned? I think you should add a comment in the code to explain what is happening here.

genesis/engine/mesh.py
duburcqa: Are you able to load both transparently without having to specify extra morph parameters? This is important.

duburcqa: And I think support mesh-like USD as input of add_entity is important.

genesis/engine/entities/rigid_entity/rigid_entity.py
duburcqa: Sure. But here we are dealing with rigid body entities only. Do you think it will be the same function that will parse every information? It does not really make sense given the current design of genesis entities i think.

genesis/engine/mesh.py
ACMLCZH: Sure. Using the 'geometry_only' option we can parse mesh-only USD

genesis/engine/scene.py
YilingQiao: 1. `for now our entities are name-less` → For existing entities, users can easiy get the handle for things they want to manipulation. 
2. To import a USD and manipulate the robot or access the objects inside, I think we should provide a convenient way for them to retrieve the entities.
3. If we only returns a list, users might access one of the entity by `list_entities[0]`, but I'm not sure if the order of the list is well defined, is it possible two different environments / runs returns the same environment.

genesis/options/morphs.py
ACMLCZH: But in `FileMorph` the annotation of `file` appears `Any`, so I add this guard, but I don't know why it comes like that. And I found `Path()` passed to `FileMorph` in unittests or somewhere, so we should at least add `str()`.

duburcqa: Ok let’s do this.

genesis/engine/mesh.py
duburcqa: I don’t think it is a good approach. add_stage should be able to parse both from only and articulated bodies automatically without specifying any option. Moreover, we should keep supporting passing usd to add_entity as mesh and raise an exception stating that add_stage should be used if incompatible (ie multiple entities or articulated).

ACMLCZH: In the latest commit `geometry_only` is removed. Now we can use `gs.morph.USD` in `add_stage` to parse articulations and in `add_entity` to parse any mesh or articulations with specifying any options. Previously `geometry_only` is just a guard to preserve the same behaviors as #2067 .

duburcqa: Excellent! Yeah I don't think we need this guard anymore. USD is still fresh, we can afford doing breaking changing before the next major release (e.g. 1.0).

duburcqa:
> It is truly renaming, but each renamed file would have huge differences in this PR, maybe more like rewriting.

What is the purpose behind it? Do you think it was important for consistency or something? Because renaming files should be well motivated for just because you like it more. Note that I do not suggest it is actually the case, I'm just trying to understand the rationale behind it.

genesis/engine/scene.py
YilingQiao: After discussion with @duburcqa , we will add name system to entities (see this PR https://github.com/Genesis-Embodied-AI/Genesis/pull/2303) So for this place, let's keep return type a dict, where key is the name of the entity.

genesis/utils/usd/usd_rigid_entity.py
```diff
@@ -0,0 +1,440 @@
...
+
+def _parse_geoms(
+    context: UsdContext,
+    entity_prim: Usd.Prim,
+    link_path_to_idx: Dict[str, int],
+    morph: gs.morphs.USD,
+    surface: gs.surfaces.Surface,
+) -> List[List[Dict]]:
+    links_g_infos = [[] for _ in range(len(link_path_to_idx))]
+    parse_prim_geoms(context, entity_prim, None, links_g_infos, link_path_to_idx, morph, surface)
```
alanray-tech, line 380: What will happen if we come across:

- LinkA:UsdGeom.Mesh
  - Visuals:Prim
      - OtherMeshes
  - Collisions:Prim
      - OtherMeshes
 
 According to `parse_prim_geoms`'s impl, all the children will be both collision and visual meshes or?
 
 Though this case is rare (depends on Assets Provider). But I'd like to separate the link and its children when calling parse_prim_geoms.

ACMLCZH: For my current implementation, in this case, the parent will be parsed as both Visual and Collision geom (because no "visual" and "collision" are specified). Then the first child will be parsed as visual geom and second as collision geom.

alanray-tech: OK, I see. The patterns are modified. Make sense.

 ```python
    # Geometry Parsing Options
    collision_mesh_prim_patterns: List[str] = [r"^([cC]ollision).*"]
    visual_mesh_prim_patterns: List[str] = [r"^([vV]isual).*"]
 ```

tests/test_rigid_physics.py
ACMLCZH: image

I use a new commit `8d9621de48a84fb182da38fd88f2495e4ccf3de9` in hugging face asset. However, between this commit and the original commit `c50bfe3e354e105b221ef4eb9a79504650709dd2`, there is another commit `fe31f14aaae60ea166102b8c41478dfaa223f45d` that converts `work_table.glb` from z-up to y-up, so I have to remove this option to pass the unittest.

tests/test_mesh.py
ACMLCZH: image

This involves a remaining issue in #1051, I import the asset `RoughnessTest.usd` into Blender and then export it as `RoughnessTest.glb` and make them as twins assets for parsing comparison. However, Blender will certainly lose some precision (especially on normals, maybe Blender does some processing) when converting from usd to glb, so I have to use this tolerance.

genesis/utils/usd/usd_rigid_entity.py
duburcqa: @ACMLCZH Can you update the documentation accordingly to explain this?

ACMLCZH:
> > It is truly renaming, but each renamed file would have huge differences in this PR, maybe more like rewriting.
> 
> What is the purpose behind it? Do you think it was important for consistency or something? Because renaming files should be well motivated for just because you like it more. Note that I do not suggest it is actually the case, I'm just trying to understand the rationale behind it.

I think the modules are not necessary to be named with "parser" every time. Just like `urdf.py`, `mjcf.py` and `gltf.py`, they are not named with "parser", and none of them has an actual `Parser` class. Additionally, in `usd_geometry.py` there is no `Adapter`, either. Thus, I renamed them.

duburcqa:
> I think the modules are not necessary to be named with "parser" every time. Just like urdf.py, mjcf.py and gltf.py, they are not named with "parser", and none of them has an actual Parser class. Additionally, in usd_geometry.py there is no Adapter, either. Thus, I renamed them.

Wait. I don't get it. The name of the file is not related to the name of a class. The name of the file is an indication of its contain. Nothing more. It may contain the implementation of 5 different classes, or no classes at all and only functions, or even no function and only constants that are used at multiple places.

ACMLCZH:
> > I think the modules are not necessary to be named with "parser" every time. Just like urdf.py, mjcf.py and gltf.py, they are not named with "parser", and none of them has an actual Parser class. Additionally, in usd_geometry.py there is no Adapter, either. Thus, I renamed them.
> 
> Wait. I don't get it. The name of the file is not related to the name of a class. The name of the file is an indication of its contain. Nothing more. It may contain the implementation of 5 different classes, or no classes at all and only functions, or even no function and only constants that are used at multiple places.

The strongest motivation I want to rename them is to keep the names clean and consistent with other parsing modules like `gltf.py`, `mjcf.py` and `urdf.py`

Issue #2254, opened 2026-01-17
[Feature]: Ghost Contacts for Tactile


Milotrince:
### What feature or enhancement are you proposing?

To implement a fast rigid-body based tactile sensor, we will need to be able to query the contact status of arbitrary (point, normal, radius)s in space. These points should not affect collision resolution in the solver, but we should be able to obtain penetration depth and direction w.r.t the normal.

### Motivation

(image from https://github.com/eanswer/DiffHand )
<img width="529" height="578" alt="Image" src="https://github.com/user-attachments/assets/8f0b5b1e-2113-4258-a4e4-ea198bf6723f" />

a bad sketch of what i mean (point, normal, radius) in red, force vector in blue.
<img width="877" height="374" alt="Image" src="https://github.com/user-attachments/assets/a38231a4-a8d7-478d-ac39-96a5c5bd3aa5" />

### Potential Benefit

- Can implement fast tactile array sensors

### What is the expected outcome of the implementation work?

- [ ] A method in RigidSolver to add a contact point
- [ ] A method to query the force normals on the added points

sunkmechie:
I've been thinking about this and would like to propose that:

We should implement the ```Ghost Contacts``` as Kinematic Probes. This ensures that the collision resolution of the physics bodies remains entirely unaffected.

1. We can model each sensor element as a kinematic sphere defined by $S(p, r, \hat{n}_s, k)$:
* **$p, r$**: World position and sensitive radius.
* **$\hat{n}_s$**: Sensor Normal
* **$k$**: Stiffness coefficient (to map penetration depth to a virtual force).

2. Probes check against the existing AABB trees/grid to identify candidate geometries, we use **SDFs** if available.

3. The sensor state can be quered by using:
* **Penetration Depth**: $d_{pen} = r - d_{surf}$ 
* **Force Vector**: $\vec{F} = k \cdot d_{pen} \cdot \hat{n}_c$. [approximately modeled as a spring-mass system]
* **Penetration Area**: $$A \approx \pi (2 r d_{pen} - d_{pen}^2)$$ [approx.]

#### **Pseudo-code**
```python
def query_contact_points(probes, geometries):
    results = []

    for probe in probes:
        # 1. Initialize result for this probe
        max_penetration = 0.0
        result_force = Vector(0, 0, 0)
        total_area = 0.0
        
        # 2. Broadphase: Get geometries near the probe's bounding sphere
[...]

PR #2292, opened 2026-01-23
[FEATURE] Contact Sensor with Kinematic Probe

YilingQiao:
## Description

Add KinematicContactProbe sensor for contact detection without physics side effects.

- New sensor type that uses support function queries to detect contact
- Supports multiple probes per sensor with `probe_local_pos` and `probe_local_normal` lists
- Returns penetration depth, contact position, normal, and estimated force per probe
- Lazy evaluation: queries execute on `read()`, not every step (unless delay configured)

To solve #2254 

## How It Works

**Algorithm for each probe:**
1. Transform `probe_pos` and `probe_normal` from link-local to world frame
2. For each nearby geom, compute `support_pos = support(-probe_normal)`
   - This finds the point on the geom furthest in the opposite direction of the probe normal
3. If `distance(support_pos, probe_pos) <= radius`, compute:
   - `penetration = dot(probe_pos - support_pos, probe_normal)`

## Force Output (Not Physical)

The returned `force = stiffness * penetration * probe_normal` is a **user-defined estimate**, not derived from the physics solver (because this probe is only at kinematic level). Genesis is not a mass-spring model. The `stiffness` parameter is for convenience to convert penetration into a force-like quantity.

## Return Values

`sensor.read()` returns a `KinematicContactProbeData` NamedTuple:

| Field | Shape | Description |
|-------|-------|-------------|
| `penetration` | `(n_envs, n_probes)` or `(n_probes,)` | Penetration depth in meters. Positive when contact detected, zero otherwise. |
| `position` | `(n_envs, n_probes, 3)` or `(n_probes, 3)` | Contact point in link-local frame. The point on the contacting geom closest to the probe (in the -normal direction). |
| `normal` | `(n_envs, n_probes, 3)` or `(n_probes, 3)` | Contact normal in link-local frame. Same as `probe_local_normal` when contact detected, zero otherwise. |
| `force` | `(n_envs, n_probes, 3)` or `(n_probes, 3)` | Estimated force in link-local frame: `stiffness * penetration * probe_normal` (NOT physical). |

Try 
```
python examples/sensors/kinematic_contact_probe.py -v -t 6
```

https://github.com/user-attachments/assets/f336401c-774f-423f-8db9-95ff28256f0a

## Test plan

- [x] Basic contact detection with plane
- [x] No-contact case returns zeros
- [x] Multi-environment support
- [x] Multi-probe sensor support
- [x] Collision filtering (contype/conaffinity)
- [x] Box geometry support function
- [x] Noise and delay functionality

## Checklist:


genesis/engine/sensors/kinematic_contact_probe.py
duburcqa, line 1: I'm not convinced implementing this using gstaichi is actually faster than using torch when zerocopy is enabled. It may not be a big deal at this point but most of our sensors are notoriously slow. I recently optimised contact sensors to make them much faster using zerocopy and optimising the math behind it (reducing computational complexity).

YilingQiao: I need to query support function for this kinematic probe, which has to happen in gstaichi. Which part do you think can be replaced with torch?

genesis/engine/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,519 @@
...
+            links_pos=links_pos.contiguous(),
+            links_quat=links_quat.contiguous(),
+            radii=shared_metadata.radii.contiguous(),
+            stiffness=shared_metadata.stiffness.contiguous(),
+            links_idx=shared_metadata.links_idx.contiguous(),
+            contypes=shared_metadata.contypes.contiguous(),
+            conaffinities=shared_metadata.conaffinities.contiguous(),
+            n_probes_per_sensor=shared_metadata.n_probes_per_sensor.contiguous(),
+            sensor_cache_start=shared_metadata.sensor_cache_start.contiguous(),
+            sensor_probe_start=shared_metadata.sensor_probe_start.contiguous(),
```
duburcqa, line 208: Remove `continuous()`. If this is currently necessary, the implementation should be refactored so that it is no longer necessary but applying transpose and changing the expected memory layout in the kernel.

genesis/engine/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,519 @@
...
+            n_geoms=n_geoms,
+            geoms_state=solver.geoms_state,
+            geoms_info=solver.geoms_info,
+            rigid_global_info=solver._rigid_global_info,
+            constraint_state=solver.constraint_solver.constraint_state,
+            equalities_info=solver.equalities_info,
+            support_field_info=solver.collider._support_field._support_field_info,
+            output=shared_ground_truth_cache.contiguous(),
```
duburcqa, line 216: Same. Find a way to remove contiguous.

genesis/engine/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,519 @@
...
+
+        links_pos = solver.get_links_pos(links_idx=shared_metadata.links_idx)
+        links_quat = solver.get_links_quat(links_idx=shared_metadata.links_idx)
+
+        if solver.n_envs == 0:
+            links_pos = links_pos[None]
+            links_quat = links_quat[None]
+
+        links_pos = links_pos.reshape(B, n_sensors, 3)
+        links_quat = links_quat.reshape(B, n_sensors, 4)
```
duburcqa, line 191: Using these public getters is not very efficient. I would recommend passing `shared_metadata.links_idx` to `_kernel_kinematic_contact_probe_support_query`, along with `solver.links_state`.

genesis/engine/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,519 @@
...
+
+        self._shared_metadata.radii = concat_with_tensor(
+            self._shared_metadata.radii, self._options.radius, expand=(1,), dim=0
+        )
+        self._shared_metadata.stiffness = concat_with_tensor(
+            self._shared_metadata.stiffness, self._options.stiffness, expand=(1,), dim=0
+        )
+        self._shared_metadata.contypes = concat_with_tensor(
+            self._shared_metadata.contypes,
+            torch.tensor([self._options.contype], dtype=torch.int32, device=gs.device),
```
duburcqa, line 149: `self._options.contype` is good enough, no need to cast as tensor. This will allow to keep everything on one line and focus on what matters.

genesis/engine/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,519 @@
...
+            expand=(1,),
+            dim=0,
+        )
+        self._shared_metadata.conaffinities = concat_with_tensor(
+            self._shared_metadata.conaffinities,
+            torch.tensor([self._options.conaffinity], dtype=torch.int32, device=gs.device),
```
duburcqa, line 155: Same. No need to cast as tensor.

genesis/engine/sensors/kinematic_contact_probe.py
duburcqa: Ok I guess I missed that part. Taichi is the way to go then!

genesis/engine/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,519 @@
...
+    contypes: torch.Tensor = make_tensor_field((0,), dtype_factory=lambda: torch.int32)
+    conaffinities: torch.Tensor = make_tensor_field((0,), dtype_factory=lambda: torch.int32)
+
+    probe_sensor_idx: torch.Tensor = make_tensor_field((0,), dtype_factory=lambda: torch.int32)
+    probe_positions: torch.Tensor = make_tensor_field((0, 3))
+    probe_normals: torch.Tensor = make_tensor_field((0, 3))
+
+    n_probes_per_sensor: torch.Tensor = make_tensor_field((0,), dtype_factory=lambda: torch.int32)
+    sensor_cache_start: torch.Tensor = make_tensor_field((0,), dtype_factory=lambda: torch.int32)
+    sensor_probe_start: torch.Tensor = make_tensor_field((0,), dtype_factory=lambda: torch.int32)
```
duburcqa, line 62: I'm replacing `dtype_factory=lambda: [...]` in favour of simply `dtype=[...]`, which is now possible because sensors cannot be imported before initializing Genesis anymore.

genesis/engine/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,519 @@
...
+    def __init__(
+        self,
+        sensor_options: KinematicContactProbeOptions,
+        sensor_idx: int,
+        data_cls: Type[KinematicContactProbeData],
+        sensor_manager: "SensorManager",
+    ):
+        # Store n_probes before super().__init__() since _get_return_format() is called there
+        self._n_probes = sensor_options.n_probes
+        super().__init__(sensor_options, sensor_idx, data_cls, sensor_manager)
```
duburcqa, line 86: Skip one line between defining derived class attributes and calling base init.

genesis/engine/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,519 @@
...
+        super().build()
+
+        probe_positions = self._options.get_probe_positions()
+        probe_normals = self._options.get_probe_normals()
+        n_probes = len(probe_positions)
+        sensor_idx = self._idx
+
+        self._shared_metadata.n_probes_per_sensor = concat_with_tensor(
+            self._shared_metadata.n_probes_per_sensor,
+            torch.tensor([n_probes], dtype=torch.int32, device=gs.device),
```
duburcqa, line 101: Casting to tensor manually is not necessary, and enforcing 1D / sequence is not necessary either.

Just do:
```python
self._shared_metadata.n_probes_per_sensor = concat_with_tensor(
    self._shared_metadata.n_probes_per_sensor, n_probes, expand=(1,), dim=0
)
```

genesis/engine/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,519 @@
...
+            expand=(1,),
+            dim=0,
+        )
+
+        current_cache_start = sum(self._shared_metadata.cache_sizes[:-1]) if self._shared_metadata.cache_sizes else 0
+        self._shared_metadata.sensor_cache_start = concat_with_tensor(
+            self._shared_metadata.sensor_cache_start,
+            torch.tensor([current_cache_start], dtype=torch.int32, device=gs.device),
```
duburcqa, line 109: same.

genesis/engine/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,519 @@
...
+            expand=(1,),
+            dim=0,
+        )
+
+        current_probe_start = self._shared_metadata.total_n_probes
+        self._shared_metadata.sensor_probe_start = concat_with_tensor(
+            self._shared_metadata.sensor_probe_start,
+            torch.tensor([current_probe_start], dtype=torch.int32, device=gs.device),
```
duburcqa, line 117: same.

genesis/engine/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,519 @@
...
+            self._shared_metadata.probe_sensor_idx,
+            torch.full((n_probes,), sensor_idx, dtype=torch.int32, device=gs.device),
+            expand=(n_probes,),
+            dim=0,
+        )
+
+        positions_tensor = torch.tensor(probe_positions, dtype=gs.tc_float, device=gs.device)
+        self._shared_metadata.probe_positions = torch.cat(
+            [self._shared_metadata.probe_positions, positions_tensor], dim=0
+        )
```
duburcqa, line 132: Why are you concatenating manually instead of using our dedicated utility for this?

genesis/engine/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,519 @@
...
+
+        normals_tensor = torch.tensor(probe_normals, dtype=gs.tc_float, device=gs.device)
+        norms = normals_tensor.norm(dim=1, keepdim=True).clamp(min=1e-8)
+        normals_tensor = normals_tensor / norms
+        self._shared_metadata.probe_normals = torch.cat([self._shared_metadata.probe_normals, normals_tensor], dim=0)
```
duburcqa, line 137: Why are you concatenating manually instead of using our dedicated utility for this?

genesis/engine/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,519 @@
...
+            buffered_data,
+            shared_metadata.cur_jitter_ts,
+            shared_metadata.interpolate,
+        )
+        cls._add_noise_drift_bias(shared_metadata, shared_cache)
+        cls._quantize_to_resolution(shared_metadata.resolution, shared_cache)
+
+    def _read_internal(self, envs_idx=None, is_ground_truth=False):
+        if not self._manager._has_delay_configured(type(self)):
+            self._manager.update_sensor_type_cache(type(self))
```
duburcqa, line 241: Don't do this. You are abusing the software design. It is the responsibility of the manager to determine whether to update a given sensor type or not.

genesis/engine/sensors/sensor_manager.py
duburcqa, line 1: Move all these changes outside of this PR. It is completely unrelated and requires further discussion. First, we need to make sure this lazy-update thing is really necessary, next we need to implement this feature in a way that does not delegate more responsibility that expected to sensors.

genesis/engine/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,519 @@
...
+            support_pos = ti.Vector.zero(gs.ti_float, 3)
+
+            if geom_type == gs.GEOM_TYPE.PLANE:
+                g_pos = geoms_state.pos[i_g, i_b]
+                g_quat = geoms_state.quat[i_g, i_b]
+                geom_data = geoms_info.data[i_g]
+                plane_normal_local = gs.ti_vec3([geom_data[0], geom_data[1], geom_data[2]])
+                plane_normal_world = gu.ti_transform_by_quat(plane_normal_local, g_quat)
+                dist_to_plane = (probe_pos - g_pos).dot(plane_normal_world)
+                support_pos = probe_pos - dist_to_plane * plane_normal_world
```
duburcqa, line 484: Why are you handling `if geom_type == gs.GEOM_TYPE.PLANE:` separately from other cases defined in `_func_support_point_for_probe` ? I don't say why they could not be unified. Just rename the function if you think it is no longer appropriate.

genesis/engine/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,519 @@
...
+def _func_point_in_expanded_aabb(
+    i_g: ti.i32,
+    i_b: ti.i32,
+    geoms_state: array_class.GeomsState,
+    point: ti.types.vector(3, gs.ti_float),
+    expansion: gs.ti_float,
+):
+    aabb_min = geoms_state.aabb_min[i_g, i_b] - expansion
+    aabb_max = geoms_state.aabb_max[i_g, i_b] + expansion
+    return (point >= aabb_min).all() and (point <= aabb_max).all()
```
duburcqa, line 313: We already have a function for this, just import it from rigid solver, you are allowed to do this.

genesis/engine/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,519 @@
...
+    i_g: ti.i32,
+    i_b: ti.i32,
+    sensor_link_idx: ti.i32,
+    sensor_contype: ti.i32,
+    sensor_conaffinity: ti.i32,
+    geoms_info: array_class.GeomsInfo,
+    rigid_global_info: array_class.RigidGlobalInfo,
+    constraint_state: array_class.ConstraintState,
+    equalities_info: array_class.EqualitiesInfo,
+):
```
duburcqa, line 327: We don't already have a function that is doing exactly this? If we do, just import it directly instead of copy-pasting it here.

genesis/options/sensors/options.py
```diff
@@ -181,6 +181,133 @@ def model_post_init(self, _):
...
+        """Return the number of probes defined for this sensor."""
+        return len(self.probe_local_pos)
+
+    def get_probe_positions(self) -> list[Tuple3FType]:
+        """Return probe positions as a list of tuples."""
+        return [tuple(pos) for pos in self.probe_local_pos]
+
+    def get_probe_normals(self) -> list[Tuple3FType]:
+        """Return probe normals as a list of tuples."""
+        return [tuple(normal) for normal in self.probe_local_normal]
```
duburcqa, line 308: I don't think this should be part of the options.

tests/test_sensors.py
```diff
@@ -627,3 +627,97 @@ def test_lidar_cache_offset_parallel_env(show_viewer, tol):
...
+    )
+
+    scene.build(n_envs=0)
+    scene.step()
+
+    assert probe.read().penetration[0] > 0.0
+
+    mount.set_pos(torch.tensor([0.0, 0.0, 2.0], dtype=gs.tc_float))
+    scene.step()
+    assert probe.read().penetration[0] == 0.0
```
duburcqa, line 661: This is clearly not sufficient. You should check that the result is analytically valid on a synthetic scenario that is simple enough to do so.

tests/test_sensors.py
```diff
@@ -627,3 +627,97 @@ def test_lidar_cache_offset_parallel_env(show_viewer, tol):
...
+            stiffness=1000.0,
+        )
+    )
+
+    scene.build(n_envs=0)
+    scene.step()
+
+    assert probe.read().penetration[0] > 0.0
+
+    mount.set_pos(torch.tensor([0.0, 0.0, 2.0], dtype=gs.tc_float))
```
duburcqa, line 659: You can just do `mount.set_pos([0.0, 0.0, 2.0])`. No need to force casting to tensor. It just makes what you are doing harder to read.

tests/test_sensors.py
```diff
@@ -627,3 +627,97 @@ def test_lidar_cache_offset_parallel_env(show_viewer, tol):
...
+    """Test BOX geometry detection via support function."""
+    scene = gs.Scene(
+        sim_options=gs.options.SimOptions(dt=1e-2, substeps=1, gravity=(0.0, 0.0, 0.0)),
+        profiling_options=gs.options.ProfilingOptions(show_FPS=False),
+        show_viewer=show_viewer,
+    )
+
+    scene.add_entity(gs.morphs.Box(size=(0.5, 0.5, 0.5), pos=(0.0, 0.0, 0.25), fixed=True))
+
+    mount = scene.add_entity(gs.morphs.Sphere(radius=0.02, pos=(0.25, 0.25, 0.06), fixed=True, collision=False))
```
duburcqa, line 643: Use multiple lines systematically for scene declaration with very few exceptions. And no need to skip lines between the declaration of each entity.

tests/test_sensors.py
```diff
@@ -627,3 +627,97 @@ def test_lidar_cache_offset_parallel_env(show_viewer, tol):
         sensor_data = sensor.read()
         assert (sensor_data.distances > gs.EPS).any()
         assert (sensor_data.points.abs() > gs.EPS).any()
+
+
+@pytest.mark.required
+def test_kinematic_contact_probe_box_support(show_viewer, tol):
+    """Test BOX geometry detection via support function."""
+    scene = gs.Scene(
+        sim_options=gs.options.SimOptions(dt=1e-2, substeps=1, gravity=(0.0, 0.0, 0.0)),
```
duburcqa, line 636: One line per argument. Moreover, remove them if they are not absolutely necessary for your test.

tests/test_sensors.py
duburcqa, line 661: Just (vaguely) checking penetration is not good enough. What about all the other outputs?! See what we did for the IMU sensor, it is much more comprehensive.

YilingQiao:
> We do not add features pro-actively without at least well identified use cases. Could you clarify in the description of this PR what are the targeted use cases of this feature? Note that following our template for PR would have avoided having this question in the first place since there is already a dedicated section for motivations.

It's for resolving this request as in the PR description.
https://github.com/Genesis-Embodied-AI/Genesis/issues/2254

tests/test_sensors.py
duburcqa: I just noticed your example script. I think it is perfectly suitable for analytical validation already. You need to convert this in a unit test and do more comprehensive validation of the sensor measurement.

genesis/engine/sensors/kinematic_contact_probe.py
YilingQiao: this part introduced affinity for sensor, and only used in this sensor. We don't need this plugin to affected the main physics solver code

genesis/engine/sensors/kinematic_contact_probe.py
YilingQiao: this is a specific variants of aabb only used in this sensor, we don't need to accommodate this function in the main code?

Milotrince:
@YilingQiao thanks for this! I could take a look at addressing the review if you want?

Milotrince:
I opened a new PR here: https://github.com/Genesis-Embodied-AI/Genesis/pull/2389

PR #2294, opened 2026-01-23
[BUG FIX] Add rigid body entity hibernation mechanism.

### LINKED ISSUE #2244: [Bug]: Lack of contact islands and hibernation results in poor performance in some scenes?
### Bug Description

Hi, very impressed by how much Genesis has grown!

I was testing the engine's limits and noticed it doesn't handle scenes from Robocasa (contain equality constraints) well. Performance on high-end GPUs drops significantly even with 1 environment and maximum GPU usage. I was wondering if that's a bug? Do you guys think adding contact islands and hibernation would speed up the simulation, and by how much? And is there a timeline of when those features are coming into Genesis fully? I'd be excited to contribute to the project.

Thank you so much!

### Steps to Reproduce

Run any Robocasa scenes with use_hibernation = True and use_contact_island = True.

### Expected Behavior

Performance improvement.

### Screenshots/Videos

_No response_

### Relevant log output

```shell

```

### Environment

- OS: Ubuntu 25
- GPU/CPU: Nvidia 5080, Intel i9-9900K

Commit: 7c4ee63

### Additional Context

_No response_

YilingQiao:
### Summary
Fixes hibernation feature to work correctly in fields mode by addressing array access patterns, buffer bounds guards, and wake-up mechanisms.

To resolve #2244 

### Changes

**Bug Fixes:**
- Fix `func_aggregate_awake_entities` counter reset being inside entity loop (caused floating/rotating boxes)
- Add bounds guards in contact island construction to prevent buffer overflow
- Add `kernel_wake_up_entities_by_links` to wake hibernated entities when `set_pos`/`set_quat` is called
- Add `kernel_update_sort_buffer_for_awake_entities` to update collision detection buffer for woken entities
- Remove `hibernation_thresh_acc` - hibernation now based only on velocity threshold

**New Files:**
- `examples/hibernation.py` - Demo showing hibernation workflow with two boxes

**Test:**
- Add `test_hibernation_and_contact_islands` to `test_rigid_physics.py` (skipped in ndarray mode)

### Test Plan
```bash
python examples/hibernation.py 
```

And you can see the output
```
======================================================================
Results
======================================================================
Without hibernation: 1.324s
With hibernation:    0.103s
Speedup:             12.90x faster with hibernation
```

## Related Issue

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,8 @@ Project Page: <https://genesis-embodied-ai.github.io/>
 
 ## Quick Installation
 
+### Using pip
+
 Install **PyTorch** first following the [official instructions](https://pytorch.org/get-started/locally/).
 
 Then, install Genesis via PyPI:
@@ -93,6 +95,44 @@ cd Genesis
 pip install -e ".[dev]"
 ```
 It is recommended to systematically execute `pip install -e ".[dev]"` after moving HEAD to make sure that all dependencies and entrypoints are up-to-date.
+
+### Using uv
+
+[uv](https://docs.astral.sh/uv/) is a fast Python package and project manager.
+
+**Install uv:**
+```bash
+# On macOS and Linux
+curl -LsSf https://astral.sh/uv/install.sh | sh
+
+# On Windows
+powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/install.ps1 | iex"
+```
+
+**Quick start with uv:**
+```bash
+git clone https://github.com/Genesis-Embodied-AI/Genesis.git
+cd Genesis
+uv sync
+```
+
+Then install PyTorch for your platform:
+
+```bash
+# NVIDIA GPU (CUDA 12.6 as an example)
+uv pip install torch --index-url https://download.pytorch.org/whl/cu126
+
+# CPU only (Linux/Windows)
+uv pip install torch --index-url https://download.pytorch.org/whl/cpu
+
+# Apple Silicon (Metal/MPS)
+uv pip install torch
+```
+
+Run an example:
+```bash
+uv run examples/rigid/single_franka.py
+```
 
 ## Docker
 
--- a/genesis/utils/geom.py
+++ b/genesis/utils/geom.py
@@ -1677,9 +1677,9 @@ def _np_euler_to_R(rpy: np.ndarray, out: np.ndarray | None = None) -> np.ndarray
     """
     assert rpy.ndim >= 1
     if out is None:
-        out_ = np.empty((*rpy.shape[1:], 3, 3), dtype=rpy.dtype)
+        out_ = np.empty((*rpy.shape[:-1], 3, 3), dtype=rpy.dtype)
     else:
-        assert out.shape == (*rpy.shape[1:], 3, 3)
+        assert out.shape == (*rpy.shape[:-1], 3, 3)
         out_ = out
 
     cos_rpy, sin_rpy = np.cos(rpy), np.sin(rpy)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -183,3 +183,8 @@ markers = [
     "slow: tests that are slow to run (> 100s).",
 ]
 log_cli_level = "WARNING"
+
+[tool.uv]
+# uv respects the build-system and project tables
+# PyTorch must be installed separately before Genesis due to platform-specific builds
+# See README.md for installation instructions with different PyTorch backends


tests/test_rigid_physics.py
```diff
@@ -4264,3 +4264,86 @@ def test_pick_heterogenous_objects(show_viewer):
     post_lift_z = het_obj.get_pos()[:, 2]
     lift_deltas = (post_lift_z - pre_lift_z).cpu().numpy()
     assert np.all(lift_deltas > 0.05), f"All objects should be lifted (deltas={lift_deltas:.3f})"
+
+
+@pytest.mark.parametrize("backend", [gs.cpu, gs.gpu])
+@pytest.mark.parametrize("performance_mode", [True])
+def test_hibernation_and_contact_islands(show_viewer):
```
duburcqa: It is great to finally have a unit test for hibernation!

tests/test_rigid_physics.py
```diff
@@ -4264,3 +4264,86 @@ def test_pick_heterogenous_objects(show_viewer):
     post_lift_z = het_obj.get_pos()[:, 2]
     lift_deltas = (post_lift_z - pre_lift_z).cpu().numpy()
     assert np.all(lift_deltas > 0.05), f"All objects should be lifted (deltas={lift_deltas:.3f})"
+
+
+@pytest.mark.parametrize("backend", [gs.cpu, gs.gpu])
```
duburcqa: Why forcing running on both cpu and gpu is necessary? Maybe it is, just trying to understand

tests/test_rigid_physics.py
```diff
@@ -4264,3 +4264,86 @@ def test_pick_heterogenous_objects(show_viewer):
...
+    Test hibernation and contact island behavior.
+
+    Scenario:
+    1. Two boxes settle separately on ground -> both hibernate, 2 contact islands
+    2. Move one box above the other using set_pos (wakes it up)
+    3. Box falls and collides -> both boxes awake
+    4. Stacked boxes settle and hibernate -> 1 contact island (merged)
+    """
+    scene = gs.Scene(
+        sim_options=gs.options.SimOptions(dt=0.01),
```
duburcqa: Is dt relevant here? I don't think it is?

tests/test_rigid_physics.py
```diff
@@ -4264,3 +4264,86 @@ def test_pick_heterogenous_objects(show_viewer):
...
+
+    solver = scene.sim.rigid_solver
+    box1_idx = box1._idx_in_solver
+    box2_idx = box2._idx_in_solver
+
+    def is_hibernated(entity_idx):
+        return solver.entities_state.hibernated[entity_idx, 0] == 1
+
+    def get_n_islands():
+        return solver.constraint_solver.contact_island.n_islands[0]
```
duburcqa: I'm not convinced these helpers are clarifying anything. They may even make the error trace harder to read in case of failure.

tests/test_rigid_physics.py
```diff
@@ -4264,3 +4264,86 @@ def test_pick_heterogenous_objects(show_viewer):
...
+    )
+
+    scene.build()
+
+    solver = scene.sim.rigid_solver
+    box1_idx = box1._idx_in_solver
+    box2_idx = box2._idx_in_solver
+
+    def is_hibernated(entity_idx):
+        return solver.entities_state.hibernated[entity_idx, 0] == 1
```
duburcqa: `bool(solver.entities_state.hibernated[entity_idx, 0])` is better (it is supposed to be a bool already but using int because of taichi limitation, so let's consider it is truly a bool).

Personally, I would even go for `solver.entities_state.hibernated[entity_idx, 0]` without casting to bool, because it would work just fine and casting to bool does not matter much.

tests/test_rigid_physics.py
```diff
@@ -4264,3 +4264,86 @@ def test_pick_heterogenous_objects(show_viewer):
...
+        return solver.constraint_solver.contact_island.n_islands[0]
+
+    # Phase 1: Let boxes settle and hibernate separately
+    for step in range(500):
+        scene.step()
+        if is_hibernated(box1_idx) and is_hibernated(box2_idx):
+            break
+
+    assert is_hibernated(box1_idx), "Box 1 should be hibernated"
+    assert is_hibernated(box2_idx), "Box 2 should be hibernated"
```
duburcqa: Remove these assert messages, it is already self explanatory.

tests/test_rigid_physics.py
```diff
@@ -4264,3 +4264,86 @@ def test_pick_heterogenous_objects(show_viewer):
...
+    for step in range(500):
+        scene.step()
+        if is_hibernated(box1_idx) and is_hibernated(box2_idx):
+            break
+
+    assert is_hibernated(box1_idx), "Box 1 should be hibernated after settling"
+    assert is_hibernated(box2_idx), "Box 2 should be hibernated after settling"
+
+    # Stacked boxes should form 1 contact island
+    assert get_n_islands() == 1, f"Expected 1 contact island for stacked boxes, got {get_n_islands()}"
```
duburcqa: Remove `got {get_n_islands()}"`. It will be printed by pytest anyway.

tests/test_rigid_physics.py
```diff
@@ -4264,3 +4264,86 @@ def test_pick_heterogenous_objects(show_viewer):
...
+    # Let box1 fall and collide with box2
+    for _ in range(25):
+        scene.step()
+
+    # Both boxes should be awake shortly after collision (before they re-hibernate)
+    assert not is_hibernated(box1_idx), "Box 1 should be awake after collision"
+    assert not is_hibernated(box2_idx), "Box 2 should be awake after collision"
+
+    # Phase 3: Let stacked boxes settle and hibernate
+    for step in range(500):
```
duburcqa: 500 is too much. You need to find a way to avoid exceeding at most 200. Otherwise the test will take too much time on cheap runners.

tests/test_rigid_physics.py
duburcqa, line 4271: Why this test is not marked as required?

genesis/utils/array_class.py
```diff
@@ -481,9 +481,15 @@ def get_contact_island_state(solver, collider):
...
+    # in addition to contact edges. The chain construction is cyclic (last entity links back to first),
+    # so worst case: each entity contributes one hibernation edge, totaling n_entities hibernation edges.
+    max_hibernation_edges = n_entities if solver._use_hibernation else 0
+    max_edges = max_contact_pairs + max_hibernation_edges
+
     return StructContactIslandState(
-        ci_edges=V(dtype=gs.ti_int, shape=(max_contact_pairs, 2, _B)),
-        edge_id=V(dtype=gs.ti_int, shape=(max_contact_pairs * 2, _B)),
+        ci_edges=V(dtype=gs.ti_int, shape=(max_edges, 2, _B)),
+        edge_id=V(dtype=gs.ti_int, shape=(max_edges * 2, _B)),
```
duburcqa, line 492: Nice catch!

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -1205,6 +1213,11 @@ def check_errno(self):
             gs.raise_exception("Invalid constraint forces causing 'nan'. Please decrease Rigid simulation timestep.")
         if errno & 0b00000000000000000000000000001000:
             gs.raise_exception("Invalid accelerations causing 'nan'. Please decrease Rigid simulation timestep.")
+        if errno & 0b00000000000000000000000000010000:
```
duburcqa, line 1217: Maybe it is the right time to add some `class ErrorCode(IntEnum):` to manage error codes x)

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -458,6 +459,13 @@ def build(self):
 
         self._static_rigid_sim_config = array_class.StructRigidSimStaticConfig(**static_rigid_sim_config)
 
+        if self._static_rigid_sim_config.use_hibernation:
+            if gs.use_ndarray:
+                gs.raise_exception(
+                    "Hibernation is not yet supported with ndarray mode. "
```
duburcqa: We don't call it `ndarray mode` from user perspective, but dynamic array mode vs static array mode. See `__init__.py` as a reference.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -458,6 +459,13 @@ def build(self):
...
+                    "Set GS_ENABLE_NDARRAY=0 or use_hibernation=False."
```
duburcqa: `GS_ENABLE_NDARRAY` is a backdoor. Do not document its existence. You should rather ask users to specify `performance_mode=True`.

examples/hibernation.py
duburcqa, line 1: I'm not sure an example for something that has no visible consequence really makes sense.

Moreover, do not expose private internal attributes in official examples, `solver.entities_state` and `solver.constraint_solver` are not supposed to exist from user perspective, you cannot suggest users to just call them or you are asking for trouble.

duburcqa: Besides, I don't think any end-user care about proof-checking that some obscure hidden flag in some undocumented data structure has the right value. Ever the simulation does what it is supposed to be or it does not. If you want to show something, demonstrate the performance benefit on a pathological use case or something, because it is exactly what hibernation is about, not the value of a flag in some math implementation.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -1205,6 +1213,11 @@ def check_errno(self):
             gs.raise_exception("Invalid constraint forces causing 'nan'. Please decrease Rigid simulation timestep.")
         if errno & 0b00000000000000000000000000001000:
             gs.raise_exception("Invalid accelerations causing 'nan'. Please decrease Rigid simulation timestep.")
+        if errno & 0b00000000000000000000000000010000:
+            gs.raise_exception(
+                "Contact island buffer overflow. Too many edges in contact island construction. "
+                "This may indicate excessive hibernation edges or contact pairs."
+            )
```
duburcqa: Wait, this exception can really happen? I thought you forced the upper bound for the number of edges?  Moreover, you should clearly state what the user can do to fix the issue. "may" something is not good enough.

examples/hibernation.py
YilingQiao: change the example to show the speedup

genesis/engine/solvers/rigid/rigid_solver.py
YilingQiao: Changed the error message. It can happen
  The buffer is sized to max_contact_pairs + n_entities, but:                                    
  - max_contact_pairs is an estimate set by the user via RigidOptions.max_collision_pairs        
  - If the actual number of contacts exceeds this estimate during simulation, the buffer can     
  overflow

genesis/engine/solvers/rigid/rigid_solver.py
YilingQiao: added a TODO, maybe let's do it in a separate PR?

YilingQiao:
## Summary of Changes in Response to PR Review                           
                                                                                                 
  1. examples/hibernation.py - Rewritten                                                         
                                                                                                 
  Reviewer feedback: Don't expose private internal attributes; show performance benefit, not     
  internal flag values.                                                                          
                                                                                                 
  Changes:                                                                                       
  - Completely rewrote to demonstrate performance comparison (with vs without hibernation)       
  - No longer accesses private attributes (solver.entities_state, solver.constraint_solver)      
  - Shows measurable speedup (~10x in testing)                                                  
                                                                                                 
  2. genesis/engine/solvers/rigid/rigid_solver.py                                                
                                                                                                 
  Reviewer feedback: Fix terminology and don't document backdoor env vars.                       
                                                                                                 
  Changes:                                                                                       
  - Line 465: "ndarray mode" → "dynamic array mode"                                              
  - Line 466: Removed GS_ENABLE_NDARRAY=0, now says "Please set performance_mode=True"           
  - Line 1217-1218: Simplified buffer overflow error to actionable message                       
                                                                                                 
  3. tests/test_rigid_physics.py                                                                 
                                                                                                 
  Reviewer feedback: Multiple issues with test style and efficiency.                             
                                                                                                 
  Changes:                                                                                       
  - Added @pytest.mark.required                                                                  
  - Removed @pytest.mark.parametrize("backend", [gs.cpu, gs.gpu]) - no variation needed          
  - Removed unnecessary dt=0.01 parameter                                                        
  - Removed helper functions is_hibernated() and get_n_islands() - inlined expressions           
  - Removed self-explanatory assert messages                                                     
  - Reduced step counts from 500 → 200                                                           
                                                                                                 
  4. Kept the buffer overflow error check                                                        
                                                                                                 
  Reviewer question: "Can this exception really happen?"                                         
                                                                                                 
  Decision: Kept it - the error CAN happen if users underestimate max_collision_pairs. It's a    
  legitimate runtime safety check.

Issue #1410, opened 2025-07-14
[Feature]: Labelled entities in Genesis


MRiabov:
### What feature or enhancement are you proposing?

Hello,
As somebody who has to manage many Genesis entities and store them for longer use (think having multiple abstraction layers), I currently have to manage them through a separate `gs_entities` dict, which has to appear in 108 places in my codebase.
This isn't per se a problem, however it would be more convenient if such entities could be stored in `genesis.Scene` under labels, and `Scene.entities` would return a dict instead of a list.

This would simplify:
1. Getting link ids for constraints
2. Getting positions of bodies
3. Getting AABBs of bodies,
4. Etc. that uses `entity`.

### Motivation

As a RL researcher developing on top of Genesis and having scenes with 5+ bodies,
I want to make managing bodies simpler and take less space in my code,
Because this leads to better code and faster development cycles.

I would also add that I use `label` to denote types of bodies I'm managing. E.g. this loop:
```
for full_name, entity in gs_entities.items():
    part_name, part_type = full_name.split("@")
        if part_type in ("solid", "fixed_solid", "connector"):
             ...
       elif part_type == "control":
            ...
        elif part_type == "fastener":
            fastener_pos = torch.tensor(entity.get_pos(env_idx), device=device)
             ...
```
Maybe this means that adding some form of groups/auxillary data to entities would be desirable (however this can be easily managed by the user). Nevertheless, entity naming is already a great step to it.

### Potential Benefit

Faster development cycles.

### What is the expected outcome of the implementation work?

- `gs.add_entity()` would now have an optional `label` attribute.
- If label is added, it is possible to get it from `entities`, or `scene.get_entity()`<- `get_entity` would minimize changes necessary in codebase. 
- `label` can be queried the entity.

### Additional information

`gs_entities` appears in 108 places in my code.
<img width="345" height="547" alt="Image" src="https://github.com/user-attachments/assets/c023939b-aef5-482a-a5d2-397398136a40" />

I don't expect that this change should impact anything but providing an API for getting bodies and their labels.

duburcqa:
Fixed by https://github.com/Genesis-Embodied-AI/Genesis/pull/2303

PR #2303, opened 2026-01-25
[FEATURE] Add naming logics to entities.

YilingQiao:
## Summary

Adds auto-generated unique names and scene-level lookup APIs to all Genesis entities.

## Changes

- **Entity**: Added `name` property with auto-generation format `{morph_type}_{uid[:8]}`
- **Scene**: Added `add_entity(name=...)`, `get_entity(name/uid)`, `entity_names` property
- **Solver prefixes**: FEM/MPM/PBD/SPH/SF entities get `{solver}_{morph}_` prefix
- **File-based morphs**: Use file stem (e.g., `go2.urdf` → `go2_abc12345`)

## Usage

```python
# Auto-generated names
box = scene.add_entity(gs.morphs.Box(size=(0.1, 0.1, 0.1)))
print(box.name)  # "box_a1b2c3d4"

# Custom names
robot = scene.add_entity(gs.morphs.URDF(file="go2.urdf"), name="my_robot")

# Lookup
entity = scene.get_entity(name="my_robot")
entity = scene.get_entity(uid="a1b2")  # prefix match

# List all
print(scene.entity_names)  # ["box_a1b2c3d4", "my_robot"]
```

## Tests

```bash
pytest tests/test_entity_naming.py -v
```

29 tests covering auto/user naming, lookups, solver prefixes, file-based morphs, USD.

## Example

```bash
python examples/tutorials/entity_name.py --demo all
```

Resolves https://github.com/Genesis-Embodied-AI/Genesis/issues/1410

### DIFF
--- a/genesis/engine/entities/base_entity.py
+++ b/genesis/engine/entities/base_entity.py
@@ -26,7 +26,10 @@ class Entity(RBC):
         surface,
         name: str | None = None,
     ):
-        self._uid = gs.UID()
+        uid = gs.UID()
+        while any(entity.uid.match(uid, short_only=True) for entity in scene.entities):
+            uid = gs.UID()
+        self._uid = uid
         self._idx = idx
         self._scene: "Scene" = scene
         self._solver = solver
@@ -43,11 +46,7 @@ class Entity(RBC):
             self._name = name
         else:
             morph_name = self._get_morph_identifier()
-            while True:
-                self._name = f"{morph_name}_{self._uid.short()}"
-                if self._name not in existing_names:
-                    break
-                self._uid = gs.UID()
+            self._name = f"{morph_name}_{uid.short()}"
 
         gs.logger.info(
             f"Adding ~<{self._repr_type()}>~. idx: ~<{self._idx}>~, uid: ~~~<{self._uid}>~~~, morph: ~<{morph}>~, material: ~<{self._material}>~."


tests/test_entity_naming.py
duburcqa, line 1: What is this??? This is absolutely not standard pytest unit test. It looks like old-fashion unittest. Please modernise this to follow what we are doing everywhere else. 

Besides, I don't think we need as many as 29 unit tests to cover this feature. 5 would already be a lot.

genesis/engine/entities/base_entity.py
```diff
@@ -76,3 +78,109 @@ def material(self):
...
+        >>> scene = gs.Scene()
+        >>> box = scene.add_entity(gs.morphs.Box(), name="my_box")
+        >>> box.name
+        'my_box'
+        >>> sphere = scene.add_entity(gs.morphs.Sphere())  # auto-generated
+        >>> sphere.name
+        'sphere_a1b2c3d4'
+        """
+        if user_name is not None:
+            self._name = user_name
```
duburcqa: This is not good enough. You should ensure uniqueness. Raise an exception if user-specified yet not unique.

genesis/engine/entities/base_entity.py
```diff
@@ -76,3 +78,109 @@ def material(self):
...
+            A name in the format "{morph_identifier}_{uid_prefix}" where
+            uid_prefix is the first 8 characters of the entity's UID.
+
+        Examples
+        --------
+        Auto-generated names follow patterns based on entity type:
+
+        - Rigid primitives: "box_a1b2c3d4", "sphere_b2c3d4e5"
+        - File-based morphs: "go2_c3d4e5f6" (URDF), "ant_d4e5f6g7" (MJCF)
+        - Particle entities: "mpm_box_e5f6g7h8", "fem_sphere_f6g7h8i9"
```
duburcqa: We don't need this section I think, or maybe just one, without this fancy formatting.

genesis/engine/entities/base_entity.py
```diff
@@ -76,3 +78,109 @@ def material(self):
...
+
+        This method should be overridden in subclasses to provide
+        type-specific identifiers (e.g., "box", "sphere", "fem_box").
+
+        Returns
+        -------
+        str
+            The morph identifier. Base implementation returns "entity".
+        """
+        return "entity"
```
duburcqa: I think it should be `raise NotImplemented` instead.

genesis/engine/entities/particle_entity.py
```diff
@@ -739,6 +739,49 @@ def find_closest_particle(self, pos, envs_idx=None):
...
+        elif isinstance(morph, gs.morphs.Cylinder):
+            return "cylinder"
+        elif isinstance(morph, gs.morphs.Mesh):
+            return Path(morph.file).stem
+        elif isinstance(morph, gs.morphs.Nowhere):
+            return "emitter"
+        else:
+            return "particle"
+
+    def _get_morph_identifier(self) -> str:
```
duburcqa: I don't think it is relevant to have 2 methods for this.

genesis/engine/scene.py
```diff
@@ -340,11 +344,23 @@ def add_entity(
...
             The created entity.
+
+        Examples
+        --------
+        >>> scene = gs.Scene()
+        >>> box = scene.add_entity(gs.morphs.Box(), name="my_box")
+        >>> box.name
+        'my_box'
+        >>> sphere = scene.add_entity(gs.morphs.Sphere())  # auto-generated name
+        >>> sphere.name  # e.g., 'sphere_a1b2c3d4'
```
duburcqa: I don't think this section is useful. I'm in favour of removing it.

genesis/engine/scene.py
```diff
@@ -1464,6 +1483,75 @@ def entities(self) -> list["Entity"]:
...
+        list[str]
+            List of entity names in order of creation.
+
+        Examples
+        --------
+        >>> scene = gs.Scene()
+        >>> scene.add_entity(gs.morphs.Box(), name="box1")
+        >>> scene.add_entity(gs.morphs.Sphere(), name="sphere1")
+        >>> scene.entity_names
+        ['box1', 'sphere1']
```
duburcqa: I don't think this section is useful. I'm in favour of removing it.

genesis/engine/scene.py
```diff
@@ -1464,6 +1483,75 @@ def entities(self) -> list["Entity"]:
...
+        """
+        return [entity.name for entity in self.entities]
```
duburcqa: Please return an immutable tuple instead of a list.

genesis/engine/scene.py
```diff
@@ -1464,6 +1483,75 @@ def entities(self) -> list["Entity"]:
...
+        Returns
+        -------
+        Entity
+            The matching entity.
+
+        Raises
+        ------
+        Exception
+            If no entity is found, multiple entities match the UID prefix,
+            or neither name nor uid is provided.
```
duburcqa: Remove this section and just mention in short docstring that it will raise an exception if name does not exist.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -3285,6 +3285,61 @@ def get_mass(self):
...
+                return morph.file.name
+        elif isinstance(morph, gs.morphs.USD):
+            # Use prim_path last segment if available, otherwise file stem
+            if morph.prim_path:
+                return morph.prim_path.rstrip("/").split("/")[-1]
+            return Path(morph.file).stem
+        elif isinstance(morph, gs.morphs.Terrain):
+            return morph.name if morph.name else "terrain"
+        else:
+            return "rigid"
```
duburcqa: Do not use elif-else, just it, since all of them return directly.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -3285,6 +3285,61 @@ def get_mass(self):
...
+            return "plane"
+        elif isinstance(morph, gs.morphs.Mesh):
+            return Path(morph.file).stem
+        elif isinstance(morph, (gs.morphs.URDF, gs.morphs.MJCF, gs.morphs.Drone)):
+            # morph.file can be a URDF object (e.g., from hybrid entity) instead of a path
+            if isinstance(morph.file, str):
+                return Path(morph.file).stem
+            else:
+                # URDF object has a name attribute
+                return morph.file.name
```
duburcqa: We should update URDF parser to extract the robot name instead of getting it from file name.

Moreover, I'm not convinced using the file stem as robot name is a good idea. It is often messy. I guess it is better than nothing though...

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -3285,6 +3285,61 @@ def get_mass(self):
...
+        elif isinstance(morph, gs.morphs.USD):
+            # Use prim_path last segment if available, otherwise file stem
+            if morph.prim_path:
+                return morph.prim_path.rstrip("/").split("/")[-1]
+            return Path(morph.file).stem
```
duburcqa: Same. I'm not convinced this pattern is really nice. I guess better than nothing...

examples/tutorials/entity_name.py
duburcqa, line 1: This is way too complicated for something as simple as that. It should not exceed 70 lines.

tests/test_entity_naming.py
YilingQiao: Rewrote from 29 unittest-style tests to 4 simple pytest functions

examples/tutorials/entity_name.py
```diff
@@ -0,0 +1,53 @@
...
+    uid_prefix = str(box.uid)[:4]
+    retrieved = scene.get_entity(uid=uid_prefix)
+    print(f"Lookup by UID: scene.get_entity(uid='{uid_prefix}') -> {retrieved.name}")
+
+    # List all entity names
+    print(f"All entity names: {scene.entity_names}")
+
+    scene.build()
+    scene.step()
+    scene.destroy()
```
duburcqa: I think you could remove these 3 lines. They are not doing anything useful for this example I think.

genesis/utils/urdf.py
```diff
@@ -13,6 +14,31 @@
...
+    """
+    try:
+        path = os.path.join(get_assets_dir(), file_path)
+        tree = ET.parse(path)
+        root = tree.getroot()
+        if root.tag == "robot":
+            return root.attrib.get("name")
+    except (ET.ParseError, FileNotFoundError, OSError):
+        pass
+    return None
```
duburcqa: Having a name is mandatory, not optional. You should enforce this as a requirement and not tolerate missing attribute.

Besides, it would be great to add a reference to the official documentation: https://wiki.ros.org/urdf/XML/robot

duburcqa: If it is absolutely necessary, you may add a fallback at caller-level, but I don't like it and it should be well motivated. Still, some kind of callback for empty name should be added at caller-level.

tests/test_entity_naming.py
```diff
@@ -0,0 +1,70 @@
...
+
+import genesis as gs
+
+
+@pytest.fixture
+def scene():
+    """Create a basic scene and destroy it after the test."""
+    s = gs.Scene(show_viewer=False)
+    yield s
+    s.destroy()
```
duburcqa: All scenes are automatically destroyed already, which makes this fixture useless. Better remove it.

tests/test_entity_naming.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def test_auto_and_user_names(scene):
+    """Test auto-generated and user-specified entity names."""
+    # Auto-generated name
+    box = scene.add_entity(gs.morphs.Box(size=(0.1, 0.1, 0.1)))
+    assert box.name.startswith("box_")
+    assert len(box.name) == len("box_") + 8  # 8-char UID suffix
```
duburcqa: Here you are enforcing the naming convention. I don't know if it is desirable. And if it is, this is not restrictive enough. You should check "box" + "_" + alphanumeric matching UID via startswith to not enforce the length, which is an implementation detail.

Moreover, you should have a check that makes sure that it is unique to some extent (at least different if adding multiple identical entities, some for all formats including URDF)

genesis/engine/entities/base_entity.py
```diff
@@ -76,3 +78,54 @@ def material(self):
...
+        if user_name is not None:
+            # Validate uniqueness for user-specified names upfront
+            if user_name in scene._entity_name_registry:
+                gs.raise_exception(f"Entity name '{user_name}' already exists in scene.")
+            self._name = user_name
+        else:
+            self._name = self._generate_name()
+            # Validate uniqueness for auto-generated names
+            if self._name in scene._entity_name_registry:
+                gs.raise_exception(f"Entity name '{self._name}' already exists in scene.")
```
duburcqa: This is not good enough. If the name is automatically generated, YOU are responsible to ensuring that it is actually unique. Raising an exception is not the correct approach if the user cannot do anything about it.

duburcqa: I would suggest to remove this check and enforce uniqueness by design.

genesis/engine/entities/base_entity.py
```diff
@@ -76,3 +78,54 @@ def material(self):
...
+
+        scene._entity_name_registry[self._name] = self
+
+    def _generate_name(self) -> str:
+        """Generate a default name based on morph type and UID."""
+        uid_suffix = str(self._uid)[:8]
+        morph_name = self._get_morph_identifier()
+        return f"{morph_name}_{uid_suffix}"
```
duburcqa: This should be hard-coded in _set_name. I see no reason to expose it this way because overloading it is unsupported and considered broken.

genesis/engine/entities/base_entity.py
duburcqa: You should ensure right away that it would lead to a unique entity name and redraw if not.

genesis/engine/entities/base_entity.py
```diff
@@ -76,3 +78,54 @@ def material(self):
...
+            self._name = self._generate_name()
+            # Validate uniqueness for auto-generated names
+            if self._name in scene._entity_name_registry:
+                gs.raise_exception(f"Entity name '{self._name}' already exists in scene.")
+
+        scene._entity_name_registry[self._name] = self
+
+    def _generate_name(self) -> str:
+        """Generate a default name based on morph type and UID."""
+        uid_suffix = str(self._uid)[:8]
```
duburcqa: I would recommend using the existing `str(self._uid.short())` for consistency.

tests/test_entity_naming.py
```diff
@@ -0,0 +1,70 @@
...
+    # Non-existent name raises error
+    with pytest.raises(Exception, match="not found"):
+        scene.get_entity(name="nonexistent")
+
+
+def test_get_entity_by_uid(scene):
+    """Test retrieving entity by UID prefix."""
+    box = scene.add_entity(gs.morphs.Box(size=(0.1, 0.1, 0.1)))
+    uid_prefix = str(box.uid)[:4]
+    assert scene.get_entity(uid=uid_prefix) is box
```
duburcqa: I don't think this is ok. I think it should be either the UID, nothing else. This flexibility is just causing additional trouble without any benefit.

tests/test_entity_naming.py
```diff
@@ -0,0 +1,70 @@
...
+
+    # Non-existent UID raises error
+    with pytest.raises(Exception, match="not found"):
+        scene.get_entity(uid="zzzzzzzz")
```
duburcqa: `scene.get_entity(uid=gs.UID())` is much more sensible.

tests/test_entity_naming.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def test_entity_names_property(scene):
+    """Test scene.entity_names returns tuple in creation order."""
+    scene.add_entity(gs.morphs.Box(size=(0.1, 0.1, 0.1)), name="first")
+    scene.add_entity(gs.morphs.Sphere(radius=0.1), name="second")
+    names = scene.entity_names
+    assert isinstance(names, tuple)
```
duburcqa: I don't think it is relevant to enforce in unit test sequence container. Just do `assert tuple(names) == ("first", "second")`

duburcqa: You should call them "B" then "A" to confirm it is NOT sorted but by insertion order.

genesis/utils/mjcf.py
```diff
@@ -22,6 +22,31 @@
 MIN_TIMECONST = np.finfo(np.double).eps
 
 
+def get_model_name(file_path):
+    """
+    Extract the model name from an MJCF file.
```
duburcqa: [...] if specified.

genesis/engine/scene.py
```diff
@@ -196,6 +196,9 @@ def __init__(
         # emitters
         self._emitters = gs.List()
 
+        # entity name registry for lookup by name
+        self._entity_name_registry: dict[str, "Entity"] = {}
```
duburcqa: I don't think adding this registry is useful. We already store the sequence of entities. It is straightforward and reasonably efficient to recover the correct one without this extra attribute.

genesis/engine/scene.py
```diff
@@ -465,6 +472,9 @@ def add_entity(
 
         entity = self._sim._add_entity(morph, material, surface, visualize_contact)
 
+        # Set entity name (auto-generate if not specified)
+        entity._set_name(name, self)
```
duburcqa: I don't see any reason for this being the responsibility of the scene. You should pass the desired name as an optional constructor argument of base entity (and all derived entities). An entity cannot exist without a scene anyway as it takes the scene in input argument.

genesis/engine/scene.py
```diff
@@ -1502,6 +1512,52 @@ def entities(self) -> list["Entity"]:
...
+
+        Returns
+        -------
+        Entity
+            The matching entity.
+        """
+        if name is not None:
+            if name in self._entity_name_registry:
+                return self._entity_name_registry[name]
+            gs.raise_exception(f"Entity not found for name: '{name}'.")
```
duburcqa: The standard pattern would rather be:
```python
try:
    return self._entity_name_registry[name]
except KeyError as e:
    gs.raise_exception_from(f"Entity not found for name: '{name}'.", e)
```

That being said, I recommended not defining `_entity_name_registry`, which means that it should rather be:
```python
try:
    return next(entity for entity in scene.entities if entity.name == name)
except KeyError as e:
    gs.raise_exception_from(f"Entity not found for name: '{name}'.", e)
```

genesis/engine/scene.py
```diff
@@ -1502,6 +1512,52 @@ def entities(self) -> list["Entity"]:
...
+        elif uid is not None:
+            # Use prefix matching (startswith) instead of substring matching
+            matches = [entity for entity in self.entities if str(entity.uid).startswith(uid)]
```
duburcqa: As mentioned somewhere else, I think you should only compare full UID, not partial. Beyond that, I think UID class should overload "==" operator to automatically compare `self.full()` to `other.full()` if other is also a UID, otherwise to `other` directly. You may check that other is either a string or a UID and always return False if it is not.

duburcqa: Actually, after think twice, it may be convenient to support equality of `short` instead of full because the user only see the short name being printed in terminal output. Instead of overloading `==` operator you may just add some `match` method with optional `short_only` argument that would be false by default to allow partial comparison.

genesis/engine/entities/mpm_entity.py
```diff
@@ -676,3 +676,11 @@ def remove_particle_constraints(self, particles_mask=None):
             particles_mask = torch.ones((self._sim._B, self.n_particles), dtype=torch.bool, device=gs.device)
 
         self._solver._kernel_remove_particle_constraints(particles_mask, self._particle_start)
+
+    # ------------------------------------------------------------------------------------
+    # --------------------------------- naming methods -----------------------------------
+    # ------------------------------------------------------------------------------------
+
+    def _get_morph_identifier(self) -> str:
+        """Get the identifier string from the morph for name generation."""
```
duburcqa: Remove this docstring since you did not modified it. It will be inherited from base implementation. Same of all other solvers.

genesis/utils/mjcf.py
```diff
@@ -22,6 +22,31 @@
...
+
+    Parameters
+    ----------
+    file_path : str or Path
+        Path to the MJCF file.
+
+    Returns
+    -------
+    str or None
+        The model name from the <mujoco model="..."> attribute, or None if not found.
```
duburcqa: Could you add a reference to Mujoco XML specification for `model` attribute?
https://mujoco.readthedocs.io/en/stable/XMLreference.html#mujoco

Besides, the sentence `from the <mujoco model="..."> attribute` should not be part of the documentation for the return values, but rather part of the documentation details (ie on the very top, right after short docstring and skipping one line).

genesis/utils/urdf.py
```diff
@@ -13,6 +14,31 @@
...
+
+    Parameters
+    ----------
+    file_path : str or Path
+        Path to the URDF file.
+
+    Returns
+    -------
+    str or None
+        The robot name from the <robot name="..."> attribute, or None if not found.
```
duburcqa: The sentence `from the <robot name="..."> attribute` should not be part of the documentation for the return values, but rather part of the documentation details (ie on the very top, right after short docstring and skipping one line).

YilingQiao:
Addressed all the comments

genesis/engine/entities/base_entity.py
```diff
@@ -76,3 +81,53 @@ def material(self):
...
+            self._name = user_name
+        else:
+            # Generate name, regenerating UID if collision occurs
+            morph_name = self._get_morph_identifier()
+            while True:
+                self._name = f"{morph_name}_{self._uid.short()}"
+                if self._name not in existing_names:
+                    break
+                # Redraw UID if name collision
+                self._uid = gs.UID()
```
duburcqa: I would suggest to inline this implementation at the only single place it is used because there is only one valid way to use it.

genesis/engine/entities/fem_entity.py
```diff
@@ -1015,6 +1017,25 @@ def clear_grad(self, f: ti.i32):
             i_global = i_v + self.el_start
             self._solver.elements_el.grad[f, i_global, i_b].actu = 0
 
+    # ------------------------------------------------------------------------------------
+    # --------------------------------- naming methods -----------------------------------
+    # ------------------------------------------------------------------------------------
+
+    def _get_morph_identifier(self) -> str:
+        from pathlib import Path
```
duburcqa: Avoid import in functions. This must only be used in very specific cases (very long to import or optional dependency). Here it is not the case.

genesis/engine/scene.py
```diff
@@ -1502,6 +1506,57 @@ def entities(self) -> list["Entity"]:
...
+                return next(entity for entity in self.entities if entity.name == name)
+            except StopIteration as e:
+                gs.raise_exception_from(f"Entity not found for name: '{name}'.", e)
+        elif uid is not None:
+            # First try exact full UID match
+            for entity in self.entities:
+                if entity.uid.match(uid, short_only=False):
+                    return entity
+            # Then try short UID match (7-character prefix shown in terminal)
+            matches = [entity for entity in self.entities if entity.uid.match(uid, short_only=True)]
```
duburcqa: Just match short, it is sufficient. We already made sure that short UID is unique for entities by design.

genesis/engine/scene.py
```diff
@@ -1502,6 +1506,57 @@ def entities(self) -> list["Entity"]:
...
+            if len(matches) == 1:
+                return matches[0]
+            elif len(matches) == 0:
+                gs.raise_exception(f"Entity not found for uid: '{uid}'.")
+            else:
+                gs.raise_exception(
+                    f"Multiple entities match short uid '{uid}'. Use the full uid to uniquely identify the entity."
+                )
```
duburcqa: This is impossible You can remove this condition:
```python
if matches:
    (match,) = matches
    return match
else:
    gs.raise_exception(f"Entity not found for uid: '{uid}'.")
```

genesis/engine/scene.py
```diff
@@ -1502,6 +1506,57 @@ def entities(self) -> list["Entity"]:
...
+        else:
+            gs.raise_exception("Neither `name` nor `uid` is provided.")
```
duburcqa: For clarity, I recommend moving all sanity checks at the beginning:
```python
if not (name is None ^ uid is None):
    gs.raise_exception("Please specified either one argument between `name` or `uid`.")

if name is not None:
    [...]
else:  # uid is not None
    [...]
```

genesis/engine/scene.py
```diff
@@ -1502,6 +1506,57 @@ def entities(self) -> list["Entity"]:
...
+        Get the names of all entities in the scene.
+
+        Returns
+        -------
+        tuple[str, ...]
+            Tuple of entity names in order of creation.
+        """
+        return tuple(entity.name for entity in self.entities)
+
+    def get_entity(self, name: str | None = None, uid: str | None = None) -> "Entity":
```
duburcqa: I would recommend forcing keyword-only for `uid`.

examples/tutorials/entity_name.py
```diff
@@ -0,0 +1,49 @@
...
+    print(f"User name: '{ground.name}', URDF auto-name: '{robot.name}'")
+
+    # Lookup by name
+    retrieved = scene.get_entity(name="ground")
+    print(f"Lookup by name: scene.get_entity(name='ground') -> {retrieved.name}")
+
+    # Lookup by UID prefix
+    uid_prefix = str(box.uid)[:4]
+    retrieved = scene.get_entity(uid=uid_prefix)
+    print(f"Lookup by UID: scene.get_entity(uid='{uid_prefix}') -> {retrieved.name}")
```
duburcqa: `uid_prefix = str(box.uid)[:4]` should not be supported anymore.

I would rather recommend passing `box.uid`, or `str(box.uid.short())` if you think it is better.

genesis/engine/entities/tool_entity/tool_entity.py
```diff
@@ -44,6 +45,21 @@ def __init__(
         # for rendering purpose only
         self.latest_pos = ti.Vector.field(3, dtype=gs.ti_float, shape=(1))
 
+    # ------------------------------------------------------------------------------------
+    # --------------------------------- naming methods -----------------------------------
+    # ------------------------------------------------------------------------------------
+
+    def _get_morph_identifier(self) -> str:
+        from pathlib import Path
```
duburcqa: Move this import on top of the file.

genesis/utils/urdf.py
```diff
@@ -13,6 +14,41 @@
...
+    """
+    path = os.path.join(get_assets_dir(), file_path)
+    tree = ET.parse(path)
+    root = tree.getroot()
+    if root.tag == "robot":
+        name = root.attrib.get("name")
+        if name:
+            return name
+        raise ValueError(f"URDF file '{file_path}' is missing required 'name' attribute on <robot> element.")
+    raise ValueError(f"File '{file_path}' does not have a <robot> root element.")
```
duburcqa: `raise ValueError(f"Invalid URDF file '{file_path}'. Missing <robot> root element.")`

genesis/engine/entities/base_entity.py
YilingQiao: I would prefer to separate it to init() cleaner, but I made the change as you said

genesis/engine/scene.py
YilingQiao: you don't want to use name? That's one of the motivations of this PR... And it's more consistent with get_link

duburcqa: I mean doing this: `def get_entity(self, name: str | None = None, *, uid: str | None = None) -> "Entity":`

Nothing more. ie forcing keyword for `uid`, to make sure users are specifying uid on purpose.

genesis/engine/entities/base_entity.py
```diff
@@ -23,6 +24,7 @@ def __init__(
         solver,
         material,
         surface,
+        name: str | None = None,
     ):
         self._uid = gs.UID()
```
duburcqa: Remove this line.

genesis/engine/entities/base_entity.py
```diff
@@ -33,6 +35,20 @@ def __init__(
...
+            if name in existing_names:
+                gs.raise_exception(f"Entity name '{name}' already exists in scene.")
+            self._name = name
+        else:
+            morph_name = self._get_morph_identifier()
+            while True:
+                self._name = f"{morph_name}_{self._uid.short()}"
+                if self._name not in existing_names:
+                    break
+                self._uid = gs.UID()
```
duburcqa:
```python
            while True:
                uid = gs.UID()
                name = f"{morph_name}_{uid.short()}"
                if name not in existing_names:
                    break
            self._name = name
            self._uid = uid
```

duburcqa: The rational is that constant attributes should not be used to store temporaries. This is a bad habit. This is case it is harmless, but it may cause bugs during exception handling that would be hard to debug.

duburcqa: After thinking twice, this is overly complicated. You just need to make sure that `uid.short()` is unique. Replace the original line 29 (`self._uid = gs.UID()`) by:
```python
uid = gs.UID()
while any(entity.uid.match(uid, short_only=True) for entity in self.entities):
    uid = gs.UID()
self._uid = uid
```

then you can just do
```python
self._name = f"{morph_name}_{uid.short()}"
```

If you really want to, you can make back the entire generate name logics in its own function, but if you want to do this, in theory it should be a staticmethod that does not mutate the class because the attribute _name is supposed to be set only once, at a very specific place in init. Using a staticmethod clarify this by delegating the responsibility to set the attribute to __init__, while still breaking down complexity.

PR #2311, opened 2026-01-26
[BUG FIX] Fix 'Mesh.convert_to_zup' by applying scaling out-of-place.

duburcqa:


genesis/engine/mesh.py
```diff
@@ -447,16 +446,10 @@ def verts(self):
...
         return self._mesh.vertices
 
-    @verts.setter
-    def verts(self, verts):
-        """
-        Set the vertices of the mesh.
-        """
-        assert len(verts) == len(self.verts)
-        self._mesh.vertices = verts
-
```
ACMLCZH, line 459: After removing the verts setter, https://github.com/Genesis-Embodied-AI/Genesis/blob/cd84ab234de030ab48f2f30f579f59cc158438d2/genesis/vis/rasterizer_context.py#L562 this line seems to throw the error.

PR #2316, opened 2026-01-28
[BUG FIX] Revert out-of-place mesh scaling and implement a new fix.

duburcqa:


genesis/utils/urdf.py
```diff
@@ -128,26 +130,20 @@ def parse_urdf(morph, surface):
...
 
                 # One asset (.obj) can contain multiple meshes. Each mesh is one RigidGeom in genesis.
-                for i, tmesh in enumerate(geometry.meshes):
+                for _, tmesh in enumerate(geometry.meshes):
                     mesh = gs.Mesh.from_trimesh(
-                        tmesh,
-                        scale=scale,
-                        surface=geom_surface,
-                        metadata={"mesh_path": mesh_path},
+                        tmesh, scale=scale, surface=geom_surface, metadata={"mesh_path": mesh_path}
```
YilingQiao: we assume non-glb mesh will never have `file_meshes_are_zup=False`? Maybe it's okay

PR #2323, opened 2026-01-29
[FEATURE] Track mesh UVs in FEM/PBD solvers for rendering of deformable entities.

alelievr:
## Description
Add a list of UVs in the FEM and PBD solvers. This list of UVs is fetched using the surface indices from the solvers, exactly like the vertex position would. 

## Motivation and Context
The UVs are necessary to rebuild the tangent frame used to sample textures and aniso surfaces of deformable geometry.


genesis/engine/entities/fem_entity.py
```diff
@@ -400,6 +401,13 @@ def sample(self):
                 faces = mesh.faces
                 # For cloth, we store faces as "elements" (treating them as surface elements)
                 self.instantiate(verts, faces)
+
+                # Load UVs from mesh (1:1 mapping for cloth)
+                try:
+                    uvs = mesh.visual.uv.copy()
+                    self._uvs = uvs.astype(gs.np_float)
```
duburcqa: I would recommend:
```python
self._uvs = np.array(mesh.visual.uv, dtype=gs.np_float, copy=True)
```
or just:
```python
self._uvs = mesh.visual.uv.astype(gs.np_float, copy=True)
```
More efficient as it avoids allocating memory twice.

genesis/engine/entities/fem_entity.py
```diff
@@ -400,6 +401,13 @@ def sample(self):
...
+                except AttributeError:
+                    self._uvs = None
```
duburcqa: Could you add a comment to explain which attribute may not exist exactly, in which scenario it would be the case (and why this scenario must be supported if not obvious), and why just ignoring UV is the right approach if not obvious.

genesis/engine/entities/fem_entity.py
```diff
@@ -419,11 +427,12 @@ def sample(self):
             elif isinstance(self.morph, gs.options.morphs.Cylinder):
                 verts, elems = eu.cylinder_to_elements()
             elif isinstance(self.morph, gs.options.morphs.Mesh):
-                verts, elems = eu.mesh_to_elements(
+                verts, elems, self._uvs = eu.mesh_to_elements(
```
duburcqa, line 435: Why can you assign uvs directly while verts, elems must go through further processing?

genesis/engine/entities/fem_entity.py
```diff
@@ -419,11 +427,12 @@ def sample(self):
...
                     file=self._morph.file,
                     pos=self._morph.pos,
                     scale=self._morph.scale,
                     tet_cfg=self.tet_cfg,
+                    return_uvs=True,
```
duburcqa: Just return UVS systematically and update all places where this function is being called. Use `, _ = [...]` to assign to new unused return value.

genesis/engine/entities/fem_entity.py
```diff
@@ -477,8 +486,23 @@ def _add_to_solver(self, in_backward=False):
...
+        Copy UV coordinates to the solver's surface_render_uvs field.
+        """
+        if self._uvs is None:
+            return
+
+        # Copy UVs to solver field starting at v_start
+        uvs_np = self._uvs.astype(gs.np_float, copy=False)
+        for i, uv in enumerate(uvs_np):
+            self._solver.surface_render_uvs[self._v_start + i] = uv
+
```
duburcqa: Just extend `_add_to_solver` with the content of `_add_uvs_to_solver`.

genesis/engine/entities/fem_entity.py
```diff
@@ -477,8 +486,23 @@ def _add_to_solver(self, in_backward=False):
...
 
+    def _add_uvs_to_solver(self):
+        """
+        Copy UV coordinates to the solver's surface_render_uvs field.
+        """
+        if self._uvs is None:
+            return
+
+        # Copy UVs to solver field starting at v_start
+        uvs_np = self._uvs.astype(gs.np_float, copy=False)
```
duburcqa: This should already be enforced by `instantiate`

genesis/engine/entities/fem_entity.py
```diff
@@ -477,8 +486,23 @@ def _add_to_solver(self, in_backward=False):
...
+        for i, uv in enumerate(uvs_np):
+            self._solver.surface_render_uvs[self._v_start + i] = uv
```
duburcqa: This is extremely inefficient (one kernel call per assignment). You should modify `_kernel_add_elements` to take into account `uv`. Just pass an 0-length vector if unspecified and check this condition inside the kernel to avoid out of bound access.

genesis/engine/entities/pbd_entity.py
duburcqa, line 1: I have more or less the same remarks than for FEM.

genesis/engine/solvers/fem_solver.py
```diff
@@ -1092,11 +1099,14 @@ def get_state(self, f):
             state = None
         return state
 
-    def get_state_render(self, f):
+    def get_state_render(self, f, include_uvs=False):
```
duburcqa: Just enforce return uvs systematically now. This is private API. No backward compatibility concerns.

genesis/utils/element.py
duburcqa, line 1: Just return uvs systematically

genesis/utils/mesh.py
```diff
@@ -180,9 +180,9 @@ def get_hashkey(*args):
     return hasher.hexdigest()
 
 
-def load_mesh(file):
+def load_mesh(file, skip_texture=True):
     if isinstance(file, (str, Path)):
-        return trimesh.load(file, force="mesh", skip_texture=True)
+        return trimesh.load(file, force="mesh", skip_texture=skip_texture)
     return file
```
duburcqa, line 191: Why skipping texture would make sense in the first place. This is weird. I would suggest to just load texture systematically, and if it fails, fallback to loading without texture, with a `gs.logger.warning` stating that texture will be skipped because they failed to load.

duburcqa:
I'm ok with the general principle of this PR. @YilingQiao Do you have any concerns on your side?

duburcqa:
I think it is important to wire this in Pyrender (or LuisaRender) if this is possible, in order to make sure it is working properly and unit tested.

genesis/engine/entities/fem_entity.py
alelievr: This is because the tetrahedralization process only appends vertices and triangles to the existing buffers.
Since I'm only interested in getting the UVs for the visible mesh I can keep them as they are, and we can't really generate new UVs for the tetrahedralized mesh anyway (doesn't make sense).
In the end the uv and verts buffers don't have the same length, but it works because of the triangle index buffer (forming triangle surfaces)

genesis/utils/mesh.py
alelievr: I didn't want to increase the load time of meshes, as I'm not sure if the skip_texture=False will load images or not, if it loads images it could do a 10x increase in the load time of meshes.

YilingQiao:
This PR looks good to me.

genesis/engine/entities/fem_entity.py
duburcqa: I think it would be great if you could add a comment in the code that explains exactly what you just said.

alelievr: Ok!

genesis/engine/entities/fem_entity.py
```diff
@@ -400,6 +401,15 @@ def sample(self):
...
                 self.instantiate(verts, faces)
+
+                # Load UVs from mesh (1:1 mapping for cloth)
+                try:
+                    self._uvs = mesh.visual.uv.astype(gs.np_float, copy=True)
+                except AttributeError:
+                    # UVs are not always available in 3D file, in case they are missing we set the entity UVs to None
+                    # When UVs are None, the solver will use 0 UVs for rendering. A mesh with 0 UVs means that no tangent directions
+                    # can be recomputed, thus texture mapping and anisotropic surfaces will not work properly.
+                    self._uvs = None
```
duburcqa: * Everything must fit into 120 chars limit, including comments
* We prefer explicit check instead of try-except in this context
```python
# Load UVs from mesh (1:1 mapping for cloth).
# UVs are not always available in 3D file, in case they are missing we set the entity UVs to None when UVs are None, 
# the solver will use 0 UVs for rendering. A mesh with 0 UVs means that no tangent directions can be recomputed, 
# thus texture mapping and anisotropic surfaces will not work properly.
self._uvs = None
if mesh.visual.uv is not None:
    self._uvs = mesh.visual.uv.astype(gs.np_float, copy=True)
```

genesis/engine/entities/fem_entity.py
```diff
@@ -443,6 +457,7 @@ def _add_to_solver(self, in_backward=False):
 
         # Convert to appropriate numpy array types
         verts_numpy = tensor_to_array(self.init_positions, dtype=gs.np_float)
+        uvs_np = self._uvs if self._uvs is not None else np.zeros((0, 2), dtype=gs.np_float)
```
duburcqa, line 459: Why are you not just setting `self._uvs = np.zeros((0, 2), dtype=gs.np_float)` instead of `None` in the first place? Then you could pass `self._uvs` directly instead of going through `uvs_np`. 

Because you want the property `.uvs` to return None if not specified? If so, I'm ok, I don't have a strong opinion.

genesis/engine/entities/fem_entity.py
duburcqa: This comment is not resolved. You just renamed `_add_uvs_to_solver ` in `_add_uvs_and_faces_to_solver` :p

genesis/engine/entities/pbd_entity.py
```diff
@@ -13,6 +13,46 @@ class PBDBaseEntity(ParticleEntity):
...
+        # Get UVs from vmesh (may be None if no texture)
+        uvs = self._vmesh.uvs if self._vmesh is not None else None
+        if uvs is None or len(uvs) != self.n_vverts:
+            # No UVs available, pass empty array (kernel will skip copy)
+            uvs = np.zeros((0, 2), dtype=gs.np_float)
+
+        self._kernel_add_uvs_and_faces_to_solver(
+            uvs=uvs,
+            faces=self._vfaces,
+        )
```
duburcqa: Single line if it fits: `self._kernel_add_uvs_and_faces_to_solver(uvs=uvs, faces=self._vfaces)`

genesis/engine/entities/pbd_entity.py
```diff
@@ -13,6 +13,46 @@ class PBDBaseEntity(ParticleEntity):
...
+            self.solver.vverts_uvs[i_vv] = uvs[i_vv_]
+
+        # Copy faces to solver's global face buffer (with global vertex indices)
+        for i_vf_ in range(self.n_vfaces):
+            i_vf = i_vf_ + self._vface_start
+            self.solver.vfaces_indices[i_vf] = ti.Vector(
+                [
+                    ti.cast(faces[i_vf_][0], gs.ti_int) + self._vvert_start,
+                    ti.cast(faces[i_vf_][1], gs.ti_int) + self._vvert_start,
+                    ti.cast(faces[i_vf_][2], gs.ti_int) + self._vvert_start,
```
duburcqa: Remove cast, it should not be necessary. You should rather cast the input argument before calling this kernel.

duburcqa: I just notice that `vfaces` was casted to float by mistake... I will create a PR to fix this. On your side you just need to remove this cast.

genesis/engine/solvers/fem_solver.py
```diff
@@ -240,6 +240,13 @@ def init_surface_fields(self):
             layout=ti.Layout.SOA,
         )
 
+        # UV coordinates for rendering (per-vertex UVs, initialized to zeros)
+        self.surface_render_uvs = ti.field(
+            dtype=gs.ti_vec2,
+            shape=(max(n_vertices_max, 1),),
+            needs_grad=False,
+        )
```
duburcqa: Single line if it fits.

genesis/utils/element.py
```diff
@@ -54,12 +55,27 @@ def mesh_to_elements(file, pos=(0, 0, 0), scale=1.0, tet_cfg=dict()):
...
+        original_uvs = mesh.visual.uv.copy()
+        original_uvs = original_uvs.astype(gs.np_float)
+    except AttributeError:
+        original_uvs = None
+
+    # Build full UV array: original vertices get their UVs, interior vertices get zeros
+    n_fem = len(verts)
+    uvs = np.zeros((n_fem, 2), dtype=gs.np_float)
+    if original_uvs is not None:
+        uvs[:n_original] = original_uvs
```
duburcqa: `uvs = np.pad(original_uvs, ((0, len(verts) - n_original), (0, 0)))`

genesis/utils/mesh.py
duburcqa: You convinced me and I resolved your comment, yet you chose to do what I suggested? x) 
If you are following this path, you are still missing this part 

> with a gs.logger.warning stating that texture will be skipped because they failed to load.

genesis/assets/meshes/uv_plane.obj
duburcqa, line 1: Do not version `uv_plane.obj` if it is only used in unit tests. It should rather be added to public hugging face repo `assets` https://huggingface.co/datasets/Genesis-Intelligence/assets/tree/main.

You can push directly on main. Then, update `HUGGINGFACE_ASSETS_REVISION` in `tests/utils.py‎` accordingly.

tests/test_render.py
```diff
@@ -1543,3 +1543,90 @@ def test_add_camera_vs_interactive_viewer_consistency(add_box, renderer_type, sh
...
+        morph=gs.morphs.Plane(),
+        surface=gs.surfaces.Aluminium(
+            ior=10.0,
+        ),
+    )
+
+    # Add PBD cloth with checker texture
+    scene.add_entity(
+        morph=gs.morphs.Mesh(
+            file="meshes/uv_plane.obj",
```
duburcqa: Use `get_hf_dataset`, eg:
```python
    asset_path = get_hf_dataset(pattern="heavy_three_joint_link.xml")
    robot_1 = scene.add_entity(
        gs.morphs.MJCF(
            file=f"{asset_path}/heavy_three_joint_link.xml",
            pos=(-0.2, -0.26, 0.0),
            scale=0.3,
        ),
    )
```

genesis/utils/mesh.py
alelievr: With the old code it could cause a regression if some asset had badly configured texture before. Also if it failed I needed to have add the fallback load. I can do both tough (param + try load)

genesis/engine/entities/fem_entity.py
alelievr: > Why are you not just setting self._uvs = np.zeros((0, 2), dtype=gs.np_float) instead of None

I don't know you told me to create an empty array in the last comment, so I assumed that None would not work for passing to a kernel?

genesis/engine/entities/fem_entity.py
alelievr: Ah, yes I understood something else

genesis/utils/mesh.py
duburcqa: As you want. Either way, you still need to add a warning if `skip_texture=True` failed.

genesis/engine/entities/fem_entity.py
duburcqa: > I don't know you told me to create an empty array in the last comment, so I assumed that None would not work for passing to a kernel?

Yes, I suggested to go further and not use `None` anywhere anymore. But not sure if it is a good idea. Sounds confusing (and apparently you are :p) so let's resolve this.

genesis/utils/mesh.py
alelievr: Ok, let's leave it like that :) 
I added a warning if it fails to load with the textures, do you want another warning if it still fail to load without texture (there were no warning before)

duburcqa: No, an exception is fine in the fallback fails

duburcqa (approved):

PR #2325, opened 2026-01-29
[MISC] Track visual overwrite by surface color as mesh metadata.

duburcqa:


duburcqa:
> I was not able to retrieve the info in the gltf use case unfortunately

Fixed

PR #2328, opened 2026-01-30
[BUG FIX] Fix parsing material of primitive geometries in MJCF files.

ACMLCZH:
- Add MJCF Primitives material parsing.
- Add unittests for USD-MJCF joints/primitives comparing.
- Organize unittests in `test_usd.py` and `test_mesh.py`.

`Genesis-Intelligence/Digital_twin_asset/drawer/acrylic_drawer.xml`:
drawer

### DIFF
--- a/tests/test_usd.py
+++ b/tests/test_usd.py
@@ -473,8 +473,10 @@ def test_primitives_mjcf_vs_usd(xml_path, all_primitives_usd, scale, tol):
     """Test that MJCF and USD scenes produce equivalent Genesis entities."""
     mjcf_scene = build_mjcf_scene(xml_path, scale=scale)
     usd_scene = build_usd_scene(all_primitives_usd, scale=scale)
-    # TODO: Now parsed primitives have the same geometry for both visual and collision meshes
-    # which is different from the how we parsed in MJCF.
+    # FIXME: Now parsed primitives have the same geometry for both visual and collision meshes
+    # which is different from how we parsed in MJCF.
+    # Additionally, in MuJoCo, primitives' masses are computed directly from their analytical
+    # parameters rather than using the actual mesh volume. This should be considered in USD and URDF parsing.
     compare_scene(mjcf_scene, usd_scene, tol=tol, strict=False)
 
 


tests/test_mesh.py
```diff
@@ -2,6 +2,7 @@
 import platform
 from contextlib import nullcontext
 
+import xml.etree.ElementTree as ET
```
duburcqa, line 5: This is part of standard library. Therefore it should be added to the first group of imports (in alphabetical order).

tests/test_mesh.py
duburcqa, line 1: I don't think it makes sense to order tests by Mesh / URDF / MJCF. Don't to this. You should rather order them by features sets being tested, instead of splitting apart unit tests that are checking the same thing on different types of morph.

genesis/utils/mjcf.py
```diff
@@ -550,16 +554,15 @@ def parse_geom(mj, i_g, scale, surface, xml_path):
         gs.logger.warning(f"Unsupported MJCF geom type '{mj_geom.type}'.")
         return None
 
+    if uv is not None:
+        if mj_mat is not None:
+            uv *= mj_mat.texrepeat
+    if uv is not None or tmesh_mat is not None:
+        tmesh.visual = TextureVisuals(uv=uv, material=tmesh_mat)
```
duburcqa: Pass visual as input argument of trimesh constructor instead of overwriting this parameter manually afterward.

tests/test_usd.py
```diff
@@ -460,37 +561,66 @@ def box_plane_usd(asset_tmp_path, box_plane_mjcf: ET.ElementTree):
...
 @pytest.mark.parametrize("scale", [1.0, 2.0])
 @pytest.mark.skipif(not HAS_USD_SUPPORT, reason="USD support not available")
-def test_box_plane_mjcf_vs_usd(xml_path, box_plane_usd, scale, tol):
+def test_primitives_mjcf_vs_usd(xml_path, all_primitives_usd, scale, tol):
     """Test that MJCF and USD scenes produce equivalent Genesis entities."""
     mjcf_scene = build_mjcf_scene(xml_path, scale=scale)
-    usd_scene = build_usd_scene(box_plane_usd, scale=scale)
-    compare_scene(mjcf_scene, usd_scene, tol=tol)
+    usd_scene = build_usd_scene(all_primitives_usd, scale=scale)
+    compare_scene(mjcf_scene, usd_scene, tol=tol, strict=False)
```
duburcqa, line 480: Please add a comment in the code to explain what this check is not strict.

tests/test_usd.py
```diff
@@ -220,15 +251,25 @@ def compare_vgeoms(compared_vgeoms, usd_vgeoms, tol, strict=True):
...
+    ----------
+    compared_scene : gs.Scene
+        The compared scene
+    usd_scene : gs.Scene
+        The USD scene
+    tol : float, optional
+        Tolerance for numerical comparisons.
+    strict : bool, optional
+        Whether to compare geometric details.
+    """
```
duburcqa: Keep the docstring short of helper methods in unit tests. One line is good enough. It is not like it is part of Genesis API (private or public).

tests/test_usd.py
```diff
@@ -460,37 +426,65 @@ def box_plane_usd(asset_tmp_path, box_plane_mjcf: ET.ElementTree):
...
-def test_box_plane_mjcf_vs_usd(xml_path, box_plane_usd, scale, tol):
+def test_primitives_mjcf_vs_usd(xml_path, all_primitives_usd, scale, tol):
     """Test that MJCF and USD scenes produce equivalent Genesis entities."""
     mjcf_scene = build_mjcf_scene(xml_path, scale=scale)
-    usd_scene = build_usd_scene(box_plane_usd, scale=scale)
-    compare_scene(mjcf_scene, usd_scene, tol=tol)
+    usd_scene = build_usd_scene(all_primitives_usd, scale=scale)
+    # TODO: Now parsed primitives have the same geometry for both visual and collision meshes
+    # which is different from the how we parsed in MJCF.
+    compare_scene(mjcf_scene, usd_scene, tol=tol, strict=False)
```
duburcqa: I still don't understand why this check cannot be strict. What is the root cause? 

> TODO: Now parsed primitives have the same geometry for both visual and collision meshes
    # which is different from the how we parsed in MJCF.

Sure, but why is this supposed to cause this issue? What is part of strict and what is not? I have noticed that spatial inertia comparison is part of strict. This is problematic because we expect the same inertia for primitives in USD and MJCF.

PR #2346, opened 2026-02-02
[FEATURE] Add entity filtering options to contact sensors.

YilingQiao:
## Summary

Add filtering options to `ContactSensor` for targeted collision detection:

- `with_entity_idx`: Filter contacts to only those involving a specific entity
- `exclude_self_contact`: Exclude contacts between links of the same entity

## Changes

- Add `with_entity_idx` and `exclude_self_contact` parameters to `Contact` sensor options
- Add `needs_filtering` flag to skip filtering logic when not needed (fast path)
- Fix tensor shape handling for `n_envs=0` (non-batched mode)
- Add `test_contact_sensor_filtering` test

## Usage

```python
# Detect any contact on box1
sensor_any = scene.add_sensor(gs.sensors.Contact(entity_idx=box1.idx))

# Only detect contact between box1 and box2
sensor_filtered = scene.add_sensor(
    gs.sensors.Contact(entity_idx=box1.idx, with_entity_idx=box2.idx)
)
```

## Test

```bash
pytest tests/test_sensors.py::test_contact_sensor_filtering -v
```

### DIFF
--- a/genesis/engine/sensors/contact_force.py
+++ b/genesis/engine/sensors/contact_force.py
@@ -245,6 +245,41 @@ class ContactSensor(Sensor):
         buffered_data.set(shared_ground_truth_cache)
         cls._apply_delay_to_shared_cache(shared_metadata, shared_cache, buffered_data)
 
+    @gs.assert_built
+    def read(self, envs_idx=None, force: bool = False):
+        """
+        Read the sensor data.
+
+        Parameters
+        ----------
+        envs_idx : array_like, optional
+            The indices of the environments to read. Defaults to all environments.
+        force : bool, optional
+            If True, run collision detection before reading. Defaults to False.
+        """
+        if force:
+            solver = self._shared_metadata.solver
+            # Update geoms state from links state
+            solver._func_update_geoms(solver._scene._envs_idx)
+            # Run collision detection
+            solver.collider.clear()
+            solver.collider.detection()
+
+            # Update the ground truth cache for this sensor type
+            dtype = self._get_cache_dtype()
+            cache_slice = self._manager._cache_slices_by_type[type(self)]
+            ground_truth_slice = self._manager._ground_truth_cache[dtype][:, cache_slice]
+            self._update_shared_ground_truth_cache(self._shared_metadata, ground_truth_slice)
+
+            # Copy ground truth to regular cache (skip delay/noise for force mode)
+            self._manager._cache[dtype][:, cache_slice] = ground_truth_slice
+
+            # Invalidate cloned cache so next read fetches fresh data
+            self._manager._is_last_cache_cloned[(False, dtype)] = False
+            self._manager._is_last_cache_cloned[(True, dtype)] = False
+
+        return super().read(envs_idx)
+
     def _draw_debug(self, context: "RasterizerContext", buffer_updates: dict[str, np.ndarray]):
         """
         Draw debug sphere when the sensor detects contact.
--- a/tests/test_sensors.py
+++ b/tests/test_sensors.py
@@ -650,13 +650,13 @@ def test_contact_sensor_filtering(show_viewer):
 
     # Envs 0,1: box2 touching box1; Envs 2,3: box2 far away
     box2.set_pos(np.array([[0, 0, 0.28], [0, 0, 0.28], [0, 0, 5.0], [0, 0, 5.0]]))
-    scene.step()
 
+    # Use force=True to detect collision without scene.step()
     # Any collision (all box1 touch ground)
-    assert_array_equal(sensor_any.read().flatten(), [True, True, True, True])
+    assert_array_equal(sensor_any.read(force=True).flatten(), [True, True, True, True])
 
     # Collision with specific entity (only envs 0,1 have box1-box2 contact)
-    assert_array_equal(sensor_with_box2.read().flatten(), [True, True, False, False])
+    assert_array_equal(sensor_with_box2.read(force=True).flatten(), [True, True, False, False])
 
     # Subset of environments
-    assert_array_equal(sensor_with_box2.read([0, 2]).flatten(), [True, False])
+    assert_array_equal(sensor_with_box2.read([0, 2], force=True).flatten(), [True, False])


genesis/engine/sensors/contact_force.py
```diff
@@ -125,9 +167,72 @@ def _update_shared_ground_truth_cache(
...
+        n_contacts = link_a.shape[-1]
+
+        # Handle empty contacts case
+        if n_contacts == 0:
+            shared_ground_truth_cache[:] = False
+            return
+
+        # Add batch dimension when n_envs=0 (non-batched mode)
+        if shared_metadata.solver.n_envs == 0:
+            link_a, link_b = link_a[None], link_b[None]
```
duburcqa, line 180: This is useless I think.

genesis/engine/sensors/contact_force.py
```diff
@@ -125,9 +167,72 @@ def _update_shared_ground_truth_cache(
...
         link_a, link_b = all_contacts["link_a"], all_contacts["link_b"]
-        is_contact_a = (link_a[..., None, :] == shared_metadata.expanded_links_idx[..., None]).any(dim=-1)
-        is_contact_b = (link_b[..., None, :] == shared_metadata.expanded_links_idx[..., None]).any(dim=-1)
-        shared_ground_truth_cache[:] = is_contact_a | is_contact_b
+
+        n_contacts = link_a.shape[-1]
+
+        # Handle empty contacts case
+        if n_contacts == 0:
+            shared_ground_truth_cache[:] = False
```
duburcqa, line 175: `shared_ground_truth_cache.zero_()`

genesis/engine/sensors/contact_force.py
```diff
@@ -125,9 +167,72 @@ def _update_shared_ground_truth_cache(
...
+        # Add batch dimension when n_envs=0 (non-batched mode)
+        if shared_metadata.solver.n_envs == 0:
+            link_a, link_b = link_a[None], link_b[None]
+
+        # Base check: sensor link is involved in contact
+        # Shape: link_a is (n_envs, n_contacts), expanded_links_idx is (n_sensors,)
+        # Result shape: (n_envs, n_sensors, n_contacts) -> any over contacts -> (n_envs, n_sensors)
+        sensor_in_a = (
+            link_a[..., None, :] == shared_metadata.expanded_links_idx[..., None]
+        )  # (n_envs, n_sensors, n_contacts)
```
duburcqa, line 187: Remove inline comment `# (n_envs, n_sensors, n_contacts)`

genesis/engine/sensors/contact_force.py
```diff
@@ -125,9 +167,72 @@ def _update_shared_ground_truth_cache(
...
+        # - OR sensor link in B, other (A) passes filter
+
+        # with_entity filter for "other": -1 means no filter (accept all)
+        # Expand has_with_filter to match (n_envs, n_sensors, n_contacts)
+        has_with_3d = has_with_filter[None, :, None].expand_as(link_b_in_with)
+        other_passes_with_a = ~has_with_3d | link_b_in_with  # B is other when sensor in A
+        other_passes_with_b = ~has_with_3d | link_a_in_with  # A is other when sensor in B
+
+        # exclude_self filter for "other"
+        exclude_self_3d = has_exclude_self[None, :, None].expand_as(link_b_in_self)
```
duburcqa, line 226: I don't think it is necessary to add `_3d` suffix. Just keep using the same variable name. More common pattern.

duburcqa: Besides, I don't think `expand_as(link_b_in_self)` is necessary either. If not, remove it for efficiency.

duburcqa: For efficiency, you should rather define `include_self = ~exclude_self_3d[None, :, None]` to avoid toggling the boolean array twice for efficiency.

genesis/engine/sensors/contact_force.py
```diff
@@ -125,9 +167,72 @@ def _update_shared_ground_truth_cache(
...
+        other_passes_self_a = ~exclude_self_3d | ~link_b_in_self  # B not in self entity
+        other_passes_self_b = ~exclude_self_3d | ~link_a_in_self  # A not in self entity
+
+        # Combine: sensor in contact AND other passes all filters
+        valid_a = sensor_in_a & other_passes_with_a & other_passes_self_a
+        valid_b = sensor_in_b & other_passes_with_b & other_passes_self_b
+
+        # Any valid contact per sensor per env
+        shared_ground_truth_cache[:] = (valid_a | valid_b).any(dim=-1)
```
duburcqa, line 235: The slow path implementation is getting fairly complex. Could you jit compile it for efficiency? Just move the implementation in its own input-output standalone function and decorate it with `@torch.jit.script`. See all methods with prefix name `_tc_` defined in `genesis/utils/geom.py` for reference.

genesis/engine/sensors/contact_force.py
```diff
@@ -140,6 +245,41 @@ def _update_shared_cache(
...
+            cache_slice = self._manager._cache_slices_by_type[type(self)]
+            ground_truth_slice = self._manager._ground_truth_cache[dtype][:, cache_slice]
+            self._update_shared_ground_truth_cache(self._shared_metadata, ground_truth_slice)
+
+            # Copy ground truth to regular cache (skip delay/noise for force mode)
+            self._manager._cache[dtype][:, cache_slice] = ground_truth_slice
+
+            # Invalidate cloned cache so next read fetches fresh data
+            self._manager._is_last_cache_cloned[(False, dtype)] = False
+            self._manager._is_last_cache_cloned[(True, dtype)] = False
```
duburcqa, line 279: I'm not ok with adding this option. Update the physics should never be the responsibility of the sensors. This is completely defeating the purpose of sensors that is reading the state of the simulation. And this is so against the current design that you have to hack around and alter the manager directly...

If you want to force physics update, just add some extra function to the rigid solver, or the scene itself if you really when to make it public, that will update the entire state of the simulation, with some optional arguments to disable some update if necessary for efficiency but only if really necessary. Just call it `update`, `refresh` or something like that.

duburcqa:
We do not add features pro-actively without at least well identified use cases. Could you clarify in the description of this PR what are the targeted use cases of this feature? Note that following our template for PR would have avoided having this question in the first place since there is already a dedicated section for motivations.

PR #2350, opened 2026-02-03
[MISC] Speed up linesearch via batched alpha evals and reduced global memory access.

erizmr:
## Description

The primary bottleneck in line search is the evaluation function func_ls_point_fn, which traverses all constraints and is invoked multiple times per iteration (2+3*refinement steps)*ls_iter. To reduce both the number of evaluations and the cost per evaluation, we apply the following optimizations:

- Reduced global memory traffic. Initialization is fused with the $p_0$ evaluation, and state updates during bracket refinement are moved from global memory read/write to local read/write. This is particularly beneficial in cases with a large number of constraints, where working data no longer fits in L1/L2 cache.
- Batched alpha evaluations. During the refinement stage, the two bracket points and the midpoint are fetched once and evaluated together. While line search typically converges within the first two iterations and does not enter refinement for simpler cases, this optimization significantly benefits harder cases with slower convergence.
- Constraint loop specialization. Constraint loops are split by type, and an equality pre-sum is introduced to eliminate redundant computations.

### DIFF
--- a/genesis/engine/solvers/rigid/constraint/solver.py
+++ b/genesis/engine/solvers/rigid/constraint/solver.py
@@ -2016,14 +2016,14 @@ def func_ls_init_and_eval_p0_opt(
 
     # Return p0 result (alpha=0)
     cost = quad_total_0
-    deriv_0 = quad_total_1
-    deriv_1 = 2 * quad_total_2
-    if deriv_1 <= 0.0:
-        deriv_1 = rigid_global_info.EPS[None]
+    grad = quad_total_1
+    hess = 2 * quad_total_2
+    if hess <= 0.0:
+        hess = rigid_global_info.EPS[None]
 
     constraint_state.ls_it[i_b] = 1
 
-    return gs.ti_float(0.0), cost, deriv_0, deriv_1
+    return gs.ti_float(0.0), cost, grad, hess
 
 
 @ti.func
@@ -2088,14 +2088,14 @@ def func_ls_point_fn_opt(
         quad_total_2 = quad_total_2 + qf_2 * active
 
     cost = alpha * alpha * quad_total_2 + alpha * quad_total_1 + quad_total_0
-    deriv_0 = 2 * alpha * quad_total_2 + quad_total_1
-    deriv_1 = 2 * quad_total_2
-    if deriv_1 <= 0.0:
-        deriv_1 = rigid_global_info.EPS[None]
+    grad = 2 * alpha * quad_total_2 + quad_total_1
+    hess = 2 * quad_total_2
+    if hess <= 0.0:
+        hess = rigid_global_info.EPS[None]
 
     constraint_state.ls_it[i_b] = constraint_state.ls_it[i_b] + 1
 
-    return alpha, cost, deriv_0, deriv_1
+    return alpha, cost, grad, hess
 
 
 @ti.func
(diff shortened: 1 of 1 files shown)


genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -1886,25 +1886,44 @@ def func_cholesky_solve_tiled(
...
+    jv, quad, quad_gauss) and evaluating the cost at alpha=0 — into a single pass. It also pre-computes eq_sum, the
+    summed quadratic coefficients for equality constraints, and stores it for reuse by subsequent evaluation calls.
+
+    The key optimization is exploiting the fact that equality constraints are always active regardless of alpha. Their
+    contribution to the cost function is constant across all linesearch iterations, so it only needs to be computed once
+    during initialization. By pre-summing eq_sum here, all subsequent calls to func_ls_point_fn_opt and
+    func_ls_point_fn_3alphas_opt can skip iterating over equality constraints entirely, reducing the per-iteration
+    constraint loop from n_constraints to (n_friction + n_contact).
+
+    This optimization is most beneficial for scenes with many equality constraints (e.g. articulated robots with joint
```
duburcqa: Joint limits are not equality constraints but inequality. Equality constraints are weld or distance constraints. Almost never used in practice.

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -1924,70 +1943,151 @@ def func_ls_init(
...
     constraint_state: array_class.ConstraintState,
     rigid_global_info: array_class.RigidGlobalInfo,
 ):
+    """Evaluate linesearch cost, gradient, and curvature at a single candidate alpha.
+
+    This function computes the quadratic cost and its first two derivatives at a given step size alpha by iterating over
+    only friction and contact constraints. Equality constraints are skipped entirely by initializing the accumulators
+    from quad_gauss + eq_sum, where eq_sum was pre-computed by func_ls_init_and_eval_p0_opt during initialization.
+
+    Since equality constraints are always active (their contribution is alpha-independent in the quadratic coefficients),
```
duburcqa: Could you avoid repeating yourself and just point to the other kernel as reference?

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -1924,70 +1943,151 @@ def func_ls_init(
...
     nef = ne + constraint_state.n_constraints_frictionloss[i_b]
+    n_con = constraint_state.n_constraints[i_b]
 
-    tmp_quad_total_0, tmp_quad_total_1, tmp_quad_total_2 = gs.ti_float(0.0), gs.ti_float(0.0), gs.ti_float(0.0)
-    tmp_quad_total_0 = constraint_state.quad_gauss[0, i_b]
-    tmp_quad_total_1 = constraint_state.quad_gauss[1, i_b]
-    tmp_quad_total_2 = constraint_state.quad_gauss[2, i_b]
-    for i_c in range(constraint_state.n_constraints[i_b]):
+    # Start from quad_gauss + eq_sum (skips ne equality constraints)
+    tmp_0 = constraint_state.quad_gauss[0, i_b] + constraint_state.eq_sum[0, i_b]
```
duburcqa: Please avoid tmp_ as variable name.

genesis/engine/solvers/rigid/constraint/solver_island.py
duburcqa, line 1: I will suggest to forget about island if it is not necessary to update it. We should probably rewrite it from strath. It was never updated and unit tested. Missing all the features we added since December 2024.

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -2128,54 +2358,52 @@ def func_linesearch_batch(
...
+                            c1,
+                            c1_d0,
+                            c1_d1,
+                            _a2,
+                            c2,
+                            c2_d0,
+                            c2_d1,
+                        ) = func_ls_point_fn_3alphas_opt(
+                            i_b, alpha_0, alpha_1, alpha_2, constraint_state, rigid_global_info
                         )
```
duburcqa:
```python
(_a0, c0, c0_d0, c0_d1, _a1, c1, c1_d0, c1_d1, _a2, c2, c2_d0, c2_d1) = (
    func_ls_point_fn_3alphas_opt(
        i_b, alpha_0, alpha_1, alpha_2, constraint_state, rigid_global_info
    )
)
```

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -2198,21 +2438,41 @@ def func_linesearch_batch(
...
+                                c0_d0,
+                                c0_d1,
+                                alpha_1,
+                                c1,
+                                c1_d0,
+                                c1_d1,
+                                alpha_2,
+                                c2,
+                                c2_d0,
+                                c2_d1,
```
duburcqa: Do you think it would be possible to pack them somehow. All these arguments are stretching the code and making everything harder to read, with code duplication for c0, c1, c2... etc.

Would it be possible to store them in fixed size 3 vectors and statically loop (unrolling) over them in computations?

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -2128,54 +2358,52 @@ def func_linesearch_batch(
...
+                            c0_d0,
+                            c0_d1,
+                            _a1,
+                            c1,
+                            c1_d0,
+                            c1_d1,
+                            _a2,
+                            c2,
+                            c2_d0,
+                            c2_d1,
```
duburcqa: All these variable names are cryptic. Would it be possible to find more explicit names? And or not, at least document in comment what is `a[0-3]`, `c[0-3]`, `c[0-3]_[0-3]` and `d[0-3]_[0-3]`?

genesis/engine/solvers/rigid/constraint/solver.py
erizmr: pre-commit looks prefer the vertical version. After we use the ti.Vector as container, it is shorter and fit to one line.

genesis/engine/solvers/rigid/constraint/solver.py
erizmr: Renamed them.

genesis/engine/solvers/rigid/constraint/solver.py
erizmr: Changed to use `ti.Vector` to contain `alphas`, `costs`, `grads`, `hess`.

genesis/engine/solvers/rigid/constraint/solver.py
duburcqa: > pre-commit looks prefer the vertical version. 

Pre-commit does not "prefer" one of the other. It just pick one that is valid for whatever reason, but you can edit and use the one you like and it should not revert it.

duburcqa (approved):

PR #2351, opened 2026-02-03
[BUG FIX] Fix OpenGL context clash when combining Rasterizer-based camera sensors and interactive viewer.

YilingQiao:
## Description

- Fix OpenGL context conflict when using `RasterizerCameraSensor` with `show_viewer=True`
- Fix `RaytracerCameraSensor` crash when attached without explicit `offset_T`

**RasterizerCameraSensor**: `add_camera` creates a Camera that gets built during `visualizer.build()`, after the rasterizer is already built, so it can directly use `visualizer.rasterizer` and register immediately. `add_sensor` creates a sensor that gets built during `sim.build()`, which happens *before* `visualizer.build()`. Originally, the sensor created its own standalone Rasterizer with `viewer=None`, which made `_offscreen=True` and caused it to create a separate `OffscreenRenderer` with its own OpenGL context. When both the viewer's pyglet context and the sensor's offscreen context exist, they conflict.

**RaytracerCameraSensor**: When `offset_T` is None, the sensor now uses `pos` as the translation offset (consistent with `BaseCameraSensor.move_to_attach()`).

You can try 
- `pytest tests/test_render.py::test_rasterizer_camera_sensor_with_viewer`
- `pytest tests/test_sensor_camera.py::test_raytracer_attached_without_offset_T`

Those two tests fail on main

## Related Issue

Resolves https://github.com/Genesis-Embodied-AI/Genesis/discussions/2347

## Checklist:

### DIFF
--- a/tests/test_sensor_camera.py
+++ b/tests/test_sensor_camera.py
@@ -9,7 +9,7 @@ import genesis as gs
 from genesis.utils.misc import tensor_to_array
 from genesis.utils.geom import trans_to_T
 
-from .utils import assert_allclose, rgb_array_to_png_bytes
+from .utils import assert_allclose, assert_array_equal, rgb_array_to_png_bytes
 
 
 try:
@@ -427,48 +427,39 @@ def test_raytracer_attached_without_offset_T():
     Also checks consistency with a scene-level camera (scene.add_camera) using the same
     pose and attachment, to make sure both camera APIs produce matching output.
     """
-    CAM_RES = (64, 64)
+    CAM_RES = (128, 64)
     CAM_POS = (0.0, 0.0, 2.0)
-    CAM_LOOKAT = (0.0, 0.0, 0.0)
-    CAM_UP = (0.0, 1.0, 0.0)
-    CAM_FOV = 30.0
-    CAM_SPP = 64
-    CAM_DENOISE = False
 
     scene = gs.Scene(renderer=gs.renderers.RayTracer())
     scene.add_entity(morph=gs.morphs.Plane())
-    sphere = scene.add_entity(morph=gs.morphs.Sphere(pos=(0.0, 0.0, 0.0)))
+    sphere = scene.add_entity(morph=gs.morphs.Sphere())
 
     # Sensor camera attached WITHOUT offset_T - should use pos as offset
+    camera_common_options = dict(
+        res=CAM_RES,
+        lookat=(0.0, 0.0, 0.0),
+        up=(0.0, 1.0, 0.0),
+        fov=30.0,
+        spp=64,
+        denoise=False,
+    )
     sensor_camera = scene.add_sensor(
         gs.sensors.RaytracerCameraOptions(
-            res=CAM_RES,
+            **camera_common_options,
             pos=CAM_POS,
-            lookat=CAM_LOOKAT,
-            up=CAM_UP,
-            fov=CAM_FOV,
-            spp=CAM_SPP,
-            denoise=CAM_DENOISE,
             entity_idx=sphere.idx,
         )
     )
 
     # Scene-level camera with the same pose, attached with explicit offset_T
     scene_camera = scene.add_camera(
-        res=CAM_RES,
-        pos=CAM_POS,
-        lookat=CAM_LOOKAT,
-        up=CAM_UP,
-        fov=CAM_FOV,
-        spp=CAM_SPP,
-        denoise=CAM_DENOISE,
+        **camera_common_options,
     )
 
     scene.build()
 
     # Attach scene-level camera with equivalent offset_T
-    link = sphere.links[0]
-    scene_camera.attach(link, trans_to_T(np.array(CAM_POS)))
+    scene_camera.attach(sphere.base_link, offset_T=trans_to_T(np.array(CAM_POS)))
 
     scene.step()
 
@@ -478,14 +469,11 @@ def test_raytracer_attached_without_offset_T():
 
     scene_camera.move_to_attach()
     scene_rgb, *_ = scene_camera.render(rgb=True, force_render=True)
-    scene_rgb = tensor_to_array(scene_rgb)
-    sensor_rgb = tensor_to_array(sensor_data.rgb)
+    scene_rgb = tensor_to_array(scene_rgb, dtype=np.int32)
+    sensor_rgb = tensor_to_array(sensor_data.rgb, dtype=np.int32)
 
     # Both cameras should produce the same image
-    assert np.allclose(sensor_rgb, scene_rgb, atol=1), (
-        f"Sensor camera and scene-level camera produce different images "
-        f"(max diff={np.abs(sensor_rgb.astype(int) - scene_rgb.astype(int)).max()})"
-    )
+    assert_array_equal(sensor_rgb, scene_rgb)
 
 
 @pytest.mark.required


genesis/engine/sensors/camera.py
```diff
@@ -160,10 +160,10 @@ def destroy(self):
 
         if self.renderer is not None:
             self.renderer.destroy()
-            self.renderer = None
         if self.context is not None:
             self.context.destroy()
-            self.context = None
+        self.renderer = None
+        self.context = None
```
duburcqa: This change is useless. Revert it.

genesis/engine/sensors/camera.py
```diff
@@ -431,29 +432,47 @@ def build(self):
...
+            viewer = getattr(visualizer, "_viewer", None) if visualizer else None
+
+            if viewer is not None:
+                # Use visualizer's rasterizer - it will use viewer.render_offscreen() which handles context properly
+                self._shared_metadata.context = visualizer.context
+                self._shared_metadata.renderer = visualizer.rasterizer
+            else:
+                # No viewer - create standalone rasterizer with offscreen context
+                self._shared_metadata.context = self._create_standalone_context(scene)
+                self._shared_metadata.renderer = Rasterizer(viewer=None, context=self._shared_metadata.context)
```
duburcqa: We should support multiple OpenGL context in parallel. Here you are hiding a bug while complexifying the implementation.

genesis/engine/sensors/camera.py
```diff
@@ -699,7 +727,7 @@ def _render_current_state(self):
             antialiasing=False,
             force_render=True,
         )
-        rgb_tensor = torch.from_numpy(rgb_arr).to(dtype=torch.uint8, device=gs.device)
+        rgb_tensor = torch.from_numpy(rgb_arr.copy()).to(dtype=torch.uint8, device=gs.device)
```
duburcqa: Why are you forcing a copy. This is wrong.

tests/test_sensor_camera.py
```diff
@@ -418,6 +418,31 @@ def test_raytracer_destroy():
...
+            pos=(0.0, 0.0, 2.0),
+            entity_idx=sphere.idx,
+        )
+    )
+
+    scene.build()
+    data = camera.read()
+
+    assert data.rgb.shape == (64, 64, 3)
+    assert data.rgb.float().std() > 1.0, "RGB std too low, image may be blank"
```
duburcqa: I would rather update `test_raytracer` to add `pos=(0.0, 0.0, 2.0),` to `camera_1` instead of adding a new test? Because every single test we add is very costly (it requires building an entire scene), so it can be avoided, it is much better.

tests/test_render.py
```diff
@@ -1543,3 +1543,31 @@ def test_add_camera_vs_interactive_viewer_consistency(add_box, renderer_type, sh
         f"interactive viewer brightness ({viewer_brightness:.2f}), "
         f"but ratio is {brightness_ratio:.2f}"
     )
+
+
+@pytest.mark.required
+@pytest.mark.parametrize("renderer_type", [RENDERER_TYPE.RASTERIZER])
+@pytest.mark.skipif(not IS_INTERACTIVE_VIEWER_AVAILABLE, reason="Interactive viewer not supported on this platform.")
+@pytest.mark.xfail(sys.platform == "win32", raises=OpenGL.error.Error, reason="Invalid OpenGL context.")
```
duburcqa: Why are you doing this? You need to have very strong motivations to add some xfail. Here you are basically saying that your fix does not work on windows.

tests/test_render.py
```diff
@@ -1543,3 +1543,31 @@ def test_add_camera_vs_interactive_viewer_consistency(add_box, renderer_type, sh
...
+def test_rasterizer_camera_sensor_with_viewer(renderer):
+    """Test that RasterizerCameraSensor works correctly when interactive viewer is enabled.
+
+    This verifies that the sensor properly shares the viewer's OpenGL context instead of
+    creating a conflicting separate context.
+    """
+    from genesis.options.sensors import RasterizerCameraOptions
+
```
duburcqa: No import inside functions.

tests/test_render.py
```diff
@@ -1543,3 +1543,31 @@ def test_add_camera_vs_interactive_viewer_consistency(add_box, renderer_type, sh
...
+    scene = gs.Scene(
+        viewer_options=gs.options.ViewerOptions(run_in_thread=(sys.platform == "linux")),
```
duburcqa: `viewer_options=gs.options.ViewerOptions(run_in_thread=False),`

tests/test_sensor_camera.py
```diff
@@ -418,6 +418,31 @@ def test_raytracer_destroy():
...
+@pytest.mark.required
+@pytest.mark.parametrize("backend", [gs.cuda])
+@pytest.mark.skipif(not ENABLE_RAYTRACER, reason="RayTracer is not supported because 'LuisaRenderPy' is not available.")
+def test_raytracer_attached_without_offset_T():
+    """Test that RaytracerCameraSensor works when attached without explicit offset_T."""
+    scene = gs.Scene(renderer=gs.renderers.RayTracer())
+    scene.add_entity(morph=gs.morphs.Plane())
+    sphere = scene.add_entity(morph=gs.morphs.Sphere(pos=(0.0, 0.0, 0.0)))
+
+    # Create attached camera WITHOUT offset_T - should use pos as offset (consistent with other cameras)
```
duburcqa: Could you check consistency with other camera explicitly? That would be much better.

tests/test_render.py
```diff
@@ -1543,3 +1543,31 @@ def test_add_camera_vs_interactive_viewer_consistency(add_box, renderer_type, sh
...
+        renderer=renderer,
+        show_viewer=True,
+    )
+    scene.add_entity(morph=gs.morphs.Plane())
```
duburcqa, line 1631: Add a comment to explain that adding at least one entity is necessary to make sure the rendered image is not entirely blank, otherwise it is not possible to make sure something was actually rendered.

tests/test_sensor_camera.py
duburcqa: Could you check consistency with other camera explicitly? That would be much better.

genesis/engine/sensors/camera.py
```diff
@@ -540,6 +559,10 @@ def _apply_camera_transform(self, camera_T: torch.Tensor):
 
     def _render_current_state(self):
         """Perform the actual render for the current state."""
+        # Deferred registration for when using visualizer's rasterizer (visualizer now built)
+        if not self._camera_registered:
```
duburcqa: The `if` condition is useless since `self._register_camera_with_renderer()` is already an no-op if already registered.

genesis/engine/sensors/camera.py
```diff
@@ -431,29 +432,47 @@ def build(self):
...
+
+        # Add lights from options to the context
         for light_config in self._options.lights:
             if self._shared_metadata.lights is not None:
-                # Convert light config to rasterizer format
                 light_dict = self._convert_light_config_to_rasterizer(light_config)
-                # self._shared_metadata.lights.append(light_dict)
                 self._shared_metadata.context.add_light(light_dict)
 
         camera_wrapper = self._get_camera_wrapper()
```
duburcqa: While you are here, what is the point of `_get_camera_wrapper` ? This function seems useless. It was already the case because, but it is even worse now. Could you remove `_get_camera_wrapper` entirely and replace it direct access to `self._camera_wrapper` ? just add `self._register_camera_with_renderer()` on top of all functions that requires a valid camera.

genesis/engine/sensors/camera.py
```diff
@@ -652,7 +675,12 @@ def build(self):
 
         # Attach the visualizer camera to the link if this sensor is attached
         if self._link is not None:
-            offset_T = torch.tensor(self._options.offset_T, dtype=gs.tc_float, device=gs.device)
+            if self._options.offset_T is not None:
+                offset_T = torch.tensor(self._options.offset_T, dtype=gs.tc_float, device=gs.device)
+            else:
+                # Use pos as offset translation (consistent with BaseCameraSensor.move_to_attach)
+                pos = torch.tensor(self._options.pos, dtype=gs.tc_float, device=gs.device)
+                offset_T = trans_quat_to_T(pos, torch.tensor([1.0, 0.0, 0.0, 0.0], dtype=gs.tc_float, device=gs.device))
```
duburcqa: Just use `gu.trans_to_T`. It was introduced for this very use case.

genesis/engine/sensors/camera.py
duburcqa: While you are here, what is the point of `_get_camera_wrapper` ? This function seems useless. It was already the case because, but it is even worse now. Could you remove `_get_camera_wrapper` entirely and replace it direct access to `self._camera_wrapper` ? just add `self._register_camera_with_renderer()` on top of all functions that requires a valid camera.

genesis/engine/sensors/camera.py
duburcqa: Just use `gu.trans_to_T`. It was introduced for this very use case.

genesis/engine/sensors/camera.py
duburcqa: The `if` condition is useless since `self._register_camera_with_renderer()` is already an no-op if already registered.

genesis/engine/sensors/camera.py
duburcqa: While you are here, what is the point of `_get_camera_wrapper` ? This function seems useless. It was already the case because, but it is even worse now. Could you remove `_get_camera_wrapper` entirely and replace it direct access to `self._camera_wrapper` ? just add `self._register_camera_with_renderer()` on top of all functions that requires a valid camera.

genesis/engine/sensors/camera.py
```diff
@@ -431,29 +432,47 @@ def build(self):
...
+        # Register camera now if standalone (offscreen), or defer to first render if using visualizer's rasterizer
+        # (visualizer isn't built yet at sensor.build() time)
+        if self._shared_metadata.renderer.offscreen:
+            self._register_camera_with_renderer()
+
+        _B = max(self._manager._sim.n_envs, 1)
+        w, h = self._options.res
+        self._shared_metadata.image_cache[self._idx] = torch.zeros((_B, h, w, 3), dtype=torch.uint8, device=gs.device)
+
+    def _register_camera_with_renderer(self):
```
duburcqa: Rename this function `_ensure_camera_registered` or anything that may indicate that it is safe to call this function repeatedly. Moreover, what is the point of adding `_with_renderer`? If you think it is useful you can keep it but I'm not sure it is adding much value here.

genesis/engine/sensors/camera.py
```diff
@@ -417,6 +417,7 @@ def __init__(
         self._camera_node = None
         self._camera_target = None
         self._camera_wrapper = None
+        self._camera_registered = False
```
duburcqa: Please use `is_` (or `has_`) prefix for boolean flags.

genesis/engine/sensors/camera.py
```diff
@@ -431,29 +432,47 @@ def build(self):
...
             self._shared_metadata.image_cache = {}
 
-            # Create standalone rasterizer
-            self._shared_metadata.context = self._create_standalone_context(scene)
-
-            self._shared_metadata.renderer = Rasterizer(viewer=None, context=self._shared_metadata.context)
-            self._shared_metadata.renderer.build()
+            # Check if scene has a viewer - if so, use visualizer's rasterizer (like add_camera does)
+            # This avoids OpenGL context conflicts since visualizer's rasterizer routes through viewer.render_offscreen()
+            visualizer = getattr(scene, "visualizer", None)
```
duburcqa: `scene.visualizer` is guaranteed to exist. No point to use `getattr` here.

genesis/engine/sensors/camera.py
```diff
@@ -431,29 +432,47 @@ def build(self):
...
+            viewer = getattr(visualizer, "_viewer", None) if visualizer else None
```
duburcqa: `visualizer._viewer` is guaranteed to exist. No point to use `getattr` here. Moreover, no need to access it via visualizer, you could get it directly from the scene, which is the expected public API:
```python
viewer = scene.viewer
```

genesis/engine/sensors/camera.py
```diff
@@ -431,29 +432,47 @@ def build(self):
...
+
+            if viewer is not None:
+                # Use visualizer's rasterizer - it will use viewer.render_offscreen() which handles context properly
+                self._shared_metadata.context = visualizer.context
+                self._shared_metadata.renderer = visualizer.rasterizer
```
duburcqa: I don't think it is relevant to define proxies for `scene.viewer` and `scene.visualizer`, just use them directly wherever to need to access them.

genesis/engine/sensors/camera.py
```diff
@@ -431,29 +432,47 @@ def build(self):
...
-
-            self._shared_metadata.renderer = Rasterizer(viewer=None, context=self._shared_metadata.context)
-            self._shared_metadata.renderer.build()
+            # Check if scene has a viewer - if so, use visualizer's rasterizer (like add_camera does)
+            # This avoids OpenGL context conflicts since visualizer's rasterizer routes through viewer.render_offscreen()
+            visualizer = getattr(scene, "visualizer", None)
+            viewer = getattr(visualizer, "_viewer", None) if visualizer else None
+
+            if viewer is not None:
+                # Use visualizer's rasterizer - it will use viewer.render_offscreen() which handles context properly
```
duburcqa: The point is more about the fact that the same windowed OpenGL context will be reused, which is capable of handling both offscreen and onscreen rendering, whereas managing two contexts, with one being headless and the other windowed, is proven to be very painful and fragile.

tests/test_sensor_camera.py
YilingQiao: removed this test

tests/test_sensor_camera.py
YilingQiao: deleted this test

genesis/engine/sensors/camera.py
YilingQiao: I think we will go with this implementation for now?

duburcqa: yes

tests/test_sensor_camera.py
duburcqa: Why did you deleted this test? It was the only one failing without your bug fix for `offset_T` no? I think it is important?

Anyway, it would be great of any a unit test to check consistency with standard scene-level cameras and we don't have any at the moment.

genesis/engine/sensors/camera.py
```diff
@@ -522,35 +539,30 @@ def _update_camera_pose(self):
...
-            self._get_camera_wrapper(),
+            self._camera_wrapper,
             rgb=True,
             depth=False,
             segmentation=False,
             normal=False,
         )
 
-        rgb_tensor = torch.from_numpy(rgb_arr.copy()).to(dtype=torch.uint8, device=gs.device)
+        rgb_tensor = torch.from_numpy(rgb_arr).to(dtype=torch.uint8, device=gs.device)
```
duburcqa: The CI is currently failing with the error:
```python
>       rgb_tensor = torch.from_numpy(rgb_arr).to(dtype=torch.uint8, device=gs.device)
E       ValueError: At least one stride in the given numpy array is negative, and tensors with negative strides are not currently supported. (You can probably work around this by making a copy of your array  with array.copy().)
```

This line should be replaced by:
```python
rgb_tensor = torch.tensor(rgb_arr, dtype=torch.uint8, device=gs.device)
```
With a comment of course explaining why this is necessary.

YilingQiao: rgb_tensor = torch.tensor(rgb_arr, dtype=torch.uint8, device=gs.device)

cannot avoid that error, will use  np.ascontiguousarray, it will only copy when the array actually has negative strides

duburcqa (dismissed):

duburcqa (approved):

PR #2353, opened 2026-02-03
[BUG FIX] Fix keyboard teleop when viewer has focus via reserved_keys

Royalvice:
## Description

When the viewer window has focus, keyboard events are consumed by the pyrender/pyglet event loop, so pynput-based teleop (e.g. arrow keys in `keyboard_teleop`) does not receive them. This PR adds an optional **reserved_keys** mechanism so that application-defined keys take priority over the viewer’s shortcuts for those keys only, without disabling the viewer’s shortcuts for other keys.

**Changes:**
- **genesis/options/vis.py**: Add `ViewerOptions.reserved_keys` (optional tuple of pyglet key symbols). When set, the viewer does not use these keys for its shortcuts and tracks their state for `get_pressed_keys()`.
- **genesis/vis/viewer.py**: Pass `reserved_keys` to the pyrender Viewer; add `get_pressed_keys()` that delegates to the pyrender viewer.
- **genesis/ext/pyrender/viewer.py**: Add `reserved_keys` parameter; when non-empty, track key state for those symbols, skip default shortcut handling for them, and expose `get_pressed_keys()` and `on_deactivate()` to clear state on focus loss.
- **examples/keyboard_teleop.py**: Use `ViewerOptions(reserved_keys=RESERVED_KEYS)` for teleop keys and merge `scene.viewer.get_pressed_keys()` with pynput so teleop works with or without viewer focus; do not use `disable_keyboard_shortcuts`, so GUI shortcuts (e.g. [a], [z], [r]) remain available.

Backward compatible: if `reserved_keys` is not set, viewer behavior is unchanged.

## Motivation and Context

In `keyboard_teleop.py`, arrow keys and other teleop keys had no effect when the user was focused on the Genesis viewer, because the viewer’s event loop received the key events first and pynput did not see them. The only workaround was `disable_keyboard_shortcuts=True`, which disables all viewer shortcuts (camera reset, record, etc.). This PR fixes the bug in a non-invasive way: only keys listed in `reserved_keys` are reserved for the application; all other keys keep their default viewer shortcuts.

- **Environment**: Windows 10, Python with Genesis and pynput installed.
- **Steps**:
  1. Run `python examples/keyboard_teleop.py`.
  2. Click inside the viewer window to give it focus.
  3. Press arrow keys (↑↓←→) and n/m/j/k/u/space: end-effector and gripper should respond.
  4. Press [a], [z], [r], etc.: viewer shortcuts (camera rotation, reset, record) should still work.
  5. Optionally click outside the viewer and press the same teleop keys: pynput fallback should still work.
- All tests were run and passed.

N/A

## Checklist:

### DIFF
--- a/examples/keyboard_teleop.py
+++ b/examples/keyboard_teleop.py
@@ -11,6 +11,10 @@ k	- Rotate Clockwise
 u	- Reset Scene
 space	- Press to close gripper, release to open gripper
 esc	- Quit
+
+Uses ViewerOptions.reserved_keys so these keys are handled by the application
+(when the viewer has focus) and do not trigger the viewer's shortcuts; the
+viewer's shortcuts (e.g. [a], [z], [r]) remain available for other keys.
 """
 
 import os
@@ -19,8 +23,24 @@ import threading
 
 import genesis as gs
 import numpy as np
+import pyglet.window.key as pyglet_key
 from pynput import keyboard
 from scipy.spatial.transform import Rotation as R
+
+# Keys reserved for this example: viewer will not use them for its shortcuts
+RESERVED_KEYS = (
+    pyglet_key.UP,
+    pyglet_key.DOWN,
+    pyglet_key.LEFT,
+    pyglet_key.RIGHT,
+    pyglet_key.N,
+    pyglet_key.M,
+    pyglet_key.J,
+    pyglet_key.K,
+    pyglet_key.U,
+    pyglet_key.SPACE,
+    pyglet_key.ESCAPE,
+)
 
 
 class KeyboardDevice:
@@ -74,6 +94,7 @@ def build_scene():
             camera_lookat=(0.2, 0.0, 0.1),
             camera_fov=50,
             max_FPS=60,
+            reserved_keys=RESERVED_KEYS,
         ),
         show_viewer=True,
         show_FPS=False,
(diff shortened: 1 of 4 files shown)


YilingQiao:
In ViewerOption, there is an option called disable_keyboard_shortcuts. Could you use that? Is there a case where you want to control both the viewer and also capture other keyboard input?

Royalvice:
> In ViewerOption, there is an option called disable_keyboard_shortcuts. Could you use that? Is there a case where you want to control both the viewer and also capture other keyboard input?

Yeah, I see disable_keyboard_shortcuts
The current keyboard_teleop example has a bug: when the viewer window has focus, the custom keys (arrows, n/m/j/k/u, space, esc) are consumed by the viewer’s event loop and never reach pynput, so pressing them does nothing. Using disable_keyboard_shortcuts=True would fix that by turning off all viewer shortcuts so the app can see those keys, but then you lose every viewer shortcut ([a], [z], [r], etc.). The reserved_keys change fixes the bug in a targeted way: only the keys you list are reserved for the app and exposed via get_pressed_keys(), while all other keys still trigger the viewer’s shortcuts, so both custom teleop keys and viewer shortcuts work at the same time.

duburcqa:
This PR is conflicting with the ongoing refactoring of keyboard handling here:
https://github.com/Genesis-Embodied-AI/Genesis/pull/2004

Closing for now, and we will reopen if it does not fit your needs.

duburcqa:
This PR is conflicting with the ongoing refactoring of keyboard handling here:
https://github.com/Genesis-Embodied-AI/Genesis/pull/2004

Closing for now, and we will reopen if it does not fit your needs.

duburcqa:
This PR is conflicting with the ongoing refactoring of keyboard handling here:
https://github.com/Genesis-Embodied-AI/Genesis/pull/2004

Closing for now. Please try this PR and let us know if it fits your needs. If not, we will find a way to address this issue.

Royalvice:
@duburcqa Great! I think this PR is sufficient to solve my current problem

duburcqa:
> Great! I think this PR is sufficient to solve my current problem

Perfect! Happy to hear that 👍

PR #2357, opened 2026-02-05
[FEATURE] Add example viewer plugins.

Milotrince:
## Description

implement example `ViewerPlugin`s, continuation of https://github.com/Genesis-Embodied-AI/Genesis/pull/2004

Refactors raycaster to enable point-and-click/mouse raycasting.

https://github.com/user-attachments/assets/afee5ab6-3878-4cb1-8e4e-489d38347213
https://github.com/user-attachments/assets/568a4bd3-eb4f-429c-82cb-617dc0d5a0b2

## Checklist:

### DIFF
--- a/tests/test_viewer.py
+++ b/tests/test_viewer.py
@@ -6,10 +6,11 @@ import OpenGL.error
 import pytest
 
 import genesis as gs
+from genesis.utils.misc import tensor_to_array
 from genesis.vis.keybindings import Key, KeyAction, Keybind, KeyMod, MouseButton
 
 from .conftest import IS_INTERACTIVE_VIEWER_AVAILABLE
-from .utils import assert_allclose, rgb_array_to_png_bytes
+from .utils import assert_allclose
 
 CAM_RES = (480, 320)
 
@@ -168,10 +169,12 @@ def test_mouse_interaction_plugin():
             gravity=(0.0, 0.0, 0.0),
         ),
         viewer_options=gs.options.ViewerOptions(
+            # Forces odd resolution so that mouse clicks are centered on pixels
+            res=(2 * (CAM_RES[0] // 2) + 1, 2 * (CAM_RES[0] // 2) + 1),
             camera_pos=CAM_POS,
-            camera_lookat=(0.0, 0.0, BOX_LENGTH),  # looking to the top of the box
+            # looking to the top of the box
+            camera_lookat=(0.0, 0.0, BOX_LENGTH),
             camera_fov=CAM_FOV,
-            res=(CAM_RES[0] + 1, CAM_RES[1] + 1),  # odd resolution so that mouse clicks are centered on pixels
             run_in_thread=(sys.platform == "linux"),
         ),
         show_viewer=True,
@@ -188,7 +191,7 @@ def test_mouse_interaction_plugin():
             rho=MASS / (BOX_LENGTH**3),
         ),
     )
-    mouse_plugin = scene.viewer.add_plugin(
+    _mouse_plugin = scene.viewer.add_plugin(
         gs.vis.viewer_plugins.MouseInteractionPlugin(
             use_force=True,
             spring_const=SPRING_CONST,
@@ -260,7 +263,7 @@ def test_mouse_interaction_plugin():
         err_msg="Final x and y velocities should be near zero since dragging only in z direction.",
     )
 
-    distance_to_box = np.linalg.norm(np.array(CAM_POS) - initial_pos.numpy())
+    distance_to_box = np.linalg.norm(tensor_to_array(initial_pos) - CAM_POS)
     pixels_to_world = 2.0 * distance_to_box * np.tan(np.radians(CAM_FOV) / 2.0) / viewport_size[1]
     total_world_displacement = STEPS * DRAG_DY * pixels_to_world
 
@@ -289,6 +292,6 @@ def test_mouse_interaction_plugin():
     assert_allclose(
         final_vel[2],
         expected_vel_z,
-        rtol=0.4,
+        rtol=0.5,
         err_msg="Final z velocity does not match expected value based on spring dynamics.",
     )


duburcqa:
This is so cool!

Milotrince:
snapshot pr: https://huggingface.co/datasets/Genesis-Intelligence/snapshots/discussions/8

examples/viewer_plugin/mesh_point_selector.py
Milotrince, line 1: I left the implementation in the example file here, as an example of implementing a custom plugin and also since this one is a bit specific for general use cases to include in genesis/

Milotrince: my use case is to use the point outputs file with https://github.com/Genesis-Embodied-AI/Genesis/pull/2292 to easily create arbitrary placements of tactile sensors!

duburcqa: Sounds good!

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
```diff
@@ -0,0 +1,311 @@
...
+                self.plane_rotation_angle = 0.0
+
+    @override
+    def update_on_sim_step(self) -> None:
+        super().update_on_sim_step()
+
+        with self.lock:
+            if self.held_link:
+                from genesis.utils.raycast import Ray, plane_raycast
+
```
duburcqa: It is necessary to do the import here rather than on top of this file? If so, could you please add a comment to explain what is the issue?

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
duburcqa, line 1: Why are you using `dtype=np.float32` everywhere instead of `gs.np_float`?

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
```diff
@@ -0,0 +1,311 @@
+from threading import Lock as threading_Lock
```
duburcqa: I think `Lock` is fine. I would not recommend to alias it as `threading_Lock`.

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
```diff
@@ -0,0 +1,311 @@
...
+                    self.plane_rotation_angle = 0.0
+                    self.prev_mouse_scene_pos = ray_hit.position
+
+                    # Create drag plane perpendicular to surface normal
+                    self._update_drag_plane()
+
+                    link_pos = tensor_to_array(self.held_link.get_pos())
+                    link_quat = tensor_to_array(self.held_link.get_quat())
+                    hit_pos_arr = np.array(ray_hit.position, dtype=np.float32)
+                    self.held_point_in_local = tuple(gu.inv_transform_by_trans_quat(hit_pos_arr, link_pos, link_quat))
```
duburcqa: Why as a tuple? this is weird.

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
```diff
@@ -0,0 +1,311 @@
...
+                    self.surface_normal = np.array(ray_hit.normal, dtype=np.float32)
+                    self.plane_rotation_angle = 0.0
+                    self.prev_mouse_scene_pos = ray_hit.position
+
+                    # Create drag plane perpendicular to surface normal
+                    self._update_drag_plane()
+
+                    link_pos = tensor_to_array(self.held_link.get_pos())
+                    link_quat = tensor_to_array(self.held_link.get_quat())
+                    hit_pos_arr = np.array(ray_hit.position, dtype=np.float32)
```
duburcqa: Why `ray_hit.position` is not already a np.ndarray?

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
```diff
@@ -0,0 +1,311 @@
...
+                    self.held_point_in_local = tuple(gu.inv_transform_by_trans_quat(hit_pos_arr, link_pos, link_quat))
+
+    @override
+    def on_mouse_release(self, x: int, y: int, button: int, modifiers: int) -> EVENT_HANDLE_STATE:
+        if button == 1:  # left mouse button
```
duburcqa: There is no enum for "left mouse button" that could be used in place of `button` ?

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
```diff
@@ -0,0 +1,311 @@
...
+
+        self.held_link: "RigidLink | None" = None
+        self.held_point_in_local: tuple[float, float, float] | None = None
+        self.mouse_drag_plane: tuple[np.ndarray, float] | None = None
+        self.prev_mouse_screen_pos: tuple[int, int] = (0, 0)
+        self.prev_mouse_scene_pos: tuple[float, float, float] | None = None
+        self.surface_normal: np.ndarray | None = None
+        self.plane_rotation_angle: float = 0.0
+
+        self.lock = threading_Lock()
```
duburcqa: Are you sure you should use `Lock` instead of `RLock` ? What are you guarding against with this lock?

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
```diff
@@ -0,0 +1,311 @@
...
+        self,
+        use_force: bool = True,
+        spring_const: float = 1.0,
+        spring_damping: float = 1.0,
+        color: tuple[float, float, float, float] = (0.1, 0.6, 0.8, 0.6),
+    ) -> None:
+        super().__init__()
+        self.spring_const = spring_const
+        self.spring_damping = spring_damping
+        self.color = color
```
duburcqa: `self.color = tuple(color)` just in case?

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
```diff
@@ -0,0 +1,311 @@
...
+    def __init__(
+        self,
+        use_force: bool = True,
+        spring_const: float = 1.0,
+        spring_damping: float = 1.0,
+        color: tuple[float, float, float, float] = (0.1, 0.6, 0.8, 0.6),
+    ) -> None:
+        super().__init__()
+        self.spring_const = spring_const
+        self.spring_damping = spring_damping
```
duburcqa: Same, maybe cast to these variables to float just in case `float(spring_const)`

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
```diff
@@ -0,0 +1,311 @@
...
+    def on_mouse_press(self, x: int, y: int, button: int, modifiers: int) -> EVENT_HANDLE_STATE:
+        if button == 1:  # left mouse button
+            ray = self._screen_position_to_ray(x, y)
+            ray_hit = self._raycaster.cast(ray[0], ray[1])
+            with self.lock:
+                if ray_hit.geom and ray_hit.geom.link is not None and not ray_hit.geom.link.is_fixed:
+                    self.held_link = ray_hit.geom.link
+
+                    # Store the surface normal for rotation
+                    self.surface_normal = np.array(ray_hit.normal, dtype=np.float32)
```
duburcqa: Why ray_hit.normal is not already a np.ndarray ? What is the type?

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
```diff
@@ -0,0 +1,311 @@
...
+                    # apply displacement
+                    pos = tensor_to_array(self.held_link.entity.get_pos())
+                    pos = pos + delta_3d_pos
+                    self.held_link.entity.set_pos(torch.from_numpy(pos))
+
+    @override
+    def on_draw(self) -> None:
+        if self.scene._visualizer is not None and self.scene._visualizer.is_built:
+            self.scene.clear_debug_objects()
+            mouse_ray = self._screen_position_to_ray(*self.prev_mouse_screen_pos)
```
duburcqa: `position, direction = self._screen_position_to_ray(*self.prev_mouse_screen_pos)`

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
```diff
@@ -0,0 +1,311 @@
...
+
+    @override
+    def update_on_sim_step(self) -> None:
+        super().update_on_sim_step()
+
+        with self.lock:
+            if self.held_link:
+                from genesis.utils.raycast import Ray, plane_raycast
+
+                mouse_ray_tuple = self._screen_position_to_ray(*self.prev_mouse_screen_pos)
```
duburcqa:
```python
origin, direction = self._screen_position_to_ray(*self.prev_mouse_screen_pos)
mouse_ray = Ray(origin, direction)
```

genesis/vis/viewer_plugins/viewer_plugin.py
```diff
@@ -65,3 +67,66 @@ def on_draw(self) -> None:
...
+        self._raycaster = Raycaster(self.scene)
+        self._camera_tan_half_fov = np.tan(0.5 * self.camera.camera.yfov)
+
+    @override
+    def update_on_sim_step(self) -> None:
+        super().update_on_sim_step()
+
+        self._raycaster.update()
+
+    def _screen_position_to_ray(self, x: float, y: float) -> tuple[np.ndarray, np.ndarray]:
```
duburcqa: You should probably return a named tuple instead?

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
```diff
@@ -0,0 +1,311 @@
...
+                assert self.mouse_drag_plane is not None
+                ray_hit: "RayHit" = plane_raycast(*self.mouse_drag_plane, mouse_ray)
+
+                # If ray doesn't hit the plane, skip this update
+                if ray_hit is None:
+                    return
+
+                new_mouse_3d_pos = ray_hit.position
+                new_pos_arr = np.array(new_mouse_3d_pos, dtype=np.float32)
+                prev_pos_arr = np.array(self.prev_mouse_scene_pos, dtype=np.float32)
```
duburcqa: Same. I don't like all these cast back and forth between tuple and ndarrays. Why not just use ndarray systematically?

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
```diff
@@ -0,0 +1,311 @@
...
+                delta_3d_pos = new_pos_arr - prev_pos_arr
+                self.prev_mouse_scene_pos = new_mouse_3d_pos
+
+                if self.use_force:
+                    self._apply_spring_force(new_mouse_3d_pos, self.scene.sim.dt)
+                else:
+                    # apply displacement
+                    pos = tensor_to_array(self.held_link.entity.get_pos())
+                    pos = pos + delta_3d_pos
+                    self.held_link.entity.set_pos(torch.from_numpy(pos))
```
duburcqa: Same. You convert to ndarray only to convert back to torch. It would be much better to use `torch` tensor all along. At least cast `delta_3d_pos` to torch instead. Ideally, keep everything as torch tensor from the start.

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
```diff
@@ -0,0 +1,311 @@
...
+                            size=0.5,
+                            color=self.plane_color,
+                        )
+
+                else:
+                    if closest_hit is not None:
+                        self.scene.draw_debug_arrow(
+                            np.array(closest_hit.position, dtype=np.float32),
+                            np.array(closest_hit.normal, dtype=np.float32) * 0.25,
+                            color=self.color,
```
duburcqa: Could we avoid all these `np.array([...], dtype=np.float32)` conversion?

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
```diff
@@ -0,0 +1,311 @@
...
+                [size, -size, 0],
+                [size, size, 0],
+                [-size, size, 0],
+            ],
+            dtype=np.float32,
+        )
+        # Create double-sided faces so plane is visible from both sides
+        faces = np.array([[0, 1, 2], [0, 2, 3], [2, 1, 0], [3, 2, 0]], dtype=np.int32)
+        plane_mesh = trimesh.Trimesh(vertices=vertices, faces=faces)
+        plane_mesh.visual = trimesh.visual.ColorVisuals(vertex_colors=np.tile(np.array([color]), [len(vertices), 1]))
```
duburcqa: Why not to use our utility `create_plane` ?

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
```diff
@@ -0,0 +1,311 @@
...
+            if abs(surface_normal[0]) < 0.9:
+                cam_proj = np.array([1.0, 0.0, 0.0], dtype=np.float32)
+            else:
+                cam_proj = np.array([0.0, 1.0, 0.0], dtype=np.float32)
+            cam_proj = cam_proj - np.dot(cam_proj, surface_normal) * surface_normal
+
+        cam_proj = cam_proj / (np.linalg.norm(cam_proj) + 1e-8)
+
+        # Apply rotation around the surface normal using geom utils
+        if abs(self.plane_rotation_angle) > 1e-6:
```
duburcqa: gs.EPS instead of `1e-6`?

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
```diff
@@ -0,0 +1,311 @@
...
+
+        # If camera is looking along the normal, use an arbitrary tangent vector
+        if cam_proj_norm < 1e-3:
+            if abs(surface_normal[0]) < 0.9:
+                cam_proj = np.array([1.0, 0.0, 0.0], dtype=np.float32)
+            else:
+                cam_proj = np.array([0.0, 1.0, 0.0], dtype=np.float32)
+            cam_proj = cam_proj - np.dot(cam_proj, surface_normal) * surface_normal
+
+        cam_proj = cam_proj / (np.linalg.norm(cam_proj) + 1e-8)
```
duburcqa: Same, maybe use `gs.EPS` instead of `1e-8`?

duburcqa: This looks very much like what we do in `z_up_to_R` at first glance no?

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
```diff
@@ -0,0 +1,311 @@
...
+            ang_vel = ang_vel + arm_x_dir * impulse * inv_spherical_inertia
+
+            total_impulse[i % 3] += impulse
+            total_torque_impulse += arm_x_dir * impulse
+
+        # Apply the new force
+        total_force = total_impulse * inv_dt
+        total_torque = total_torque_impulse * inv_dt
+        force_tensor: torch.Tensor = torch.tensor(total_force, dtype=torch.float32)[None]
+        torque_tensor: torch.Tensor = torch.tensor(total_torque, dtype=torch.float32)[None]
```
duburcqa: Why not to keep using `torch.Tensor` all along instead of converting first to numpy then back to torch?

duburcqa: `[None]` should be unnecessary. Remove it.

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
```diff
@@ -0,0 +1,311 @@
...
+
+            dir = np.zeros(3, dtype=np.float32)
+            dir[i % 3] = 1.0
+
+            pos_err = np.dot(dir, pos_err_v)
+            vel_err = np.dot(dir, vel_err_v)
+            error = tau * pos_err * inv_dt + damp * vel_err
+
+            arm_x_dir = np.cross(arm_in_world, dir)
+            virtual_mass = 1.0 / (inv_mass + np.dot(arm_x_dir, arm_x_dir) * inv_spherical_inertia + 1e-24)
```
duburcqa: `gs.EPS` instead of `1e-24`?

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
```diff
@@ -0,0 +1,311 @@
...
+        inertial_quat = tensor_to_array(link.inertial_quat)
+
+        world_principal_quat = gu.transform_quat_by_quat(inertial_quat, link_quat)
+
+        arm_in_principal = gu.inv_transform_by_trans_quat(held_point_local_arr, inertial_pos, inertial_quat)
+        arm_in_world = gu.transform_by_quat(arm_in_principal, world_principal_quat)
+
+        control_point_arr = np.array(control_point, dtype=np.float32)
+        pos_err_v = control_point_arr - held_point_in_world
+        inv_mass: float = float(1.0 / link.get_mass() if link.get_mass() > 0.0 else 0.0)
```
duburcqa: I think `link.get_mass()` is guaranteed to be larger than zero. Anyway, it is `gs.EPS`, then `inv_mass` would still explode. Same issue with the inertia matrix. I would suggest to throw a warning if `mass < 1e-3` and prevent dragging such objects.

genesis/vis/viewer_plugins/plugins/mouse_interaction.py
```diff
@@ -0,0 +1,311 @@
...
+        inv_spherical_inertia: float = float(1.0 / link.inertial_i[0, 0] if link.inertial_i[0, 0] > 0.0 else 0.0)
```
duburcqa: Why are you using spherical inertia instead of full inertia matrix? This simplification seems very inaccurate and unnecessary.

If it turns out being necessary, you should rather do `float(1.0 / np.mean(np.diag(link.inertial_i)))`.

duburcqa: Besides, I'm wondering whether it would make more sense to use `invweight` or `meaninertia`, or even the true apparent link inertia instead of the constant link-local isolated inertia.

tests/test_viewer.py
```diff
@@ -150,3 +149,55 @@ def toggle_flag(idx):
...
+
+
+@pytest.mark.required
+@pytest.mark.skipif(not IS_INTERACTIVE_VIEWER_AVAILABLE, reason="Interactive viewer not supported on this platform.")
+def test_mouse_interaction_plugin(png_snapshot):
+    scene = gs.Scene(
+        sim_options=gs.options.SimOptions(
+            dt=1e-2,
+            substeps=1,
+        ),
```
duburcqa, line 170: Remove these options if they are not necessary to keep only what truly matters in unit tests.

tests/test_viewer.py
```diff
@@ -150,3 +149,55 @@ def toggle_flag(idx):
...
+    for _ in range(30):
+        x += dx
+        y += dy
+        pyrender_viewer.dispatch_event("on_mouse_drag", x, y, dx, dy, 1, 0)
+        scene.step()
+
+    rgb_arr, *_ = pyrender_viewer.render_offscreen(
+        pyrender_viewer._camera_node, pyrender_viewer._renderer, rgb=True, depth=False, seg=False, normal=False
+    )
+    assert rgb_array_to_png_bytes(rgb_arr) == png_snapshot
```
duburcqa: You should probably take multiple screenshot and make sure that all the images are different?

tests/test_viewer.py
duburcqa: by the way, what happens if gravity is 0.0? It may be easier to check that the force being applied matches expectation? If you use spheres instead of boxes, it should be easier to validate the analytical formula by comparing the current linear and angular acceleration with the expected one?

genesis/utils/raycast.py
```diff
@@ -0,0 +1,463 @@
...
+
+
+@ti.func
+def get_triangle_vertices(
+    i_f: ti.i32,
+    i_b: ti.i32,
+    faces_info: ti.template(),
+    verts_info: ti.template(),
+    fixed_verts_state: ti.template(),
+    free_verts_state: ti.template(),
```
duburcqa: Are you sure that you should use `ti.template()` instead of `array_clas.V_ANNOTATION`?

genesis/utils/raycast.py
```diff
@@ -0,0 +1,463 @@
...
+    ray_start: ti.template(),
+    ray_dir: ti.template(),
+    max_range: ti.f32,
+    i_b: ti.i32,
+    bvh_nodes: ti.template(),
+    bvh_morton_codes: ti.template(),
+    faces_info: ti.template(),
+    verts_info: ti.template(),
+    fixed_verts_state: ti.template(),
+    free_verts_state: ti.template(),
```
duburcqa: Are you sure that you should use `ti.template()` instead of `array_clas.V_ANNOTATION`?

genesis/utils/raycast.py
```diff
@@ -0,0 +1,463 @@
...
+                if stack_idx < ti.static(STACK_SIZE - 2):
+                    node_stack[stack_idx] = node.left
+                    node_stack[stack_idx + 1] = node.right
+                    stack_idx += 2
+
+    return hit_face, closest_distance, hit_normal
+
+
+@ti.func
+def ray_triangle_intersection(ray_start, ray_dir, v0, v1, v2):
```
duburcqa: No annotations for this one?

genesis/utils/raycast.py
```diff
@@ -0,0 +1,463 @@
...
+@ti.kernel
+def kernel_update_verts_and_aabbs(
+    geoms_info: array_class.GeomsInfo,
+    geoms_state: array_class.GeomsState,
+    verts_info: array_class.VertsInfo,
+    faces_info: array_class.FacesInfo,
+    free_verts_state: array_class.VertsState,
+    fixed_verts_state: array_class.VertsState,
+    static_rigid_sim_config: ti.template(),
+    aabb_state: ti.template(),
```
duburcqa: `abb_state` is always a ti.field and never ti.types.ndarray ? Otherwise you should use `array_clas.V_ANNOTATION` ?

genesis/utils/raycast.py
```diff
@@ -0,0 +1,463 @@
...
+    free_verts_state: ti.template(),
+    verts_info: ti.template(),
+    faces_info: ti.template(),
+    bvh_nodes: ti.template(),
+    bvh_morton_codes: ti.template(),
+    ray_start: ti.types.ndarray(ndim=1),  # (3,)
+    ray_direction: ti.types.ndarray(ndim=1),  # (3,)
+    max_range: ti.f32,
+    envs_idx: ti.types.ndarray(ndim=1),  # [n_envs]
+    result: array_class.RaycastResult,
```
duburcqa: Are you sure that you should use `ti.template()` instead of `array_clas.V_ANNOTATION`?

duburcqa:
> snapshot pr: https://huggingface.co/datasets/Genesis-Intelligence/snapshots/discussions/8

Merged!

examples/viewer_plugin/mesh_point_selector.py
duburcqa, line 1: Are these examples running on the CI? They should if not already the case!

genesis/utils/raycast_ti.py
duburcqa, line 1: I don't think using `_ti` suffix is a good idea. Why do you think it is useful?

Milotrince: these require `array_class` which means they cant be imported at top level because `array_class` cannot be imported before `gs.init`. i moved `Ray` and `RayHit` and `plane_raycast` to `raycast.py` which does not import `array_class`

genesis/utils/raycast.py
Milotrince: I updated `verts_state` and others to use the appropriate array_class. but we do use `ti.template()` in LBVH for `bvh_nodes` and `bvh_morton_codes` so I left those

tests/test_viewer.py
Milotrince: I think taking one screenshot and validating the force applied should be sufficient for this test

duburcqa: Why would it be sufficient? There is no proof that anything happened.

genesis/utils/raycast.py
duburcqa: Perfect.

genesis/utils/raycast_ti.py
duburcqa: Ok I see, there are 2 files. Ok then it makes sense.

tests/test_viewer.py
Milotrince: the scene is initialized with the object in the center of the screen and after dragging the snapshot shows that the object is moved?

duburcqa: > the scene is initialized with the object in the center of the screen and after dragging the snapshot shows that the object is moved?

ok but this is human-based. The less we need to rely on human-validation in unit test the better. Here you know what you need to pay attention to, but what about the future maintainers if your unit test starts failing for whatever reason in 1 year?

Milotrince: ok, updated snapshot: https://huggingface.co/datasets/Genesis-Intelligence/snapshots/discussions/9

tests/test_viewer.py
Milotrince: is the current test ok with checking against two screenshots instead of calculating the expected velocity?

tests/test_viewer.py
duburcqa: You don’t need 2 snapshots for this. Just need to store and ndarrays and check that the diff is non-zero.

tests/test_viewer.py
duburcqa: Calculating the expected velocity is both more strict (you check that it does what it is really supposed to do) and more robust (snapshot are horrendous to maintain because the rendering output depending on the device and driver being used).

ideally no snapshot should be used at all. They are only here in last resort. In this case I don’t think it is necessary.

Besides, you need to disable shadows on the CI if you want to use snapshots because they are mint supported during CPU emulation.

tests/test_viewer.py
```diff
@@ -147,3 +148,145 @@ def toggle_flag(idx):
...
+    num_tau = total_sim_time * np.sqrt(SPRING_CONST / MASS)
+    velocity_fraction = 1.0 - (1.0 + num_tau) * np.exp(-num_tau)
+    expected_vel_z = avg_mouse_velocity * velocity_fraction
+
+    assert_allclose(
+        final_vel[2],
+        expected_vel_z,
+        rtol=0.4,  # the forces from mouse spring are approximate, so use a large tolerance
+        err_msg="Final z velocity does not match expected value based on spring dynamics.",
+    )
```
Milotrince, line 297: @duburcqa I updated the mouse interaction test to be lifting up a box. with using a sphere I couldnt get the mouse drag plane normal to be perfectly z-up without setting any private variables on the plugin so this is about as accurate as i could get it to be. any suggestions?

duburcqa: Let's go with this for now if the unit tests, and revise this later.

duburcqa: Could you add a FIXME to remember us that we should investigate whether it is impossible to improve tolerance?

PR #2360, opened 2026-02-06
[MISC] Cleanup unit test infra.

duburcqa:


duburcqa:
snapshots need to be renamed since the path as changed.

PR #2389, opened 2026-02-10
[FEATURE] Add KinematicContactProbe sensor.

Milotrince:
## Description
This PR replaces https://github.com/Genesis-Embodied-AI/Genesis/pull/2292

I implement KinematicContactProbe as:
1. `solver.collider.get_contacts()`
2. Filter contacts by links the probes are attached to
3. For each probe, raycast against geoms in contact in the direction of -probe_normal
image

4. For probes with no rayhit, compute collision of probe sphere vs geoms, use distance to contact point projected along normal as penetration depth
image

## Related Issue
Resolves https://github.com/Genesis-Embodied-AI/Genesis/issues/2254

## Motivation and Context
With contact information from the solver, we get contacts only at support points.
The kinematic probes should be able to get information about "contact" at arbitrary points.

updated `test_sensors.py`

https://github.com/user-attachments/assets/e18bbf79-0c1b-42be-8b69-b93fa30159bf

## Checklist:

### DIFF
--- a/examples/sensors/kinematic_contact_probe.py
+++ b/examples/sensors/kinematic_contact_probe.py
@@ -93,7 +93,7 @@ def main():
         gs.morphs.Cylinder(
             radius=PUSHER_SIZE,
             height=PUSHER_SIZE,
-            pos=tuple(pusher_start),
+            pos=pusher_start,
         ),
         surface=gs.surfaces.Default(
             color=(0.15, 0.55, 0.95, 1.0),
--- a/genesis/engine/sensors/kinematic_contact_probe.py
+++ b/genesis/engine/sensors/kinematic_contact_probe.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, NamedTuple, Sequence, Type
+from typing import TYPE_CHECKING, NamedTuple, Type
 
 import numpy as np
 import quadrants as qd
@@ -345,10 +345,10 @@ class KinematicContactProbe(
 
         self._shared_metadata.total_n_probes += n_probes
 
-        if isinstance(self._options.radius, Sequence):
-            radii_tensor = torch.tensor(self._options.radius, dtype=gs.tc_float, device=gs.device)
-        else:
+        if isinstance(self._options.radius, float):
             radii_tensor = torch.full((n_probes,), self._options.radius, dtype=gs.tc_float, device=gs.device)
+        else:
+            radii_tensor = torch.tensor(self._options.radius, dtype=gs.tc_float, device=gs.device)
 
         self._shared_metadata.radii = concat_with_tensor(
             self._shared_metadata.radii, radii_tensor, expand=(n_probes,), dim=0
--- a/genesis/options/sensors/options.py
+++ b/genesis/options/sensors/options.py
@@ -1,17 +1,20 @@
-from typing import TYPE_CHECKING, Sequence
+from typing import TYPE_CHECKING
 
 import numpy as np
-from pydantic import Field
+from pydantic import Field, conlist
 
 import genesis as gs
 
 from ..options import Options
 from .raycaster import DepthCameraPattern, RaycastPattern
 
-Tuple3FType = tuple[float, float, float]
-MaybeTuple3FType = float | Tuple3FType
-Matrix3x3Type = tuple[tuple[float, float, float], tuple[float, float, float], tuple[float, float, float]]
-MaybeMatrix3x3Type = Matrix3x3Type | MaybeTuple3FType
+Vec3FType = conlist(float, min_length=3, max_length=3)
+Vec4FType = conlist(float, min_length=4, max_length=4)
+Vec3FArrayType = conlist(Vec3FType, min_length=1)
+FArrayType = conlist(float, min_length=1)
+MaybeVec3FType = float | Vec3FType
+Matrix3x3Type = conlist(conlist(float, min_length=3, max_length=3), min_length=3, max_length=3)
+MaybeMatrix3x3Type = Matrix3x3Type | MaybeVec3FType
 
 if TYPE_CHECKING:
     from genesis.engine.scene import Scene
@@ -62,16 +65,16 @@ class RigidSensorOptionsMixin:
         The global entity index of the RigidEntity to which this sensor is attached. -1 or None for static sensors.
     link_idx_local : int, optional
         The local index of the RigidLink of the RigidEntity to which this sensor is attached.
-    pos_offset : tuple[float, float, float]
+    pos_offset : array-like[float, float, float], optional
         The positional offset of the sensor from the RigidLink.
-    euler_offset : tuple[float, float, float]
+    euler_offset : array-like[float, float, float], optional
         The rotational offset of the sensor from the RigidLink in degrees.
     """
 
     entity_idx: int | None = -1
     link_idx_local: int = 0
-    pos_offset: Tuple3FType = (0.0, 0.0, 0.0)
-    euler_offset: Tuple3FType = (0.0, 0.0, 0.0)
+    pos_offset: Vec3FType = (0.0, 0.0, 0.0)
+    euler_offset: Vec3FType = (0.0, 0.0, 0.0)
 
     def validate(self, scene: "Scene"):
         from genesis.engine.entities import RigidEntity
@@ -93,14 +96,14 @@ class NoisySensorOptionsMixin:
 
     Parameters
     ----------
-    resolution : float | tuple[float, ...], optional
+    resolution : float | array-like[float, ...], optional
         The measurement resolution of the sensor (smallest increment of change in the sensor reading).
         Default is 0.0, which means no quantization is applied.
-    bias : float | tuple[float, ...], optional
+    bias : float | array-like[float, ...], optional
         The constant additive bias of the sensor.
-    noise : float | tuple[float, ...], optional
+    noise : float | array-like[float, ...], optional
         The standard deviation of the additive white noise.
-    random_walk : float | tuple[float, ...], optional
+    random_walk : float | array-like[float, ...], optional
         The standard deviation of the random walk, which acts as accumulated bias drift.
     jitter : float, optional
         The jitter in seconds modeled as a a random additive delay sampled from a normal distribution.
@@ -110,10 +113,10 @@ class NoisySensorOptionsMixin:
         Otherwise, the sensor data at the closest time step will be used. Default is False.
     """
 
-    resolution: float | tuple[float, ...] = 0.0
-    bias: float | tuple[float, ...] = 0.0
-    noise: float | tuple[float, ...] = 0.0
-    random_walk: float | tuple[float, ...] = 0.0
+    resolution: float | FArrayType = 0.0
+    bias: float | FArrayType = 0.0
+    noise: float | FArrayType = 0.0
+    random_walk: float | FArrayType = 0.0
     jitter: float = 0.0
     interpolate: bool = False
 
@@ -132,12 +135,12 @@ class Contact(RigidSensorOptionsMixin, SensorOptions):
     ----------
     debug_sphere_radius : float, optional
         The radius of the debug sphere. Defaults to 0.05.
-    debug_color : float, optional
+    debug_color : array-like[float, float, float, float], optional
         The rgba color of the debug sphere. Defaults to (1.0, 0.0, 1.0, 0.5).
     """
 
     debug_sphere_radius: float = 0.05
-    debug_color: tuple[float, float, float, float] = (1.0, 0.0, 1.0, 0.5)
+    debug_color: Vec4FType = (1.0, 0.0, 1.0, 0.5)
 
 
 class ContactForce(RigidSensorOptionsMixin, NoisySensorOptionsMixin, SensorOptions):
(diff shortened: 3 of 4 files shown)


tests/test_sensors.py
```diff
@@ -627,3 +627,112 @@ def test_lidar_cache_offset_parallel_env(show_viewer, tol):
...
+    assert_allclose(sphere_data.normal, sphere_ground_truth.normal, tol=gs.EPS)
+    assert_allclose(sphere_data.force, sphere_ground_truth.force, tol=gs.EPS)
+
+    zeros_3 = (0.0, 0.0, 0.0)
+    assert_allclose(true_data.penetration, (PENETRATION, 0.0, 0.0), tol=tol)
+    assert_allclose(true_data.position, ((0.0, 0.0, BOX_SIZE / 2 - PENETRATION), zeros_3, zeros_3), tol=tol)
+    assert_allclose(true_data.normal, ((0.0, 0.0, 1.0), zeros_3, zeros_3), tol=tol)
+    assert_allclose(true_data.force, ((0.0, 0.0, STIFFNESS * PENETRATION), zeros_3, zeros_3), tol=tol)
+    # TODO: This is failing, probably due to how the sensor uses support points?
+    assert_allclose(sphere_data.penetration, PENETRATION, tol=tol)
```
Milotrince: will un-draft the PR once I address this

genesis/engine/solvers/rigid/collider/narrowphase.py
```diff
@@ -47,6 +43,19 @@ class CCD_ALGORITHM_CODE(IntEnum):
...
+def func_point_in_geom_aabb(
+    i_g: ti.i32,
+    i_b: ti.i32,
+    geoms_state: array_class.GeomsState,
+    point: ti.types.vector(3, gs.ti_float),
+    expansion: gs.ti_float = 0.0,
+):
+    aabb_min = geoms_state.aabb_min[i_g, i_b] - expansion
+    aabb_max = geoms_state.aabb_max[i_g, i_b] + expansion
+    return (point >= aabb_min).all() and (point <= aabb_max).all()
```
YilingQiao: you changed >< to >=, <=, is this intentional?

genesis/engine/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,448 @@
...
+@register_sensor(KinematicContactProbeOptions, KinematicContactProbeMetadata, KinematicContactProbeData)
+@ti.data_oriented
+class KinematicContactProbe(
+    RigidSensorMixin[KinematicContactProbeMetadata],
+    NoisySensorMixin[KinematicContactProbeMetadata],
+    Sensor[KinematicContactProbeMetadata],
+):
+    """Kinematic contact probe measuring penetration depth along the probe normal on collisions."""
+
+    _update_on_read: bool = True
```
YilingQiao: it's not used

genesis/engine/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,448 @@
...
+            for side in ti.static(range(2)):
+                contact_link = c_link_a if side == 0 else c_link_b
+                i_g = c_geom_b if side == 0 else c_geom_a
+
+                # Is this contact relevant to this sensor?
+                if contact_link == sensor_link_idx and func_point_in_geom_aabb(
+                    i_g, i_b, geoms_state, probe_pos, radius
+                ):
+                    # Update vertices only for geoms that passed filtering
+                    func_update_verts_for_geom(
```
YilingQiao: Potential data race. Different probes can call this function from different threads. Not sure if it will cause bug, but it's certainly expensive? Will it be more efficient if you update verts for all geoms when using this sensor?

genesis/options/sensors/options.py
```diff
@@ -1,6 +1,7 @@
 from typing import TYPE_CHECKING, Sequence
 
 import numpy as np
+import torch
```
YilingQiao: unused

genesis/engine/solvers/rigid/collider/broadphase.py
```diff
@@ -59,12 +77,7 @@ def func_check_collision_valid(
         i_lb = geoms_info.link_idx[i_gb]
 
         # Filter out collision pairs that are involved in dynamically registered weld equality constraints
-        for i_eq in range(rigid_global_info.n_equalities[None], constraint_state.ti_n_equalities[i_b]):
-            if equalities_info.eq_type[i_eq, i_b] == gs.EQUALITY_TYPE.WELD:
-                i_leqa = equalities_info.eq_obj1id[i_eq, i_b]
-                i_leqb = equalities_info.eq_obj2id[i_eq, i_b]
-                if (i_leqa == i_la and i_leqb == i_lb) or (i_leqa == i_lb and i_leqb == i_la):
-                    is_valid = False
+        is_valid = func_check_weld_constraint(i_la, i_lb, i_b, rigid_global_info, constraint_state, equalities_info)
```
YilingQiao: why modify this?

genesis/engine/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,448 @@
...
+    sensor_cache_start: ti.types.ndarray(),
+    sensor_probe_start: ti.types.ndarray(),
+    collider_state: array_class.ColliderState,
+    geoms_state: array_class.GeomsState,
+    geoms_info: array_class.GeomsInfo,
+    fixed_verts_state: array_class.VertsState,
+    free_verts_state: array_class.VertsState,
+    verts_info: array_class.VertsInfo,
+    faces_info: array_class.FacesInfo,
+    default_raycast_depth: ti.f32,
```
YilingQiao: gs.ti_float

examples/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,225 @@
...
+        print("  [\\]         Reset pusher position")
+    else:
+        print(f"Running headless for {args.seconds}s ...")
+    print()
+
+    # ── Simulation loop ────────────────────────────────────────────────
+    steps = int(args.seconds / scene.sim_options.dt) if not args.vis else None
+    step = 0
+
+    try:
```
YilingQiao, line 186: do we need this `try`?

genesis/engine/sensors/kinematic_contact_probe.py
duburcqa: Race condition is not possible for our Taichi kernels. But definitely useless to call this update multiple times of it is the case.

genesis/engine/sensors/kinematic_contact_probe.py
duburcqa: Actually we use raw taichi types in signatures at every places for now. This is legacy because in the past Genesis was not guaranteed to be initialized before importing taichi kernels modules.

genesis/engine/sensors/kinematic_contact_probe.py
Milotrince: `func_update_verts_for_geom` already does a `geoms_state.verts_updated` check. since in most cases there are few geoms in contact, updating all geoms once is probably more expensive than updating the ones we need?

genesis/engine/solvers/rigid/collider/broadphase.py
Milotrince: earlier i used this check for the probe kernel but I dont need it now since we have the filter on contacts from the collider. will remove this change

examples/sensors/kinematic_contact_probe.py
Milotrince: its not critical but we have this pattern in other examples as well, clean break on keyboard interrupt

genesis/engine/sensors/kinematic_contact_probe.py
duburcqa: sure, but calling a kernel may be costly even if it does nothing, if the GPU queue is already empty.

Anyway, I'm in favour of starting with something simple and optimising at a later stage, rather than keeping this PR is the pipe forever.

Milotrince: we're already in a kernel, this is calling a ti func so that cost is not incurred right?

duburcqa: Ah yes indeed! in that case it is already perfect I think.

duburcqa: Ok After reading the code more carefully, @YilingQiao is right. there is a race condition issue because you are parallelising over both envs and probes, and probes could be attached to the same geometries. The way it is currently guarded by `verts_updated` is not thread-safe.

duburcqa: I think you should do what @YilingQiao suggest and start updating first from all geoms. If you think this may be too costly, then pass a pre-computed list of geometry indices to update as input argument of the kernel. You should be able to parallelise over envs + updated geometries, so it should be very fast.

examples/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,226 @@
...
+            y = (j - centre) * spacing
+            z = platform_height / 2  # top surface in link-local frame
+
+            r = PROBE_RADIUS + i * (PROBE_RADIUS / 10.0)
+
+            positions.append((x, y, z))
+            normals.append((0.0, 0.0, 1.0))
+            radii.append(float(r))
+
+    return positions, normals, radii
```
duburcqa, line 44: What about vectoring this?
```python
i = np.repeat(np.arange(grid_n), grid_n)
j = np.tile(np.arange(grid_n), grid_n)
x = (i - centre) * spacing
y = (j - centre) * spacing
z = np.full_like(x, platform_height / 2)  # top surface in link-local frame
positions = np.stack([x, y, z], axis=-1)
normals = np.tile([0.0, 0.0, 1.0], (grid_n * grid_n, 1))
radii = PROBE_RADIUS + i * (PROBE_RADIUS / 10.0)
```

examples/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,226 @@
...
+    parser.add_argument("--cpu", action="store_true", help="Run on CPU instead of GPU")
+    parser.add_argument("--seconds", "-t", type=float, default=3.0, help="Seconds to simulate (headless mode)")
+    args = parser.parse_args()
+
+    gs.init(backend=gs.cpu if args.cpu else gs.gpu, precision="32", logging_level="info")
+
+    scene = gs.Scene(
+        sim_options=gs.options.SimOptions(
+            gravity=(0.0, 0.0, -9.81),
+        ),
```
duburcqa: Remove this block if useless.

examples/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,226 @@
...
+        )
+    )
+
+    pusher_start = np.array([0.0, 0.0, PLATFORM_HEIGHT + PUSHER_SIZE / 2 - 0.02], dtype=np.float32)
+
+    pusher = scene.add_entity(
+        gs.morphs.Cylinder(
+            radius=PUSHER_SIZE,
+            height=PUSHER_SIZE,
+            pos=tuple(pusher_start),
```
duburcqa: No need to cast to tuple.

genesis/engine/solvers/rigid/collider/utils.py
```diff
@@ -19,6 +21,19 @@ def func_det3(
...
 
 
+@ti.func
+def func_point_in_geom_aabb(
+    i_g: ti.i32,
+    i_b: ti.i32,
+    geoms_state: array_class.GeomsState,
+    point: ti.types.vector(3, ti.f32),
+    expansion: ti.f32 = 0.0,
+):
```
duburcqa: It would be nice to "match" the argument order of `func_is_geom_aabbs_overlap` right below.

genesis/options/sensors/options.py
```diff
@@ -181,6 +181,70 @@ def model_post_init(self, _):
             gs.raise_exception(f"resolution must be a float or tuple of 3 floats, got: {self.resolution}")
 
 
+class KinematicContactProbe(RigidSensorOptionsMixin, NoisySensorOptionsMixin, SensorOptions):
+    """
+    Kinematic contact probe for detecting contact without affecting physics simulation.
+
+    The returned "force" is non-physical estimate: F = stiffness * penetration * probe_normal.
```
duburcqa: Could you elaborate on the meaning of "non-physical estimate". For an uninitiated like me this is very disturbing.

genesis/options/sensors/options.py
```diff
@@ -181,6 +181,70 @@ def model_post_init(self, _):
...
+
+    Note
+    ----
+    If the probe is attached to a fixed entity, it will not detect contacts with other fixed entities.
+
```
duburcqa, line 193: Could you add some reference about what is "kinematic contact probe" and how it relates to tactile array? You had a new illustration on your PR / issue. It would be nice to reference it here maybe? Or point to the official documentation and add a section with a video and nice illustration in genesis-doc?

genesis/options/sensors/options.py
```diff
@@ -181,6 +181,70 @@ def model_post_init(self, _):
...
+            gs.raise_exception(f"radius must be non-negative, got: {self.radius}")
+        if self.stiffness < 0:
+            gs.raise_exception(f"stiffness must be non-negative, got: {self.stiffness}")
+
+        probe_local_pos = self._validate_probe_arrays(self.probe_local_pos)
+        probe_local_normal = self._validate_probe_arrays(self.probe_local_normal)
+        norms = np.linalg.norm(probe_local_normal, axis=1)
+        bad_idx = np.where(norms < gs.EPS)[0]
+        if bad_idx.size > 0:
+            idx = int(bad_idx[0])
```
duburcqa:
```python
is_invalid_mask = norms < gs.EPS
if np.any(is_invalid_mask):
    invalid_idx, _ = np.where(is_invalid_mask)
    gs.raise_exception(f"probe_local_normal[{invalid_idx}] must be non-zero, got: {probe_local_normal[is_invalid_mask]}")
```

genesis/engine/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,445 @@
...
+    stiffness: torch.Tensor = make_tensor_field((0,))
+
+    probe_sensor_idx: torch.Tensor = make_tensor_field((0,), dtype=torch.int32)
+    probe_positions: torch.Tensor = make_tensor_field((0, 3))
+    probe_normals: torch.Tensor = make_tensor_field((0, 3))
+    probe_max_raycast_range: float = 0.0
+
+    n_probes_per_sensor: torch.Tensor = make_tensor_field((0,), dtype=torch.int32)
+    sensor_cache_start: torch.Tensor = make_tensor_field((0,), dtype=torch.int32)
+    sensor_probe_start: torch.Tensor = make_tensor_field((0,), dtype=torch.int32)
```
duburcqa: `dtype=gs.tc_int`

genesis/engine/sensors/kinematic_contact_probe.py
```diff
@@ -0,0 +1,445 @@
...
+        # Store n_probes before super().__init__() since _get_return_format() is called there
+        self._n_probes = len(sensor_options.probe_local_pos)
+
+        super().__init__(sensor_options, sensor_idx, data_cls, sensor_manager)
+
+        self._debug_objects: list["Mesh | None"] = []
+        self._probe_local_pos = torch.tensor(self._options.probe_local_pos, dtype=gs.tc_float, device=gs.device)
+        self._probe_local_normal = torch.tensor(self._options.probe_local_normal, dtype=gs.tc_float, device=gs.device)
+        norms = self._probe_local_normal.norm(dim=1, keepdim=True).clamp(min=gs.EPS)
+        self._probe_local_normal /= norms
```
duburcqa: `self._probe_local_normal /= self._probe_local_normal.norm(dim=1, keepdim=True).clamp(min=gs.EPS)`

examples/sensors/kinematic_contact_probe.py
Milotrince: I do need to cast to tuple here since np arrays are not of type Sequence. we should add support for pydantic np.ndarray type

duburcqa: We have strict pydantic now? I see.

PR #2401, opened 2026-02-13
[BUG FIX] Fix handling of unspecified spatial inertia properties in USD.

### LINKED ISSUE #2397: [Bug]: USD Example crashes on first step due to NaN
### Bug Description

The basic USD example causes nans, at the first step after scene.build. Even decreasing the rigid dt as instructed still causes the issue
I have tried setting the as small as 0.0002

### Steps to Reproduce

run the `examples/usd/import_stage.usd` 
with the config
```
            dt=0.025,
            use_gjk_collision=True,
            constraint_solver=gs.constraint_solver.Newton,
            enable_self_collision=False,
        ),
```

### Expected Behavior

.

### Screenshots/Videos

_No response_

### Relevant log output

```shell
File "/home/vybhav/Genesis/genesis/engine/scene.py", line 986, in step
    self._sim.step()
  File "/home/vybhav/Genesis/genesis/engine/simulator.py", line 270, in step
    self.rigid_solver.check_errno()
  File "/home/vybhav/Genesis/genesis/engine/solvers/rigid/rigid_solver.py", line 1209, in check_errno
    gs.raise_exception("Invalid constraint forces causing 'nan'. Please decrease Rigid simulation timestep.")
  File "/home/vybhav/Genesis/genesis/utils/misc.py", line 43, in raise_exception
    raise gs.GenesisException(msg)
genesis.GenesisException: Invalid constraint forces causing 'nan'. Please decrease Rigid simulation timestep.

[Genesis] [13:33:36] [ERROR] GenesisException: Invalid constraint forces causing 'nan'. Please decrease Rigid simulation timestep.
```

### Environment

    OS: Ubuntu 22.04
    GPU/CPU RTX 3050 Mobile
    GPU-driver version NVIDIA-SMI 550.163.01
    CUDA / CUDA-toolkit 12.4
    omniverse-kit enabled
    -python 3.11 (venv)

latest main, major version: 3.14

### Additional Context

possibly an issue with the asset itself or spawn config, as i was able to spawn other USD Entities.
When i had the same issue, i was able to mitigate it by commenting the JointAnimator.animate, but it is no longer the case
@ACMLCZH

alanray-tech:
## Summary

This PR fixes #2397 where USD examples were crashing due to NaN errors caused by invalid default values in USD Physics MassAPI attributes.

## Changes

- **Added helper functions** in `genesis/utils/usd/usd_utils.py` to handle invalid MassAPI default values:
  - `usd_center_of_mass_to_numpy()` - detects invalid default `(-inf, -inf, -inf)` and returns `None` to trigger recomputation
  - `usd_principal_axes_to_numpy()` - detects invalid default `(0, 0, 0, 0)` quaternion and returns `None` to trigger recomputation
  - `usd_diagonal_inertia_to_numpy()` - handles default `(0, 0, 0)` (ignored value) and returns `None` to trigger recomputation
  - `usd_mass_to_float()` - handles default `0` (ignored value) and returns `None` to trigger recomputation

- **Updated `genesis/utils/usd/usd_rigid_entity.py`** to use the new helper functions and ensure `inertial_i` is always a valid numpy array (diagonal matrix) to avoid AttributeError in downstream code.

- **Added test** `test_massapi_invalid_defaults_mjcf_vs_usd` to verify that USD MassAPI with invalid defaults produces equivalent results to MJCF when mass properties are computed from geometry.

- **Added documentation** `.github/contributing/USD_PARSER.md` with guidelines for developing and fixing the USD parser.

## Problem

USD Physics MassAPI defines some attributes with invalid default values:
- `centerOfMass`: default `(-inf, -inf, -inf)` - invalid, indicates should be computed from geometry
- `principalAxes`: default `(0, 0, 0, 0)` - invalid quaternion (identity should be `(1, 0, 0, 0)`)
- `diagonalInertia`: default `(0, 0, 0)` - valid but means ignored/computed from geometry
- `mass`: default `0` - valid but means ignored/computed from geometry

When these invalid defaults were parsed, they caused NaN errors in the simulation because the physics engine tried to use invalid mass properties.

## Solution

The fix detects these invalid default values and returns `None` from the helper functions, allowing the downstream physics engine to recompute mass properties from geometry. This matches the intended behavior of USD Physics MassAPI.

## Testing

Run the USD tests to verify the fix:

```bash
uv run pytest tests/test_usd.py -v
```

The new test `test_massapi_invalid_defaults_mjcf_vs_usd` verifies that:
- USD files with invalid MassAPI defaults are parsed correctly
- Mass properties are recomputed from geometry
- Results match equivalent MJCF files

## Related Issues

Fixes #2397

### DIFF
--- a/genesis/utils/usd/usd_rigid_entity.py
+++ b/genesis/utils/usd/usd_rigid_entity.py
@@ -14,6 +14,7 @@ from .usd_utils import (
     get_attr_value_by_candidates,
     usd_center_of_mass_to_numpy,
     usd_diagonal_inertia_to_numpy,
+    usd_inertia_to_numpy,
     usd_mass_to_float,
     usd_pos_to_numpy,
     usd_principal_axes_to_numpy,
@@ -45,9 +46,7 @@ def _parse_joint_axis_pos(
     context: UsdContext, joint: UsdPhysics.Joint, child_link: Usd.Prim, is_body1: bool
 ) -> Tuple[str, np.ndarray, np.ndarray]:
     joint_pos_attr = joint.GetLocalPos1Attr() if is_body1 else joint.GetLocalPos0Attr()
-    joint_pos = usd_pos_to_numpy(joint_pos_attr.Get() if joint_pos_attr.HasValue() else None)
-    if joint_pos is None:
-        joint_pos = gu.zero_pos()
+    joint_pos = usd_pos_to_numpy(joint_pos_attr.Get()) if joint_pos_attr.HasValue() else gu.zero_pos()
     T = context.compute_transform(child_link)
     joint_pos = gu.transform_by_T(joint_pos, T)
     Q, S = context.compute_gs_transform(child_link)
@@ -103,13 +102,7 @@ def _parse_link(
         principal_axes_attr = mass_api.GetPrincipalAxesAttr()
         l_info["inertial_quat"] = usd_principal_axes_to_numpy(principal_axes_attr.Get())
         inertia_attr = mass_api.GetDiagonalInertiaAttr()
-        diagonal_inertia = usd_diagonal_inertia_to_numpy(inertia_attr.Get())
-        # Always set inertial_i as a diagonal matrix to avoid None values
-        # If invalid, use zeros which will trigger recomputation in rigid_entity.py
-        if diagonal_inertia is not None:
-            l_info["inertial_i"] = np.diag(diagonal_inertia)
-        else:
-            l_info["inertial_i"] = np.diag(gu.zero_pos())
+        l_info["inertial_i"] = usd_inertia_to_numpy(inertia_attr.Get())
         mass_attr = mass_api.GetMassAttr()
         l_info["inertial_mass"] = usd_mass_to_float(mass_attr.Get())
 
--- a/genesis/utils/usd/usd_utils.py
+++ b/genesis/utils/usd/usd_utils.py
@@ -116,6 +116,30 @@ def usd_principal_axes_to_numpy(usd_quat: Gf.Quatf) -> np.ndarray | None:
     return quat
 
 
+def usd_inertia_to_numpy(inertia: Gf.Vec3f) -> np.ndarray | None:
+    """
+    Convert USD diagonal inertia to numpy diagonal matrix, handling default and invalid values.
+
+    The USD Physics MassAPI defines diagonalInertia with default value (0, 0, 0),
+    which is valid but means the inertia should be ignored/computed from geometry.
+    This function returns None for the default ignored value or invalid values (non-finite).
+
+    Parameters
+    ----------
+    inertia : Gf.Vec3f
+        USD diagonal inertia attribute value.
+
+    Returns
+    -------
+    np.ndarray | None
+        Valid diagonal inertia matrix (3x3), or None if default ignored or invalid.
+    """
+    diagonal_inertia = usd_diagonal_inertia_to_numpy(inertia)
+    if diagonal_inertia is None:
+        return None
+    return np.diag(diagonal_inertia)
+
+
 def usd_diagonal_inertia_to_numpy(usd_pos: Gf.Vec3f) -> np.ndarray | None:
     """
     Convert USD diagonal inertia to numpy array, handling default and invalid values.
@@ -138,8 +162,8 @@ def usd_diagonal_inertia_to_numpy(usd_pos: Gf.Vec3f) -> np.ndarray | None:
     # Default is (0, 0, 0) which means ignored - only return if non-zero and valid
     if np.allclose(inertia, 0):
         return None
-    # Check for invalid values (negative or inf/nan)
-    if not np.all(inertia >= 0):
+    # Check for invalid values (non-finite)
+    if not all(math.isfinite(e) for e in inertia):
         return None
     return inertia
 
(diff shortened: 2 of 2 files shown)


genesis/utils/usd/usd_rigid_entity.py
```diff
@@ -40,7 +44,10 @@
 def _parse_joint_axis_pos(
     context: UsdContext, joint: UsdPhysics.Joint, child_link: Usd.Prim, is_body1: bool
 ) -> Tuple[str, np.ndarray, np.ndarray]:
-    joint_pos = usd_pos_to_numpy((joint.GetLocalPos1Attr() if is_body1 else joint.GetLocalPos0Attr()).Get())
+    joint_pos_attr = joint.GetLocalPos1Attr() if is_body1 else joint.GetLocalPos0Attr()
+    joint_pos = usd_pos_to_numpy(joint_pos_attr.Get() if joint_pos_attr.HasValue() else None)
+    if joint_pos is None:
+        joint_pos = gu.zero_pos()
```
duburcqa: Just do:
```python
joint_pos = usd_pos_to_numpy(joint_pos_attr.Get()) if joint_pos_attr.HasValue() else gu.zero_pos()
```

genesis/utils/usd/usd_rigid_entity.py
```diff
@@ -90,10 +97,21 @@ def _parse_link(
...
-        l_info["inertial_quat"] = usd_quat_to_numpy(mass_api.GetPrincipalAxesAttr().Get())
-        l_info["inertial_i"] = np.diag(usd_pos_to_numpy(mass_api.GetDiagonalInertiaAttr().Get()))
-        l_info["inertial_mass"] = float(mass_api.GetMassAttr().Get() or 0.0)
+
+        com_attr = mass_api.GetCenterOfMassAttr()
+        l_info["inertial_pos"] = usd_center_of_mass_to_numpy(com_attr.Get())
+        principal_axes_attr = mass_api.GetPrincipalAxesAttr()
+        l_info["inertial_quat"] = usd_principal_axes_to_numpy(principal_axes_attr.Get())
+        inertia_attr = mass_api.GetDiagonalInertiaAttr()
+        diagonal_inertia = usd_diagonal_inertia_to_numpy(inertia_attr.Get())
```
duburcqa: Just implement `usd_inertia_to_numpy(inertia_attr.Get())` that calls `np.diag` internally.

genesis/utils/usd/usd_rigid_entity.py
```diff
@@ -90,10 +97,21 @@ def _parse_link(
...
+        # Always set inertial_i as a diagonal matrix to avoid None values
+        # If invalid, use zeros which will trigger recomputation in rigid_entity.py
```
duburcqa: Just use None. It is better.

genesis/utils/usd/usd_rigid_entity.py
```diff
@@ -90,10 +97,21 @@ def _parse_link(
...
+        if diagonal_inertia is not None:
+            l_info["inertial_i"] = np.diag(diagonal_inertia)
+        else:
+            l_info["inertial_i"] = np.diag(gu.zero_pos())
```
duburcqa: `l_info["inertial_i"] = usd_inertia_to_numpy(inertia_attr.Get())`

genesis/utils/usd/usd_utils.py
```diff
@@ -25,18 +26,151 @@
...
+    np.ndarray | None
+        Valid diagonal inertia as numpy array, or None if default ignored or invalid.
+    """
+    inertia = usd_pos_to_numpy(usd_pos)
+    # Default is (0, 0, 0) which means ignored - only return if non-zero and valid
+    if np.allclose(inertia, 0):
+        return None
+    # Check for invalid values (negative or inf/nan)
+    if not np.all(inertia >= 0):
+        return None
```
duburcqa: I would rather check `all(math.isfinite(e) for e in inertia)`. Because we do not want to hind random negative values.

genesis/utils/usd/usd_utils.py
```diff
@@ -25,18 +26,151 @@
...
+    Returns
+    -------
+    float | None
+        Valid mass value, or None if default ignored or invalid.
+    """
+    # Default is 0 which means ignored
+    if usd_mass <= 0:
+        return None
+    # Check for invalid values (inf or nan)
+    if np.isinf(usd_mass) or np.isnan(usd_mass):
```
duburcqa: `if math.isfinite(usd_mass)`

genesis/utils/usd/usd_utils.py
```diff
@@ -25,18 +26,151 @@
...
+        USD mass attribute value.
+
+    Returns
+    -------
+    float | None
+        Valid mass value, or None if default ignored or invalid.
+    """
+    # Default is 0 which means ignored
+    if usd_mass <= 0:
+        return None
```
duburcqa, line 191: Do not do this. It hides user-specified invalid values.

genesis/utils/usd/usd_rigid_entity.py
alanray-tech: https://github.com/Genesis-Embodied-AI/Genesis/blob/bc408865d0dbb7d2a6fa2b768e3ed7f66d17ff8c/genesis/engine/entities/rigid_entity/rigid_entity.py#L729

A mistake in rigid_entity.py.

When inertial_i is None:
l_info.get("inertial_i") is None → True
Due to short-circuit, np.diag(l_info["inertial_i"]) <= 0.0 is not evaluated
The or expression evaluates to True (boolean)
.any() is called on True → AttributeError
The parentheses group the or expression, so .any() is applied to the boolean result, not the numpy array.

I think it can be fixed in this PR?

duburcqa: Nice spot. PR to fix this in 5min!

duburcqa: Merged! just need to rebase and it should be fine.

PR #2399, opened 2026-02-13
[MISC] Migrate from GsTaichi to Quadrants (part 1).

hughperkins:
## Description

Note: methodology:
1. run https://github.com/Genesis-Embodied-AI/perso_hugh/blob/main/bin/rename_files_for_quadrants.py
2. commit everything except RELEASE.md, which I feel like we probably dont want to rename things in :)

Follow-up PR: https://github.com/Genesis-Embodied-AI/Genesis/pull/2409

PR #2406, opened 2026-02-14
[MISC] Decompose solver and use parallel Hessian and Cholesky to significantly speedup contact-rich and dynamic simulations.

hughperkins:


genesis/engine/solvers/rigid/constraint/solver_breakdown.py
duburcqa, line 1: Now that 'decomp' keyword is available again, could we switch back to this one?

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -1,5 +1,7 @@
 from typing import TYPE_CHECKING
+import os
```
duburcqa: Wrong order.

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -1,5 +1,7 @@
...
 
+from frozendict import frozendict
```
duburcqa: wrong order.

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -20,6 +23,8 @@
 
 IS_OLD_TORCH = tuple(map(int, torch.__version__.split(".")[:2])) < (2, 8)
 
+GS_SOLVER_DECOMPOSE = os.environ.get("GS_SOLVER_DECOMPOSE", "1") == "1"
+
```
duburcqa: Why is this optional? I'm not a fan of this kind of hidden magic flag. Only add them if necessary.

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
```diff
@@ -0,0 +1,191 @@
...
+            solver.func_terminate_or_update_descent_batch(
+                i_b,
+                rigid_global_info=rigid_global_info,
+                constraint_state=constraint_state,
+                static_rigid_sim_config=static_rigid_sim_config,
+            )
+
+
+def register_decomposed_solver_body() -> None:
+    print("register_decomposed_solver_body", "gs.backend", gs.backend)
```
duburcqa: Same. Remove this registration function if it can be avoid.

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
```diff
@@ -0,0 +1,191 @@
...
+
+    @solver.func_solve_body.register(is_compatible=lambda *args, **kwargs: gs.backend in {gs.cuda})
+    def func_solve_decomposed_macrokernels(
```
duburcqa: Why '_macrokernels' suffix? This function is local to this file, just call it `func_solve_decomposed`.

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
duburcqa: Remove print. Using logger at debug level.

tests/conftest.py
```diff
@@ -635,7 +635,7 @@ def initialize_genesis(request, monkeypatch, tmp_path, backend, precision, perfo
         yield
     finally:
         gs.destroy()
-        # Double garbage collection is over-zealous since Quadrants 2.2.1 but let's do it anyway
+        # Double garbage collection is over-zealous since gstaichi 2.2.1 but let's do it anyway
```
duburcqa, line 653: revert;

pyproject.toml
```diff
@@ -10,9 +10,10 @@ readme = "README.md"
 requires-python = ">=3.10,<3.14"
 dependencies = [
     "psutil",
-    "quadrants==0.4.0",
+    "quadrants==0.4.1b4",
```
duburcqa: We need a new release.

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
hughperkins: It's not available again. Because we have backwards compatiblity in place for decomp.

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -2938,13 +2998,30 @@ def func_solve_iter(
...
+) -> None: ...
+
+
+if GS_SOLVER_DECOMPOSE:
+    solver_breakdown.register_decomposed_solver_body()
+
+
+@func_solve_body.register(is_compatible=lambda *args, **kwargs: True)
+@qd.kernel(fastcache=gs.use_fastcache)
+def func_solve_body_monolith(
```
duburcqa, line 3012: `_monolith` is not clear to me. What are you referring to with this keyword?

genesis/engine/solvers/rigid/constraint/solver.py
hughperkins: I needed it in order to compare performance with/without decomp on Eden. I dont remember why I couldnt simply use `main` to compare. But anyway, in the future, we won't be able to use `main`, to compare, because main will contain the perf dispatch.

But I confess to not having used this flag for a few days. I can remove it if we want. I can always put it back in a dev branch, if I need it.

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
hughperkins: dont think it can be avoided. end up with weird issues with import sequencing. this is the cleanest, easiest approach I feel.

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
hughperkins: to answer your question, it's because we had 'microkernel' vs 'macrokernel' for a while.

but sure, can change it now that macro won.

tests/conftest.py
hughperkins: Quadrants 2.2.1 doesnt exist. might never exist.

hughperkins: you yourself asked for this change, https://github.com/Genesis-Embodied-AI/Genesis/pull/2399#pullrequestreview-3805150423

pyproject.toml
hughperkins: yeah 👍

genesis/engine/solvers/rigid/constraint/solver.py
hughperkins: monolithic kernel vs decomposed kernels.

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
duburcqa: I see. Ok let's keep this then.

genesis/engine/solvers/rigid/constraint/solver.py
hughperkins: Screenshot 2026-02-16 at 12 55 01

genesis/engine/solvers/rigid/constraint/solver.py
duburcqa: > But I confess to not having used this flag for a few days. I can remove it if we want. I can always put it back in a dev branch, if I need it.

Yes I would like please.

genesis/engine/solvers/rigid/constraint/solver.py
duburcqa: What is the difference with 'macrokernel' that you are using somewhere else?

duburcqa: (To clarify, it is not a big deal.)

hughperkins: macrokernel and monolith are the two optinos:
- monolith: one single kernel, that does everything
- macrokernel: one kernel per fucntion, eg linesearch, hessian, etc

hughperkins: (macrokernel vs microkernel:
- microkernel: one single quadrants kernel, containing multiple gpu kernels
- macrokernel: one gpu kernel per function, eg linesearch, hessian etc

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
duburcqa: I don't like this pattern. I can help you to fix the import sequencing if you want.

genesis/engine/solvers/rigid/constraint/solver.py
hughperkins: monlith:
```
@ti.func
def linesearch

@ti.kernel
def solver:
    for b in range(B):
         linesearch(b)
         hessian(b)
          ....
```

macrokernels:
```
@ti.kernel
def linesearch:
    for b in range(B):


def solver:
    linesearch()
    hessian()
          ....
```

microkernels:
```
@ti.func
def linesearch:
    for b in range(B):

@ti.kernel()
def solver:
    linesearch()
    hessian()
          ....
```

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
hughperkins: I've actually been thinking that using patterns like this will make things cleaner, and eventually allow us to add annotation types in geom.py for example.

But sure, we can debate that another time, and maybe in the context of geom.py.

Yes, if you can provide a branch that handles this in a way you are most comfortalbe with, that sounds useful 🙌

genesis/engine/solvers/rigid/constraint/solver.py
duburcqa: Wait, macrokernel is not a kernel?

tests/conftest.py
duburcqa: You are right... Sorry about that.

genesis/engine/solvers/rigid/constraint/solver.py
hughperkins: macrokernels is a function that calls multiple kernels.

hughperkins: This is why I had to add in the python calling support into perf dispatch, taht you were asking for.

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
duburcqa: Here it is:
```diff
diff --git a/genesis/engine/solvers/rigid/constraint/__init__.py b/genesis/engine/solvers/rigid/constraint/__init__.py
index 70afdb95..cf1b56a9 100644
--- a/genesis/engine/solvers/rigid/constraint/__init__.py
+++ b/genesis/engine/solvers/rigid/constraint/__init__.py
@@ -5,4 +5,5 @@ Contains constraint solving, island detection, and backward pass.
 """

 from .solver import ConstraintSolver
+from .solver_breakdown import *
 from .solver_island import ConstraintSolverIsland
diff --git a/genesis/engine/solvers/rigid/constraint/solver.py b/genesis/engine/solvers/rigid/constraint/solver.py
index e3a3d864..f02d9269 100644
--- a/genesis/engine/solvers/rigid/constraint/solver.py
+++ b/genesis/engine/solvers/rigid/constraint/solver.py
@@ -13,7 +13,6 @@ from genesis.engine.solvers.rigid.abd import func_solve_mass_batch
 from genesis.utils.misc import qd_to_torch

 from . import backward as backward_constraint_solver
-from . import solver_breakdown
 from . import noslip as constraint_noslip
 from ..collider.contact_island import ContactIsland

@@ -3010,10 +3009,6 @@ def func_solve_body(
 ) -> None: ...


-if GS_SOLVER_DECOMPOSE:
-    solver_breakdown.register_decomposed_solver_body()
-
-
 @func_solve_body.register(is_compatible=lambda *args, **kwargs: True)
 @qd.kernel(fastcache=gs.use_fastcache)
 def func_solve_body_monolith(
diff --git a/genesis/engine/solvers/rigid/constraint/solver_breakdown.py b/genesis/engine/solvers/rigid/constraint/solver_breakdown.py
index ea25adc7..81a82517 100644
--- a/genesis/engine/solvers/rigid/constraint/solver_breakdown.py
+++ b/genesis/engine/solvers/rigid/constraint/solver_breakdown.py
@@ -189,3 +189,6 @@ def register_decomposed_solver_body() -> None:
                 rigid_global_info,
                 static_rigid_sim_config,
             )
+
+
+register_decomposed_solver_body()
```

duburcqa: Note that I'm calling 'register_decomposed_solver_body' right after defining it to reduce the size of the diff and focus on import sequencing.

genesis/engine/solvers/rigid/constraint/solver.py
duburcqa: Sure. I'm just confused by the terminology. I thought a macrokernel was a kernel in the first place.

hughperkins: Fair enough :)

Yeah, macrokernels refers to the concept, rather than to the top level function. At least, that's how I'm thinking about it. The top level function is just a python function, which calls the macro-kernels, is how I think about it.

In fact, in alignment iwth this, a macrokernel *is* a kernel. Just the top level python funciton that calls the macrokernels is not.

hughperkins: monolith: a python kernel, that does everything. and a 1:1 mapping between the quadrants monolithic kernel and the single gpu kernel inside it.

microkernels: a singl quadrants kernel calls multiple gpu microkernels

macrokernels: a single python funciton calls multiple quadrants kernels, which each contain a single gpu kernel. (but the decomposition is over the macrokernels, not over the kernels inside them).

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
hughperkins: old debug statement. this is already handled on quadrants side. => removing.

hughperkins: `from .solver_breakdown import *` is a code smell I feel?

hughperkins: I'm not sure what you are intending to achieve here. I thought your intended goal was to remove the registration function?

hughperkins:
Addressed, or responded to, all comments, I think?

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
duburcqa: > from .solver_breakdown import * is a code smell I feel?

Absolutely not. Why would it be? You can replace it by `from .solver_breakdown import func_solve_body_monolith` if you prefer. None of this matters.

> I'm not sure what you are intending to achieve here. I thought your intended goal was to remove the registration function?

I did. As a said "Note that I'm calling 'register_decomposed_solver_body' right after defining it to reduce the size of the diff and focus on import sequencing.". Just remove the function yourself at this point there is no blocker after applying my patch.

hughperkins: > Absolutely not. Why would it be? You can replace it by from .solver_breakdown import func_solve_body_monolith if you prefer. None of this matters.

Why would we import func_solve_body_monolith? We arent calling it from solver.py. We arent even calling func_solve_body_deocmposed. We are only calling func_solve_body, which is inside solver.py:

Screenshot 2026-02-18 at 14 43 17

hughperkins: What we need to happen is that the register call happens after we have declared func_solve_body.

We could hack around with import ordering to do that. But I think that is not sustaibnale or easy to maintain.

I feel the clean solution is to call 'register_decomposed', from solver, after declaring func_solve_body.

duburcqa: I'm not ok.

duburcqa: In may not be sustainable but it is not a problem for today.

duburcqa: > Why would we import func_solve_body_monolith? 

Then just do * as a suggest. Same result.

hughperkins: Same question. Why would we import *anyting* at all from _breakdown? We arent calling anything in breakdown from solver.py.

hughperkins: "In may not be sustainable but it is not a problem for today." => what problem are you trying to solve today by not having a registration function?

duburcqa: > what problem are you trying to solve today by not having a registration function?

We don't do this anywhere. I don't want to start it today. It has side effects I don't want to worry about for now. There are other priorities than this.

hughperkins: removed registration in 07292b19

duburcqa (approved):

PR #2407, opened 2026-02-14
[MISC] Fix benchmarks.

duburcqa:


.github/workflows/production.yml
```diff
@@ -87,6 +87,7 @@ jobs:
         if: github.event_name == 'pull_request'
         run: |
           srun ${SRUN_COMMON} bash -s <<'EOF'
+            set -e
```
hughperkins, line 90: oh whoops, the silent failures was my fault.

hughperkins:
The changes to production.yml and to g1_fall look good to me.

PR #2419, opened 2026-02-17
[BUG FIX] Tune GJK to avoid false negative collision detection.

SonSang:
## Description

In GJK collision detection, we compute (relative) reprojection error before finalizing the witness points in the collision. If this error is too large, we cannot rely on the estimated witness points. However, even though this value can usually reach upto `5e-4`, we have been using too tight value `1e-5`. Therefore, the algorithm ignored collisions in several test cases even though the estimated collisions were precise. To fix this issue, increase the error threshold to `1e-4`.

## Checklist:

### DIFF
--- a/tests/test_rigid_physics_analytical_vs_gjk.py
+++ b/tests/test_rigid_physics_analytical_vs_gjk.py
@@ -501,15 +501,15 @@ def test_sphere_capsule_vs_gjk(backend, monkeypatch, tmp_path: Path, show_viewer
     3. Run ALL GJK scenarios (patched kernel with its own empty cache)
     """
     test_cases = [
-        # (sphere_pos, capsule_pos, capsule_euler, should_collide, description, skip_gpu)
-        ((0, 0, 0.4), (0, 0, 0), (0, 0, 0), True, "sphere_above_capsule_top", False),
-        ((0.18, 0, 0), (0, 0, 0), (0, 0, 0), True, "sphere_close_to_capsule", False),
-        ((0.17, 0.17, 0), (0, 0, 0), (0, 0, 0), False, "sphere_near_cylinder", False),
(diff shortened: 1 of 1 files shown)


genesis/engine/solvers/rigid/collider/gjk.py
```diff
@@ -93,9 +93,9 @@ def __init__(self, rigid_solver):
             simplex_max_degeneracy_sq=1e-5**2,
             polytope_max_faces=polytope_max_faces,
             # This value has been experimentally determined based on the examples that we currently have (e.g. pyramid,
-            # tower, ...). We observed the error usually reaches around 5e-4, so we set the threshold to 1e-5 to be
+            # tower, ...). We observed the error usually reaches around 5e-4, so we set the threshold to 1e-4 to be
```
hughperkins, line 96: if the error is up to 5e-4,shouldnt the threshold be *larger* than this, eg 1e-3?

tests/test_rigid_physics_analytical_vs_gjk.py
```diff
@@ -508,7 +508,7 @@ def test_sphere_capsule_vs_gjk(backend, monkeypatch, tmp_path: Path, show_viewer
         ((0.35, 0, 0.35), (0, 0, 0), (0, 45, 0), False, "sphere_near_cap", False),
         ((0.15, 0, 0), (0, 0, 0), (0, 0, 0), True, "sphere_touching_cylinder", False),
         ((0, 0, 0), (0, 0, 0), (0, 0, 0), True, "sphere_at_capsule_center", False),
-        ((0.15, 0, 0.3), (0, 0, 0), (0, 0, 0), True, "sphere_near_capsule_cap", True),
+        ((0.15, 0, 0.3), (0, 0, 0), (0, 0, 0), True, "sphere_near_capsule_cap", False),
```
hughperkins: 🙌

PR #2421, opened 2026-02-17
[MISC] Add torch profile facilities to rigid benchmarks.

hughperkins:


tests/conftest.py
```diff
@@ -416,13 +416,68 @@ def pytest_addoption(parser):
         else SUPPRESS
     )
     parser.addoption("--mem-monitoring-filepath", type=str, help=help_text)
+    parser.addoption("--profile", action="store_true", default=False, help="Enable PyTorch profiling for benchmarks.")
+    parser.addoption("--profile-wait", type=int, default=5, help="Number of steps to skip before profiling.")
+    parser.addoption("--profile-warmup", type=int, default=1, help="Number of warmup steps for profiling.")
+    parser.addoption("--profile-active", type=int, default=3, help="Number of active profiling steps.")
```
duburcqa: Are you sure it is really necessary to expose this many options?
I have the feeling just adding `--profile-num-steps` that would be equal to 0 by default should be good enough?

duburcqa:
I'm OK with this PR. No major blocker for merging this.

tests/conftest.py
hughperkins: These tests are not stable over time. I generally need to wait a few seconds before sampling, else not representative.

I'm also pondering I might need to add `repeat` tbh. Maybe I just add `repeat` now, and then it's done?

hughperkins: (happy to factorize out these options to a separate fucntion in a separae file. like `profiling.add_profiling_options(parser)` )

hughperkins: (could also factorize out the body of the other function being added here, into that same file. I think that will keep conftest.py cleanre, and make both of us happier?)

duburcqa: > These tests are not stable over time. I generally need to wait a few seconds before sampling, else not representative.

Yes, I agree. I'm not saying you should not wait a few seconds before sampling, I'm wondering whether it is relevant to let the user specify this or their is a good default that will do the job in all cases.

duburcqa: > (could also factorize out the body of the other function being added here, into that same file. I think that will keep conftest.py cleanre, and make both of us happier?)

This is not my concern. My concern is avoiding exposing tweakable that may be unnecessary because there is a single good value or the default is always fine. I don't know if it is the case here, that's why I'm asking.

hughperkins: There isnt really, because each test runs at a different steps per second. And there isnt an option to profile after eg 5 seconds.

hughperkins: If your concern is that this means there are too many command-line options, one option could be only to add the profiling options if we provide an env var, eg GS_PROFILING=1.

duburcqa: Not too many options. Adding options that have a single correct value.

> There isnt really, because each test runs at a different steps per second. 

If each test must use some specific value, I think this value should be hard-coded in the test run itself then.

hughperkins: That would make this pr expensive to merge, and to maintain, and sprinkle dev stuff throughout hte benchmarks. I think that if anyone is running the profiler, t hey have fairly detailed knowledge on which step they want to profile.

Also note that the distribution is NOT stable. So in reality, we will likely want to profile multiple steps (which is why we want `repeat` available).

hughperkins: perhaps a compromise is simply to make the step optoin mandatory/requried, without a default value, so it's clear that the user needs to provide an appropriate value to it?

duburcqa: Let's go for the compromise.

hughperkins: Cool.

Also, I'm not sure if you noticed that I added an env var GS_PROFILING or not? Nor what you feel about this?

If we are going to use that env var to enable profiling, so I've removed `--profile`, since using the env var already implies this I feel.

This also allows me to directly make `--profile-wait` a mandatory argparse vairalb, (rather than have to manually enforce this, using a conditional and throw exception)

tests/conftest.py
```diff
@@ -416,13 +418,25 @@ def pytest_addoption(parser):
...
 
 
+@pytest.fixture(scope="session")
+def pytorch_profiler(pytestconfig):
+    if os.environ.get("GS_PROFILING", "0") == "1":
+        for res in profiling.pytorch_profiler(pytestconfig):
+            yield res
+    else:
+        noop = lambda: None  # noqa: E731
+        yield None, noop
```
duburcqa: Use nullcontext; See `examples/speed_benchmark/timers.py`

tests/profiling.py
```diff
@@ -0,0 +1,71 @@
...
+    print(f"PyTorch profiling enabled (wait={wait}, warmup={warmup}, active={active})")
+    prof.__enter__()
+
+    yield prof, prof.step
+
+    prof.__exit__(None, None, None)
+
+    trace_path = "profile_trace.json"
+    prof.export_chrome_trace(str(trace_path))
+    print(f"Chrome trace exported to: {trace_path}")
```
duburcqa, line 67: I don't get it. What not just use the context as it is designed to be used instead of calling __enter__ and __exit__ manually?

tests/test_rigid_benchmarks.py
```diff
@@ -313,7 +313,8 @@ def write(self, items):
 
 
 @pytest.fixture
-def go2(solver, n_envs, gjk):
+def go2(solver, n_envs, gjk, pytorch_profiler):
+    _, profiler_step = pytorch_profiler
```
duburcqa: Why not to yield only 'profiler_step' since the other argument is used? 
I would recommend doing this and extending the API later if necessary.

tests/conftest.py
duburcqa: Why not to move this fixture in profiling since you already created a dedicated file? It would not work?

tests/conftest.py
duburcqa: I don't have a strong opinion. We can keep this as a backdoor for now. It more or less makes sense.

tests/conftest.py
hughperkins: I dont know. I assumed all fixtures have to be in conftest.py 🤔

hughperkins: This isnt a context manager though. It's a function that does nothing. 🤔

hughperkins: (res is the `profiler.step` function)

hughperkins: Doesnt seem to work? :

Screenshot 2026-02-19 at 16 42 32

duburcqa: > I dont know. I assumed all fixtures have to be in conftest.py

That is not the case for sure. Look at `tests/test_bvh.py`.

tests/test_rigid_benchmarks.py
```diff
@@ -313,7 +313,8 @@ def write(self, items):
 
 
 @pytest.fixture
-def go2(solver, n_envs, gjk):
+def go2(solver, n_envs, gjk, pytorch_profiler):
+    profiler_step = pytorch_profiler
```
duburcqa: I hate just renaming variables. Just call your fixture `pytorch_profiler_step`, or `profiler_calling` or something.

tests/test_rigid_benchmarks.py
```diff
@@ -563,26 +570,38 @@ def _franka(solver, n_envs, gjk, is_collision_free, is_randomized, accessors):
...
+    )
 
 
 @pytest.fixture
-def franka_random(solver, n_envs, gjk):
-    return _franka(solver, n_envs, gjk, is_collision_free=False, is_randomized=True, accessors=False)
+def franka_random(solver, n_envs, gjk, pytorch_profiler):
+    profiler_step = pytorch_profiler
+    return _franka(
+        solver, n_envs, gjk, is_collision_free=False, is_randomized=True, accessors=False, profiler_step=profiler_step
```
duburcqa: This is even worse.

tests/conftest.py
duburcqa: If it is not possible to move it in profiler for some reason, then just add a comment to state this and we are good.

tests/test_rigid_benchmarks.py
hughperkins: its a relic from changing from `_, profiler_step =` to just `profiler_step = `

hughperkins:
addressed comments I think?

hughperkins:
Thanks! 🙌

PR #2427, opened 2026-02-18
[FEATURE] Add IPC Robot-Cloth Coupling.

Roushelfy:


genesis/options/solvers.py
```diff
@@ -347,7 +357,10 @@ class IPCCouplerOptions(BaseCouplerOptions):
     disable_genesis_contact: bool = True
     disable_ipc_ground_contact: bool = False
```
ACMLCZH, line 358: The name and function of "disable_genesis_contact" is somehow confusing. Should it be used for disable extra ABD-ABD collision in RigidSolver? However it disables the entire RigidSolver collision,  affecting other rigid entities not in IPC. I think the logic is somehow similar to "disabling FEM Solver's pre-coupling step", but they are inconsistent in implementation.

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -887,15 +1209,30 @@ def _add_rigid_geoms_to_ipc(self):
...
+
                         external_kinetic_attr = merged_mesh.instances().find(builtin.external_kinetic)
                         if external_kinetic_attr is not None:
                             external_kinetic_view = view(external_kinetic_attr)
-                            external_kinetic_view[:] = 1
+                            # Don't set external_kinetic for IPC-driven free base
+                            if not is_free_base_ipc_driven:
+                                external_kinetic_view[:] = 1
+                            else:
+                                external_kinetic_view[:] = 0
```
ACMLCZH, line 1232: Does it imply `ipc_only` rigid entities are all `external_kinetic = 1`? Will it force `ipc_only` Entities must be `fixed=True` otherwise it will cause error in IPC?

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -887,15 +1209,30 @@ def _add_rigid_geoms_to_ipc(self):
                             mass_density=entity_rho,
                         )
 
-                        # Set external_kinetic=1 for all ABD objects
+                        # Set external_kinetic attribute:
+                        # - For free_base_driven_by_ipc=True: keep external_kinetic=0 for non-fixed base link
+                        # - Otherwise: set external_kinetic=1 for all ABD objects
```
ACMLCZH, line 1214: Just a question: For entities with `external_kinetic=0` (and no STC), IPC doesn't seem to read the current-step Genesis state (positions/velocities). Does that mean the corresponding ABD states inside IPC won't be updated to reflect post-coupling changes on the Genesis side? If so, could this lead to state inconsistency between Genesis and IPC?

genesis/options/solvers.py
ACMLCZH, line 308: `gravity` is defined but never used?

PR #2436, opened 2026-02-19
[MISC] Propagate interactive viewer exceptions when running in thread.

hughperkins:


genesis/ext/pyrender/viewer.py
```diff
@@ -1244,6 +1248,9 @@ def run(self):
             except AttributeError:
                 # The graphical window has been closed manually
                 pass
+            except Exception as e:
```
duburcqa, line 1251: Remove this. It is not applicable here.

genesis/vis/visualizer.py
```diff
@@ -202,7 +202,11 @@ def update(self, force=True, auto=None):
             if self._viewer.is_alive():
                 self._viewer.update(auto_refresh=auto, force=force)
             else:
-                gs.raise_exception("Viewer closed.")
+                pyrender_viewer = self._viewer._pyrender_viewer
```
duburcqa, line 205: This is not the right place to do this. I would suggest to “refactor” this part to move the responsibility of handling exception in viewer. This means not checking if viewer is alive anymore in this function.

Issue #230, opened 2024-12-22
Is real-time visualization possible in Google Colab? Attempted frame-by-frame image conversion as an alternative


ardkyer:
# Issue Description
When running Genesis in Google Colab environment, real-time visualization is not possible due to OpenGL restrictions. Using show_viewer=True option results in an error.

# Alternative Approach
I implemented visualization by saving images frame by frame. Here's a minimal working example:

```
import genesis as gs
import numpy as np
from PIL import Image
import os

gs.init(backend=gs.cuda)

# Create a directory to save the results
save_dir = "/content/simulation_frames"
os.makedirs(save_dir, exist_ok=True)

scene = gs.Scene(show_viewer=False)

# Add objects
plane = scene.add_entity(gs.morphs.Plane())
box = scene.add_entity(
    gs.morphs.Box(
        pos=(0, 0, 2),
        size=(0.5, 0.5, 0.5)
    )
)

# Camera settings
cam = scene.add_camera(
    res=(320, 240),
    pos=(2.0, 2.0, 2.0),
    lookat=(0, 0, 0),
    fov=30,
)

scene.build()

# Render and save images
for i in range(30):
    scene.step()
    if i % 5 == 0:
        print(f"Step {i}/30")
        rgb_data = cam.render(rgb=True)
        
        # rgb_data[0] contains the actual image data
        img_array = rgb_data[0]
        img = Image.fromarray((img_array * 255).astype(np.uint8))
        img.save(f"{save_dir}/frame_{i:03d}.png")
        print(f"Frame {i} saved to {save_dir}")
```

# Results

While real-time visualization is not possible, frame-by-frame image saving works successfully
FPS varies by environment, but achieves about 10-15 FPS on A100 GPU
Saved images can be found in the '/content/simulation_frames' directory

# Questions

Are there any other methods to achieve real-time visualization in Colab? #166 
Are there ways to improve the performance of the current approach?
Is real-time visualization possible in other cloud environments?

# Environment Details

Google Colab
GPU: NVIDIA A100-SXM4-40GB

SuanCai-Fish:
I have the same issue. Is it possible to use SSH to access an interactive window on a remote server? From my tests, it seems not to work. I ran the same installation program on both a remote server and a local server. The local server with a display had no issues, but the remote server couldn't show the interactive window.

ardkyer:
[...]

Issue #82, opened 2024-12-19
Remote Dev


zihan987:
I want to deploy Genesis on a server and access it remotely using SSH from my laptop. Would this be inconvenient?

YilingQiao:
We can use headless rendering.
https://genesis-world.readthedocs.io/en/latest/user_guide/getting_started/visualization.html#camera-headless-rendering

PR #2439, opened 2026-02-20
[FEATURE] Add web-based viewer for remote display support.

Ozzey:
## Description

- Added `genesis/vis/streaming/` with `WebRTCStreamer` and `GenesisCameraVideoTrack` (aiortc + aiohttp + av).
- Added optional `webrtc` extra in `pyproject.toml`.
- Added examples:
  - `examples/streaming/webrtc_camera.py`
  - `examples/streaming/webrtc_manipulation_env.py`
- Added docs:
  - `WEBRTC_STREAMING.md`
  - `examples/streaming/README.md`
- Added/updated tests:
  - `tests/test_webrtc_streaming.py` with optional-import behavior and route checks.

## Related Issues
Related to:
- https://github.com/Genesis-Embodied-AI/Genesis/issues/230
- https://github.com/Genesis-Embodied-AI/Genesis/issues/82

## Motivation and Context
Genesis currently supports headless simulation (`show_viewer=False`), but remote real-time visualization over SSH/cloud (e.g., Colab/remote dev servers) is difficult when interactive viewer windows are unavailable.

This PR adds an MVP browser-based WebRTC stream for camera output in headless mode, which is useful for monitoring simulation and visual inference workflows (e.g., RL training/eval) without a local display.

### Unit tests
```bash
uv run pytest tests/test_webrtc_streaming.py
```
Result:
```bash
2 passed
```

Lint/format
```bash
uv run --with ruff ruff check genesis/vis/streaming examples/streaming tests/test_webrtc_streaming.py
uv run --with ruff ruff format --check genesis/vis/streaming examples/streaming tests/test_webrtc_streaming.py
```
Manual test (headless)
```bash
pip install -e ".[webrtc]"
python3 examples/streaming/webrtc_camera.py --host 0.0.0.0 --port 8000
# open http://<server-ip>:8000
```
Verified:

- browser receives live stream
- state chips update (Peer/ICE)
- pressing Close Stream closes browser stream and exits server process

### Screenshots:
Screenshot 2026-02-20 at 1 24 57 PM

## Checklist:

- [X] I read the **CONTRIBUTING** document.
- [X] I followed the `Submitting Code Changes` section of **CONTRIBUTING** document.
- [X] I tagged the title correctly (including BUG FIX/FEATURE/MISC/BREAKING)
- [X] I tested my changes and added instructions on how to test it for reviewers.

- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.


YilingQiao:
thank you! I'm taking a look

YilingQiao:
I cannot see the rendering. See the screenshot.

webrtc

Coding agent me it's because
```
  1. force_render=False — stale/empty frames                                                                   
                                                                                                               
  On line 118, the track calls:                                                                                
  self._camera.render(rgb=True, ..., force_render=False)                                                       
                             
  With force_render=False, the rasterizer only re-renders if it detects the scene has changed. But in the
  example, scene.step() and camera.render() are both running as cooperative async tasks in the same event loop
  thread. The timing is unpredictable — recv() can fire before scene.step() has run, so the rasterizer sees no
  change and returns a stale (black) buffer.

  2. Event loop starvation

  Both scene.step() and camera.render() are synchronous blocking calls sharing the same asyncio event loop:

  # Task 1: stepping
  async def _step_scene(scene, step_fps):
      while True:
          scene.step()          # blocks event loop
          await asyncio.sleep(step_dt)

  # Task 2: frame delivery (inside aiortc)
  async def recv(self):
      await self._sleep_until_next_frame()
      frame_array = await self._render_frame_with_retry()  # blocks event loop

  When scene.step() blocks for a while, the aiortc recv() task can't run, so frames are delayed. When recv()
  blocks on camera.render(), stepping stalls. They starve each other, causing inconsistent frame delivery and
  long gaps where the browser shows black (waiting for a video keyframe it missed).

```

Any idea?

Ozzey:
> I cannot see the rendering. See the screenshot.
> 
> webrtc
> Coding agent me it's because
> 
> ```
>   1. force_render=False — stale/empty frames                                                                   
>                                                                                                                
>   On line 118, the track calls:                                                                                
>   self._camera.render(rgb=True, ..., force_render=False)                                                       
>                              
>   With force_render=False, the rasterizer only re-renders if it detects the scene has changed. But in the
>   example, scene.step() and camera.render() are both running as cooperative async tasks in the same event loop
>   thread. The timing is unpredictable — recv() can fire before scene.step() has run, so the rasterizer sees no
>   change and returns a stale (black) buffer.
> 
>   2. Event loop starvation
> 
>   Both scene.step() and camera.render() are synchronous blocking calls sharing the same asyncio event loop:
> 
>   # Task 1: stepping
>   async def _step_scene(scene, step_fps):
>       while True:
>           scene.step()          # blocks event loop
>           await asyncio.sleep(step_dt)
> 
>   # Task 2: frame delivery (inside aiortc)
>   async def recv(self):
>       await self._sleep_until_next_frame()
>       frame_array = await self._render_frame_with_retry()  # blocks event loop
> 
>   When scene.step() blocks for a while, the aiortc recv() task can't run, so frames are delayed. When recv()
>   blocks on camera.render(), stepping stalls. They starve each other, causing inconsistent frame delivery and
>   long gaps where the browser shows black (waiting for a video keyframe it missed).
> ```
> 
> Any idea?

It seems chromium based browsers can fire ```ontrack``` with ```event.streams``` empty, there was no playable stream attached and the video stayed black. This doesn't happen in Safari so I didn't face this issue during testing.

I have modified the code now and fixed the bug + tested on few different browsers. Please test it and let me know if there are any other bugs/issues here.

tests/test_webrtc_streaming.py
```diff
@@ -0,0 +1,111 @@
...
+from pathlib import Path
+
+import numpy as np
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def initialize_genesis():
+    # Override tests/conftest.py autouse fixture: this unit test only validates the streaming app wiring.
+    yield
```
duburcqa: This is not how it works. Just do:
```python
@pytest.mark.parametrize("backend", [None])  # Disable genesis initialization at worker level
```

tests/test_webrtc_streaming.py
duburcqa, line 1: I don't think any of this is really useful. Just add one integration test popping a simple simulation (falling box), then record frames from the browser using `playwright-python` (https://github.com/microsoft/playwright-python). Then check that:
* the UX is ok
* the scene image is consistent with a local camera screenshot without going through the web browser
* the box is falling (just run a few frames, no more than 10, with a BIG timestep, like 0.05 or even 0.1).

examples/streaming/README.md
```diff
@@ -0,0 +1,100 @@
...
+git add genesis/vis/streaming examples/streaming tests/test_webrtc_streaming.py \
+  WEBRTC_STREAMING.md README.md pyproject.toml
+git commit -m "[FEATURE] Add headless WebRTC streaming MVP with examples and docs"
+```
+
+5. Push branch:
+
+```bash
+git push -u origin feature/webrtc-streaming-mvp
+```
```
duburcqa: Remove this section. it is not useful.

duburcqa: Remove this section. it is not useful.

examples/streaming/webrtc_camera.py
```diff
@@ -0,0 +1,109 @@
...
+    print(f"WebRTC viewer URL: {url}")
+
+    step_task = asyncio.create_task(_step_scene(scene, args.step_fps))
+    try:
+        await streamer.wait_for_shutdown_request()
+    finally:
+        step_task.cancel()
+        await asyncio.gather(step_task, return_exceptions=True)
+        await streamer.stop()
+        scene.destroy()
```
duburcqa, line 88: Having to wrap everything as async functions is destroying user experience. This is not acceptable.

PR #2445, opened 2026-02-22
[MISC] Speed up solver via linesearch/update_constraint/func_init optimizations

erizmr:
## Related Issue

## Checklist:

### DIFF
--- a/genesis/engine/couplers/ipc_coupler/coupler.py
+++ b/genesis/engine/couplers/ipc_coupler/coupler.py
@@ -10,18 +10,16 @@ import numpy as np
 
 import genesis as gs
 import genesis.utils.geom as gu
-from genesis.options.solvers import RigidOptions
+from genesis.engine.materials.FEM.cloth import Cloth
+from genesis.options.solvers import IPCCouplerOptions, RigidOptions
 from genesis.repr_base import RBC
-from genesis.utils.misc import tensor_to_array, qd_to_numpy, geometric_mean, harmonic_mean
+from genesis.utils.misc import geometric_mean, harmonic_mean, qd_to_numpy, tensor_to_array
 
 if TYPE_CHECKING:
-    from genesis.engine.entities import RigidEntity, FEMEntity
-    from genesis.engine.entities.rigid_entity import RigidLink, RigidJoint
+    from genesis.engine.entities import FEMEntity, RigidEntity
+    from genesis.engine.entities.rigid_entity import RigidJoint, RigidLink
     from genesis.engine.simulator import Simulator
-    from genesis.engine.solvers import RigidSolver, FEMSolver
-
-from genesis.engine.materials.FEM.cloth import Cloth
-from genesis.options.solvers import IPCCouplerOptions
+    from genesis.engine.solvers import FEMSolver, RigidSolver
 
 # Check if libuipc is available
 try:
@@ -34,7 +32,6 @@ except ImportError:
 if TYPE_CHECKING or UIPC_AVAILABLE:
     import polyscope as ps
     from uipc.backend import SceneVisitor
-    from uipc.core import Engine, World, Scene, AffineBodyStateAccessorFeature, ContactElement, Object, SubsceneElement
     from uipc.constitution import (
         AffineBodyConstitution,
         AffineBodyPrismaticJoint,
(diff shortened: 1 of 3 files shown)


genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -2262,6 +2262,12 @@ def update_bracket_no_eval_local(
     return flag, p_alpha, p_cost, p_grad, p_hess, p_next_alpha
 
 
+@qd.func
+def _log_scale(min_value: gs.qd_float, max_value: gs.qd_float, num_values: qd.i32, i: qd.i32) -> gs.qd_float:
+    step = (qd.log(max_value) - qd.log(min_value)) / qd.max(1.0, gs.qd_float(num_values - 1))
+    return qd.exp(qd.log(min_value) + gs.qd_float(i) * step)
```
duburcqa, line 2268: This should be moved in generic quadrants kernel utils

genesis/utils/array_class.py
```diff
@@ -1876,6 +1876,7 @@ class StructRigidSimStaticConfig(metaclass=AutoInitMeta):
     n_entities: int = -1
     n_links: int = -1
     n_geoms: int = -1
+    iterations: int = -1
```
duburcqa, line 1879: Remove this. It is not the right place. We do consider dynamically changing variables as pre-compiled constant. Moreover, "iterations" is a terrible variable name.

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
```diff
@@ -1,55 +1,420 @@
-import quadrants as ti
+import quadrants as qd
 
 import genesis as gs
 import genesis.utils.array_class as array_class
 from genesis.engine.solvers.rigid.constraint import solver
 
+LS_PARALLEL_K = 16
+LS_PARALLEL_MIN_STEP = 1e-6
```
duburcqa, line 8: This should depends on the desired precision. It should not be hard-coded this way.

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
```diff
@@ -1,55 +1,420 @@
...
+LS_PARALLEL_N_REFINE = 1  # number of successive refinement passes in parallel linesearch
```
duburcqa, line 9: These 2 options should probably be added to the rigid solver options.

genesis/utils/array_class.py
erizmr: This is for removing the reading from field `rigid_global_info.iterations[None]` in python scope, i.e., removing the implicit cpu-gpu sync. I have two other solutions in mind: 1. Store as a Python attribute on `rigid_global_info` 2. add a new parameter such `n_solver_iterations` to `func_solve_body`, which one would you prefer? Or feel free to share a three

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
erizmr: I can add them to the rigid solver options. The `LS_PARALLEL_K` is similar to the `BLOCK_DIM` in `func_hessian_direct_tiled` impl, which determines the parallelism and shared memory pattern of the kernel, which is not runtime changing variable.

`LS_PARALLEL_MIN_STEP` is a starting low bound of the linesearch alpha search range. It is more reasonable to set it as a constant in my opinion.

genesis/engine/solvers/rigid/constraint/solver.py
erizmr: This is only for linesearch use. I intend to add a note to state it can be moved in generic quadrants kernel utils and do follow up PRs on it. How do you think?

genesis/utils/array_class.py
duburcqa: I need to think about it. At this point none of them seems acceptable.

duburcqa: Could you share more details regarding what this attribute is storing and how it is used? Is it changing at every simulation step or is it a constant? How does it works?

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
duburcqa: I see. Can you document them? I think it is weird to pick a constant that does not depends on genesis precision and/or linesearch tolerance. Does it make sense to use a threshold that is smaller than desired linesearch tolerance? You should document all of this.

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
duburcqa: If it does not make sense to make this parameters tunable then it is great, but you need to document why you think so.

genesis/utils/array_class.py
duburcqa: If it is a constant, the usual approach is to store another Python native attribute with _ prefix in the solver and pass it as input argument of Python-native functions only. We are doing this for instance to skip kernel calls for specialized collision detection if useless, eg box-box.

genesis/engine/solvers/rigid/constraint/solver.py
duburcqa: If think it should either be inlined or move in generic quadrants kernel utils. No good reason to wait.

erizmr:
To make the PRs easier for review, intend to split it into small PRs:
1. https://github.com/Genesis-Embodied-AI/Genesis/pull/2518
2. https://github.com/Genesis-Embodied-AI/Genesis/pull/2524
3. https://github.com/Genesis-Embodied-AI/Genesis/pull/2521
4. https://github.com/Genesis-Embodied-AI/Genesis/pull/2523

PR #2455, opened 2026-02-24
[MISC] Refactor internal implementation of IPC coupler.

duburcqa:
* getattr / hasattr should be prohibited in favor of None initialization and isinstance check
* external / non-owned instances are being hijacked to added attributes on the fly. This is strictly forbidden.
* Quadrants kernels are used whereas interop data are numpy arrays on CPU. It would be much faster and simpler to keep everything as torch tensors or numpy arrays from the start. Numpy implementation should be fully vectorized. Numba (single thread, CPU backend) should be used for computation heavy "kernels".
* Data preallocation based on max size is prohibited. The exact memory size that is needed should be allocated. Nothing more.
* Plain dict for packing attributes is prohibited. Strongly typed named data structures should be used instead, either via dataclasses or just simple NamedTuples.
* Local imports in functions is prohibited, unless strictly necessary to avoid circular dependencies.
* There are many prints that are commented out. This is prohibited. They should be converted to logging debug traces.
* There are code duplication everywhere. This should be avoided.
* Some variable names are not following our naming conventions. See Rigid engine for reference. You can also have a like to the description of PR#1053 (https://github.com/Genesis-Embodied-AI/Genesis/pull/1053).
* Many helpers that are defined in genesis/utils/misc.py and genesis/utils/geom.py are not used. This is a major flaw. In particular, 'detach().cpu().numpy()' should never be used and 'tensor_to_array' should be prefered.
* Calling Quadrants data instance method 'to_numpy()' is forbidden. 'qd_to_(torch|numpy)' should be used instead. Take advantage of torch-based zero-copy as much as possible to update the internal state of other solvers. Beware, numpy-based zero-copy is only supported on CPU backend, that is why relying on torch is probably preferable.
* Catch-all exceptions is prohibited, except for exception forwarding in multi-processing process. Moreover, critical errors should not be replaced by warnings. If something breaks the physics, it should just raise an exception, eventually via re-raise to clarify the error message.
* Direct use of numpy / torch / quadrants data types is prohibited, unless strictly necessary due to external interfaces enforcing specific dtypes. Our types are defined in genesis/__init__.py. You may also take inspiration / relying on our quadrants dataclasses dtypes defined in genesis/utils/array_class.py.
* Defining local-scope functions should be avoided unless very well motivated.
* attributes-like instances methods (eg has_any_rigid_coupling) should be converted to properties.
* scene.sim.coupler should not be considered part of public API. It cannot be expected from the user to call 'set_link_ipc_collision'. It should be replaced by some entity-level material options.
* Warnings for nan is not acceptable. Simulation should be halted. The correct way to do this is leveraging our existing errno mechanism. Define a new error code if necessary.
* Remove all legacy code. We do not handle deprecation / legacy compatibility for now since this coupler is still experimental.
* Unused helpers should be removed (eg has_coupling_type). No need to implement private helpers proactively if they would be easy to add later on.
* Prefer using dev getters/setters API in various solvers over directly accessing quadrants fields, mainly for maintainability concerns. Quadrants fields may be renamed or even removed, while getters/setters should be considered stable dev API.
* Batching based on batched / non-batched should be avoided. Deal with it at argument-level 'envs_idx=env_idx if batched else None'.
* Do not call getters/setters repeatedly for even env idx. This is inefficient. Deal with all envs at once as much as you can, even if it is necessary to fallback to for-loop for IPC-related getters/setters. Notably, calling 'qd_to_numpy' in tight loop is not acceptable. It should only be done once. Moreover, if it is absolutely necessary, then you should specify the slice to extract in input argument of 'qd_to_numpy' instead of slicing manually afterward.
* Avoid generic 'all_' prefix/suffix for containers if possible but rather be specific, eg 'joints_xanchor', 'links_inertia_i', 'geoms_pos', 'entities_quat', 'verts_idx'.
* Avoid pure varilable name assignment, eg 'mass_mat_env = mass_mat_all'.
* Prefer clear over fresh allocation for mutable containers, eg 'self.abd_data_by_link.clear()' instead of 'self.abd_data_by_link = {}'.
* Never use `flatten()`, `reshape((-1,))`, or `ravel()` in place of squashing. Moreover, prefer `[..., 0, :]` over `squash(axis=-2)`.
* Don't use np.asarray unless strictly necessary, ie for public API with no control on the data source. For internal dev or private code path, it should NEVER be the case. 
* Don't use np.asarray unless strictly necessary, ie for public API with no control on the data source. For internal dev or private code path, it should NEVER be the case.

### DIFF
--- a/genesis/engine/couplers/ipc_coupler/coupler.py
+++ b/genesis/engine/couplers/ipc_coupler/coupler.py
@@ -64,7 +64,7 @@ if UIPC_AVAILABLE:
     )
     from uipc.backend import SceneVisitor
     from uipc.unit import MPa
-    
+
 from .data import ABDLinkEntry, ArticulatedEntityData, ContactForceEntry, ForceBatch, IPCCouplingData, ArticulationData
 from .utils import (
     find_target_link_for_fixed_merge,
--- a/tests/test_ipc.py
+++ b/tests/test_ipc.py
@@ -192,8 +192,7 @@ def test_joints(n_envs, coupling_type, joint_type, fixed, show_viewer):
     assert (0, moving_link_idx) in scene.sim.coupler._link_to_abd_slot
     if coupling_type == "two_way_soft_constraint":
         assert moving_link_idx in scene.sim.coupler.abd_data_by_link
-        envs_data = scene.sim.coupler.abd_data_by_link[moving_link_idx]
-        assert all(envs_data[i] is not None for i in envs_idx)
+        assert not any(env_data is None for env_data in scene.sim.coupler.abd_data_by_link[moving_link_idx])
     elif coupling_type == "external_articulation":
         entity_idx = scene.sim.rigid_solver.entities.index(robot)
         art_data = scene.sim.coupler._articulated_entities[entity_idx]


genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -29,14 +36,47 @@
...
+        label_surface,
+        label_triangle_orient,
+        linemesh,
+        merge,
+        tetmesh,
+        trimesh as uipc_trimesh,
+        flip_inward_triangles,
+    )
+    from uipc.backend import SceneVisitor
+    from uipc.unit import MPa
```
duburcqa, line 66: These imports should be moved below this block to avoid false positive.

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -710,123 +296,40 @@ def _init_ipc(self):
...
+        # Each entry: (element_a, element_b, friction_mu, enabled)
+        opts = self.options
+        abd = self._ipc_abd_contact
+        fem = self._ipc_fem_contact
+        cloth = self._ipc_cloth_contact
+        ground = self._ipc_ground_contact
+        no_col = self._ipc_no_collision_contact
+        contact_mu = opts.contact_friction_mu
+        fem_mu = opts.fem_fem_friction_mu
+        resistance = opts.contact_resistance
```
duburcqa: Remove these proxies. They are irrelevant.

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -846,7 +349,7 @@ def _init_ipc(self):
                             self._ipc_scene_subscenes[i], self._ipc_scene_subscenes[j], False
                         )
 
-        self.abd_data_by_link = {}
+        self.abd_data_by_link = {}  # Reset on each build
```
duburcqa: Prefer clear over fresh allocation  for mutable containers.

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -981,7 +483,7 @@ def _add_rigid_geoms_to_ipc(self):
                 geom_type = rigid_solver.geoms_info.type[i_g]
                 if geom_type == gs.GEOM_TYPE.PLANE:
                     # Handle planes as static IPC geometry (no coupling needed)
-                    pos = rigid_solver.geoms_info.pos[i_g].to_numpy()
+                    pos = geoms_pos[i_g]
```
duburcqa, line 530: Beware `rigid_solver.geoms_info` may be batched depending on rigid options.

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -1001,7 +503,7 @@ def _add_rigid_geoms_to_ipc(self):
                     scene_subscenes[i_b].apply_to(plane_geom)
 
                 plane_obj.geometries().create(plane_geom)
-                rigid_solver._mesh_handles[f"rigid_plane_{i_b}_{geom_idx}"] = plane_geom
+                self._rigid_mesh_handles[f"rigid_plane_{i_b}_{geom_idx}"] = plane_geom
```
duburcqa: This does not match our naming conventions. it should be `self._rigid_meshes_handle`

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -1104,30 +604,26 @@ def _add_rigid_geoms_to_ipc(self):
...
+                        rigid_mesh = uipc_trimesh(transformed_verts.astype(np.float64, copy=False), geom_faces.astype(np.int32, copy=False))
 
                         # Store uipc mesh (SimplicialComplex) for merging
                         link_geoms[target_link_idx]["meshes"].append((i_g, rigid_mesh))
 
-                    except Exception as e:
-                        gs.logger.warning(f"Failed to convert trimesh to tetmesh for geom {i_g}: {e}")
-                        continue
+                    except (ValueError, RuntimeError) as e:
+                        gs.raise_exception(f"Failed to create IPC surface mesh for geom {i_g}: {e}")
```
duburcqa: Prefer `gs.raise_exception_from(f"Failed to create IPC surface mesh for geom {i_g}", e)`

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -1104,30 +604,26 @@ def _add_rigid_geoms_to_ipc(self):
...
-
-                except Exception as e:
-                    gs.logger.warning(f"Failed to process geom {i_g}: {e}")
-                    continue
+                        link_geoms[target_link_idx]["link_world_pos"] = tensor_to_array(
+                            rigid_solver.get_links_pos(links_idx=target_link_idx, envs_idx=i_b)
+                        ).flatten()
+                        link_geoms[target_link_idx]["link_world_quat"] = tensor_to_array(
+                            rigid_solver.get_links_quat(links_idx=target_link_idx, envs_idx=i_b)
+                        ).flatten()
```
duburcqa: Never use `flatten()` or `reshape((-1,))` in place of squashing. Moreover, prefer `[..., 0, :]` over `squash(axis=-2)`.

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -1104,30 +604,26 @@ def _add_rigid_geoms_to_ipc(self):
...
+
+                except (ValueError, RuntimeError, KeyError) as e:
+                    gs.raise_exception(f"Failed to process geom {i_g} for IPC: {e}")
```
duburcqa: Same.

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -1406,72 +839,56 @@ def _init_abd_state_accessor(self):
...
+        ------
+        RuntimeError
+            If AffineBodyStateAccessorFeature is not available in the libuipc version.
+        """
+        self._abd_state_feature = self._ipc_world.features().find(AffineBodyStateAccessorFeature)
+        if self._abd_state_feature is None:
+            raise RuntimeError(
+                "AffineBodyStateAccessorFeature not available. "
+                "Please update libuipc to a version that supports this feature."
+            )
```
duburcqa: No need to do this. Just assume it is available. Users are supposed to install the correct version.

genesis/engine/materials/rigid.py
```diff
@@ -46,6 +46,11 @@ class Rigid(Material):
         coupling_link_filter : tuple of str or None, optional
             Tuple of link names to include in IPC coupling. Only supported with coupling_mode='two_way_soft_constraint'.
             If None, all links participate. Use this to filter to specific links. Default is None.
+        coupling_collision_enabled : bool, optional
+            Whether IPC collision is enabled for this entity's links. Default is True.
```
duburcqa: This is not consistent with our naming conventions. It should rather be `enable_coupling_collision`.

Moreover, it should be clarify that this option is only supported by IPC.

genesis/engine/materials/rigid.py
```diff
@@ -46,6 +46,11 @@ class Rigid(Material):
...
+        coupling_collision_links : tuple of str or None, optional
+            If set, only these links are affected by ``coupling_collision_enabled``.
+            If None, the setting applies to ALL coupled links of this entity. Default is None.
```
duburcqa, line 55: Clarify that this option is only supported by IPC.

genesis/engine/materials/rigid.py
```diff
@@ -46,6 +46,11 @@ class Rigid(Material):
         coupling_link_filter : tuple of str or None, optional
             Tuple of link names to include in IPC coupling. Only supported with coupling_mode='two_way_soft_constraint'.
             If None, all links participate. Use this to filter to specific links. Default is None.
+        coupling_collision_enabled : bool, optional
```
duburcqa: I wonder if disabling collision for a given entity is any different from specifying 'coupling_mode=None'. If it is equivalent, only one of the 2 alternatives should be kept.

genesis/engine/couplers/ipc_coupler/utils.py
```diff
@@ -371,3 +371,178 @@ def build_link_transform_matrix(pos_3, quat_4):
...
+        Mass matrix submatrix in column-major (Fortran) order
+    """
+    # Build absolute DOF indices
+    abs_indices = np.array([dof_start + d for d in joint_dof_indices[:n_joints]], dtype=int)
+
+    # Extract submatrix using fancy indexing
+    sub = mass_mat[np.ix_(abs_indices, abs_indices)]
+
+    # Return flattened in column-major order (Fortran order = transpose then flatten)
+    return sub.T.flatten()
```
duburcqa: This is inefficient because fancy indexing forces a copy. You should probably use our helper 'indices_to_mask' in gs.utils.misc.

duburcqa: As already mentioned. Avoid `.flatten()`

genesis/engine/couplers/ipc_coupler/utils.py
```diff
@@ -371,3 +371,178 @@ def build_link_transform_matrix(pos_3, quat_4):
...
+    torques = np.cross(r, forces)  # (n, 3)
+
+    # Accumulate using np.add.at (unbuffered, like atomic add)
+    np.add.at(out_forces, (link_indices, env_indices), forces)
+    np.add.at(out_torques, (link_indices, env_indices), torques)
+
+    return out_forces, out_torques
+
+
+def extract_joint_mass_matrix(mass_mat, dof_start, joint_dof_indices, n_joints):
```
duburcqa: passing n_joints as input argument seems irrelevant.

genesis/engine/couplers/ipc_coupler/utils.py
duburcqa: Wait, the mass matrix is symmetric, what the point of transpose?

genesis/engine/couplers/ipc_coupler/utils.py
```diff
@@ -371,3 +371,178 @@ def build_link_transform_matrix(pos_3, quat_4):
...
+
+    # M_affine = S @ A (rotation part of transform)
+    A = abd_transforms[:, :3, :3]
+    M = np.einsum("nij,njk->nik", S, A)
+    out[:, 3:] = M.reshape(n, 9)
+
+    return out
+
+
+def compute_coupling_forces(
```
duburcqa, line 384: This looks like something that should be jitted with numba.

genesis/engine/couplers/ipc_coupler/utils.py
```diff
@@ -371,3 +371,178 @@ def build_link_transform_matrix(pos_3, quat_4):
...
     t.rotate(uipc_quat)
     return t.matrix().copy()
+
+
+# ============================================================
+# Numpy computation functions (replacing Quadrants kernels)
+# ============================================================
+
+
+def compute_external_force_12d(contact_forces, contact_torques, abd_transforms):
```
duburcqa, line 337: Same. probably worth jitting.

genesis/engine/couplers/ipc_coupler/utils.py
```diff
@@ -371,3 +371,178 @@ def build_link_transform_matrix(pos_3, quat_4):
     uipc_quat = Quaternion(quat_4)
     t.rotate(uipc_quat)
     return t.matrix().copy()
```
duburcqa: WTF is this? You geom helpers for this. Not UIPC.

genesis/engine/couplers/ipc_coupler/utils.py
duburcqa: Why local imports are still there?

genesis/engine/couplers/ipc_coupler/utils.py
```diff
@@ -104,7 +104,7 @@ def is_robot_entity(entity):
     """Heuristic: treat URDF/MJCF/Drone morphs as robots."""
     try:
         return isinstance(entity.morph, (gs.morphs.URDF, gs.morphs.MJCF, gs.morphs.Drone))
-    except Exception:
+    except AttributeError:
         return False
```
duburcqa: Just do `return isinstance(entity.morph, (gs.morphs.URDF, gs.morphs.MJCF, gs.morphs.Drone))`. There is no point of handling exception here. Moreover, `gs.morphs.Drone` derives from `gs.morphs.URDF`. Finally, USD is missing...

duburcqa: Take inspiration of what is done here to determine what is a robot and what is not: https://github.com/Genesis-Embodied-AI/Genesis/blob/aee46811f6b42107771ed3ea1ec8010ae026a31e/genesis/engine/entities/rigid_entity/rigid_entity.py#L756

genesis/engine/couplers/ipc_coupler/data.py
```diff
@@ -81,43 +21,36 @@ def __init__(self, max_links):
...
-        self.entity_indices = qd.field(dtype=qd.i32, shape=max_entities)
-        self.entity_env_indices = qd.field(dtype=qd.i32, shape=max_entities)
-        self.entity_n_dofs = qd.field(dtype=qd.i32, shape=max_entities)
-        self.entity_n_joints = qd.field(dtype=qd.i32, shape=max_entities)
-        self.entity_dof_start = qd.field(dtype=qd.i32, shape=max_entities)  # DOF start index in rigid solver
+        self.n_entities = n_entities
+        self.entity_indices = np.zeros(n_entities, dtype=np.int32)
+        self.entity_n_dofs = np.zeros(n_entities, dtype=np.int32)
+        self.entity_n_joints = np.zeros(n_entities, dtype=np.int32)
+        self.entity_dof_start = np.zeros(n_entities, dtype=np.int32)
```
duburcqa: Why not use `gs.np_int` here?

genesis/engine/couplers/ipc_coupler/data.py
```diff
@@ -1,78 +1,18 @@
...
-        self.stored_qpos_size = qd.field(dtype=qd.i32, shape=max_links)  # Number of dofs per entity
-        self.stored_qpos_start = qd.field(dtype=qd.i32, shape=max_links)  # DOF start index per entity
-
-
 class IPCCouplingData:
-    """Data-oriented class for IPC coupling force computation using numpy arrays."""
+    """Pre-allocated numpy arrays for coupling force computation."""
 
     def __init__(self, max_links):
-        # Pre-allocated numpy buffers for coupling force computation
```
duburcqa, line 72: Why remove the first line of this comment?

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -3150,7 +2227,7 @@ def _add_articulated_entities_to_ipc(self):
                 if self._use_subscenes:
                     try:
                         scene_subscenes[i_b].apply_to(articulation_geo)
-                    except Exception:
+                    except (TypeError, RuntimeError):
                         # Some uipc builds may not expose subscene binding for articulation geometry types.
                         pass
```
duburcqa: I don't think it is acceptable to ignore this error. Just plain and simple crash without catching this exception, or re-raise with a clean error message.

genesis/engine/couplers/ipc_coupler/coupler.py
duburcqa, line 1: Please apply pre-commit formatting.

genesis/engine/couplers/ipc_coupler/coupler.py
duburcqa: You should rather get all pos / quat at once, and assign subblocks. Same as we did elsewhere. It may be convenient to directly access rigid solver state to avoid dealing with batched / non-batched: `rigid_solver.links_state.pos`. Initially, the idea was to only rely on public getters, but there is a mix already, so at this point we may just go all in and only access rigid solver state directly...

genesis/engine/couplers/ipc_coupler/coupler.py
duburcqa: I think it would be more appropriate to use our existing geom utils.

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -1383,9 +817,8 @@ def animate_rigid_link(info):
 
                         link_obj_counter += 1
 
-                except Exception as e:
-                    gs.logger.warning(f"Failed to create IPC object for link {link_idx}: {e}")
-                    continue
+                except (ValueError, RuntimeError, KeyError) as e:
+                    gs.raise_exception(f"Failed to create IPC object for link {link_idx}: {e}")
```
duburcqa: As already mentioned, use `gs.raise_exception_from`

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -1507,23 +924,72 @@ def _collect_coupling_config_from_materials(self):
...
+                        if force_attr is not None:
+                            force_vector = force_data[key]
+                            view(force_attr)[:] = force_vector.reshape(-1, 1)
+
+                        if is_constrained_attr is not None:
+                            view(is_constrained_attr)[:] = 1
+                    else:
+                        if force_attr is not None:
+                            view(force_attr)[:] = np.zeros((12, 1), dtype=np.float64)
+
```
duburcqa: Do not skip line at the end of branches.

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -1507,23 +924,72 @@ def _collect_coupling_config_from_materials(self):
...
+            except (ValueError, RuntimeError, KeyError) as e:
+                gs.raise_exception(f"Error setting IPC animation target for link {link_idx}, env {env_idx}: {e}")
```
duburcqa: `gs.raise_exception_from`

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -1536,95 +1002,63 @@ def has_any_rigid_coupling(self) -> bool:
...
+        """Look up IPC transform for a link/env pair, or return None."""
+        if self.abd_data_by_link is None:
+            return None
+        env_data = self.abd_data_by_link.get(link_idx)
+        if env_data is None:
+            return None
+        entry = env_data.get(env_idx)
+        if entry is None:
+            return None
+        return entry.get("transform")
```
duburcqa: Are all these branches really necessary? It sounds like a bad code design. Either you get something that exists or it raises an exception. Nothing more. The code flow should be adapted to avoid having to deal with all these fallbacks...

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -1536,95 +1002,63 @@ def has_any_rigid_coupling(self) -> bool:
...
-        # Store collision settings
-        if entity_idx not in self._link_collision_settings:
-            self._link_collision_settings[entity_idx] = {}
-
-        for link_idx in target_links:
-            self._link_collision_settings[entity_idx][link_idx] = enabled
-
-        status = "enabled" if enabled else "disabled"
-        gs.logger.info(f"Entity {entity_idx}: {len(target_links)} link(s) set to collision {status}")
+        rigid_solver = self.rigid_solver
```
duburcqa: I don't think defining this proxy is useful.

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -1536,95 +1002,63 @@ def has_any_rigid_coupling(self) -> bool:
...
+        is_batched = self.sim._B > 1
```
duburcqa: If this proxy is only used once, just inline it.

duburcqa: Moreover, prefer, `scene.n_envs` for `sim._B` if available.

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -2485,79 +1769,40 @@ def _apply_abd_coupling_forces(self, entity_set=None):
...
-                    continue
+        # Group by environment and apply
+        env_batches = {}
+        for i in range(n_items):
+            env_idx = int(env_indices[i])
+            if env_idx not in env_batches:
+                env_batches[env_idx] = {"link_indices": [], "forces": [], "torques": []}
+            env_batches[env_idx]["link_indices"].append(int(link_indices[i]))
+            env_batches[env_idx]["forces"].append(out_forces[i])
+            env_batches[env_idx]["torques"].append(out_torques[i])
```
duburcqa: Stop using plain dict. As mentioned in guidelines of this PR:
* Plain dict for packing attributes is prohibited. Strongly typed named data structures should be used instead, either via dataclasses or just simple NamedTuples.

examples/IPC_Solver/ipc_objects_falling.py
```diff
@@ -102,11 +102,20 @@ def main():
...
+            on_trace_ready=torch.profiler.tensorboard_trace_handler("./benchmark"),
+            schedule=torch.profiler.schedule(wait=0, warmup=0, active=1),
+            record_shapes=False,
+            profile_memory=False,
+            with_stack=True,
+            with_flops=False,
+        ):
+        horizon = 20 if "PYTEST_VERSION" not in os.environ else 5
+        for i in range(horizon):
+            scene.step()
```
duburcqa: Revert this.

genesis/engine/materials/rigid.py
```diff
@@ -46,6 +46,13 @@ class Rigid(Material):
         coupling_link_filter : tuple of str or None, optional
             Tuple of link names to include in IPC coupling. Only supported with coupling_mode='two_way_soft_constraint'.
             If None, all links participate. Use this to filter to specific links. Default is None.
+        enable_coupling_collision : bool, optional
+            Whether IPC collision is enabled for this entity's links. Only used by the IPC coupler.
+            Unlike ``coupling_mode=None`` (which removes the entity from IPC entirely), setting this to
+            False keeps the entity in IPC for coupling forces but disables contact response. Default is True.
```
duburcqa, line 52: Wait, the whole point of IPC is dealing with collisions (soft bodies in the first place but also rigid optionally). I don't know what would be the point of adding an entity to IPC but not dealing with collisions...

genesis/engine/couplers/ipc_coupler/utils.py
```diff
@@ -364,10 +361,199 @@ def build_link_transform_matrix(pos_3, quat_4):
...
+    joint_dof_indices : array-like of int
+        DOF indices for each joint (local to entity)
+
+    Returns
+    -------
+    np.ndarray, shape (n_joints * n_joints,)
+        Mass matrix submatrix in column-major (Fortran) order
+    """
+    abs_indices = np.array([dof_start + d for d in joint_dof_indices], dtype=int)
+    sub = mass_mat[np.ix_(abs_indices, abs_indices)]
```
duburcqa: dof indices for a given joint should be contiguous. It would be better to pass a `slice` rather than an actual sequence of indices. Moreover, my original comment still stands:

This is inefficient because fancy indexing forces a copy. You should probably use our helper 'indices_to_mask' in gs.utils.misc.

genesis/engine/couplers/ipc_coupler/utils.py
```diff
@@ -340,17 +342,12 @@ def decompose_transform_matrix(transform_4x4):
...
     """
-    pos = transform_4x4[:3, 3]
-    rot_mat = transform_4x4[:3, :3]
-    quat_xyzw = R.from_matrix(rot_mat).as_quat()
-    quat_wxyz = np.array([quat_xyzw[3], quat_xyzw[0], quat_xyzw[1], quat_xyzw[2]])
-    return pos, quat_wxyz
+    return gu.T_to_trans_quat(transform_4x4)
 
 
 def build_link_transform_matrix(pos_3, quat_4):
```
duburcqa: Remove this method entirely. It is irrelevant for one line.

genesis/engine/couplers/ipc_coupler/utils.py
duburcqa: Remove this method entirely. It is irrelevant for one line.

genesis/engine/couplers/ipc_coupler/utils.py
```diff
@@ -101,11 +108,8 @@ def compute_link_to_link_transform(rigid_solver, from_link_idx, to_link_idx):
 
 
 def is_robot_entity(entity):
-    """Heuristic: treat URDF/MJCF/Drone morphs as robots."""
-    try:
-        return isinstance(entity.morph, (gs.morphs.URDF, gs.morphs.MJCF, gs.morphs.Drone))
-    except Exception:
-        return False
+    """Check if entity is a robot (URDF/MJCF/USD morph)."""
+    return isinstance(entity.morph, (gs.morphs.URDF, gs.morphs.MJCF, gs.morphs.USD))
```
duburcqa: Take inspiration of what is done here to determine what is a robot and what is not: https://github.com/Genesis-Embodied-AI/Genesis/blob/aee46811f6b42107771ed3ea1ec8010ae026a31e/genesis/engine/entities/rigid_entity/rigid_entity.py#L756

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -1536,95 +994,59 @@ def has_any_rigid_coupling(self) -> bool:
...
+        """Return (active_env_indices, articulation_slots_by_env) with env0 fallback."""
+        active_env_indices = art_data.get("active_env_indices", [art_data.get("env_idx", 0)])
+        articulation_slots_by_env = art_data.get(
+            "articulation_slots_by_env", {art_data.get("env_idx", 0): art_data["articulation_slot"]}
+        )
+        return active_env_indices, articulation_slots_by_env
+
+    def _get_abd_link_transform(self, link_idx, env_idx):
+        """Look up IPC transform for a link/env pair, or return None."""
+        entry = self.abd_data_by_link.get(link_idx, {}).get(env_idx)
```
duburcqa: Using `.get(env_idx)` for an integer index is weird. Why not just pre-allocated a tuple with all the data for each env from the start?

PR #2460, opened 2026-02-24
[MISC] Get rid of now unnecessary kernel workaround for Apple Metal backend.

hughperkins:


pyproject.toml
```diff
@@ -10,7 +10,7 @@ readme = "README.md"
 requires-python = ">=3.10,<3.14"
 dependencies = [
     "psutil",
-    "quadrants==0.4.1",
+    "quadrants==0.4.2b1",
```
duburcqa: Need official release and rebase then we can merge.

PR #2462, opened 2026-02-25
[MISC] More IPC coupler cleanup.

duburcqa:


genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -303,9 +315,8 @@ def _init_ipc(self):
             UIPCLogger.set_level(UIPCLogger.Level.Error)
             UIPCTimer.disable_all()
 
-        # Create workspace directory for IPC output.
-        # Use process ID to avoid collisions under pytest-xdist.
-        workspace = os.path.join(tempfile.gettempdir(), f"genesis_ipc_workspace_{os.getpid()}")
+        # Create workspace directory for IPC output, named after scene UID.
+        workspace = os.path.join(tempfile.gettempdir(), f"genesis_ipc_{self.sim.scene.uid}")
```
ACMLCZH, line 319: Does Genesis itself have own helper function to manage tmp folder like cache folder? `utils/particles.py` also uses lib `tempfile`.

genesis/engine/couplers/ipc_coupler/coupler.py
ACMLCZH, line 969:
```suggestion
                if collision_link_names is not None:
                    collision_links = [entity.get_link(name=name) for name in collision_link_names]
                else:
                    collision_links = entity.links
                for link in collision_links:
                    self._link_collision_settings.setdefault(entity_idx, {})[link.idx] = False
```

ACMLCZH, line 340: Why are `_ipc_dsb` and `_ipc_nks` not inserted? AI tells me this may be a bug.

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -643,18 +654,13 @@ def _add_rigid_geoms_to_ipc(self):
...
-                        rigid_mesh = uipc_trimesh(
-                            transformed_verts.astype(np.float64, copy=False), geom_faces.astype(np.int32, copy=False)
-                        )
-
-                        # Store uipc mesh (SimplicialComplex) for merging
-                        link_geoms[target_link_idx]["meshes"].append((i_g, rigid_mesh))
+                    # Create uipc trimesh directly (dim=2, surface mesh for ABD)
+                    rigid_mesh = uipc_trimesh(
+                        transformed_verts.astype(np.float64, copy=False), geom_faces.astype(np.int32, copy=False)
+                    )
```
ACMLCZH, line 660: ABD may require watertight mesh to compute its volume? Should we check the meshes before sending them into libuipc?

genesis/engine/couplers/ipc_coupler/coupler.py
duburcqa: No. temporary folders is not a thing in Genesis. We are doing our best to avoid it because it is a bad practice.

genesis/engine/couplers/ipc_coupler/coupler.py
duburcqa: Probably

genesis/engine/couplers/ipc_coupler/coupler.py
duburcqa: A simple assert would be great indeed

PR #2482, opened 2026-02-28
[MISC] Aggregate speed benchmark wandb uploads into a single run.

hughperkins:
Instead of creating a separate wandb run per benchmark test case, upload all speed/FPS results via upload_benchmarks_table_to_wandb.py after all benchmarks complete, matching the memory benchmark approach.


hughperkins:
uploads to wandb at least seem to be working: https://wandb.ai/genesis-ai-company/genesis-benchmarks-2/runs/zwg6g1ej/overview?nw=nwusergenesisai_hughperkins

PR #2490, opened 2026-03-03
[MISC] IPC Coupler Collide Logic Change and Cleanup

ACMLCZH:
### Behavior Change
  - **Fix rigid solver collision delegation**: correctly skip `ipc_only` link pairs and respect `collision_links` filter. `ipc_only` entities do not track poses in Genesis, so we should not apply any Rigid Solver physics to `ipc_only`. `ipc_only` and non-IPC entity collisions will be disabled. Other collisions remain unchanged.
  - **Rename material fields**: `coupling_tpe` → `ipc_coup_type`, `coupling_link_filter` → `coup_links`, `enable_coupling_collision` → `enable_ipc_collision`,
  `coupling_collision_links` → `ipc_collision_links` to follow `coup_*`/`ipc_*` naming convention
  - **Add IPC coupler options**: `show_debug_info`, `show_ipc_gui`; validate `contact_resistance > 0`
  - **Add `collision_links`**: positive filter for per-link rigid solver collision control
  - **Add `ipc_collision_links`**: positive filter for per-link IPC collision control

  ### Code Refactor
  - **Clean up IPC coupler internals**: unify `_by_link`/`_by_entity` naming convention for member dicts, make public fields private (`coupling_entries` → `_coupling_entries`,
  `coupling_data` → `_coupling_data`, `articulation_data` → `_articulation_data_by_entity`), remove dead storage fields from `ArticulatedEntityData` and `IPCCoupler`
  - **Refactor code structure**: eager constitution creation before loops instead of lazy `if None` checks, separate env-independent joint info from env-specific geometry creation in
  `_add_articulation_entities_to_ipc`, add `_init_ipc_gui` using `sys.platform` for EGL backend selection
  - **Expand test coverage**: 52 IPC tests including `test_find_target_links` for `merge_fixed_links=False` correctness, IPC GUI test, cloth biaxial stretching, collision delegation,
  and collision link filtering

### DIFF
--- a/genesis/engine/couplers/ipc_coupler/coupler.py
+++ b/genesis/engine/couplers/ipc_coupler/coupler.py
@@ -210,10 +210,15 @@ class IPCCoupler(RBC):
                 continue
 
             self._ipc_coup_types[entity] = ipc_coup_type = getattr(COUPLING_TYPE, ipc_coup_type.upper())
-            if ipc_coup_type == COUPLING_TYPE.EXTERNAL_ARTICULATION and entity.n_joints == 0:
-                gs.raise_exception(
-                    f"Rigid entity {i_e} has no joint. Coupling type 'external_articulation' is not supported."
-                )
+            if ipc_coup_type == COUPLING_TYPE.EXTERNAL_ARTICULATION:
+                if not entity.base_link.is_fixed:
+                    gs.raise_exception(
+                        f"Rigid entity {i_e} is not fixed. Coupling type 'external_articulation' is not supported."
+                    )
+                if entity.n_joints == 0:
+                    gs.raise_exception(
+                        f"Rigid entity {i_e} has no joint. Coupling type 'external_articulation' is not supported."
+                    )
             gs.logger.debug(f"Rigid entity {i_e}: coupling type '{ipc_coup_type.name.lower()}'")
 
             # Resolve link filter from material
@@ -411,6 +416,11 @@ class IPCCoupler(RBC):
                 # ---- Collect geom meshes ----
                 meshes = []
                 for source_link in source_links:
+                    if not source_link.geoms:
+                        gs.raise_exception(
+                            f"Rigid link {source_link.idx} has no collision geometry. Coupling type "
+                            "'external_articulation' is not supported."
+                        )
                     for geom in source_link.geoms:
                         if geom.type == gs.GEOM_TYPE.PLANE:
                             if entity_ipc_coup_type != COUPLING_TYPE.IPC_ONLY:
--- a/genesis/engine/entities/fem_entity.py
+++ b/genesis/engine/entities/fem_entity.py
@@ -890,10 +890,15 @@ class FEMEntity(Entity):
             envs_idx : array_like, optional
                 List of environment indices to apply the constraints to. If None, applies to all environments.
         """
+        from genesis.engine.couplers import IPCCoupler
+
         if self._solver._use_implicit_solver and not self._solver._enable_vertex_constraints:
             gs.raise_exception(
                 "This feature is disabled. Please set 'enable_vertex_constraints=True' when using FEM implicit solver."
             )
+
+        if isinstance(self.sim.coupler, IPCCoupler):
+            gs.raise_exception("This method is only supported by IPC coupler.")
 
         if not self._solver._constraints_initialized:
             self._solver.init_constraints()
--- a/tests/test_ipc.py
+++ b/tests/test_ipc.py
@@ -384,9 +384,11 @@ def test_link_filter_strict():
 
 @pytest.mark.required
 @pytest.mark.parametrize("n_envs", [0, 2])
-@pytest.mark.parametrize("ipc_coup_type", ["two_way_soft_constraint", "external_articulation"])
+@pytest.mark.parametrize(
+    "ipc_coup_type, fixed",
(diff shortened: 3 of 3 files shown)


genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -457,110 +459,84 @@ def _add_rigid_geoms_to_ipc(self) -> None:
                 if not meshes:
                     continue
 
-                # Merge meshes belonging to this target link
-                merged_mesh = meshes[0] if len(meshes) == 1 else uipc.geometry.merge(meshes)
-                uipc.geometry.label_surface(merged_mesh)
+                # ---- Merge meshes and apply world transform ----
+                abd_geom = meshes[0] if len(meshes) == 1 else uipc.geometry.merge(meshes)
```
duburcqa, line 471: This variable name is way worst than the original.

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -374,13 +383,13 @@ def _add_rigid_geoms_to_ipc(self) -> None:
         for link in self.rigid_solver.links:
             entity = link.entity
 
-            coupling_type = self._coupling_types.get(entity)
-            if coupling_type is None:
+            ipc_coup_type = self._ipc_coup_types.get(entity)
```
duburcqa, line 389: prefixing with ipc makes no sense here.

genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -389,30 +398,26 @@ def _add_rigid_geoms_to_ipc(self) -> None:
...
                 i_e = entity._idx_in_solver
 
-                # Inner loop: iterate geoms of all source links
+                # ---- Collect geom meshes ----
                 meshes = []
                 for source_link in source_links:
                     for geom in source_link.geoms:
                         if geom.type == gs.GEOM_TYPE.PLANE:
-                            if entity_coupling_type != COUPLING_TYPE.IPC_ONLY:
+                            if entity_ipc_coup_type != COUPLING_TYPE.IPC_ONLY:
```
duburcqa, line 426: Same. This is just worst than before.

duburcqa:
I had a look to your PR. Please remove all the API changes and variable renames and let's discuss this in a follow up.

genesis/engine/materials/FEM/base.py
```diff
@@ -47,8 +47,8 @@ def __init__(
 
         if friction_mu < 0:
             gs.raise_exception("`friction_mu` must be non-negative.")
-        if contact_resistance is not None and contact_resistance < 0:
-            gs.raise_exception("`contact_resistance` must be non-negative.")
+        if contact_resistance is not None and contact_resistance <= 0:
+            gs.raise_exception("`contact_resistance` must be positive.")
```
duburcqa: strictly positive

genesis/options/solvers.py
```diff
@@ -286,6 +286,11 @@ class IPCCouplerOptions(BaseCouplerOptions):
         For external_articulation with non-fixed base: whether base link is fully driven by IPC physics.
         When False, base link uses SoftTransformConstraint controlled by Genesis. When True, base link
         is fully driven by IPC physics. Defaults to False.
+    show_debug_info : bool, optional
+        Enable verbose libuipc logging, performance timers, and solver state dump to disk.
```
duburcqa: Remove this option. You should relying on debug logging.

genesis/options/solvers.py
```diff
@@ -286,6 +286,11 @@ class IPCCouplerOptions(BaseCouplerOptions):
...
+        Also enabled when Genesis logger level is DEBUG. Defaults to False.
+    show_ipc_gui : bool, optional
+        Enable the libuipc built-in GUI viewer for visualizing the IPC scene. Defaults to False.
```
duburcqa: Explicitly mention that this is a dev option for debug. You should also prefix it with '_'.

genesis/options/solvers.py
duburcqa: Regarding "solver state dump to disk.", the a secret env variable for this. We do not want to leak this on the user side.

ACMLCZH:
> I'm not ok with the rename of the options to prefix with 'ipc_'. These are not supposed to be IPC. They currently are IPC-only, but material options are not supposed to be coupler specific. We need to consider them generic and usable by any coupler.

I have thought about those naming. I think:
- `need_coup` is whether this entity should be added to the coupler's simulation.
- `coup_links` is when `need_coup=True`, whether only part of links are in the coupler simulation.
- `enable_coup_collision` is whether this entity should be added to the coupler's collision simulation.
- `coup_collision_links` is when `enable_coup_collision=True`, whether only part of links are in the coupler's collision simulation.
- `ipc_coup_type` is actually IPC-specific.

However, when in other Couplers, collision simulation is the whole simulation, so other Couplers may need either `need_coup` or `enable_coup_collision`. Only IPC Coupler uses both, so I made one of them `ipc_*`.
Would you like me to use `enable_coup_collision` and `coup_collision_links`? @duburcqa

genesis/engine/couplers/ipc_coupler/coupler.py
ACMLCZH: I think `abd_geom` here is consistent with other locations `*_geom`. And `merged_mesh` only applies for branch `len(meshes)>1`. Do we have better names?

PR #2495, opened 2026-03-03
[MISC] IPC Collider Logic Change

ACMLCZH:
### Behavior Change
  - **Fix rigid solver collision delegation**: correctly skip `ipc_only` link pairs and respect `collision_links` filter. `ipc_only` entities do not track poses in Genesis, so we should not apply any Rigid Solver physics to `ipc_only`. `ipc_only` and non-IPC entity collisions will be disabled. Other collisions remain unchanged.
  - **Rename material fields**: `coupling_tpe` → `coup_type`, `coupling_link_filter` → `coup_links`, `enable_coupling_collision` → `enable_coup_collision`,
  `coupling_collision_links` → `coup_collision_links` to follow `coup_*` naming convention
  - **Add IPC coupler options**: `show_ipc_gui`; validate `contact_resistance > 0`
  - **Add `collision_links`**: positive filter for per-link rigid solver collision control
  - **Add `ipc_collision_links`**: positive filter for per-link IPC collision control

### DIFF
--- a/tests/test_ipc.py
+++ b/tests/test_ipc.py
@@ -317,7 +317,7 @@ def test_ipc_rigid_ground_clearance(n_envs, show_viewer):
 
 @pytest.mark.required
 def test_needs_coup():
-    """needs_coup=False takes precedence over coup_type — entity excluded from IPC."""
+    """needs_coup=False excludes entity from IPC; coup_type with needs_coup=False raises."""
     scene = gs.Scene(
         coupler_options=gs.options.IPCCouplerOptions(),
         show_viewer=False,
@@ -325,7 +325,7 @@ def test_needs_coup():
     scene.add_entity(gs.morphs.Plane(), material=gs.materials.Rigid(needs_coup=False))
     scene.add_entity(
         morph=gs.morphs.Box(size=(0.1, 0.1, 0.1), pos=(0, 0, 0.5)),
-        material=gs.materials.Rigid(needs_coup=False, coup_type="ipc_only"),
+        material=gs.materials.Rigid(needs_coup=False),
     )
     scene.build()
     assert scene.sim.coupler._coupling_types == {}


genesis/engine/couplers/ipc_coupler/coupler.py
```diff
@@ -414,10 +430,7 @@ def _add_rigid_geoms_to_ipc(self) -> None:
                 meshes = []
                 for source_link in source_links:
                     if not source_link.geoms:
-                        gs.raise_exception(
-                            f"Rigid link {source_link.idx} has no collision geometry. Coupling type "
-                            "'external_articulation' is not supported."
-                        )
+                        continue
```
duburcqa: Remove this. It is not supported. Unless I'm missing something or it has been fixed in IPC? By the way, the version of IPC should be checked and a clear exception raised if not compatible with Genesis.

genesis/engine/materials/rigid.py
```diff
@@ -34,25 +34,32 @@ class Rigid(Material):
             Maximum resolution of the SDF grid. Must be >= sdf_min_res. Default is 128.
         gravity_compensation : float, optional
             Compensation factor for gravity. 1.0 cancels gravity. Default is 0.
-        coupling_type : str or None, optional
-            IPC coupling mode for this entity. Valid values:
-              - None: Entity not processed by IPC. Entity is completely ignored by IPC coupler.
+        collision_links : tuple of str or None, optional
+            Tuple of link names whose geoms participate in rigid solver collision detection.
+            If None, all links participate. When set, geoms belonging to links NOT in this tuple
+            are excluded from rigid solver collision pair computation. Default is None.
```
duburcqa: This is not just for IPC coupling?

genesis/engine/materials/rigid.py
```diff
@@ -34,25 +34,32 @@ class Rigid(Material):
...
+        coup_type : str or None, optional
+            Coupling mode for this entity. Valid values:
+              - None: Entity not processed by coupler. Entity is completely ignored by the coupler.
```
duburcqa: Mention that this option is only used by the IPC coupler.

genesis/engine/materials/rigid.py
```diff
@@ -34,25 +34,32 @@ class Rigid(Material):
...
-            Whether IPC collision is enabled for this entity's links. Only used by the IPC coupler.
-            Unlike ``coupling_type=None`` (which removes the entity from IPC entirely), setting this to
-            False keeps the entity in IPC for coupling forces but disables contact response. Default is True.
-        coupling_collision_links : tuple of str or None, optional
-            If set, only these links are affected by ``enable_coupling_collision``. Only used by the IPC coupler.
-            If None, the setting applies to ALL coupled links of this entity. Default is None.
+        coup_links : tuple of str or None, optional
+            Tuple of link names to include in coupling. When set, only the named links participate
+            in coupling; other links are excluded. Only supported with needs_coup=True and
+            ``two_way_soft_constraint`` type in IPC. Default is None.
```
duburcqa, line 51: Mention that this option is only used by the IPC coupler.

genesis/engine/materials/rigid.py
```diff
@@ -34,25 +34,32 @@ class Rigid(Material):
...
+        enable_coup_collision : bool, optional
+            Whether coupler collision is enabled for this entity's links.
+            Unlike ``coup_type=None`` (which removes the entity from the coupler entirely), setting this to
+            False keeps the entity in the coupler for coupling forces but disables contact response. Default is True.
```
duburcqa, line 55: Mention that this option is only used by the IPC coupler.

genesis/engine/materials/rigid.py
```diff
@@ -34,25 +34,32 @@ class Rigid(Material):
...
+        coup_collision_links : tuple of str or None, optional
+            Tuple of link names whose geoms participate in coupler collision.
+            Only effective when ``enable_coup_collision=True``. If None, all coupled links have collision.
+            When set, only the named links get coupler collision; other links are marked no-collision.
+            Default is None.
```
duburcqa, line 60: Mention that this option is only used by the IPC coupler.

genesis/engine/materials/rigid.py
duburcqa: Remove None. it is redundant with `need_coup`.

genesis/engine/materials/rigid.py
```diff
@@ -34,25 +34,32 @@ class Rigid(Material):
             Maximum resolution of the SDF grid. Must be >= sdf_min_res. Default is 128.
         gravity_compensation : float, optional
             Compensation factor for gravity. 1.0 cancels gravity. Default is 0.
-        coupling_type : str or None, optional
-            IPC coupling mode for this entity. Valid values:
-              - None: Entity not processed by IPC. Entity is completely ignored by IPC coupler.
+        collision_links : tuple of str or None, optional
```
duburcqa: Why did you introduce this option? I don't like it. It was never necessary so far, why it is suddenly useful?

genesis/engine/materials/rigid.py
ACMLCZH: `coup_links` is also used by other couplers. See `genesis/engine/entities/rigid_entity/rigid_entity.py` `needs_coup` in geom.

genesis/options/solvers.py
```diff
@@ -286,6 +286,10 @@ class IPCCouplerOptions(BaseCouplerOptions):
         For external_articulation with non-fixed base: whether base link is fully driven by IPC physics.
         When False, base link uses SoftTransformConstraint controlled by Genesis. When True, base link
         is fully driven by IPC physics. Defaults to False.
+    _show_ipc_gui : bool, optional
+        [Dev/debug] Enable the libuipc built-in polyscope GUI viewer for inspecting the IPC scene.
+        Not a public option — prefixed with ``_`` to indicate internal use.
```
duburcqa: `[Dev/debug]` + `_` is clear enough. Remove the second sentence.

tests/test_ipc.py
```diff
@@ -1491,3 +1583,59 @@ def test_cloth_uniform_biaxial_stretching(E, nu, strech_scale, n_envs, show_view
...
+        coupler_options=gs.options.IPCCouplerOptions(_show_ipc_gui=True),
+        show_viewer=False,
+    )
+    scene.add_entity(
+        gs.morphs.Box(size=(0.1, 0.1, 0.1), pos=(0, 0, 0.5)),
+        material=gs.materials.Rigid(coup_type="ipc_only"),
+    )
+    scene.build()
+    assert scene.sim is not None
+    assert scene.sim.coupler._ipc_gui is not None
```
duburcqa: You are not checking anything. If you cannot check that it works, remove this test.

genesis/engine/materials/rigid.py
ACMLCZH: `need_coup` is default to `True`, so we must also set a default `coup_type`. `ipc_only`, `two_way_soft_constraint` and `external_articulation`, which is the best default value?
But I actually think `need_coup=False` may be the best value for IPCCoupler.

genesis/engine/couplers/ipc_coupler/coupler.py
ACMLCZH: I think this should be under `if not meshes:`. What do you think?
And it is not place to process `external_articulation`?

duburcqa: I don't care where it is as long as it is still raising an exception. In this case, just remove this block entirely instead of adding continue.

genesis/engine/materials/rigid.py
duburcqa: I see. That's nice!

genesis/engine/materials/rigid.py
duburcqa: What about allowing `None` for `coup_type `, but use it to automatically select the best default depending on the type of entity? (fixed base robot, floating base robot, object).

Regarding `need_coup`, I'm fine with complexifying the logics and using `False` by default for IPC and `True` for other couplers. I think you have no other choice than also allowing `None` for this one if you want to do this.

tests/test_ipc.py
```diff
@@ -507,6 +524,73 @@ def test_single_joint(n_envs, coupling_type, joint_type, fixed, show_viewer):
...
+
+    if coup_type == "external_articulation":
+        art_data = coupler.articulation_data[robot]
+        assert len(art_data.articulation_slots_by_env) == 1
+        # All 7 revolute joints should be present (fixed joint is skipped)
+        assert len(art_data.joints_child_link) == 7
+
+    # Verify simulation runs without error
+    for _ in range(5):
+        scene.step()
```
duburcqa: I don't think this is useful. `build` already perform a single step. If something was broken for IPC, it would instantly fail. If you cannot check anything after stepping just remove these steps.

tests/test_ipc.py
```diff
@@ -1491,3 +1583,43 @@ def test_cloth_uniform_biaxial_stretching(E, nu, strech_scale, n_envs, show_view
...
+            coup_collision_links=("moving",),
+        ),
+    )
+    assert isinstance(robot, RigidEntity)
+    scene.build()
+    assert scene.sim is not None
+
+    # Verify the collision settings were applied
+    coupler = cast("IPCCoupler", scene.sim.coupler)
+    collision_settings = coupler._coupling_collision_settings.get(robot, {})
```
duburcqa: `collision_settings = coupler._coupling_collision_settings[robot]`

duburcqa (approved):

PR #2500, opened 2026-03-04
[Misc] Add dexterous hand benchmark.

hughperkins:


hughperkins:
cc @erizmr for info.

tests/test_rigid_benchmarks.py
```diff
@@ -762,6 +762,144 @@ def g1_fall(solver, n_envs, gjk, pytorch_profiler_step):
...
+        },
+    ]
+
+    coacd_opts = gs.options.misc.CoacdOptions(
+        threshold=0.05, max_convex_hull=20, preprocess_mode="auto",
+        preprocess_resolution=100, resolution=1000, mcts_nodes=20,
+        mcts_iterations=100, mcts_max_depth=3, pca=False, merge=True,
+        decimate=True, max_ch_vertex=256, extrude=False, extrude_margin=0.1,
+        apx_mode="ch", seed=0,
+    )
```
duburcqa: What is this?

tests/test_rigid_benchmarks.py
```diff
@@ -762,6 +762,144 @@ def g1_fall(solver, n_envs, gjk, pytorch_profiler_step):
...
+
+    scene = gs.Scene(
+        sim_options=gs.options.SimOptions(substeps=25, dt=step_dt, gravity=(0, 0, -9.81)),
```
duburcqa: All arguments on one line, and only what matters.

tests/test_rigid_benchmarks.py
```diff
@@ -762,6 +762,144 @@ def g1_fall(solver, n_envs, gjk, pytorch_profiler_step):
...
+        rigid_options=gs.options.RigidOptions(
+            enable_mujoco_compatibility=False,
```
duburcqa: Remove this option. it is confusing. `enable_mujoco_compatibility` is not really a user-facing option. We should probably make it private.

tests/test_rigid_benchmarks.py
```diff
@@ -762,6 +762,144 @@ def g1_fall(solver, n_envs, gjk, pytorch_profiler_step):
...
+    for cfg in hand_configs:
+        hand = scene.add_entity(gs.morphs.URDF(
+            file=str(shadow_hand_path / "shadow_hand" / cfg["urdf"]),
+            pos=cfg["pos"], quat=cfg["quat"], fixed=False,
+        ))
+        hands.append((hand, {name: cfg["dofs"][i] for i, name in enumerate(JOINT_NAMES)}))
+
+    scene.add_entity(gs.morphs.Mesh(
+        file=str(dex_path / "dex" / "table.glb"),
+        pos=(0.1, 0.0, 0.485403), euler=(0, 0, 90), fixed=True, coacd_options=coacd_opts,
```
duburcqa: one line.

tests/test_rigid_benchmarks.py
```diff
@@ -762,6 +762,144 @@ def g1_fall(solver, n_envs, gjk, pytorch_profiler_step):
...
+    ))
+    drill = scene.add_entity(gs.morphs.Mesh(
+        file=str(dex_path / "dex" / "drill_1.glb"),
+        pos=(0.15, 0.1, 0.87), euler=(90, 0, 225), fixed=False, coacd_options=coacd_opts,
```
duburcqa: one line.

tests/test_rigid_benchmarks.py
hughperkins: used for the drill and the table, per my undrsatnding.

tests/test_rigid_benchmarks.py
```diff
@@ -762,6 +762,144 @@ def g1_fall(solver, n_envs, gjk, pytorch_profiler_step):
...
+    time_start = time.time()
+    scene.build(n_envs=n_envs)
+    compile_time = time.time() - time_start
+
+    wrist_stiffness = 20
+    finger_force = 0.6
+    drill_stiffness = 20
+
+    for hand, default_dof in hands:
+        kp = torch.tensor([wrist_stiffness] * 6 + [40.0] * (hand.n_dofs - 6), dtype=torch.float32, device=gs.device)
```
duburcqa: Remove convert to torch.

tests/test_rigid_benchmarks.py
```diff
@@ -762,6 +762,144 @@ def g1_fall(solver, n_envs, gjk, pytorch_profiler_step):
...
+        hand.set_dofs_kp(kp)
+        hand.set_dofs_kv(2.0 * torch.sqrt(kp))
```
duburcqa: np.sqrt is just fine.

tests/test_rigid_benchmarks.py
```diff
@@ -762,6 +762,144 @@ def g1_fall(solver, n_envs, gjk, pytorch_profiler_step):
...
+            "quat": (-0.45215, -0.31265, 0.76087, 0.34480),
+            "dofs": LEFT_DOFS,
+            "urdf": "shadow_hand_left_woarm.urdf",
+        },
+        {
+            "pos": (0.16257, 0.24658, 1.28047),
+            "quat": (0.74525, 0.46466, -0.45186, -0.15660),
+            "dofs": RIGHT_DOFS,
+            "urdf": "shadow_hand_right_woarm.urdf",
+        },
```
duburcqa: Could you use `dict()` instead?

tests/test_rigid_benchmarks.py
```diff
@@ -762,6 +762,144 @@ def g1_fall(solver, n_envs, gjk, pytorch_profiler_step):
...
+        pos=(0.15, 0.1, 0.87), euler=(90, 0, 225), fixed=False, coacd_options=coacd_opts,
+    ))
+
+    time_start = time.time()
+    scene.build(n_envs=n_envs)
+    compile_time = time.time() - time_start
+
+    wrist_stiffness = 20
+    finger_force = 0.6
+    drill_stiffness = 20
```
duburcqa: Move these constants on top, in upper case.

tests/test_rigid_benchmarks.py
hughperkins: I'm not sure I follow. This is a standard way of writing dicts? What do you see as the advantages of using dict() instead?

tests/test_rigid_benchmarks.py
hughperkins: done

tests/test_rigid_benchmarks.py
```diff
@@ -762,6 +762,144 @@ def g1_fall(solver, n_envs, gjk, pytorch_profiler_step):
...
+        show_FPS=False,
+    )
+
+    hands = []
+    for cfg in hand_configs:
+        hand = scene.add_entity(gs.morphs.URDF(
+            file=str(shadow_hand_path / "shadow_hand" / cfg["urdf"]),
+            pos=cfg["pos"], quat=cfg["quat"], fixed=False,
+        ))
+        hands.append((hand, {name: cfg["dofs"][i] for i, name in enumerate(JOINT_NAMES)}))
```
duburcqa: This is useless. Just store the actual list on hands and do  a zip later. Moreover, computing `{name: cfg["dofs"][i] for i, name in enumerate(JOINT_NAMES)}`  is convoluted, and just makes things unclear.

Just do:

```python
    for hand, default_dof in zip(hands, (LEFT_DOFS, RIGHT_DOFS)):
        [...]
        hand.set_dofs_position(
            default_dof, dofs_idx_local=[hand.get_joint(name).dofs_idx_local[0] for name in JOINT_NAMES]
        )
```

tests/test_rigid_benchmarks.py
```diff
@@ -762,6 +762,144 @@ def g1_fall(solver, n_envs, gjk, pytorch_profiler_step):
...
+        kp = torch.tensor([wrist_stiffness] * 6 + [40.0] * (hand.n_dofs - 6), dtype=torch.float32, device=gs.device)
+        hand.set_dofs_kp(kp)
+        hand.set_dofs_kv(2.0 * torch.sqrt(kp))
+        hand.set_dofs_position(
+            tuple(default_dof.values()),
+            dofs_idx_local=[hand.get_joint(name).dofs_idx_local[0] for name in default_dof.keys()],
+        )
+        hand.control_dofs_position(torch.clamp(hand.get_dofs_position(), *hand.get_dofs_limit()))
+
+    finger_dofs = {hand: list(range(6, hand.n_dofs)) for hand, _ in hands}
```
duburcqa: `slice(6, hand.n_dofs)`

tests/test_rigid_benchmarks.py
```diff
@@ -762,6 +762,144 @@ def g1_fall(solver, n_envs, gjk, pytorch_profiler_step):
...
+    drill.set_dofs_kv(2.0 * torch.sqrt(drill_xy_kp), dofs_idx_local=[0, 1])
+    drill_xy_target = drill.get_dofs_position()[:, :2]
+
+    num_steps = 0
+    is_recording = False
+    qd.sync()
+    time_start = time.time()
+    while True:
+        for hand, _ in hands:
+            hand.control_dofs_position(base_xy_targets[hand], dofs_idx_local=[0, 1])
```
duburcqa: Move this outside the while loop.

tests/test_rigid_benchmarks.py
```diff
@@ -762,6 +762,144 @@ def g1_fall(solver, n_envs, gjk, pytorch_profiler_step):
...
+            random_forces[hand].uniform_(-finger_force, finger_force)
+            hand.control_dofs_force(random_forces[hand], dofs_idx_local=finger_dofs[hand])
+        drill.control_dofs_position(drill_xy_target, dofs_idx_local=[0, 1])
```
duburcqa: Same. Outside while loop.

tests/test_rigid_benchmarks.py
duburcqa: I need to understand why this is needed.

tests/test_rigid_benchmarks.py
duburcqa: Nothing. Just use option-like formatting for things that are options. It makes things clearer for me. it is immediately clear that all keys are arguments.

duburcqa: And it is basically the pattern that is used everywhere else in Genesis in such case.

tests/test_rigid_benchmarks.py
hughperkins: The table is not convex, nor the drill.

tests/test_rigid_benchmarks.py
hughperkins: tbh, I think I prefer dict() approach. I hate quotation marks. I *suspect* that I tried using it on a PR many years ago, and got made to turn it into {} sytnax 🙃 . Anyway, will change to dict() approach.

tests/test_rigid_benchmarks.py
duburcqa: So what? Not being convex does not require setting such options yourself, thanks god, that would be a nightmare.

hughperkins: removed

duburcqa: I'm not saying it should be removed either. You just need to motivate what all these parameters are doing. Typically, what is usually important is to specify:
```python
    coacd_opts = gs.options.misc.CoacdOptions(
        threshold=0.05, 
    )
```

All the other options are too deep and require to much expertise.

I would suggest that you check the resulting collision meshs with the original parameters, without any custom options, and with what I suggest. Then share screenshot here if you notice any difference. Because the collision geometries are critical for the benchmark to be representative, so just removing this block entirely is concerning.

hughperkins: Here we go:
Screenshot 2026-03-06 at 19 57 07
Screenshot 2026-03-06 at 19 57 20
Screenshot 2026-03-06 at 19 57 28

hughperkins: Screenshot 2026-03-06 at 19 59 18

hughperkins: Screenshot 2026-03-06 at 19 59 57
Screenshot 2026-03-06 at 20 00 03
Screenshot 2026-03-06 at 20 00 08
Screenshot 2026-03-06 at 20 00 14

duburcqa: Can you enable collision display? `vis_mode="collision"` at entity level

duburcqa: Otherwise it is hard to tell anything I'm afraid :/

hughperkins: I see. here you go:

Screenshot 2026-03-07 at 13 33 56
Screenshot 2026-03-07 at 13 34 07

duburcqa: I need to see the comparison with and without custom parameters to determine whether it is important to add them. The 3 variants:
* Default
* threshold only
* The entire set of parameters

hughperkins: with full parmaetrs:

Screenshot 2026-03-07 at 15 48 55
Screenshot 2026-03-07 at 15 49 08

hughperkins: Just threshold:

Screenshot 2026-03-07 at 15 52 00
Screenshot 2026-03-07 at 15 52 12

hughperkins: I feel like the one with no parameters is the best?

hughperkins: (and maybe I should PR that onto Yian's script, or ask him why he isnt using default params)

duburcqa: Thank you! Hard to define what is "the best". I tend to think that the default is sufficiently accurate when much less convex. The main objective here is to add a new benchmark that complements the existing ones to progressively give a comprehensive view of the current performance. If the perf stats that you are getting for the default decomposition are aligned with what you get with custom default, I would keep it simple and remove the custom decomposition, also because it is more likely to be consistent with what most users do.

duburcqa: In any case, it would be nice to add a comment pointing to this point in the benchmark code so that we remember convex decomposition parameters play a role in performance.

hughperkins: added link to this discussion into the code

PR #2503, opened 2026-03-05
[BUG FIX] Fix mass getter for PBD entities.

YilingQiao:
## Description

self.solver._kernel_get_mass(mass, envs_idx) should be self.solver._kernel_get_mass(self._particle_start, self._n_particles, mass, envs_idx)

### DIFF
--- a/genesis/engine/entities/particle_entity.py
+++ b/genesis/engine/entities/particle_entity.py
@@ -688,6 +688,7 @@ class ParticleEntity(Entity):
         """
         raise NotImplementedError
 
+    @gs.assert_built
     def get_mass(self, envs_idx=None):
         """
         Return the total mass of the entity.
@@ -695,7 +696,7 @@ class ParticleEntity(Entity):
         Parameters
         ----------
         envs_idx : None | int | array_like, shape (M,), optional
-            The indices of the environments to set. If None, all environments will be considered. Defaults to None.
+            The indices of the environments to query. If None, all environments will be considered. Defaults to None.
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


genesis/engine/entities/particle_entity.py
```diff
@@ -688,23 +688,24 @@ def get_particles_active(self, envs_idx=None):
...
 
         Returns
         -------
         mass : torch.Tensor, shape (M,)
             The computed total mass.
         """
         envs_idx = self._scene._sanitize_envs_idx(envs_idx)
-        mass = torch.empty((len(envs_idx),), dtype=gs.tc_float, device=gs.device)
-        self.solver._kernel_get_mass(mass, envs_idx)
+        mass = torch.zeros((len(envs_idx),), dtype=gs.tc_float, device=gs.device)
```
YilingQiao, line 707: torch.empty(...) here should be torch.zeros(...) since the SPH solver's kernel uses mass[i_b_] += which accumulates

tests/test_pbd.py
```diff
@@ -152,6 +152,34 @@ def test_cloth_attach_fixed_point(n_envs, material_type, show_viewer, tol):
...
+            scale=1.0,
+        ),
+        material=gs.materials.PBD.Cloth(),
+    )
+    scene.build(n_envs=n_envs)
+
+    mass = cloth.get_mass()
+    expected_n = max(n_envs, 1)
+    assert mass.shape == (expected_n,), f"Expected shape ({expected_n},), got {mass.shape}"
+    assert (mass > 0).all(), f"Expected positive mass, got {mass}"
```
duburcqa, line 180: Not a great check but it is better than nothing.

genesis/engine/entities/particle_entity.py
duburcqa: Adding a comment to clarify this is the code would have been nice.

PR #2504, opened 2026-03-05
[FEATURE] Add avatar (kinematic) entity and refactor rigid entity.

duburcqa:


genesis/engine/states/solvers.py
```diff
@@ -84,6 +84,43 @@ def s_global(self):
...
+        self.scene = scene
+        self._s_global = s_global
+
+        _B = scene.sim.kinematic_solver._B
+        args = {
+            "dtype": gs.tc_float,
+            "requires_grad": scene.requires_grad,
+            "scene": self.scene,
+        }
+        self.qpos = gs.zeros((_B, scene.sim.rigid_solver.n_qs), **args)
```
YilingQiao: you are using scene.sim.rigid_solver.* instead of scene.sim.kinematic_solver.* is this expected?

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -0,0 +1,1167 @@
...
+        self._queried_states.clear()
+
+    # ------------------------------------------------------------------------------------
+    # ----------------------------------- render -----------------------------------------
+    # ------------------------------------------------------------------------------------
+
+    def update_geoms_render_T(self):
+        kernel_update_geoms_render_T(
+            self._geoms_render_T,
+            geoms_state=self.geoms_state,
```
YilingQiao: self.geoms_state and self._geoms_render_T are only defined on RigidSolver, not KinematicSolver.

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -0,0 +1,1167 @@
...
+        This method is called after _init_link_fields to update the per-environment inertial properties
+        for entities with heterogeneous morphs.
+        """
+        for entity in self._entities:
+            # Skip non-heterogeneous entities
+            if not entity._enable_heterogeneous:
+                continue
+
+            # Get the number of variants for this entity
+            n_variants = len(entity.variants_geom_start)
```
YilingQiao: entity.variants_geom_start which doesn't exist on KinematicEntity. Maybe it's fine, if we won't use heterogeneous KinematicEntity.

Kashu7100: I think heterogeneous KinematicEntity is desirable for manipulation tasks

duburcqa: Good remark, this is a bug.

genesis/engine/solvers/kinematic_solver.py
duburcqa: Good catch!

genesis/engine/states/solvers.py
duburcqa: No!

PR #2518, opened 2026-03-09
[MISC] Fix Python-scope field read to avoid GPU-CPU sync.

erizmr:


genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -31,7 +31,9 @@ def __init__(self, rigid_solver: "RigidSolver"):
         self._para_level = rigid_solver._para_level
 
         self._solver_type = rigid_solver._options.constraint_solver
-        self.iterations = rigid_solver._options.iterations
+        self._n_iterations = (
+            rigid_solver._options.iterations
```
duburcqa, line 35: Cast to int just in case

erizmr:
i think the comments are related to https://github.com/Genesis-Embodied-AI/quadrants/pull/404? but not this PR?

hughperkins:
yup

hughperkins:
Seems like some serious regressions on this PR, in the benchmark report?

erizmr:
yes, @duburcqa pointed out this is induced by the int argument caching https://github.com/Genesis-Embodied-AI/quadrants/pull/404

hughperkins:
> yes, @duburcqa pointed out this is induced by the int argument caching [Genesis-Embodied-AI/quadrants#404](https://github.com/Genesis-Embodied-AI/quadrants/pull/404)

Oh nice! 🙌

duburcqa:
@erizmr @hughperkins After running some manual benchmark locally, I can confirm that:
* I can reproduce the performance regression locally using the latest release of Quadrants from PyPI
* I do not observe any performance regression anymore with the latest main of Quadrants.

erizmr:
Update to use the Quadrants pre-release 0.4.4b1

hughperkins:
Review from opus 4.6:

### Me: comment on the changes in this branch erizmr:mingrui/260309/solver_opt_cpu_gpu_sync since it was forked from origin/main

  The branch is tightly focused — 4 files, 11 insertions, 4 deletions. There are two logical changes:


  1. CPU-GPU Sync Avoidance (the main optimization)

  In solver_breakdown.py, func_solve_decomposed is a Python-scope function (not a GPU kernel) that orchestrates multiple kernel launches per iteration. The loop was:

  iterations = rigid_global_info.iterations[None]
  for _it in range(iterations):

  Accessing rigid_global_info.iterations[None] reads a GPU field from Python, which forces a full CPU-GPU synchronization — the CPU blocks until all pending GPU work
  drains. This happens every solver step, defeating the async kernel-launch pipeline that makes the decomposed solver fast.
  The fix caches the value as self._n_iterations = int(rigid_solver._options.iterations) at init time and threads it through as a plain Python int parameter. This is
  correct and clean.
  Note that func_solve_body_monolith (line 3037) also reads rigid_global_info.iterations[None], but that's fine — it's inside a @qd.kernel, so the read happens on the GPU
   side with no sync penalty. The branch correctly adds the _n_iterations parameter to its signature for consistency but doesn't need to use it there.
  One thing to watch: this assumes iterations is constant for the lifetime of the solver. No external code accesses constraint_solver.iterations, and the island solver
  (solver_island.py) maintains its own separate self.iterations on a different class. So the rename from self.iterations to self._n_iterations is safe.


  2. `quadrants` bump from 0.4.3 to 0.4.4

  Straightforward dependency update. Likely needed for the branch to work (or just bundled in).

  ────────────────────────────────────────

  Overall: This is a clean, well-scoped optimization. The CPU-GPU sync in a Python-scope loop is a real performance problem and the fix is the standard approach. The test
   update in test_grad.py correctly propagates the new parameter to the monkeypatched call site. No concerns.

### Me:   what si the 'golden source of truth' for iterations value?

  The golden source of truth is RigidOptions.iterations in genesis/options/solvers.py (default: 50). From there it flows to two places:
  1. `rigid_global_info.iterations` — a GPU-side scalar field, created in genesis/utils/array_class.py line 166 via V_SCALAR_FROM(dtype=gs.qd_int,
     value=solver._options.iterations). This lives on the GPU and is what func_solve_body_monolith and func_solve_decomposed were reading with
     rigid_global_info.iterations[None].
  2. `constraint_solver.iterations` (or ._n_iterations on the branch) — a Python-side copy, set at init time from rigid_solver._options.iterations. The island solver
     similarly stores its own copy at solver_island.py line 28.

  Both are initialized once from the same source (rigid_solver._options.iterations) and neither is ever mutated afterward. The backward.py path reads from yet another
  copy — static_rigid_sim_config.iterations — which is a template/compile-time constant also sourced from the same options.
  So RigidOptions.iterations is the single source, and these are all read-only copies propagated at init time. The branch's optimization of caching a Python int is safe
  because the value never changes.

PR #2519, opened 2026-03-09
[MISC] Make torch profiling per test, rather than per process.

hughperkins:
## Description

- make torch profiling per test, rather than per process, so we can run multiple benchmarks in parallel, with profiling activated, and get profiles from all the benchmarks
- also add a qd.sync() at each profiling step, to ensure we dont have 'bleed through' between different steps.

## Related Issue

## Checklist:

### DIFF
--- a/tests/profiling.py
+++ b/tests/profiling.py
@@ -107,7 +107,7 @@ def pytorch_profiler(pytestconfig, request):
             f.write(f"=== Kernel Summary (sorted by {sort_by}) ===\n")
             f.write(table + "\n")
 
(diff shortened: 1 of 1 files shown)


tests/profiling.py
duburcqa, line 1: I'm fine on principle, but I don't like the idea of hijacking stderr to print stuffs that are not errors.

hughperkins: addressed in f964daf0

hughperkins:
Thanks!

PR #2523, opened 2026-03-09
[FEATURE] Add parallel linesearch for constraint solver.

erizmr:


erizmr:
> **🔴 Benchmark Regression Detected ➡️ [Report](https://github.com/Genesis-Embodied-AI/Genesis/runs/67022914613)**

perf dispatch re-benchmarking induces this, updated to re-run

hughperkins:
> > **🔴 Benchmark Regression Detected ➡️ [Report](https://github.com/Genesis-Embodied-AI/Genesis/runs/67022914613)**
> 
> perf dispatch re-benchmarking induces this, updated to re-run

I'd like to see some simple standalone reproduction of how you feel perf dispatch repeat induces high latency, ideally.

hughperkins:
> > > **🔴 Benchmark Regression Detected ➡️ [Report](https://github.com/Genesis-Embodied-AI/Genesis/runs/67022914613)**
> > 
> > 
> > perf dispatch re-benchmarking induces this, updated to re-run
> 
> I'd like to see some simple standalone reproduction of how you feel perf dispatch repeat induces high latency, ideally.

This would have a few benefits:
- easier to discuss and reason about
- easier to fix (if it is a perf dispatch bug)

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
```diff
@@ -228,9 +843,83 @@ def func_solve_decomposed(
...
+    constraint_state,
+    rigid_global_info,
+    static_rigid_sim_config,
+    _n_iterations,
+):
+    """Decomposed solver with sequential iterative linesearch.
+
+    Same kernel-per-step structure as func_solve_decomposed but uses the sequential
+    Newton-guided linesearch (1 thread per env) instead of the parallel grid search
+    (K threads per env). Better for scenes with many constraints per env where the
```
hughperkins: what is your reasoning for why many constraints per environment would make using a single thread per environment faster than using K threads per environment?

erizmr: No. This sentence is wrong. Thanks for catching. I will remove it. And also as per our discussion, we would keep the code simpler, i.e., remove this fallback option.

erizmr: Fixed in bd18461d2188dd8c8ac1ec267f26a97dd387578c.

hughperkins: Yeah, I agree with making the code simpler (though I might have to put this code back in the future potentially 😅  Depending on results on other benchmarks, and once we have migrated to cuda graph, so decomp is no longer slow).

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
```diff
@@ -1,62 +1,634 @@
...
+def _kernel_parallel_linesearch_p0(
+    dofs_info: array_class.DofsInfo,
     entities_info: array_class.EntitiesInfo,
     dofs_state: array_class.DofsState,
     constraint_state: array_class.ConstraintState,
     rigid_global_info: array_class.RigidGlobalInfo,
-    static_rigid_sim_config: ti.template(),
+    static_rigid_sim_config: qd.template(),
 ):
+    """Fused mv + jv + snorm + quad_gauss + eq_sum + p0_cost.
```
hughperkins: 1. This appears to be the main parallel line search entry point, is that right?

Can we add a title line to this doc string to make this obvious? (I searched the entire file looking for the entry point, before deciding it's probably this function).

2. Per our earlier discussion the algo is:
- run one pass of parallel linesearch, with [name of config var] candidates, logarithmically spaced from alpha = [name of config var] to 1
- calculate first-order gradient
- if gradient < threshold == 1e-4 (?) then we are done: finish
- else: run a newton step, and finish

Could you document this algo in an appropraite place please? (e.g. in this doc string).

Also, I looked through the function trying to find where it calculates the gradient, and compares with threshold, and it wasn't obvious to me. Maybe add a one line comment just in front of each of the 4 events listed above?

hughperkins: Note: personally I like to break functions into sub functions, with descriptive names. I find it easier to read.

BUT I think that @duburcqa is comfortable with much longer functions.

So I'm just stating this preference, but I will defer to @duburcqa 's point of view on this.

hughperkins:
Benchmarks look good to me.

hughperkins:
Wait, the old code was simpler, and giving a benefit in the benchmarks. Do we need the more complicated code in fact?

erizmr:
> Wait, the old code was simpler, and giving a benefit in the benchmarks. Do we need the more complicated code in fact?

The previous version misses the newton correction when porting from the dev branch, now the updated one is the full version, which shows better performance than previous version as expected, shown in https://github.com/Genesis-Embodied-AI/Genesis/runs/68644518864.

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
erizmr: There is a short description in docstring in _kernel_parallel_linesearch_eval, but can add more comments.

hughperkins:
Per Opus there are three dead functions:

```
  Decomposed (func_solve_decomposed) uses, in a Python for-loop:
  • _kernel_parallel_linesearch_p0 — fused mv + jv + snorm + quad_gauss + eq_sum + p0_cost
  • _kernel_parallel_linesearch_eval — grid search + bisection + apply alpha

  Monolith (func_solve_body_monolith in solver.py) uses:
  • solver.func_linesearch_and_apply_alpha — sequential Newton-guided linesearch

  Unused:
  • _kernel_parallel_linesearch_mv
  • _kernel_parallel_linesearch_jv
  • _kernel_linesearch
```

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -366,6 +366,10 @@ def _build_static_config(self):
             sparse_solve=self._options.sparse_solve,
             integrator=self._integrator,
             solver_type=self._options.constraint_solver,
+            prefer_parallel_linesearch=(
+                0 if self._enable_mujoco_compatibility
+                else {None: -1, False: 0, True: 1}[self._options.prefer_parallel_linesearch]
+            ),
```
duburcqa: This should be implemented directly in option class.

tests/conftest.py
```diff
@@ -682,6 +682,22 @@ def initialize_genesis(request, monkeypatch, tmp_path, backend, precision, perfo
...
+        _orig_init = RigidOptions.__init__
+
+        def _patched_init(self, **data):
+            if data.get("prefer_parallel_linesearch") is None:
+                # Only enable parallel linesearch on CUDA; other GPU archs
+                # (e.g. Metal) fall back to iterative since decomp uses shared memory.
+                data["prefer_parallel_linesearch"] = gs.backend == gs.cuda
+            _orig_init(self, **data)
+
+        monkeypatch.setattr(RigidOptions, "__init__", _patched_init)
```
duburcqa: I would advise to rather monkeypatch `_model_post_init`. Less invasive.

erizmr:
> Per Opus there are three dead functions:
> 
> ```
>   Decomposed (func_solve_decomposed) uses, in a Python for-loop:
>   • _kernel_parallel_linesearch_p0 — fused mv + jv + snorm + quad_gauss + eq_sum + p0_cost
>   • _kernel_parallel_linesearch_eval — grid search + bisection + apply alpha
> 
>   Monolith (func_solve_body_monolith in solver.py) uses:
>   • solver.func_linesearch_and_apply_alpha — sequential Newton-guided linesearch
> 
>   Unused:
>   • _kernel_parallel_linesearch_mv
>   • _kernel_parallel_linesearch_jv
>   • _kernel_linesearch
> ```

Unused functions removed.

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -3117,10 +3119,13 @@ def func_solve_body(
 ) -> None: ...
 
 
-@func_solve_body.register(is_compatible=lambda *args, **kwargs: True)
+@func_solve_body.register(
+    is_compatible=lambda *args, **kwargs: _get_static_config(*args, **kwargs).prefer_parallel_linesearch != 1
```
duburcqa, line 3140: `prefer_parallel_linesearch != True`

genesis/options/solvers.py
```diff
@@ -526,6 +532,21 @@ def __init__(self, *, contact_resolve_time: float | None = None, **data):
...
+                raise ValueError(
+                    "prefer_parallel_linesearch=True is incompatible with enable_mujoco_compatibility=True. "
+                    "Mujoco compatibility requires the iterative (Newton-guided) linesearch."
+                )
+            self.prefer_parallel_linesearch = False
+
+    @property
+    def prefer_parallel_linesearch_int(self) -> int:
+        """Map prefer_parallel_linesearch to int for static config: None → -1, False → 0, True → 1."""
+        return {None: -1, False: 0, True: 1}[self.prefer_parallel_linesearch]
```
duburcqa: Why a 3-state int ?

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -366,6 +366,7 @@ def _build_static_config(self):
             sparse_solve=self._options.sparse_solve,
             integrator=self._integrator,
             solver_type=self._options.constraint_solver,
+            prefer_parallel_linesearch=self._options.prefer_parallel_linesearch_int,
```
duburcqa: Could you convert to 3-state int here rather than defining the property `prefer_parallel_linesearch_int` ? I feel like it is an implementation details that has nothing to do in public API of the option class.

genesis/options/solvers.py
duburcqa: Ok now I get it. I'm fine with it.

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
duburcqa: I'm fine.

genesis/engine/solvers/rigid/constraint/solver.py
duburcqa: Ignore this.

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
hughperkins: @duburcqa just to confirm, you are saying you are fine with the long function right? Or are you saying, you feel existing documentation is suffiicent?

genesis/engine/solvers/rigid/rigid_solver.py
erizmr: Fixed in 6e7a8e005d8409ec778573d2c7665787f323ab15

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
duburcqa: I'm fine with the current state of the PR, which means long functions.

tests/conftest.py
```diff
@@ -682,6 +682,22 @@ def initialize_genesis(request, monkeypatch, tmp_path, backend, precision, perfo
...
+        from genesis.options.solvers import RigidOptions
+
+        _orig_model_post_init = RigidOptions.model_post_init
+
+        def _patched_model_post_init(self, __context):
+            _orig_model_post_init(self, __context)
+            if self.prefer_parallel_linesearch is None:
+                # Only enable parallel linesearch on CUDA; other GPU archs
+                # (e.g. Metal) fall back to iterative since decomp uses shared memory.
+                self.prefer_parallel_linesearch = gs.backend == gs.cuda
```
duburcqa: Metal does support shared memory

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
hughperkins: Cool. @erizmr please could you add the documentation I requested above, ie:

1. This appears to be the main parallel line search entry point, is that right?

Can we add a title line to this doc string to make this obvious? (I searched the entire file looking for the entry point, before deciding it's probably this function).

2. Per our earlier discussion the algo is:
- run one pass of parallel linesearch, with [name of config var] candidates, logarithmically spaced from alpha = [name of config var] to 1
- calculate first-order gradient
- if gradient < threshold == 1e-4 (?) then we are done: finish
- else: run a newton step, and finish

Could you document this algo in an appropraite place please? (e.g. in this doc string).

Also, I looked through the function trying to find where it calculates the gradient, and compares with threshold, and it wasn't obvious to me. Maybe add a one line comment just in front of each of the 4 events listed above?

erizmr:
> Please move the doc for linesearch under the linesearch entrypoint function.

Please check https://github.com/Genesis-Embodied-AI/Genesis/pull/2523/changes/550de9780e193c42c623bc372a2cad705675a30c

tests/conftest.py
erizmr: Disabled metal backend for this algorithm temporally, will support after quadrants support it, see bbf3ceaee5115797274b5c330d08efc5e4dadf2a.

genesis/options/solvers.py
```diff
@@ -526,6 +532,16 @@ def __init__(self, *, contact_resolve_time: float | None = None, **data):
...
 
+    def model_post_init(self, __context):
+        super().model_post_init(__context)
+        if self.enable_mujoco_compatibility:
+            if self.prefer_parallel_linesearch is True:
+                raise ValueError(
+                    "prefer_parallel_linesearch=True is incompatible with enable_mujoco_compatibility=True. "
+                    "Mujoco compatibility requires the iterative (Newton-guided) linesearch."
+                )
+            self.prefer_parallel_linesearch = False
```
duburcqa: You should add a check for compatible backends.

erizmr: Do you means excluding `metal` here for parallel linesearch? The iterative linesearch should work for metal backend as per our previous tests. The parallel linesearch is excluded by `enable_mujoco_compatibility` here anyway, do we still the check for compatible backends?

tests/conftest.py
```diff
@@ -682,6 +682,21 @@ def initialize_genesis(request, monkeypatch, tmp_path, backend, precision, perfo
...
+
+        _orig_model_post_init = RigidOptions.model_post_init
+
+        def _patched_model_post_init(self, __context):
+            _orig_model_post_init(self, __context)
+            if self.prefer_parallel_linesearch is None:
+                # FIXME: Also enable for gs.metal once Quadrants supports atomics on Apple Metal.
+                self.prefer_parallel_linesearch = gs.backend == gs.cuda
+
+        monkeypatch.setattr(RigidOptions, "model_post_init", _patched_model_post_init)
```
duburcqa: Patching must be guarded by 'benchmarks' marker.

duburcqa: Pass `config: pytest.Config` as input argument of the fixture, then do:

```python
    expr = Expression.compile(config.option.markexpr)
    is_benchmarks = expr.evaluate(MarkMatcher.from_markers((pytest.mark.benchmarks,)))
```

genesis/options/solvers.py
duburcqa: > Do you means excluding metal here for parallel linesearch?

Yes.

tests/conftest.py
erizmr: Fixed in b11bac498086c7d579e7943a1fcd40502ff919a8

PR #2525, opened 2026-03-09
[BUG FIX] More robust collision detection on 32bits precision.

SonSang:
## Description

Previously GJK algorithm could not detect collision between two spheres in the script below.

The main reason was in maximum reprojection error in EPA algorithm, which was also a problem of #2419 . In #2419 we increased the error bound to solve this issue, but it happened again in this case, even with the increased error bound.

In this PR, rather than increasing the error bound again, we fix this issue by introducing another `absolute` error bound for this reprojection error. The existing error bound is `relative`, and when it comes to smooth geometries like sphere or capsule, this relative error could be large even though the absolute error is pretty small (e.g. `1e-7`). Therefore, in this PR we change to ignore the contact only when both the absolute and relative error goes over the error bound.

```
#!/usr/bin/env python3
"""
Reproduce the diagonal_3d sphere-sphere GJK miss.

Two spheres: A at origin (r=0.10), B at (0.08, 0.06, 0.06) (r=0.08).
Distance = sqrt(0.08² + 0.06² + 0.06²) ≈ 0.1166, combined_r = 0.18.
Expected penetration ≈ 0.063.

Usage:
    python repro_diagonal_3d.py          # headless
    python repro_diagonal_3d.py -v       # with viewer
    python repro_diagonal_3d.py --gjk    # use GJK instead of MPR
"""

import argparse
import xml.etree.ElementTree as ET
from pathlib import Path
import tempfile
import numpy as np
import genesis as gs


def create_sphere_mjcf(name, radius):
    mjcf = ET.Element("mujoco", model=name)
    ET.SubElement(mjcf, "compiler", angle="degree")
    ET.SubElement(mjcf, "option", timestep="0.01", gravity="0 0 0")
    worldbody = ET.SubElement(mjcf, "worldbody")
    body = ET.SubElement(worldbody, "body", name=name, pos="0 0 0")
    ET.SubElement(body, "geom", type="sphere", size=f"{radius}")
    ET.SubElement(body, "joint", name=f"{name}_joint", type="free")
    return mjcf


def main():
    parser = argparse.ArgumentParser()
    parser.add_argument("-v", action="store_true", help="Show viewer")
    parser.add_argument("-c", action="store_true", help="Use CPU backend")
    parser.add_argument("--gjk", action="store_true", help="Use GJK instead of MPR")
    args = parser.parse_args()

    pos_a = (0, 0, 0)
    pos_b = (0.08, 0.06, 0.06)
    r_a = 0.10
    r_b = 0.08
    combined_r = r_a + r_b
    dist = np.linalg.norm(np.array(pos_a) - np.array(pos_b))

    label = "GJK" if args.gjk else "MPR"
    print(f"pos_a={pos_a}  r_a={r_a}")
    print(f"pos_b={pos_b}  r_b={r_b}")
    print(f"dist={dist:.6f}  combined_r={combined_r}")
    print(f"expected_pen={combined_r - dist:.6f}")
    print(f"method={label}")
    print()

    backend = gs.cpu if args.c else gs.gpu
    gs.init(backend=backend, precision="32", seed=0, logging_level="INFO")

    with tempfile.TemporaryDirectory() as tmp_dir:
        tmp = Path(tmp_dir)

        scene = gs.Scene(
            show_viewer=args.v,
                use_gjk_collision=args.gjk,
                gravity=(0, 0, 0),
            ),
        )

        ma = create_sphere_mjcf("sphere_a", r_a)
        pa = tmp / "sa.xml"
        ET.ElementTree(ma).write(pa)
        ea = scene.add_entity(gs.morphs.MJCF(file=pa))

        mb = create_sphere_mjcf("sphere_b", r_b)
        pb = tmp / "sb.xml"
        ET.ElementTree(mb).write(pb)
        eb = scene.add_entity(gs.morphs.MJCF(file=pb))

        scene.build()

        quat = [1, 0, 0, 0]
        ea.set_qpos((*pos_a, *quat))
        ea.zero_all_dofs_velocity()
        eb.set_qpos((*pos_b, *quat))
        eb.zero_all_dofs_velocity()

        scene.step()
        contacts = scene.rigid_solver.collider.get_contacts(as_tensor=False, to_torch=False)
        n_contacts = len(contacts["geom_a"])

        print(f"--- {label} ---")
        print(f"  contacts: {n_contacts}")
        if n_contacts > 0:
            pen = contacts["penetration"][0]
            nrm = np.array(contacts["normal"][0])
            pos = np.array(contacts["position"][0])
            print(f"  penetration: {pen:.6f}")
            print(f"  normal: ({nrm[0]:.4f}, {nrm[1]:.4f}, {nrm[2]:.4f})")
            print(f"  position: ({pos[0]:.4f}, {pos[1]:.4f}, {pos[2]:.4f})")
        else:
            print("  NO COLLISION DETECTED")

        if args.v:
            print("\nViewer open. Close window to exit.")
            while True:
                scene.step()


if __name__ == "__main__":
    main()
```

## Related Issue

## Motivation and Context

GJK did not detect collision for a corner case.

The script provided above now runs without error -- GJK detects collision.

## Checklist:

### DIFF
--- a/genesis/engine/solvers/rigid/collider/gjk.py
+++ b/genesis/engine/solvers/rigid/collider/gjk.py
@@ -96,7 +96,7 @@ class GJK:
             # tower, ...). We observed the error usually reaches around 5e-4, so we set the threshold to 1e-4 to be
             # safe. However, this value could be further tuned based on the future examples.
             polytope_max_rel_reprojection_error=1e-4,
-            polytope_max_abs_reprojection_error=1e-6,
+            polytope_max_abs_reprojection_error=1e-4,
             # The values are matching MuJoCo for compatibility. Increasing this value could be useful for detecting
             # contact manifolds even when the normals are not perfectly aligned, but we observed that it leads to more
             # false positives and thus not a perfect solution for the multi-contact detection.
--- a/tests/test_rigid_physics_analytical_vs_gjk.py
+++ b/tests/test_rigid_physics_analytical_vs_gjk.py
@@ -1,33 +1,31 @@
 """
 Unit test comparing analytical capsule-capsule contact detection with GJK.
 
-This test creates a modified version of narrowphase.py in a temporary file that
-forces capsule-capsule and sphere-capsule collisions to use GJK instead of
-analytical methods, allowing direct comparison between the two approaches.
+This test creates a modified version of narrowphase.py in a temporary file that forces capsule-capsule and
+sphere-capsule collisions to use GJK instead of analytical methods, allowing direct comparison between the two
+approaches.
 
 # errno
 
-We abuse errno in this test, because it is considerably easier, and needs much less code, than
-attempting to add a new tensor into one of the existing structures, and have that work for both
-ndarray and field, via monkey-patching.
+We abuse errno in this test, because it is considerably easier, and needs much less code, than attempting to add a
+new tensor into one of the existing structures, and have that work for both ndarray and field, via monkey-patching.
 
-errno is NOT designed for how we use it. Nevertheless with a couple of reasonable-ish assumptions
-we can work with it.
+errno is NOT designed for how we use it. Nevertheless with a couple of reasonable-ish assumptions we can work with it.
 
-Assumption 1: when code runs normally and correctly, nothing in Genesis production code (not including
-test code) will ever set bit 16 of errno to any value except 0.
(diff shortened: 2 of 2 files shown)


duburcqa:
Your PR is great, but it would be nice to write a unit test out of you reproduction script. I will do it myself if I have time.

Issue #2400, opened 2026-02-13
[Feature]: Performance of convex_vs_convex kernel?


gpu-daal:
### Feature Description

Kernel "func_narrow_phase_convex_vs_convex"  is parallelized over the number of environments, but why not over the number of collision pairs in total? 
If I have a training algorithm and have 2048 environments, that's only 2 SM blocks in use out of 50+. Why such restrictive GPU utilization? If I have 100 collision pairs per environment with 2048 environments, those 2048 threads would do 100 loop iterations, which is not ideal for weak GPU threads, and puts more pressure into registers.

### Release version or Commit ID

Commit bc40886

duburcqa:
This specific issue is already a work in progress. I cannot share any information at the moment. But I can tell you it is not easy to speed but by parallelising over collision pairs because GJK is memory-bound already.

duburcqa:
Closing for now, since we are already tracking this internally.

duburcqa:
Fixed by https://github.com/Genesis-Embodied-AI/Genesis/issues/2400!

PR #2527, opened 2026-03-10
[MISC] Speed up rigid collision detection on GPU for generic convex meshes.

hughperkins:
### Challenges we want to solve

- not enough threads used for narrowphase kernel, when the number of environemtns is less than the number of gpu cores
- gjk memory linear in nuber of environemtns, since gjk threads locked to number of environments
- thread divergence since different threads in each warp have to handle:
    - mpr
    - gjk
    - no contact (braodphase output a contact, but rapidlly discoverer, during mpr or gjk, that wasnt a real contact)
- cannot increase number of threads for mpr (which is compute bound) without incraseing allocation of gjk memory
- cannot sequence mpr and gjk, since mpr is compute bound, and gjk is memory i/o bound, and we wanat them running in parallel, to hide each other

### Solution
```
- split narrowphase into two kernels:
    - kernel 1 does NOT do any EPA, only mpr and initial gjk contact detection
          - classifies each contact into:
                - not a real contact
                - mpr contact
                - gjk contact
      - kernel 2 handles:
           - for mpr: multicontacts 1-4, and enqueing all contacts (including contact 0)
           - for gjjk: all contacts, 0-4, and enqueing them
```

Note: this is ONLY for GPU. CPU continues to use the old narrowphase.

### What this solution enables

- completley free to parametrize number of gjk threads as one wishes (parameterization not implemented in this PR - it uses a fixed heuristic for now - but would trivial to vary the number of gjk trheads to low, to save memory, or high, to go faster)
- gjk memory usage indepdenent of number of environemtns
- cores on gpu fully utilized by narrowphase
- uses many more threads for mpr than before
     - and, similar to gjk, trivial to parametrize as one wants (parametrization not impelmetned in this PR - it uses a fixed heuristic for now - but would be trivial to expose this parameization to the user)

### Good points about this solution, compared to some other solutions considered

- does not need streams 🙌
    - not that we couldn't use streams, or that streams are bad, but we don't yet have streams, and using streams immediately would massively increase the risk
    - we can consider migrating/experimenting with streams in the future (might reduce register pressure => better occupancy)
 - continues to parallelize gjk and mpr threads (despite no streams)
 - can be used for both use_gjk = False, and use_gjk = True
 - does not need perf_dispatch
      - not that perf_dispatch is necessarily bad, but it has some downsides which are nice to be able to avoid
           - no wasted cycles from trying a sub-optimal kernel sometimes
           - no complications during profiling /benchmarking / bugfixing when a bug has made the optimal kernel abnormally slow, and so the sub-optimal kernel is chosen instead, giving weird benchmarking results

### Challenges faced by the solution
```
- solver becmae really slow
     - this was actually caused by multiple different issues
            - initially, we didn't allow upgrades of multicontacts 1-4 from mpr to gjk
                  - there are pathological cases, notably with box pyramid, where a benign contact0 penetration, eg 0.000032, becomes 0.25 (entire box), during a perturbation
                        - gjk upgade handles this
                        - without upgrade, solver chokes on these garbage inputs
                        - fixed by allowing mpr to gjk upgrade in kernel 2
            - solver runs MUCH faster when inputs are sorted approximately in the order that braodphase does, i.e. spatialaly
                  - and the threading structure of kernel 1 and 2 completely trashes the sort order
                  - fixed by adding a sort kernel at the end of narrowphase
                       - (this ahs the bonus that will likely solve the same problem that a broadphase kernle I have was facing)
solver becoming slow was really the main challenge. Other stuff was simply experimentation with an appropriate heuristic for choosing the number of threads for gjk vs mpr in kernel 2, and for contact 0 detection in kernel 1.
```

## Details

- split mpr gjk narrowphase kernel into a few kernels, in order to accelerate narrowphase on gpu:
```
- func_reset_narrowphase_work_queues:
    - clears the mpr/gjk work queues

- kernel1:
     - if use_mpr:
           - run mpr contact0
                - if contact:
                      - if penetration < threshold (assuming warm start available):
                            - add to mpr work queue
                            - note: does NOT add to contacts
                      - else:
                            - add to gjk work queue
     elif use_gjk:
           - run gjk contact0, WITHOUT EPA
           - if contact:
                - add to gjk work queue

- kernel2:
    - for thread in total_threads:
         if thread_id < num_gjk_threads:
              - process items from the gjk work queue
                  - all multicontacts handled by single thread
                  - enqueues all contacts in a single atomic batch
                  - handles detecting dupe perturbations locally
         else:
              - process items from the mpr work queue
                  - all multicontacts handled by single thread
                  - if gjk upgrade needed:
                       - add to end of gjk work queue
                  - enqueues all contacts in a single atomic batch
                  - handles detecting dupe perturbations locally

- func_prepare_gjk_rerun:
    - move gjk work queue pointers to handle the new gjk work queue elements

- kernel2 (again):
    for thread in gjk threads:
        - process new items in gjk work queue

- func_sort_contacts:
    - sort contacts by contact x-position (in order to somewhat recoever sweep sequence from broadphase)
```

Other changes:
- box-box and box-plane no longer go through convex-convex (or kernel1)

Resolves https://github.com/Genesis-Embodied-AI/Genesis/issues/2400

### DIFF
--- a/genesis/utils/array_class.py
+++ b/genesis/utils/array_class.py
@@ -691,7 +691,7 @@ def get_collider_state(
         contact_cache=get_contact_cache(solver, n_possible_pairs),
         broad_collision_pairs=V_VEC(2, dtype=gs.qd_int, shape=(max(max_collision_pairs_broad, 1), _B)),
         contact_data=get_contact_data(solver, max_contact_pairs, requires_grad),
-        diff_contact_input=get_diff_contact_input(solver, max(max_contact_pairs, 1), is_active=True),
+        diff_contact_input=get_diff_contact_input(_B, max(max_contact_pairs, 1), True, requires_grad),
         narrowphase_work_queues=get_narrowphase_work_queues(
             max(max_collision_pairs_broad * _B, 1) if collider_static_config.has_non_box_plane_convex_convex else 1
         ),


genesis/engine/solvers/rigid/collider/narrowphase.py
duburcqa, line 1: There is no way to share more code with the original implementation?

genesis/engine/solvers/rigid/collider/narrowphase.py
```diff
@@ -962,6 +970,1045 @@ def func_convex_convex_contact(
...
+                    gjk_info,
+                    gjk_static_config,
+                    support_field_info,
+                    diff_contact_input,
+                    errno,
+                )
+        else:
+            # MPR partition: pull from mpr_queue
+            for _iter in range(max_items_per_thread):
+                idx = qd.atomic_add(collider_state.narrowphase_work_queues.mpr_work_counter[0], 1)
```
duburcqa, line 1637: You should probably define aliases for `collider_state.narrowphase_work_queues`. This is hard to read otherwise.

genesis/engine/solvers/rigid/collider/collider.py
```diff
@@ -6,13 +6,21 @@
...
 import trimesh
 
 import genesis as gs
+
+# None = GS_SORT not set, sort contacts after narrowphase (default)
+# True = GS_SORT=1, sort contacts after narrowphase
+# False = GS_SORT=0, skip contact sorting
+_SORT: bool | None = None
+if os.environ.get("GS_SORT") is not None:
+    _SORT = os.environ.get("GS_SORT", "").lower() in ("1", "true", "yes")
```
duburcqa: Do you need this in production?

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -7,6 +8,15 @@
...
 import genesis as gs
+
+
+class _ForceSolver(enum.Enum):
+    NONE = "none"
+    MONOLITH = "monolith"
+    DECOMPOSED = "decomposed"
+
+
+_FORCE_SOLVER = _ForceSolver(os.environ.get("GS_FORCE_SOLVER", "none"))
```
duburcqa: Do you need this in production?

genesis/engine/solvers/rigid/collider/collider.py
hughperkins: Only during dev. Can just hack the code in a branch during dev, it is true. Can remove it.

genesis/engine/solvers/rigid/constraint/solver.py
hughperkins: No, I dont even use this. I use the branch we were discussing in Quadrants. Will remove.

genesis/engine/solvers/rigid/collider/narrowphase.py
hughperkins: There's a "more factorized" branch at https://github.com/Genesis-Embodied-AI/Genesis/pull/2529/changes , but I put "more factorized" in quotation marks, because the +/- stats are almost identical.

genesis/utils/array_class.py
```diff
@@ -698,7 +755,12 @@ def get_collider_info(solver, n_vert_neighbors, collider_static_config, **kwargs
 @qd.data_oriented
 class StructColliderStaticConfig(metaclass=AutoInitMeta):
     has_terrain: bool
-    has_convex_convex: bool
+    # True when the scene has convex-convex collision pairs not handled by
+    # func_narrow_phase_convex_specializations (box-box, plane-box). Computed once
+    # at scene build time by iterating all geom pairs in collider._init_static_config().
+    # On GPU, the split-kernel narrowphase path runs (kernel1 + kernel2 + sort).
+    # On CPU, falls back to the monolithic func_narrow_phase_convex_vs_convex.
+    needs_kernel1: bool
```
hughperkins: Note: happy to rename this to eg `needs_mpr_gjk`

hughperkins: oh wait, it's not just mpr-gjk, it's also plane-box I think. But it's stuff that needs at least some mpr or gjk support things I think?

hughperkins: oh, actually, in this branch, it's also capsule too 🤔  (I had a branch where I moved capsule specializations out, https://github.com/hughperkins/Genesis/commits/hp/narrow-mpr-gjk-part-evict-specializations/ but it got slower, so I ditched that, for now).

hughperkins: (it was barely slower though, by the way, so if we had a strong reason to want to move capsule out, we could do that (probably in a separate pr))

genesis/engine/solvers/rigid/collider/narrowphase.py
hughperkins: cannot create aliases for sub dataclassses currently:

```
(genesis-world) ~/git/genesis (hp/narrow-mpr-gjk-part|…7) $ cat ~/ais/squid/tmp/test_nested_alias.py
import genesis as gs
gs.init(backend=gs.cpu)
import quadrants as qd
from dataclasses import dataclass

@dataclass(frozen=True)
class Inner:
    x: qd.types.ndarray()

@dataclass(frozen=True)
class Outer:
    inner: Inner

outer = Outer(inner=Inner(x=qd.ndarray(dtype=qd.i32, shape=(4,))))

@qd.kernel
def works(o: Outer):
    for i in range(4):
        o.inner.x[i] = i

@qd.kernel
def fails(o: Outer):
    for i in range(4):
        inn = o.inner  # QuadrantsTypeError
        inn.x[i] = i

works(outer)
print("direct access works")
fails(outer)
print("alias works")
(genesis-world) ~/git/genesis (hp/narrow-mpr-gjk-part|…7) $ python ~/ais/squid/tmp/test_nested_alias.py
[Genesis] [08:47:48] [INFO] ╭─────────────────────────────────────────────────────────────────────────────────────╮
[Genesis] [08:47:48] [INFO] │┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉ Genesis ┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉│
[Genesis] [08:47:48] [INFO] ╰─────────────────────────────────────────────────────────────────────────────────────╯
[Genesis] [08:47:48] [INFO] Running on [Apple M4] with backend gs.cpu. Device memory: 32.00 GB.
[Genesis] [08:47:48] [INFO] 🚀 Genesis initialized. 🔖 version: 0.4.1, 🎨 theme: dark, 🌱 seed: None, 🐛 debug: False, 📏 precision: 32, 🔥 performance: False, 💬 verbose: INFO
direct access works
  File "/Users/hugh/ais/squid/tmp/test_nested_alias.py", line 29, in <module>
    fails(outer)
  File "/Users/hugh/git/genesis/.venv/lib/python3.10/site-packages/quadrants/lang/_quadrants_callable.py", line 96, in __call__
    return self.wrapper.__call__(*args, **kwargs)
  File "/Users/hugh/git/genesis/.venv/lib/python3.10/site-packages/quadrants/lang/kernel_impl.py", line 145, in wrapped_func
    raise type(e)("\n" + str(e)) from None
quadrants.lang.exception.QuadrantsTypeError:
quadrants stack trace:
===
  File "/Users/hugh/git/genesis/.venv/lib/python3.10/site-packages/quadrants/lang/ast/ast_transformer_utils.py", line 40, in __call__
    res = method(ctx, node)
  File "/Users/hugh/git/genesis/.venv/lib/python3.10/site-packages/quadrants/lang/ast/ast_transformer.py", line 151, in build_Assign
    values = node.value.ptr if is_static_assign else impl.expr_init(node.value.ptr)
  File "/Users/hugh/git/genesis/.venv/lib/python3.10/site-packages/quadrants/lang/util.py", line 292, in wrapped
    return func(*args, **kwargs)
  File "/Users/hugh/git/genesis/.venv/lib/python3.10/site-packages/quadrants/lang/impl.py", line 122, in expr_init
    Expr(rhs).ptr, _qd_core.DebugInfo(get_runtime().get_current_src_info())
  File "/Users/hugh/git/genesis/.venv/lib/python3.10/site-packages/quadrants/lang/expr.py", line 46, in __init__
    self.ptr = make_constant_expr(arg, dtype).ptr
  File "/Users/hugh/git/genesis/.venv/lib/python3.10/site-packages/quadrants/lang/expr.py", line 139, in make_constant_expr
    raise QuadrantsTypeError(f"Invalid constant scalar data type: {type(val)}")
quadrants.lang.exception.QuadrantsTypeError: Invalid constant scalar data type: <class 'type'>

===

Your code:
File "/Users/hugh/ais/squid/tmp/test_nested_alias.py", line 24, in fails:
        inn = o.inner  # QuadrantsTypeError
        ^^^^^^^^^^^^^
Invalid constant scalar data type: <class 'type'>


[Genesis] [08:47:48] [ERROR] QuadrantsTypeError:
quadrants stack trace:
===
  File "/Users/hugh/git/genesis/.venv/lib/python3.10/site-packages/quadrants/lang/ast/ast_transformer_utils.py", line 40, in __call__
    res = method(ctx, node)
  File "/Users/hugh/git/genesis/.venv/lib/python3.10/site-packages/quadrants/lang/ast/ast_transformer.py", line 151, in build_Assign
    values = node.value.ptr if is_static_assign else impl.expr_init(node.value.ptr)
  File "/Users/hugh/git/genesis/.venv/lib/python3.10/site-packages/quadrants/lang/util.py", line 292, in wrapped
    return func(*args, **kwargs)
  File "/Users/hugh/git/genesis/.venv/lib/python3.10/site-packages/quadrants/lang/impl.py", line 122, in expr_init
    Expr(rhs).ptr, _qd_core.DebugInfo(get_runtime().get_current_src_info())
  File "/Users/hugh/git/genesis/.venv/lib/python3.10/site-packages/quadrants/lang/expr.py", line 46, in __init__
    self.ptr = make_constant_expr(arg, dtype).ptr
  File "/Users/hugh/git/genesis/.venv/lib/python3.10/site-packages/quadrants/lang/expr.py", line 139, in make_constant_expr
    raise QuadrantsTypeError(f"Invalid constant scalar data type: {type(val)}")
quadrants.lang.exception.QuadrantsTypeError: Invalid constant scalar data type: <class 'type'>

===

Your code:
File "/Users/hugh/ais/squid/tmp/test_nested_alias.py", line 24, in fails:
        inn = o.inner  # QuadrantsTypeError
        ^^^^^^^^^^^^^
Invalid constant scalar data type: <class 'type'>

[Genesis] [08:47:48] [INFO] 💤 Exiting Genesis and caching compiled kernels...
```

genesis/engine/solvers/rigid/collider/collider.py
hughperkins: removed in 25073b51

genesis/engine/solvers/rigid/constraint/solver.py
hughperkins: fixed in aed8a8e6

hughperkins:
As predicted from my broadphase experiments last night, noslip failing, after I added a new 'mixed' test:
```
gw46] [ 22%] FAILED tests/test_rigid_physics.py::test_noslip_iterations_mixed
```
I need ot update the sort I believe.

hughperkins:
Updated the sort, and new noslip_iterations_mixed test passing now.

hughperkins:
When I ran with new sort last night, here are the results I got, for ndarray:

Screenshot 2026-03-11 at 11 40 19

Not as good as with the original sort, but not awful.

hughperkins:
I'm going to:
- put back the original sort (not grouped)
- only use these kernels when we are not using noslip (otherwise use the original narrowphase kernel)

genesis/engine/solvers/rigid/collider/narrowphase.py
duburcqa: I'm sad.

hughperkins: Yeah, we have a ticket to enable such things at some point, but probably not the highest priority right now.

I might add it to my list of 'potential ai background tasks' though.

duburcqa: Note that the official syntax is this one from what I know:

```python
@qd.kernel
def fails(o: Outer):
    inn = qd.static(o.inner)
    for i in range(4):
        inn.x[i] = i
```

But it is still failing :

```python
File "<ipython-input-1-3e327ffd585a>", line 23, in fails:
    inn = qd.static(o.inner)
          ^^^^^^^^^^^^^^^^^^
  File "/Users/alexis.duburcq/workspace/src/quadrants/python/quadrants/lang/ast/ast_transformers/call_transformer.py", line 185, in _expand_Call_dataclass_args
    child_name = create_flat_name(arg.id, field.name)
                                  ^^^^^^
AttributeError: 'Attribute' object has no attribute 'id'

The above exception was the direct cause of the following exception:

  File "/Users/alexis.duburcq/workspace/src/quadrants/python/quadrants/lang/ast/ast_transformer_utils.py", line 40, in __call__
    res = method(ctx, node)
          ^^^^^^^^^^^^^^^^^
  File "/Users/alexis.duburcq/workspace/src/quadrants/python/quadrants/lang/ast/ast_transformer.py", line 431, in build_Call
    return CallTransformer.build_Call(ctx, node, build_stmt, build_stmts)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alexis.duburcq/workspace/src/quadrants/python/quadrants/lang/ast/ast_transformers/call_transformer.py", line 293, in build_Call
    added_args, node_args = CallTransformer._expand_Call_dataclass_args(ctx, node.args)
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alexis.duburcq/workspace/src/quadrants/python/quadrants/lang/ast/ast_transformers/call_transformer.py", line 187, in _expand_Call_dataclass_args
    raise RuntimeError(f"Exception whilst processing {field.name} in {type(dataclass_type)}") from e
RuntimeError: Exception whilst processing x in <class 'type'>
```

hughperkins: Note that that syntax is not officially supported, or tested, to be clear, by Quadrants. But good that it works for you, sometimes.

hughperkins: (I mean, it might work for data_oriented, but it isn't supported or tested for dataclasses.dataclass, at this time)

duburcqa:
> only use these kernels when we are not using noslip (otherwise use the original narrowphase kernel)

I don't like this idea. This would make the code more complicated to understand and follow. If it is really necessary it could be discussed, but I want to make sure there is no other option.

If I understand correctly, your PR changes the contact points that are being detected? Could you try running `python examples/collision/tower.py --object duck --vis` and `python examples/collision/tower.py --vis` to make sure the simulation is stable?

hughperkins:
> > only use these kernels when we are not using noslip (otherwise use the original narrowphase kernel)
> 
> I don't like this idea. This would make the code more complicated to understand and follow. If it is really necessary it could be discussed, but I want to make sure there is no other option.
> 
> If I understand correctly, your PR changes the contact points that are being detected? Could you try running `python examples/collision/tower.py --object duck --vis` and `python examples/collision/tower.py --vis` to make sure the simulation is stable?

Changes the ordering of contacts. The contacts themselves are otherwise the same (to within 1ULP).

hughperkins:
> > only use these kernels when we are not using noslip (otherwise use the original narrowphase kernel)
> 
> I don't like this idea. This would make the code more complicated to understand and follow. If it is really necessary it could be discussed, but I want to make sure there is no other option.
> 
> If I understand correctly, your PR changes the contact points that are being detected? Could you try running `python examples/collision/tower.py --object duck --vis` and `python examples/collision/tower.py --vis` to make sure the simulation is stable?

Duck:

Screenshot 2026-03-15 at 06 57 17
Screenshot 2026-03-15 at 06 57 22

hughperkins:
> ython examples/collision/tower.py --vis

tower:

Screenshot 2026-03-15 at 06 58 28
Screenshot 2026-03-15 at 06 58 34
Screenshot 2026-03-15 at 06 58 41

hughperkins:
I can enable on noslip too if you want. The existing noslip test (before this PR) still passes. But I'm concerned it might affect teleoperation (which isnt tested by existing noslip tests). So I would prefer to disable noslip in this PR for now.

hughperkins:
Appears grouped vs ungrouped results are ~identical. This is FPS:

Screenshot 2026-03-15 at 07 58 03

I'll switch to using grouped, and remove the noslip exception.

hughperkins:
actually, those two branches I just compared are identical code 🤔

duburcqa:
Let's remove the nonslip exception and merge. The original unit test was not robust at all, no wonder it was failing after your update. But I don't think it has anything to do with noslip per se, just this unit test was very sensitive to the contact points being detected. The current test is even more strict than better, but operating collision detection issues from noslip solver validation. We already have tests for contact detection, if they pass then I'm good.

genesis/engine/solvers/rigid/collider/collider.py
```diff
@@ -93,11 +100,14 @@ def __init__(self, rigid_solver: "RigidSolver"):
 
         if self._collider_static_config.has_nonconvex_nonterrain:
             self._sdf.activate()
-        if self._collider_static_config.has_convex_convex:
+        if self._collider_static_config.needs_kernel1:
```
duburcqa: `needs_kernel1` is not a good name. Could you use something more explicit?

genesis/engine/solvers/rigid/collider/collider.py
```diff
@@ -93,11 +100,14 @@ def __init__(self, rigid_solver: "RigidSolver"):
...
             self._gjk.activate()
-        if self._collider_static_config.has_terrain or self._collider_static_config.has_convex_convex:
+        if self._collider_static_config.has_terrain or self._collider_static_config.needs_kernel1:
             self._support_field.activate()
 
+        if self._use_split_kernels:
+            self._init_kernel2_gjk_state()
```
duburcqa: Same for kernel 2. Could you replace kernel 1 vs kernel 2 by something that describes what they do?

genesis/engine/solvers/rigid/collider/collider.py
hughperkins: needs_mpr_gjk?

genesis/engine/solvers/rigid/collider/collider.py
```diff
@@ -194,7 +204,41 @@ def _init_collision_fields(self) -> None:
         # 'contact_data_cache' is not used in Quadrants kernels, so keep it outside of the collider state / info
         self._contact_data_cache: dict[tuple[bool, bool], dict[str, torch.Tensor | tuple[torch.Tensor]]] = {}
 
-        self.reset()
+        # Kernel 1 & 2 scratch states only needed when split-kernel narrowphase is active
+        if self._use_split_kernels:
+            gpu_props = torch.cuda.get_device_properties(gs.device)
+            gpu_cuda_cores = gpu_props.multi_processor_count * 128
+            self._kernel1_n_chunks = max(1, -(-gpu_cuda_cores // self._solver._B))
```
duburcqa: I don't like this pattern. You math.ceil, math.floor or similar if possible.

genesis/engine/solvers/rigid/collider/collider.py
```diff
@@ -194,7 +204,41 @@ def _init_collision_fields(self) -> None:
...
+            self._kernel1_grid_size = self._solver._B * self._kernel1_n_chunks
+            self._kernel1_mpr_state = array_class.get_mpr_state(self._kernel1_grid_size)
+            self._kernel1_gjk_state = array_class.get_gjk_state_contact_only(self._kernel1_grid_size)
+
+            gjk_only = self._collider_static_config.ccd_algorithm in (
+                CCD_ALGORITHM_CODE.GJK,
+                CCD_ALGORITHM_CODE.MJ_GJK,
+            )
```
duburcqa: I double checked, this should fit on one line:
```python
            gjk_only = self._collider_static_config.ccd_algorithm in (CCD_ALGORITHM_CODE.GJK, CCD_ALGORITHM_CODE.MJ_GJK)
```

And even if it does not, prefer:
```python
            gjk_only = self._collider_static_config.ccd_algorithm in (
                CCD_ALGORITHM_CODE.GJK, CCD_ALGORITHM_CODE.MJ_GJK
            )
```

genesis/engine/solvers/rigid/collider/collider.py
```diff
@@ -194,7 +204,41 @@ def _init_collision_fields(self) -> None:
...
+        class _SolverProxy:
+            def __init__(self, _B, requires_grad):
+                self._B = _B
+                self._requires_grad = requires_grad
+                self._static_rigid_sim_config = type("_", (), {"requires_grad": requires_grad})()
+
+        proxy = _SolverProxy(self._kernel2_n_gjk_threads, self._solver._static_rigid_sim_config.requires_grad)
+        self._kernel2_gjk_state = array_class.get_gjk_state(
+            proxy, self._solver._static_rigid_sim_config, self._gjk._gjk_info, True
+        )
```
duburcqa: What is this solver proxy? I don't like this pattern at all.

genesis/utils/array_class.py
duburcqa: yes please rename. Use anything that makes sense.

genesis/utils/array_class.py
```diff
@@ -1042,6 +1104,67 @@ def get_gjk_state(solver, static_rigid_sim_config, gjk_info, is_active):
...
+    func_gjk never accesses them.
+    """
+
+    class _Proxy:
+        pass
+
+    full = _Proxy()
+    full._B = _B
+    dummy = _Proxy()
+    dummy._B = 1
```
duburcqa: I hate this proxy stuff.

duburcqa: Just pass all the parameters you need as input arguments of these helper functions, without using this proxy structure.

duburcqa: It will affect a larger part of the code but I would much prefer that compared to this kind of pseudo monkeypatching.

tests/test_rigid_physics.py
```diff
@@ -4098,6 +4098,153 @@ def test_noslip_iterations(scale, friction, show_viewer, tol):
...
+        "contacts processed in a different order produce slightly different friction "
+        "forces that accumulate and cause the simulation to diverge."
+    ),
+    strict=False,
+)
+def test_noslip_iterations_mixed(scale, box_box_detection, show_viewer, tol, asset_tmp_path):
+    """Noslip test with box primitives plus a distant mesh cube, forcing the convex-convex kernel to run."""
+    mesh_path = str(asset_tmp_path / f"noslip_mixed_box_{scale}.obj")
+    tmesh = trimesh.creation.box(extents=(scale, scale, scale))
+    tmesh.export(mesh_path, file_type="obj")
```
duburcqa: Please parameterise the original unit test instead of duplicating it.

We have model_name and/or xml_path fixtures for this. Look at other unit tests, eg 'test_urdf_parsing_undefined_inertia'.

tests/test_rigid_physics.py
```diff
@@ -4098,6 +4098,153 @@ def test_noslip_iterations(scale, friction, show_viewer, tol):
...
+@pytest.mark.xfail(
+    reason=(
+        "Fragile: adding a distant non-contacting mesh entity changes the broadphase "
+        "pair list, which reorders box-box contacts within the same geom pair. The PGS "
+        "noslip solver with only 5 iterations is sensitive to this ordering — the same "
+        "contacts processed in a different order produce slightly different friction "
+        "forces that accumulate and cause the simulation to diverge."
+    ),
+    strict=False,
+)
```
duburcqa: Is this still real with the new unit test?

tests/test_rigid_physics.py
```diff
@@ -4098,6 +4098,153 @@ def test_noslip_iterations(scale, friction, show_viewer, tol):
         assert box_z < -scale
 
 
+@pytest.mark.slow
+@pytest.mark.required
+@pytest.mark.parametrize("scale", [0.1, 10.0])
+@pytest.mark.parametrize("box_box_detection", [False, True])
```
duburcqa: This is irrelevant here.

tests/test_rigid_physics.py
```diff
@@ -4098,6 +4098,153 @@ def test_noslip_iterations(scale, friction, show_viewer, tol):
         assert box_z < -scale
 
 
+@pytest.mark.slow
+@pytest.mark.required
+@pytest.mark.parametrize("scale", [0.1, 10.0])
```
duburcqa: A scale of 10.0 is completely unrealistic. See new noslip unit test.

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -7,6 +6,8 @@
 from frozendict import frozendict
 
 import genesis as gs
+
+
```
duburcqa: Only one line skip here

hughperkins:
actual grouped vs ungrouped:

Screenshot 2026-03-15 at 08 58 47

Seems within variance.

I'll switch to grouped sort, to avoid risk of breaking noslip.

hughperkins:
(switched to block sort, and removed the conditional on 'noslip' condition)

hughperkins:
(weird that changing to block sort isnt visible as a commit in the conversation. anyway, it's visible in commits)

hughperkins:
> Let's remove the nonslip exception and merge. The original unit test was not robust at all, no wonder it was failing after your update. But I don't think it has anything to do with noslip per se, just this unit test was very sensitive to the contact points being detected. The current test is even more strict than better, but operating collision detection issues from noslip solver validation. We already have tests for contact detection, if they pass then I'm good.

Note. the original test was NOT failing after my update. The new tests I created were failing.

hughperkins:
> > Let's remove the nonslip exception and merge. The original unit test was not robust at all, no wonder it was failing after your update. But I don't think it has anything to do with noslip per se, just this unit test was very sensitive to the contact points being detected. The current test is even more strict than better, but operating collision detection issues from noslip solver validation. We already have tests for contact detection, if they pass then I'm good.
> 
> Note. the original test was NOT failing after my update. The new tests I created were failing.

The original test is very weak though. It only covers the very narrow situation of 3 boxes. It does not consider adding meshes to the scene, which my new tests do.

genesis/engine/solvers/rigid/collider/narrowphase.py
```diff
@@ -962,6 +970,1045 @@ def func_convex_convex_contact(
...
+            for i in range(n_con):
+                i_c = start_idx + i
+                pos_i = qd.Vector(
+                    [local_contact_pos[i, 0], local_contact_pos[i, 1], local_contact_pos[i, 2]],
+                    dt=gs.qd_float,
+                )
+                normal_i = qd.Vector(
+                    [local_normal[i, 0], local_normal[i, 1], local_normal[i, 2]],
+                    dt=gs.qd_float,
+                )
```
duburcqa: It fits on one line.

genesis/engine/solvers/rigid/collider/narrowphase.py
```diff
@@ -962,6 +970,1045 @@ def func_convex_convex_contact(
...
+        collider_state.narrowphase_work_queues.gjk_i_pair[idx] = i_pair
+    else:
+        start_idx = qd.atomic_add(collider_state.n_contacts[i_b_env], n_con)
+        if start_idx + n_con <= collider_info.max_contact_pairs[None]:
+            for i in range(n_con):
+                i_c = start_idx + i
+                pos_i = qd.Vector(
+                    [local_contact_pos[i, 0], local_contact_pos[i, 1], local_contact_pos[i, 2]],
+                    dt=gs.qd_float,
+                )
```
duburcqa, line 1310:
```python
                pos_i = qd.Vector(
                    [local_contact_pos[i, 0], local_contact_pos[i, 1], local_contact_pos[i, 2]], dt=gs.qd_float
                )
```

genesis/engine/solvers/rigid/collider/narrowphase.py
```diff
@@ -962,6 +970,1045 @@ def func_convex_convex_contact(
...
+                        normal = gjk_state.normal[i_b_scratch, 0]
+                    used_gjk = True
+
+    return is_col, normal, contact_pos, penetration, used_gjk
+
+
+@qd.func
+def func_kernel2_mpr_multicontact(
+    i_b_scratch,
+    i_b_env,
```
duburcqa: `i_b` is used for `i_b_env` specifically. Nothing else. You should rather define `i_scratch` and `i_b`

genesis/engine/solvers/rigid/collider/narrowphase.py
```diff
@@ -962,6 +970,1045 @@ def func_convex_convex_contact(
...
+                            (contact_pos - 0.5 * penetration * normal) - contact_pos_0,
+                            gu.qd_inv_quat(qrot),
+                        )
+                        + contact_pos_0
+                    )
+                    contact_point_b = (
+                        gu.qd_transform_by_quat(
+                            (contact_pos + 0.5 * penetration * normal) - contact_pos_0,
+                            qrot,
+                        )
```
duburcqa: one line.

genesis/engine/solvers/rigid/collider/narrowphase.py
```diff
@@ -962,6 +970,1045 @@ def func_convex_convex_contact(
...
+                    collider_state.contact_cache.normal[i_pair, i_b_env] = qd.Vector.zero(gs.qd_float, 3)
+            elif not gjk_multi_done and multi_contact and is_col:
+                if qd.static(
+                    collider_static_config.ccd_algorithm not in (CCD_ALGORITHM_CODE.MJ_MPR, CCD_ALGORITHM_CODE.MJ_GJK)
+                ):
+                    contact_point_a = (
+                        gu.qd_transform_by_quat(
+                            (contact_pos - 0.5 * penetration * normal) - contact_pos_0,
+                            gu.qd_inv_quat(qrot),
+                        )
```
duburcqa, line 1502:
```python
                        gu.qd_transform_by_quat(
                            (contact_pos - 0.5 * penetration * normal) - contact_pos_0, gu.qd_inv_quat(qrot)
                        )
```

genesis/engine/solvers/rigid/collider/narrowphase.py
```diff
@@ -962,6 +970,1045 @@ def func_convex_convex_contact(
...
+            for i in range(n_con):
+                i_c = start_idx + i
+                pos_i = qd.Vector(
+                    [local_contact_pos[i, 0], local_contact_pos[i, 1], local_contact_pos[i, 2]],
+                    dt=gs.qd_float,
+                )
+                normal_i = qd.Vector(
+                    [local_normal[i, 0], local_normal[i, 1], local_normal[i, 2]],
+                    dt=gs.qd_float,
+                )
```
duburcqa, line 1549: Formatting.

genesis/engine/solvers/rigid/collider/narrowphase.py
duburcqa: I see. It is an official feature for Taichi: https://docs.taichi-lang.org/docs/syntax_sugars#aliases

For quadrants, I don't know, there is no online doc. It would be great to support it at some point.

genesis/engine/solvers/rigid/collider/narrowphase.py
```diff
@@ -962,6 +970,1045 @@ def func_convex_convex_contact(
...
+                is_col, normal, contact_pos, penetration = capsule_contact.func_sphere_capsule_contact(
+                    i_ga=i_ga,
+                    i_gb=i_gb,
+                    ga_pos=ga_pos,
+                    ga_quat=ga_quat,
+                    gb_pos=gb_pos,
+                    gb_quat=gb_quat,
+                    geoms_info=geoms_info,
+                    rigid_global_info=rigid_global_info,
+                )
```
duburcqa, line 1844: Avoid keyword arguments for consistency.

genesis/engine/solvers/rigid/collider/narrowphase.py
```diff
@@ -962,6 +970,1045 @@ def func_convex_convex_contact(
...
+        collider_state.narrowphase_work_queues.gjk_work_counter[0] = k1_size
+        collider_state.narrowphase_work_queues.gjk_queue_size[0] = k1_size + k2_count
+        collider_state.narrowphase_work_queues.mpr_queue_size[0] = 0
+        collider_state.narrowphase_work_queues.mpr_work_counter[0] = 0
+
+
+@qd.func
+def func_enqueue_for_multicontact(
+    collider_state: array_class.ColliderState,
+    prefer_gjk,
```
duburcqa: Move this kind of boolean flag at the end of the parameter list. This way, you can specify keyword when calling this function without having to do it for the other parameters
```python
func_enqueue_for_multicontact(
    [...],
    prefer_gjk=True,
)
```
This is what we are doing elsewhere and I like this pattern.

genesis/engine/solvers/rigid/collider/narrowphase.py
hughperkins: > I see. It is an official feature for Taichi: https://docs.taichi-lang.org/docs/syntax_sugars#aliases

Interesting

genesis/engine/solvers/rigid/collider/narrowphase.py
hughperkins: I prefer keyword arguments, and would prefer to migrate gradually to use them. You prefer positional arguments?

genesis/engine/solvers/rigid/collider/collider.py
```diff
@@ -93,11 +96,14 @@ def __init__(self, rigid_solver: "RigidSolver"):
 
         if self._collider_static_config.has_nonconvex_nonterrain:
             self._sdf.activate()
-        if self._collider_static_config.has_convex_convex:
+        if self._collider_static_config.needs_mpr_gjk:
```
duburcqa: Could you clarify what is the difference between `has_convex_convex` and `needs_mpr_gjk`?

Did you change something or you just don't like the variable name? I'm not in favor of changing this variable name if the behaviour is the same. The current pattern is `has_<geom_type_1>_<geom_type_2>`. `has_any_terrain` was renamed `has_terrain` a few days ago. This was a mistake but I merged anyway to speedup the merging process because we are in a rush. But I don't like it.

In you want to distinguish convex primitive from convex mesh (ie generic vs specialisation), I would suggest `has_convex_mesh_convex_mesh`.

genesis/engine/solvers/rigid/collider/narrowphase.py
duburcqa: > I prefer keyword arguments, and would prefer to migrate gradually to use them. 

I'm not in favor of doing this. It is too verbose.

> You prefer positional arguments?

Yes.

duburcqa: > would prefer to migrate gradually to use the

There is not plan of migrating gradually at the moment.

genesis/engine/solvers/rigid/collider/collider.py
hughperkins: renamed to `needs_convex_convex_except_box_plane`.

really I think something like `needs_contact0_kernel` would be more intuitive, since this is what it physically means.

hughperkins: (and this would also allow us to move capsule spec outside of contact0 kernel, without having to rename this variable)

tests/test_rigid_physics.py
hughperkins: removed

genesis/engine/solvers/rigid/collider/narrowphase.py
hughperkins: updated

genesis/engine/solvers/rigid/collider/collider.py
hughperkins: related thread https://github.com/Genesis-Embodied-AI/Genesis/pull/2527#discussion_r2936687650

genesis/engine/solvers/rigid/collider/narrowphase.py
hughperkins: what do you mean? This passes pre-commit run -a

hughperkins: did the one line change

genesis/engine/solvers/rigid/collider/narrowphase.py
hughperkins: i_b_env can't become i_b because i_b is already used as a loop variable in the calling kernels. The parameter i_b_env represents the environment index  while i_b in the kernels is the thread/scratch index

hughperkins: well, perhaps i_b_env should be i_b, and i_b should be i_tid. Doing this rename

hughperkins: (I know that 'i_tid' is lierally 'index of thread index', but it seemed more intuitivev to me than `i_t`, and meanwhile respects the pattern `i_[something]`.)

hughperkins: (open to your suggestion for how to name `i_tid`)

genesis/utils/array_class.py
hughperkins: Here is what Opus says:

"  Here's the trade-off:
  1. Full refactor (what reviewer wants): Change all helpers to take _B instead of solver. ~12 function sigs + ~38 call sites. Touches a lot of existing code.
  2. Minimal fix: Just change get_gjk_state to take _B and requires_grad, keep helpers using solver but pass a types.SimpleNamespace(_B=_B) internally. Removes _SolverProxy from
     collider.py but still uses a namespace internally.
  3. Middle ground: Change get_gjk_state signature to take _B and requires_grad, and refactor just the helpers that are called from get_gjk_state_contact_only (the new code) to
     take _B directly. But that means the same helpers have inconsistent signatures.

  "I'd recommend option 1 — it's what the reviewer asked for, and while it touches more code, each change is mechanical (solver → _B). Want me to go ahead with that?"

Thoughts? Can you confirm this is what you intend?

genesis/engine/solvers/rigid/collider/collider.py
hughperkins: renaming `needs_convex_convex_except_box_plane` to `has_non_box_plane_convex_convex`

hughperkins: (done)

hughperkins: (but in the future we might need to rename to `has_non_box_plane_non_capsule_capsule_non_capsule_sphere_non_sphere_sphere_convex_convex`... )

hughperkins: I could call it `has_needs_contact0_convex_convex` perhaps? This would be a compromise between starting with `has_`, containing `convex_convex`, and making clear that the meaning is 'contains a convex convex that is handled by contact0 kernel'.

hughperkins: actually, I'm going to rename to `has_needs_contact0_convex_convex`

hughperkins: wait, chagned my mind, going back to `has_non_box_plane_convex_convex`

tests/test_rigid_physics.py
```diff
@@ -4016,21 +4016,24 @@ def test_mesh_primitive_COM(show_viewer, tol):
 @pytest.mark.required
 @pytest.mark.parametrize("scale", [0.04, 1.0])
 @pytest.mark.parametrize("friction", [0.5, 2.0])
+@pytest.mark.parametrize("mesh_boxes", [False, True])
+@pytest.mark.parametrize("add_faraway_mesh", [False, True])
```
duburcqa: Could you force this to True? The matrix is already large and this test is quite long. Moreover, False does not provide any additional information over true.

duburcqa (dismissed):
Just need to update the unit test and we merge.

hughperkins:
🙌

PR #2534, opened 2026-03-12
[MISC] Faster initialisation of collision pairs validity mask.

ACMLCZH:
- Replace two O(N^2) Python loops in `Collider._init_static_config` and `_compute_collision_pair_idx` with numpy-vectorized filtering. It is previously a great bottleneck at build time.
- Pre-compute per-geom properties (link idx, root idx, is_fixed, contype/conaffinity, etc.) into numpy arrays and use broadcasting for bulk pair evaluation
- Self-collision checks (adjacency, neutral overlap) remain as a targeted Python loop over only same-root pairs, with lazy vertex computation for only the geoms that need it
- Algorithm flags (`has_terrain`, `has_convex_convex`, etc.) are now computed inline during pair filtering, removing the redundant second O(N^2) pass

   **Benchmark** on a 138-body scene (1203 geoms after convex decomposition):
   | Version | Build time | Collision pairs |
   |---------|-----------|-----------------|
   | Before  | 98.6s     | 750             |
   | After   | 28.4s     | 750             |

   **3.5x speedup**, identical output.

### DIFF
--- a/.github/workflows/alarm.yml
+++ b/.github/workflows/alarm.yml
@@ -31,7 +31,9 @@ jobs:
 
       - name: Install deps
         run: |
-          python -m pip install --quiet --upgrade wandb frozendict
+          # Pin wandb<0.25: 0.25.x changed api.runs() to no longer populate
+          # run.config in batch queries, breaking alarm.py's config["revision"] lookup.
+          python -m pip install --quiet --upgrade 'wandb<0.25' frozendict
 
       - name: Download speed artifacts from triggering run
         id: dl_speed


genesis/engine/solvers/rigid/collider/collider.py
```diff
@@ -137,45 +137,16 @@ def _init_static_config(self) -> None:
...
         # Note that updating any of them will trigger recompilation.
         self._collider_static_config = array_class.StructColliderStaticConfig(
-            has_terrain=has_any_vs_terrain,
-            has_convex_convex=has_convex_vs_convex,
-            has_convex_specialization=has_convex_specialization,
-            has_nonconvex_nonterrain=has_nonconvex_vs_nonterrain,
+            has_terrain=pair_flags["has_terrain"],
+            has_convex_convex=pair_flags["has_convex_convex"],
+            has_convex_specialization=pair_flags["has_convex_specialization"],
+            has_nonconvex_nonterrain=pair_flags["has_nonconvex_nonterrain"],
```
duburcqa: Guys. stop using dict to pass around structured information. That is not a good practice. Just return tuples as before, and if you need something more sophisticated, use named tuple.

genesis/engine/solvers/rigid/collider/collider.py
```diff
@@ -360,7 +373,53 @@ def _compute_collision_pair_idx(self):
...
+        has_any_vs_terrain = bool(np.any((valid_type_a == TERRAIN) | (valid_type_b == TERRAIN)))
+        has_convex_vs_convex = bool(np.any(valid_convex_a & valid_convex_b))
+
+        if self._solver._options.box_box_detection:
+            has_convex_specialization = bool(
+                np.any(
+                    (valid_type_a == TERRAIN)
+                    | (valid_type_a == BOX)
+                    | (valid_type_b == TERRAIN)
+                    | (valid_type_b == BOX)
```
duburcqa: np.isin(

genesis/engine/solvers/rigid/collider/collider.py
```diff
@@ -360,7 +373,53 @@ def _compute_collision_pair_idx(self):
...
+        # --- Build collision_pair_idx and count ---
+        valid_indices = np.where(valid)[0]
+        n_possible_pairs = len(valid_indices)
+        collision_pair_idx = np.full((n_geoms, n_geoms), fill_value=-1, dtype=gs.np_int)
+        collision_pair_idx[row[valid_indices], col[valid_indices]] = np.arange(n_possible_pairs, dtype=gs.np_int)
+
+        # --- Compute algorithm flags from valid pairs ---
+        TERRAIN = gs.GEOM_TYPE.TERRAIN
+        BOX = gs.GEOM_TYPE.BOX
+        PLANE = gs.GEOM_TYPE.PLANE
```
duburcqa: Remove these aliases. We don't do this in Genesis.

genesis/engine/solvers/rigid/collider/collider.py
```diff
@@ -242,115 +229,142 @@ def _compute_collision_pair_idx(self):
...
+        same_root = geom_root_idx[row] == geom_root_idx[col]
+        if not self._solver._enable_self_collision:
+            valid &= ~same_root
+
+        # Weld constraint filtering
+        if weld_pairs:
+            link_min = np.minimum(link_a, link_b)
+            link_max = np.maximum(link_a, link_b)
+            is_weld = np.array(
+                [(int(link_min[i]), int(link_max[i])) in weld_pairs for i in range(len(row))], dtype=bool
```
duburcqa: cast to int is irrelevant

genesis/engine/solvers/rigid/collider/collider.py
```diff
@@ -242,115 +229,142 @@ def _compute_collision_pair_idx(self):
...
+        geoms_verts: dict[int, np.ndarray] = {}
+        if needs_neutral_check:
+            self_root_geom_idxs = set()
+            self_root_indices = np.where(valid & same_root)[0]
+            for idx in self_root_indices:
+                self_root_geom_idxs.add(int(row[idx]))
+                self_root_geom_idxs.add(int(col[idx]))
+            # Compute vertices only for geoms involved in self-collision pairs,
+            # shrunk by 0.1% to avoid false positive when detecting self-collision
+            for gi in self_root_geom_idxs:
```
duburcqa: There should be a better way vectorise this, eventually using `np.unique`

genesis/engine/solvers/rigid/collider/collider.py
```diff
@@ -242,115 +229,142 @@ def _compute_collision_pair_idx(self):
...
+                verts = tensor_to_array(geoms[gi].get_verts())
+                verts = verts.reshape((-1, *verts.shape[-2:]))
+                centroid = verts.mean(axis=1, keepdims=True)
+                verts = centroid + (1.0 - 1e-3) * (verts - centroid)
+                geoms_verts[gi] = verts
+
+        if needs_self_check:
+            self_root_indices = np.where(valid & same_root)[0]
+            for idx in self_root_indices:
+                i_ga, i_gb = int(row[idx]), int(col[idx])
```
duburcqa: Stop cast to int systematically. This is irrelevant.

PR #2550, opened 2026-03-14
[FEATURE] Add ProximitySensor.

### LINKED ISSUE #2549: [Feature]: ProximitySensor

Sensor which returns closest points and distance to other entity(s)

### Motivation

enables something like: https://arxiv.org/abs/2504.05287
Image

### Potential Benefit

training dexterous policies

- [ ] ProximitySensor implementation: is attached to a rigid entity link, takes probe_local_pos and track_link_idx,  outputs per probe: distance (scalar) and nearest_point (3D, world frame) to the closest point on any mesh of the tracked links. If no mesh is within max_range, report max_range and the probe position as nearest point.
- [ ] Tests

### Additional information

Capacitive-based proximity sensors: https://youtu.be/nfkQoAKzq0U?si=b4Qpk2Mw9fBqrBon&t=101
Isaac has ProximitySensor as well https://docs.isaacsim.omniverse.nvidia.com/5.1.0/sensors/isaacsim_sensors_physics_proximity.html

Milotrince:
## Related Issue

Resolves https://github.com/Genesis-Embodied-AI/Genesis/issues/2549

https://github.com/user-attachments/assets/0d91a87a-67be-4b30-8849-1f86d1fb664a

## Checklist:

### DIFF
--- a/genesis/engine/sensors/proximity.py
+++ b/genesis/engine/sensors/proximity.py
@@ -1,0 +1,304 @@
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, NamedTuple, Type
+
+import numpy as np
+import quadrants as qd
+import torch
+
+import genesis as gs
+import genesis.utils.array_class as array_class
+import genesis.utils.geom as gu
+from genesis.engine.solvers.rigid.abd.forward_kinematics import func_update_all_verts
+from genesis.options.sensors import ProximityOptions
+from genesis.utils.misc import concat_with_tensor, make_tensor_field, tensor_to_array
+from genesis.utils.raycast_qd import get_triangle_vertices
+
+from .base_sensor import (
+    NoisySensorMetadataMixin,
+    NoisySensorMixin,
+    RigidSensorMetadataMixin,
+    RigidSensorMixin,
+    Sensor,
+    SharedSensorMetadata,
+)
+from .kinematic_tactile import _func_closest_point_on_triangle
+from .sensor_manager import register_sensor
+
+if TYPE_CHECKING:
+    from genesis.options.sensors import SensorOptions
+    from genesis.utils.ring_buffer import TensorRingBuffer
+    from genesis.vis.rasterizer_context import RasterizerContext
+
+    from .sensor_manager import SensorManager
+
+
+@qd.kernel(fastcache=gs.use_fastcache)
+def _kernel_proximity(
+    probe_positions_local: qd.types.ndarray(),
+    probe_sensor_idx: qd.types.ndarray(),
+    links_idx: qd.types.ndarray(),
+    sensor_cache_start: qd.types.ndarray(),
+    sensor_probe_start: qd.types.ndarray(),
+    n_probes_per_sensor: qd.types.ndarray(),
+    track_link_start: qd.types.ndarray(),
+    track_link_end: qd.types.ndarray(),
+    track_link_flat: qd.types.ndarray(),
+    max_range: qd.types.ndarray(),
+    static_rigid_sim_config: qd.template(),
+    links_state: array_class.LinksState,
+    links_info: array_class.LinksInfo,
+    geoms_info: array_class.GeomsInfo,
+    geoms_state: array_class.GeomsState,
+    faces_info: array_class.FacesInfo,
+    verts_info: array_class.VertsInfo,
+    fixed_verts_state: array_class.VertsState,
+    free_verts_state: array_class.VertsState,
+    output: qd.types.ndarray(),
+):
+    total_n_probes = probe_positions_local.shape[0]
+    n_batches = output.shape[0]
+
+    func_update_all_verts(
+        geoms_info, geoms_state, verts_info, free_verts_state, fixed_verts_state, static_rigid_sim_config
+    )
+
+    for i_b, i_p in qd.ndrange(n_batches, total_n_probes):
+        i_s = probe_sensor_idx[i_p]
+        sensor_link_idx = links_idx[i_s]
+        link_pos = links_state.pos[sensor_link_idx, i_b]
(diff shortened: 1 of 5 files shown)


genesis/engine/sensors/proximity.py
```diff
@@ -0,0 +1,306 @@
...
+
+from .base_sensor import (
+    NoisySensorMetadataMixin,
+    NoisySensorMixin,
+    RigidSensorMetadataMixin,
+    RigidSensorMixin,
+    Sensor,
+    SharedSensorMetadata,
+)
+from .kinematic_tactile import _func_closest_point_on_triangle
```
Milotrince, line 24: should we move `_func_closest_point_on_triangle` somewhere else? if so, where?

duburcqa: We need to have a better project architecture for quadrants utils / helpers.

duburcqa:
I'm not sure adding such "sensor" makes sense. I'm ok if it is useful for anybody but it is definitely not a real sensor because not hardware is able to measure proximity the way you implemented it. In my view, our sensors are supposed to reflect the behaviour of real hardware, not just be convenience facilities to return pre-processed information that may be useful for policy training.

genesis/options/sensors/options.py
```diff
@@ -289,6 +291,48 @@ def model_post_init(self, context: Any) -> None:
         self.noise = self.acc_noise + self.gyro_noise + self.mag_noise
 
 
+class ProximityOptions(RigidSensorOptionsMixin, NoisySensorOptionsMixin, SensorOptions):
```
duburcqa: You can remove 'SensorOptions'. It is redundant.

genesis/options/sensors/options.py
```diff
@@ -289,6 +291,48 @@ def model_post_init(self, context: Any) -> None:
...
+    max_range: PositiveFloat = 10.0
+
+    debug_sphere_radius: PositiveFloat = 0.008
+    debug_color: UnitIntervalVec4Type = (0.2, 0.6, 1.0, 0.6)
+
+    def validate_scene(self, scene: "Scene"):
+        super().validate_scene(scene)
+        if scene.sim is None:
+            return
+        solver = scene.sim.rigid_solver
```
duburcqa: Remove this proxy, it is useless.

tests/test_sensors.py
```diff
@@ -1046,3 +1046,112 @@ def test_elastomer_displacement_sensor_box_sphere(show_viewer, tol, n_envs):
...
+    DISTANCE = 0.15
+    MAX_RANGE = 10.0
+    BOX_PROBE_POS = [(0.0, 0.0, 0.0), (0.0, 0.0, 0.05)]
+    SPHERE_PROBE_POS = [(0.0, 0.0, SPHERE_RADIUS)]
+
+    scene = gs.Scene(
+        sim_options=gs.options.SimOptions(
+            gravity=(0.0, 0.0, 0.0),
+        ),
+        profiling_options=gs.options.ProfilingOptions(show_FPS=False),
```
duburcqa: One line per option.

tests/test_sensors.py
```diff
@@ -1046,3 +1046,112 @@ def test_elastomer_displacement_sensor_box_sphere(show_viewer, tol, n_envs):
...
+    sphere1 = scene.add_entity(
+        gs.morphs.Sphere(radius=SPHERE_RADIUS, pos=(0.0, 0.0, DISTANCE)),
+    )
+    sphere2 = scene.add_entity(
+        gs.morphs.Sphere(radius=SPHERE_RADIUS, pos=(0.0, 0.0, DISTANCE * 2.0)),
+    )
+    # Not tracked objects
+    sphere3 = scene.add_entity(
+        gs.morphs.Sphere(radius=SPHERE_RADIUS, pos=(0.0, DISTANCE / 2.0, 0.0)),
+    )
```
duburcqa, line 1104: One line per option.

Milotrince:
> I'm not sure adding such "sensor" makes sense. I'm ok if it is useful for anybody but it is definitely not a real sensor because not hardware is able to measure proximity the way you implemented it. In my view, our sensors are supposed to reflect the behaviour of real hardware, not just be convenience facilities to return pre-processed information that may be useful for policy training.

Technically there are capacitive-based proximity sensors: https://youtu.be/nfkQoAKzq0U?si=b4Qpk2Mw9fBqrBon&t=101
but yes the implementation is not accurate since we're not simulating electrostatic fields. Still, I think it can come in handy in cases beyond my own. Plus, Isaac has ProximitySensor as well https://docs.isaacsim.omniverse.nvidia.com/5.1.0/sensors/isaacsim_sensors_physics_proximity.html

duburcqa:
> Technically there are capacitive-based proximity sensors: https://youtu.be/nfkQoAKzq0U?si=b4Qpk2Mw9fBqrBon&t=101

Wow nice! Could you mention this in the documentation of the sensor?

> but yes the implementation is not accurate since we're not simulating electrostatic fields. Still, I think it can come in handy in cases beyond my own. 

This is not what I mean. You do not have to implement the internal physics of course, but the raw output signal has to match the real sensor. The output being post-processed in the same way as a real sensor. For instance, we return linear acceleration and angular velocity for IMUs, while the raw measurement is beam deformation. But all integrated IMU sensors return linear acceleration and angular velocity. However, returning the quaternion directly in Genesis sensor would be a mistake.

duburcqa (dismissed):

Milotrince:
The calibration process for the ground truth proximity/nearest distances to what a real capacitive proximity sensor is a bit unclear since the signal can vary a lot on object material properties. Also of course returning world position values of the nearest points is unrealistic, but useful for sim applications..

maybe we can only return the distances in `read()`, store the ground truth nearest points in the sensor in a different field?

duburcqa (approved):

PR #2571, opened 2026-03-18
[FEATURE] Improve numerical stability of simulation by aligning inertial axes of free joints.

duburcqa:
Follow up: https://github.com/Genesis-Embodied-AI/Genesis/pull/2569


genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -2101,6 +2082,91 @@ def _add_by_info(self, l_info, j_infos, g_infos, morph, surface):
...
+                inertia_in_link = inertia_R @ l_info["inertial_i"] @ inertia_R.T
+                R_principal = uu.principal_axes_rot(inertia_in_link)
+                principal_quat = gu.R_to_quat(R_principal)
+                global_com = inertia_pos
+                # Update inertia to diagonalized form in the new (aligned) link frame
+                l_info["inertial_pos"] = gu.zero_pos()
+                l_info["inertial_quat"] = gu.identity_quat()
+                l_info["inertial_i"] = R_principal.T @ inertia_in_link @ R_principal
+            else:
+                # Compute COM and principal axes from convexified collision geometry
```
nimrod-gileadi, line 2123: At least in MuJoCo models I've worked with, it's very common to use the visual mesh for inertia, and set the mass on the collision geometry to zero, because:

- Sometimes you make a simplified model where only the robot's feet are colliding, for example.
-  The inertia computation works fine with non-convex meshes, so sometimes you have an inaccurate primitive collision geometry but still use the complex mesh for inertia.

So you shouldn't check contype and conaffinity - you should just look at the geom's mass or density.

PR #2576, opened 2026-03-19
[FEATURE] Disable reporting when vis and add vis to benchmarks

hughperkins:
## Description

- disable reporting when --vis set
- add --vis to benchmarks
- add --vis-n-envs

## Related Issue

## Checklist:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -456,6 +456,7 @@ def pytest_addoption(parser: pytest.Parser) -> None:
     )
     if IS_INTERACTIVE_VIEWER_AVAILABLE:
         parser.addoption("--vis", action="store_true", default=False, help="Enable interactive viewer.")
+        parser.addoption("--vis-n-envs", type=int, default=None, help="Max environments to render with --vis.")
     parser.addoption("--dev", action="store_true", default=False, help="Enable genesis debug mode.")
     supported, _reason = is_mem_monitoring_supported()
     help_text = (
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -51,6 +51,8 @@
     # Assuming headless server if tkinter is not installed unless DISPLAY env var is available on Linux
     if sys.platform.startswith("linux"):
         has_display = bool(os.environ.get("DISPLAY"))
+    elif sys.platform == "darwin":
+        has_display = True
```
duburcqa, line 55: I don't think this is correct. If tkinter does not work, this Mac is unlikely to have a working display.

tests/conftest.py
```diff
@@ -454,6 +456,7 @@ def pytest_addoption(parser: pytest.Parser) -> None:
     )
     if IS_INTERACTIVE_VIEWER_AVAILABLE:
         parser.addoption("--vis", action="store_true", default=False, help="Enable interactive viewer.")
+        parser.addoption("--vis-n-envs", type=int, default=None, help="Max environments to render with --vis.")
```
duburcqa, line 459: I don't want this option to be added. First, we already have too many options. Seconds, modifying all the unit tests to support this would be a nightmare, and partially supporting this option is not acceptable.

hughperkins: Ok. but more than 16 envs, is unwatchable. Thoughts on an approach like eg an env var `GS_TEST_VIS_N_ENVS` ?

hughperkins:
Will replace with alternative approach.

PR #2577, opened 2026-03-19
[MISC] Refactor rigid benchmarks into reusable scene factories.

hughperkins:


tests/test_rigid_benchmarks.py
```diff
@@ -945,31 +698,242 @@ def dex_hand(solver, n_envs, gjk, pytorch_profiler_step):
...
+        scene,
+        step,
+        dict(
+            compile_time=compile_time,
+            step_dt=step_dt,
+            duration_warmup=20.0,
+            duration_record=5.0,
+            needs_sync=True,
+        ),
+    )
```
duburcqa, line 724: What about using a NamedTuple for this? No strong opinion but I think it would be better.

PR #2604, opened 2026-03-25
[MISC] Unpin wandb in CI.

hughperkins:


hughperkins:
Let's go ahead and merge this. Worst case we can revert if it fails horribly.

duburcqa:
Was this tested locally?

The CI is not passing.

hughperkins:
> Was this tested locally?
> 
> The CI is not passing.

It's only failing linter.

hughperkins:
It's tested implicitly by this script https://github.com/Genesis-Embodied-AI/cmp-tooling/blob/179f003472ea675ca7e4e0b50efb3c14784fb657/bench_cluster_wandb.py#L58 , which uses the same appraoch, and has been working for me for several days.

hughperkins:
fixed linter

hughperkins:
(I also had a standalone script to reproduce the bug, and fix the bug , at some point, but I'm not sure if I have that anywher now, but the above script is sufficient I feel).

hughperkins:
Thanks!

🤞

PR #2607, opened 2026-03-26
[FEATURE] Add support of broadphase all-vs-all to speedup simulation on GPU backend.

hughperkins:
## Description

- add configurable option `broadphase_traversal` with options:
    - SAP
    - ALL_VS_ALL

Default None, which does:
- sap if:
     - on cpu, or
     - hibernate available, or
     - scene contains heterogeneous objects
 - else all vs all

## Related Issue

## Checklist:

### DIFF
--- a/genesis/engine/solvers/rigid/collider/collider.py
+++ b/genesis/engine/solvers/rigid/collider/collider.py
@@ -261,8 +261,8 @@ class Collider:
         geoms = self._solver.geoms
 
         if n_geoms == 0:
-            empty_int = np.array([], dtype=gs.np_int)
-            return 0, np.full((0, 0), fill_value=-1, dtype=gs.np_int), empty_int, empty_int, False, False, False, False
+            empty_pairs = np.empty((0, 2), dtype=gs.np_int)
+            return 0, np.full((0, 0), fill_value=-1, dtype=gs.np_int), empty_pairs, False, False, False, False
 
         # Links delegated to IPC coupler (skip pair only when BOTH are IPC-handled)
         ipc_delegated_link_idxs = set()


genesis/constants.py
```diff
@@ -62,6 +63,18 @@ class constraint_solver(IntEnum):
...
 
+# rigid solver broadphase traversal strategy
+class broadphase_traversal(IntEnum):
+    SAP = 0
+    ALL_VS_ALL = 1
+
+
+# rigid solver broadphase filter chain (combinable bitmask)
+class broadphase_filter(_IntFlag):
+    SPHERE = 1
```
hughperkins: Note: it seems reasonable to me to:
- remove SPHERE from this enum, for this PR, since this PR doesn't use it, and then
- remove this enum, at that point, since it wouldn't be being used in any way

My experiments to date have not shown any benefit on any other filter combinations over simply using `AABB`.

hughperkins: I've removed both SPHERE and this enum.

genesis/engine/solvers/rigid/collider/collider.py
```diff
@@ -661,19 +679,40 @@ def detection(self) -> None:
...
-            self._collider_info,
-            self._solver._errno,
-        )
+        if self._solver._static_rigid_sim_config.broadphase_traversal == gs.broadphase_traversal.ALL_VS_ALL:
+            func_broad_phase_all_vs_all_clear(
+                self._solver.links_state,
+                self._solver.links_info,
+                self._collider_state,
+                self._solver._static_rigid_sim_config,
+            )
```
duburcqa: Why clear is necessary for all_vs_all but not the original broad phase?

genesis/utils/array_class.py
```diff
@@ -753,6 +757,9 @@ def get_collider_info(solver, n_vert_neighbors, collider_static_config, **kwargs
         max_collision_pairs=V(dtype=gs.qd_int, shape=()),
         max_contact_pairs=V(dtype=gs.qd_int, shape=()),
         max_collision_pairs_broad=V(dtype=gs.qd_int, shape=()),
+        n_valid_pairs=V_SCALAR_FROM(dtype=gs.qd_int, value=n_valid_pairs),
+        valid_pairs_a=V(dtype=gs.qd_int, shape=(max(n_valid_pairs, 1),)),
+        valid_pairs_b=V(dtype=gs.qd_int, shape=(max(n_valid_pairs, 1),)),
```
duburcqa: I don't like introducing redundant information. Please either remove 'collision_pair_idx' or 'valid_pairs_a + valid_pairs_b', but do not keep both.

genesis/options/solvers.py
```diff
@@ -521,11 +524,19 @@ class RigidOptions(Options):
...
+    broadphase_traversal: gs.broadphase_traversal | None = None
+
     def __init__(self, *, contact_resolve_time: float | None = None, **data):
         super().__init__(**data)
         if contact_resolve_time is not None:
             gs.logger.warning("'contact_resolve_time' is deprecated. Use 'constraint_timeconst' instead.")
 
+    def model_post_init(self, context):
+        if self.broadphase_traversal == gs.broadphase_traversal.ALL_VS_ALL:
+            if self.use_hibernation:
```
duburcqa: Use a single condition.

genesis/options/solvers.py
```diff
@@ -469,6 +469,9 @@ class RigidOptions(Options):
     use_gjk_collision: bool, optional
         Whether to use GJK for collision detection instead of MPR. More stable but much slower. Defaults to
         `sim_options.requires_grad`.
+    broadphase_traversal : gs.broadphase_traversal, optional
+        Broadphase traversal strategy. ``SAP`` (sweep-and-prune) or ``ALL_VS_ALL``.
+        Defaults to ``None`` (auto: ``SAP`` on CPU, ``ALL_VS_ALL`` on GPU).
```
duburcqa: Is it relevant to give the should to the user? Will the result of collision detection be any different? If not, I'm strongly in favour of burying as an implementation details inside Rigid solver internal flags.

duburcqa: I'm in favour of forcing 'SAP' if mujoco compatibility is enabled (without documenting this).

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -348,6 +348,13 @@ def build(self):
         self._func_vel_at_point = func_vel_at_point
         self._func_apply_coupling_force = func_apply_coupling_force
 
+    def _resolve_broadphase_traversal(self):
+        if self._options.broadphase_traversal is not None:
+            return self._options.broadphase_traversal
+        if gs.backend == gs.cpu or self._use_hibernation or self._enable_heterogeneous:
```
duburcqa, line 363: Why is it not enabled for heterogeneous entities? If it is because this feature is not supported by ALL vs ALL, then I'm not convinced we should be handled this way rather than raising an exception.

Anyway, if you want to do this, and explicit option is kept, I would advice to update the documentation for auto mode stating that it while use by default what is going to give the best performance while being support.

duburcqa: By the way, you should document somewhere in the code why it is resolved this way.

duburcqa: Know that I'm thinking about it, I think you should probably fallback to SAP by default if the number of collision pairs gets very large, unless I'm missing something. Like, is ALL vs ALL still more efficient if there is 5000 candidate collision pairs? I would use 'examples/collision/tower.py' as benchmark.

genesis/options/solvers.py
duburcqa: This documentation is not good enough by the way. If you want to expose options, you must give some pointers to the user for them to adjust these parameters. What are supposed to be SAP and ALL_VS_ALL? What is the benefit of one compared to the user? Are you expecting them to tune it? If so, when it is likely to be relevant? Just 2 sentences, no need to write a report.


By the way, we stick to 120 chars linewidth even in documentation, and only moving to new line if necessary.

genesis/utils/array_class.py
duburcqa: If keeping both is necessary for efficiency, this should be documented.

duburcqa (changes requested):
If you want to expose this as an option, you need to make sure we have proper unit test coverage for it.

If the results between both variants are supposed to match and it is only a performance, only one unit test making sure that it is actually the case would be fine. Otherwise, it is more complicated. In theory all the unit tests validating the physics should be parameterised for both. In this case, it sounds overkill. But it would be great to use parameterisation for the pure physics-related unit tests. There are only like 10 of them.

But from what I understand, your new variant falls into the first category so it should be easier.

genesis/utils/array_class.py
duburcqa: Note that we are not trying to minimise changes in every PRs. We have unit tests for this. But rather to avoid increasing the maintenance burden and technical debt. In this case, getting rid of redundant 'collision_pair_idx' does induce more changes but also reduces the final maintenance burden, so I see this as a plus overall.

genesis/engine/solvers/rigid/collider/collider.py
hughperkins: since SAP is single-threaded (per env), then it can clear within the same kernel.

All vs all is multi-threaded, and it is not easily possible to sync thread writes to global memory without a new kernel launch.

genesis/utils/array_class.py
hughperkins: We need both:
- SAP needs collision_pair_idx because it provides fast look up of (geom_index_1, geom_index_2)
- all vs all needs the compact list of valid geom pairs, which it iterates over

genesis/options/solvers.py
hughperkins: When using AABB filter, the results of broadphase won't be any different.

No matter what filters are used, the output of narrowphase will be unchanged..

However, all vs all is quadratic in time wrt number of geoms, which might not work well for certain scenes.

> I'm in favour of forcing 'SAP' if mujoco compatibility is enabled (without documenting this).

Makes sense to me.

> This documentation is not good enough by the way. If you want to expose options, you must give some pointers to the user for them to adjust these parameters.

This appears to contradict the earlier position of not providing the option to the user. Though on the whole, as a user, I would prefer to:
1. have reasonable defaults/heuristics, but
2. have the option available
I feel

(like eg cpus 99.999% of the time, we just let them 'do their thing', but it's still nice to be able to turn hyper-threading on or off, and play around with that, sometimes, I feel).

genesis/engine/solvers/rigid/rigid_solver.py
hughperkins: > Why is it not enabled for heterogeneous entities? If it is because this feature is not supported by ALL vs ALL, then I'm not convinced we should be handled this way rather than raising an exception.

why should hibernate automatically trigger SAP whereas heterogeneous raises an exception?

note that a big reason to handle these things automatically is because in earlier version of this PR, I modified the tests instead, to change the option, but having it detect automaically is much cleaner I feel.

but also: by default it will be using all vs all on gpu. As a user, would you expect that running Genesis on GPU, with heterogeneous objects, would throw an exception? 🤔

hughperkins:
> If you want to expose this as an option, you need to make sure we have proper unit test coverage for it.
> 
> If the results between both variants are supposed to match and it is only a performance, only one unit test making sure that it is actually the case would be fine. Otherwise, it is more complicated. In theory all the unit tests validating the physics should be parameterised for both. In this case, it sounds overkill. But it would be great to use parameterisation for the pure physics-related unit tests. There are only like 10 of them.
> 
> But from what I understand, your new variant falls into the first category so it should be easier.

results between all vs all and sap should match. They are both ultimately filtering on AABB bounding boxes. We are just using a different approach to find the overlapping bounding boxes.

Note that the *ordering* of the outputs will differ. But since we added a sort after narrowphase, the input to the solver will be the same either way.

genesis/engine/solvers/rigid/rigid_solver.py
hughperkins: > Know that I'm thinking about it, I think you should probably fallback to SAP by default if the number of collision pairs gets very large, unless I'm missing something. Like, is ALL vs ALL still more efficient if there is 5000 candidate collision pairs? I would use 'examples/collision/tower.py' as benchmark.

It's quadratic in number of geoms, so SAP will be faster for large numbers of collision pairs. The exact cut-off will be hardware specific.

hughperkins: (will also be scene dependent, since the list of valid geom pairs is created at scene-build time. Some scenes with large nubmer of geoms might have fewer valid geom pairs than other scenes with fewer geoms).

hughperkins:
> If you want to expose this as an option, you need to make sure we have proper unit test coverage for it.
> 
> If the results between both variants are supposed to match and it is only a performance, only one unit test making sure that it is actually the case would be fine. Otherwise, it is more complicated. In theory all the unit tests validating the physics should be parameterised for both. In this case, it sounds overkill. But it would be great to use parameterisation for the pure physics-related unit tests. There are only like 10 of them.
> 
> But from what I understand, your new variant falls into the first category so it should be easier.

Which tests are the 'pure physics related unit tests'?

genesis/engine/solvers/rigid/collider/collider.py
duburcqa: I don't get it. You should simply call `func_broad_phase_all_vs_all_clear` in the preamble of `func_broad_phase_all_vs_all` (if it was a Quadrants function of course). It would still be another GPU kernel, but a single Quadrants kernel, which is what I was referring to.

genesis/utils/array_class.py
hughperkins: (What we could do I suppose is to create a subclass of ColliderInfo:

collider_info.sap_info, for sap things, and 

collider_info.all_vs_all_info, for all_vs_all things

This would keep the structs cleaner and factorized.

Thoughts?
)

hughperkins: oh, just saw 'If keeping both is necessary for efficiency, this should be documented.' => will look into this.

duburcqa: > all vs all needs the compact list of valid geom pairs, which it iterates over

It is really faster than looping over all possible pairs and using the fast look up (geom_index_1, geom_index_2) ? Conversely, what would SAP need collision_pair_idx ? It is not possible to iterate over all valid geom pairs directly?

genesis/options/solvers.py
duburcqa: The standpoint is only exposing options for which we are expecting advanced users to tinker with. Otherwise, it is just noise and detrimental, both for users and use as maintainers to ensure proper coverage. In this particular case, I'm slightly inclined to think we should not expose any option, because it is only about performance and it seems that the heuristics should be correct 99% of the time, and just marginally slower the rest of the time.

genesis/utils/array_class.py
hughperkins: This isn't a loop. We are spawning gpu threads. gpu threads *MUST* be spawned from a compact list, otherwise we get thread divergence, wasted threads etc, and its sllooowwww

If you prefer, I can store a compact list of the valid indices into collision_pair_idx, if that will make you feel more comfortable?

genesis/engine/solvers/rigid/rigid_solver.py
duburcqa: > why should hibernate automatically trigger SAP whereas heterogeneous raises an exception?

You are right, both should be treated the same. It is just not clear to me why you need special handling for hibernation and heterogeneous entities. But the idea is simple: ideally default should be as simple as possible, and defaults should be properly documented (if too complex, we can be vague, as long as it is not misleading or plain untrue). 

> It's quadratic in number of geoms, so SAP will be faster for large numbers of collision pairs. The exact cut-off will be hardware specific.

Sure, it is hardware specific. But anyway, let's just be conservative and set a limit that is reasonable. It won't be optimal in all cases, but it is not a big deal. It cannot be worst than no limit at all, if set to something like 16k I guess. No need for an extensive study. That is what I did for parallel Hessian and it worked out reasonably well.

genesis/options/solvers.py
hughperkins: We dont have a 99% correct heuristic. As you point out, when there are tons of valid geom pairs, all vs all will be slow.

genesis/engine/solvers/rigid/rigid_solver.py
duburcqa: > (will also be scene dependent, since the list of valid geom pairs is created at scene-build time. Some scenes with large nubmer of geoms might have fewer valid geom pairs than other scenes with fewer geoms).

Yes it is, but it is not a problem. Being scene dependent is not an issue. As long as it is the same for everybody running the exact same scene.

genesis/utils/array_class.py
hughperkins: for SAP, yes, doing an O(1) lookup is obviusly faster than an O(N*N) table scan.

duburcqa:
> results between all vs all and sap should match. They are both ultimately filtering on AABB bounding boxes. We are just using a different approach to find the overlapping bounding boxes.

Ok perfect. Then it falls in the easy category: only one unit test making sure that it is actually the case would be fine. This is only if you decide to expose this feature as an option.

> Which tests are the 'pure physics related unit tests'?

Answering this question would require me to go through all our unit tests one by one to determine this. Since it is not necessary in this case, let's postpone this analysis to when it will be truly necessary?

genesis/engine/solvers/rigid/rigid_solver.py
hughperkins: we cannot get a precalculated list of valid collision pairs for heterogeneous pairs. We would have to have a per-env list. since heterogeneous objects are currently a minority usecases (AFAIK), I'm not intending to handle this in this PR.

A similar argument applies to hibernation.

hughperkins:
> > results between all vs all and sap should match. They are both ultimately filtering on AABB bounding boxes. We are just using a different approach to find the overlapping bounding boxes.
> 
> Ok perfect. Then it falls in the easy category: only one unit test making sure that it is actually the case would be fine. This is only if you decide to expose this feature as an option.
> 
> > Which tests are the 'pure physics related unit tests'?
> 
> Answering this question would require me to go through all our unit tests one by one to determine this. Since it is not necessary in this case, let's postpone this analysis to when it will be truly necessary?

Why would not exposiong the feature as an option obviate the need for such a unit test? We will still have both forms of broadphase either way, one on cpu and one on gpu. Or, do you mean, since we have both gpu and cpu tests, so we are already exercising both branches?

genesis/engine/solvers/rigid/rigid_solver.py
duburcqa: I see. Good point. Just keep the current logics then. As long as it is documented. Still, you should raise an exception if the choice of the algorithm is exposed as an option and the user explicitly request ALL-vs-ALL in cases where it is not supported, if not already the case. This should happen as early as possible, in _model_post_init.

EDIT: I noticed it is already the case for hibernation, so you just need to handle heterogeneous entities. And in this case, you can only raise the exception during build.

hughperkins: >  Still, you should raise an exception if the choice of the algorithm is exposed as an option and the user explicitly request ALL-vs-ALL in cases where it is not supported

Good point

hughperkins: > As long as it is documented

You mean, in a user-visible docstring?

genesis/utils/array_class.py
duburcqa: > for SAP, yes, doing an O(1) lookup is obviusly faster than an O(N*N) table scan.

Sure it is faster, but is it 1% or 20%? if it is 1%, I would rather keep it simple.

> If you prefer, I can store a compact list of the valid indices into collision_pair_idx, if that will make you feel more comfortable?

it is not about feeling comfortable or not. I just want to make sure we are not introducing a new field that could be avoided. If you tell me that it is impossible to factories the code to use the same representation of the valid collision geometries without causing a significant slowdown, then I'm fine obviously.

genesis/options/solvers.py
duburcqa: Sure, but then, why not just set a threshold to automatic select SAP if the number of valid collision pairs is large? The default broad phase algorithm is selected during build already, so you should be able to take into account the actual number of valid collision pairs to select which algorithm to use.

duburcqa: Actually 99% is way too high, but like basically anything above 51% should be sufficient to switch the default.

genesis/utils/array_class.py
hughperkins: > Sure it is faster, but is it 1% or 20%? if it is 1%, I would rather keep it simple.

Using a table scan, when we are striving for maximum performance, to save an obvious 2d index, makes me recoil viscerally in horror.

But sure, we can run the experiment.

But even if it turns out to be only 1%, I feel that 1% is important. There are only a hundred 1% between 0% and 100%. Every 1% is important I feel.

hughperkins: Like, the amount of work I typically need to do to save 5% is typically 100s or 1000s of lines of code. A couple of lines of code saving 1% I would say is *exceptional* good value.

genesis/options/solvers.py
hughperkins: because:
1. we dont know what the threshold should be
2. figuring it out would take a bunch of time, and we are short on time
3. the number will be hardware specific, so we'd need to run the experiment on a range of current hardware, and on all future hardware.

Sure, we can come up with some sensible 'best guess', but it will be wrong much of the time, and the user should be able to tune it I feel.

hughperkins: (or use perf dispatch)

duburcqa:
> Why would not exposiong the feature as an option obviate the need for such a unit test? We will still have both forms of broadphase either way, one on cpu and one on gpu. Or, do you mean, since we have both gpu and cpu tests, so we are already exercising both branches?

It is a good question. It does not. It is just a tradeoff between cost and return on investment. If there is no exposed option to select a specific algorithm, testing the various algorithms will require monkeypatching or similar, which may be quite involving. It is doable, but the return on investment must be significant. I think it is the case for analytical contact detection specializations, but I'm not convinced here. Moreover, since the selection is automatic and deterministic for a given scene, then we know for sure that every single scene that are currently running properly in the unit tests will keep running fine for all users, since we have both gpu and cpu tests, so we already exercising everything branches that users may exercise on their side even if they tweak the options (for this specific feature at least).

hughperkins:
> Moreover, since the selection is automatic and deterministic for a given scene

I'm not sure what you mean by 'deterministic for a given scene'. It depends on the hardware.

duburcqa:
> I'm not sure what you mean by 'deterministic for a given scene'. It depends on the hardware.

No It does not. Why are you saying it does? Here is your own code:
```python
    def _resolve_broadphase_traversal(self):
        if self._options.broadphase_traversal is not None:
            return self._options.broadphase_traversal
        if gs.backend == gs.cpu or self._use_hibernation or self._enable_heterogeneous:
            return gs.broadphase_traversal.SAP
        return gs.broadphase_traversal.ALL_VS_ALL
```

hughperkins:
> > I'm not sure what you mean by 'deterministic for a given scene'. It depends on the hardware.
> 
> No It does not. Why are you saying it does? Here is your own code:
> 
> ```python
>     def _resolve_broadphase_traversal(self):
>         if self._options.broadphase_traversal is not None:
>             return self._options.broadphase_traversal
>         if gs.backend == gs.cpu or self._use_hibernation or self._enable_heterogeneous:
>             return gs.broadphase_traversal.SAP
>         return gs.broadphase_traversal.ALL_VS_ALL
> ```

`cpu` is a type of hardware I feel?

genesis/utils/array_class.py
hughperkins: Here are the benchmark results using table scan instead of index lookup:

results (9)

Branch here https://github.com/Genesis-Embodied-AI/Genesis/compare/main...hughperkins:Genesis:hp/sap-table-scan?expand=1 (happy to make draft PR if you want, in order to trigger official benchmarks?)

duburcqa:
It is the type of backend, not hardware. On MacBook Pro, both CPU and GPU corresponds to the same hardware in my view.

genesis/options/solvers.py
duburcqa: Your current implementation is even worse anyway. Any best guess would be better what no threshold at all. Like 16k if you want to be conservative. Even 64k or 256k if you are still not sure.

hughperkins:
> It is the type of backend, not hardware. On MacBook Pro, both CPU and GPU corresponds to the same hardware in my view.

It's just semantics, and we can define any. term to mean anything we would like. I would opine that defining 'cpu' and 'gpu' to be the same hardware would be an unusual definition for 'hardware'. I think maybe 'platform' or 'os' might be closer to the meaning you are using for 'hardware' I feel?

genesis/engine/solvers/rigid/rigid_solver.py
duburcqa: Yes.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -348,6 +348,13 @@ def build(self):
         self._func_vel_at_point = func_vel_at_point
         self._func_apply_coupling_force = func_apply_coupling_force
 
+    def _resolve_broadphase_traversal(self):
```
duburcqa, line 351: You forgot to take into account collision pairs that are involved in dynamically registered weld equality constraints:
```python
        # Filter out collision pairs that are involved in dynamically registered weld equality constraints
        for i_eq in range(rigid_global_info.n_equalities[None], constraint_state.qd_n_equalities[i_b]):
            if equalities_info.eq_type[i_eq, i_b] == gs.EQUALITY_TYPE.WELD:
                i_leqa = equalities_info.eq_obj1id[i_eq, i_b]
                i_leqb = equalities_info.eq_obj2id[i_eq, i_b]
                if (i_leqa == i_la and i_leqb == i_lb) or (i_leqa == i_lb and i_leqb == i_la):
                    is_valid = False
```

duburcqa: Hum.. actually no since you are calling `func_check_collision_valid`. My bad.

genesis/options/solvers.py
hughperkins: I'm not sure I udnerstand what you are saying? The current implemetnation shows favorable speedsup on the benchmarks, https://github.com/Genesis-Embodied-AI/Genesis/runs/69006028702

Screenshot 2026-03-29 at 11 24 54
Screenshot 2026-03-29 at 11 24 56
Screenshot 2026-03-29 at 11 25 01
Screenshot 2026-03-29 at 11 25 03

duburcqa: I would rather stick to what we have rather than making performance worst for existing scenarios. So I would much prefer a best guess that disable ALL vs ALL more often than necessary, rather than enable it in cases where it impedes performance.

duburcqa: > I'm not sure I udnerstand what you are saying? The current implemetnation shows favorable speedsup on the benchmarks

I'm saying that your implementation, that lacks threshold on the number of valid collision pairs, is strictly worst and selecting the most efficient kernel than taking into account a best guess threshold.

hughperkins: I think we are arguing two things in parallel:
1. should we expose an option to the users?
2. what heuristic should we use

We can both agree that your proposed new heuristic is better than the current one.

However, I would opine it is not sufficient that we can remove control from the users, and not expose the ability to switch between implementations as an option.

hughperkins: (on the whole, I woud prefer a simpler heuristic for now tbh, until such time as we encounter a concrete situation where we need a more complicated heuristic).

genesis/utils/array_class.py
```diff
@@ -717,11 +717,16 @@ class StructColliderInfo(metaclass=BASE_METACLASS):
...
+    # (i_ga, i_gb) -> dense pair index, or -1 if invalid. Used by SAP broadphase, narrowphase, and contact cache.
     collision_pair_idx: V_ANNOTATION
     max_possible_pairs: V_ANNOTATION
     max_collision_pairs: V_ANNOTATION
     max_contact_pairs: V_ANNOTATION
     max_collision_pairs_broad: V_ANNOTATION
+    # Compact list of valid collision pairs. Used by all-vs-all broadphase to dispatch valid pairs to GPU threads.
+    n_valid_pairs: V_ANNOTATION
+    valid_pairs_a: V_ANNOTATION
+    valid_pairs_b: V_ANNOTATION
```
duburcqa: Could you store this as a pair directly:
```python
valid_collision_pairs: V_ANNOTATION
```

```python
valid_collision_pairs=V(dtype=gs.qd_ivec2, shape=(max(n_valid_pairs, 1),)),
```

duburcqa:
>  I think maybe 'platform' or 'os' might be closer to the meaning you are using for 'hardware' I feel?

By hardware I mean the manufacturer's reference of the CPU or GPU depending on the backend. The output of Genesis's "Running on ..." init message:
image

The output of `_, device_name, _, _ = get_device(gs.backend)`

duburcqa:
> I'm not sure what you mean by 'deterministic for a given scene'. It depends on the hardware.

What I mean 'deterministic for a given scene' is genesis init arguments + all scene options + all entities being added to the scene.

duburcqa:
What usually means "depending on the hardware" means that if you send me your script and I run it on my machine, then I will get a different output, not because of floating point precision, but because a different algorithm that is not equivalent is being used.

genesis/engine/solvers/rigid/rigid_solver.py
duburcqa: By the way, it would be great to add a comment in the code that explains why hibernation and heterogeneous entities are not supported for now (I mean, what would be required to support them, especially heterogeneous entities, because this feature is probably going to become wildly used in the future).

hughperkins:
> > I'm not sure what you mean by 'deterministic for a given scene'. It depends on the hardware.
> 
> What I mean 'deterministic for a given scene' is genesis init arguments + all scene options + all entities being added to the scene.

I dont think we should have that as a goal. If we are running on some ancient 3070, it is logical to expect that we won't be able to run the same things as on a Blackwell. For example, cuda graph is not supported on 3070.

duburcqa:
I don’t see the relation between CUDA graph not being supported on all platforms and running algorithms that are not equivalent if not supported.

hughperkins:
> I don’t see the relation between CUDA graph not being supported on all platforms and running algorithms that are not equivalent if not supported.

Ok. Well, there's a hazy line.

No matter what algorithm we use for broadphase, the output of narrowphase should be identical. If it's not, that's an error.

hughperkins:
> > I don’t see the relation between CUDA graph not being supported on all platforms and running algorithms that are not equivalent if not supported.
> 
> Ok. Well, there's a hazy line.
> 
> No matter what algorithm we use for broadphase, the output of narrowphase should be identical. If it's not, that's an error.

Arguably then, *any* broadphase algorithm could be considered functionally identical, from the point of view of the solver, which is what we most care about I think?

duburcqa:
> Arguably then, any broadphase algorithm could be considered functionally identical, from the point of view of the solver, which is what we most care about I think?

Exactly!

genesis/options/solvers.py
duburcqa: How about a middle ground: you keep your option, and you add some large threshold in the heuristics?

hughperkins: Could do. But how would we decide objecdtively on what is 'large'? After all, even dex_hand and box 6 are showing a benefit.

If you can find a scene that does hit this limit we could base on that. Maybe coffee beans?

Otherwise, I think it would be better to be agile, and deal with that issue when we hit it.

genesis/engine/solvers/rigid/rigid_solver.py
hughperkins: I don't know what would be required to support them in the future. I would prefer not to spend time thinking about that for now please.

genesis/options/solvers.py
duburcqa: On my Mac, SAP is faster for this scene:

```python
import genesis as gs

gs.init(backend=gs.gpu, precision="32", performance_mode=True)

scene = gs.Scene(
        dt=0.004,
    ),
        max_collision_pairs=150,
        # broadphase_traversal=gs.broadphase_traversal.SAP,
    ),
    show_viewer=False,
)

# create pyramid of boxes
box_width, box_length, box_height = 0.25, 2.0, 0.1
num_stacks = 50
for i in range(num_stacks):
    if i % 2 == 0:  # horizontal stack
        box_size = (box_width, box_length, box_height)
        box_pos_0 = (-0.4 * box_length, 0, i * (box_height - 1e-3) + 0.5 * box_height)
        box_pos_1 = (+0.4 * box_length, 0, i * (box_height - 1e-3) + 0.5 * box_height)
    else:  # vertical stack
        box_size = (box_length, box_width, box_height)
        box_pos_0 = (0, -0.4 * box_length, i * (box_height - 1e-3) + 0.5 * box_height)
        box_pos_1 = (0, +0.4 * box_length, i * (box_height - 1e-3) + 0.5 * box_height)
    for box_pos in (box_pos_0, box_pos_1):
        scene.add_entity(
            gs.morphs.Box(
                size=box_size,
                pos=box_pos,
            ),
        )

scene.build(n_envs=950)

for i in range(10):
    scene.step()
```

On Nvidia GPU it is crashing even for 50 envs:
```python
Exception ignored in atexit callback: <function destroy at 0x7e2afbb3b130>
  File "/mnt/home/duburcqa/workspace/src/genesis/genesis/__init__.py", line 404, in destroy
    qd.reset()
  File "/mnt/home/duburcqa/.venv/lib/python3.10/site-packages/quadrants/lang/misc.py", line 203, in reset
    impl.reset()
  File "/mnt/home/duburcqa/.venv/lib/python3.10/site-packages/quadrants/lang/impl.py", line 557, in reset
    pyquadrants.clear()
  File "/mnt/home/duburcqa/.venv/lib/python3.10/site-packages/quadrants/lang/impl.py", line 533, in clear
    self._prog.finalize()
RuntimeError: [cuda_driver.h:operator()@93] CUDA Error CUDA_ERROR_ILLEGAL_ADDRESS: an illegal memory access was encountered while calling stream_synchronize (cuStreamSynchronize)
```

genesis/engine/solvers/rigid/rigid_solver.py
duburcqa: > I don't know what would be required to support them in the future. I would prefer not to spend time thinking about that for now please.

Then do not think about what would be required to support them in the future. Still, you must mention why it is not supported at the moment. It is not acceptable to disable a feature without any explanation. You justify to be able to yourself here.

hughperkins: added a comment

hughperkins: Oh, "in a user visible docstring" 🤔 

Oh wait, taht was a separate comment, but I still need to do that.

duburcqa: > added a comment

Thank you very much. It is very informative and helpful for me as a maintainer.

> Oh, "in a user visible docstring" Oh wait, taht was a separate comment, but I still need to do that.

Yes, still visible doctring update missing.

hughperkins: wait, which function would this 'user visible docstring' be attached to? Maybe the options struct I guess?

hughperkins: added docstring to `class broadphase_traversal(IntEnum):`

genesis/options/solvers.py
hughperkins: Got Opus to run a grid search:

Screenshot 2026-03-29 at 22 40 21

It appears that for box pyramid:
- beyond 7, the Jacobian explodes
- below 7, all vs all is always faster

hughperkins: wait. that doesnt look right. going to question opus on that.

hughperkins: from Opus:

  The setting is being applied correctly:
  • Requested SAP → config uses SAP (0)
  • Requested ALL_VS_ALL → config uses ALL_VS_ALL (1)

  So the FPS being nearly identical is a real result, not a bug. At these pyramid sizes (3-6 cubes, meaning 6-21 boxes), the broadphase is not the bottleneck — the constraint
  solver dominates runtime. The number of valid geometry pairs is small enough that both SAP's sequential sweep and ALL_VS_ALL's parallel pair checking finish nearly instantly.
  The broadphase difference would likely only show up with many more geometries where the O(n_geoms log n_geoms) vs O(n_valid_pairs) distinction matters. Let me cancel that
  diagnostic job.

  $ ssh -A cluster 'scancel 272977' 2>&1; true 622ms

  Summary: The broadphase_traversal setting is correctly applied (verified). The near-identical FPS is genuine — with only 6-21 boxes, broadphase is negligible compared to
  constraint solving. To see a meaningful SAP vs ALL_VS_ALL difference, you'd likely need scenes with many more geometries (e.g. hundreds of objects) or fewer n_envs so larger
  n_cubes values don't exceed the Jacobian size limit.
  Want me to re-run with fewer n_envs (e.g. 512 or 1024) so we can test larger pyramid sizes where broadphase might matter more?

hughperkins: lets try this: "can you run for a new scene you create where you simply position a number of blocks sapced out along the x axis, so they are not colliding with each other?"

hughperkins: Ok, Opus wrote hte following script:

```
"""Benchmark FPS vs num_geoms: SAP vs ALL_VS_ALL broadphase.

Uses a few dynamic boxes + many fixed boxes to create lots of valid geometry
pairs while keeping DOFs low enough for 4096 envs.

Usage:
    python bench_broadphase.py
    python bench_broadphase.py --fixed-boxes 50,100,200,500,1000

Outputs:
    ~/ais/pelican/tmp/bench_broadphase_results.csv
    ~/ais/pelican/tmp/bench_broadphase_plot.png
"""

import argparse
import atexit
import csv
import json
import sys
import textwrap
from pathlib import Path

sys.path.insert(0, str(Path(__file__).resolve().parent.parent / "cmp-tooling"))

from common import (
    CONTAINER_IMAGE, REMOTE_HOME, REMOTE_WORKSPACE,
    ensure_ssh_agent, log, run_in_container, run_in_container_bg,
    ssh, ssh_output, wait_for_jobs,
)

REMOTE_GENESIS = f"{REMOTE_WORKSPACE}/genesis"
REMOTE_TMP = f"{REMOTE_WORKSPACE}/tmp"
BRANCH = "hp/broad-all-v-all-mar26"
FORK = "hughperkins"

WORKER_SCRIPT = textwrap.dedent(r"""
    import argparse, json, sys, time
    import numpy as np

    parser = argparse.ArgumentParser()
    parser.add_argument("--n-fixed", type=int, required=True)
    parser.add_argument("--n-dynamic", type=int, default=5)
    parser.add_argument("--broadphase", type=str, required=True)
    parser.add_argument("--n-envs", type=int, default=4096)
    parser.add_argument("--warmup", type=float, default=10.0)
    parser.add_argument("--record", type=float, default=5.0)
    args = parser.parse_args()

    import genesis as gs
    gs.init(backend=gs.gpu, precision="32", logging_level="warning")

    broadphase = getattr(gs.broadphase_traversal, args.broadphase)
    scene = gs.Scene(
        rigid_options=gs.options.RigidOptions(dt=0.01, broadphase_traversal=broadphase),
        show_viewer=False, show_FPS=False,
    )

    box_size = 0.25
    spacing = 2.0

    for i in range(args.n_dynamic):
        scene.add_entity(gs.morphs.Box(
            size=[box_size, box_size, box_size],
            pos=(i * spacing, 0.0, 0.5),
        ))

    for i in range(args.n_fixed):
        row = i // 50
        col = i % 50
        scene.add_entity(gs.morphs.Box(
            size=[box_size, box_size, box_size],
            pos=(col * spacing, (row + 1) * spacing, 0.5),
            fixed=True,
        ))

    n_geoms = args.n_dynamic + args.n_fixed
    n_valid_pairs = args.n_dynamic * args.n_fixed + args.n_dynamic * (args.n_dynamic - 1) // 2
    print(f"n_fixed={args.n_fixed} n_dynamic={args.n_dynamic} n_geoms={n_geoms} "
          f"n_valid_pairs~={n_valid_pairs} broadphase={args.broadphase}", file=sys.stderr, flush=True)

    t0 = time.time()
    scene.build(n_envs=args.n_envs)
    compile_time = time.time() - t0

    cfg = scene.rigid_solver._static_rigid_sim_config
    print(f"  actual broadphase={cfg.broadphase_traversal}", file=sys.stderr, flush=True)

    import quadrants as qd
    qd.sync()

    print(f"  warming up ({args.warmup}s)...", file=sys.stderr, flush=True)
    t_start = time.time()
    while time.time() - t_start < args.warmup:
        scene.step()
    qd.sync()

    print(f"  recording ({args.record}s)...", file=sys.stderr, flush=True)
    t_start = time.time()
    num_steps = 0
    while time.time() - t_start < args.record:
        scene.step()
        num_steps += 1
    qd.sync()
    elapsed = time.time() - t_start
    fps = int(num_steps * max(args.n_envs, 1) / elapsed)

    print(f"  FPS={fps} compile={compile_time:.1f}s", file=sys.stderr, flush=True)
    print(json.dumps(dict(
        n_fixed=args.n_fixed, n_dynamic=args.n_dynamic, n_geoms=n_geoms,
        n_valid_pairs=n_valid_pairs, broadphase=args.broadphase,
        n_envs=args.n_envs, compile_time=round(compile_time, 2),
        fps=fps, num_steps=num_steps, elapsed=round(elapsed, 3),
    )), flush=True)
""").strip()


def parse_args():
    p = argparse.ArgumentParser(description=__doc__,
                                formatter_class=argparse.RawDescriptionHelpFormatter)
    p.add_argument("--fixed-boxes", default="20,50,100,200,500",
                   help="Comma-separated list of fixed box counts")
    p.add_argument("--n-dynamic", type=int, default=5)
    p.add_argument("--n-envs", type=int, default=4096)
    p.add_argument("--warmup", type=float, default=10.0)
    p.add_argument("--record", type=float, default=5.0)
    p.add_argument("--partition", default="rtx-mid")
    p.add_argument("--time", type=int, default=120, help="Allocation time in minutes")
    return p.parse_args()


def main():
    args = parse_args()
    fixed_list = [int(x) for x in args.fixed_boxes.split(",")]
    broadphases = ["SAP", "ALL_VS_ALL"]
    container_name = "pelican-broadphase"
    job_name = "pelican-broadphase"
    branch_safe = BRANCH.replace("/", "-")
    remote_genesis_branch = f"{REMOTE_WORKSPACE}/genesis-{branch_safe}"

    ensure_ssh_agent()

    job_ids = []

    def cleanup():
        for jid in job_ids:
            log(f"Cancelling job {jid}...")
            ssh(f"scancel {jid}", check=False)

    atexit.register(cleanup)

    # --- Prepare genesis worktree on cluster ---
    log("Preparing genesis worktree on cluster...")
    fork_url = f"git@github.com:{FORK}/Genesis.git"
    ssh(f"cd {REMOTE_GENESIS} && "
        f"(git remote get-url {FORK} >/dev/null 2>&1 || git remote add {FORK} {fork_url}) && "
        f"git fetch {FORK}")
    ssh(f"if [ -d {remote_genesis_branch} ]; then "
        f"git -C {remote_genesis_branch} checkout {FORK}/{BRANCH}; "
        f"else cd {REMOTE_GENESIS} && "
        f"git worktree add {remote_genesis_branch} {FORK}/{BRANCH}; fi")
    branch_commit = ssh_output(f"git -C {remote_genesis_branch} rev-parse --short HEAD")
    log(f"Branch commit: {branch_commit}")

    # --- Write worker script to cluster ---
    log("Writing worker script to cluster...")
    ssh(f"mkdir -p {REMOTE_TMP}")
    ssh(f"cat > {REMOTE_TMP}/bench_broadphase_worker.py << 'WORKER_EOF'\n{WORKER_SCRIPT}\nWORKER_EOF")

    # --- Allocate node ---
    log(f"Allocating 1-GPU node on {args.partition} ({args.time}min)...")
    ssh(f"nohup salloc --gpus=1 --cpus-per-gpu=16 "
        f"--job-name={job_name} --partition={args.partition} "
        f"--time={args.time} --no-shell "
        f"> /tmp/salloc_{job_name}.txt 2>&1 &")

    log("Waiting for allocation...")
    jobs = wait_for_jobs([job_name])
    jobid = jobs[job_name]
    job_ids.append(jobid)
    log(f"Got job {jobid}")

    # --- Create container ---
    log("Creating container...")
    ssh(f"srun --overlap --jobid={jobid} --container-image={CONTAINER_IMAGE} "
        f"--container-name={container_name} --container-mounts={REMOTE_HOME}:{REMOTE_HOME} "
        f"--no-container-mount-home --export=NVIDIA_DRIVER_CAPABILITIES=all "
        f"echo 'Container {container_name} created'")

    # --- Build environment ---
    log("Running production_build.sh...")
    p = run_in_container_bg(
        jobid, container_name,
        f"cd {remote_genesis_branch} && rm -rf /venv && bash .github/workflows/scripts/production_build.sh",
        "[build] ",
    )
    rc = p.wait()
    if rc != 0:
        raise RuntimeError("production_build.sh failed")
    log("Environment built")

    # --- Run benchmarks ---
    results = []
    total = len(fixed_list) * len(broadphases)
    idx = 0
    for n_fixed in fixed_list:
        for bp in broadphases:
            idx += 1
            log(f"[{idx}/{total}] n_fixed={n_fixed} broadphase={bp}...")
            rc, output = run_in_container(
                jobid, container_name,
                f"source /venv/bin/activate && "
                f"cd {remote_genesis_branch} && "
                f"GS_ENABLE_NDARRAY=1 python {REMOTE_TMP}/bench_broadphase_worker.py "
                f"--n-fixed {n_fixed} --n-dynamic {args.n_dynamic} "
                f"--broadphase {bp} --n-envs {args.n_envs} "
                f"--warmup {args.warmup} --record {args.record}",
                prefix=f"{bp}-{n_fixed}",
            )
            if rc != 0:
                log(f"  FAILED (exit code {rc})")
                results.append(dict(
                    n_fixed=n_fixed, n_dynamic=args.n_dynamic,
                    n_geoms=n_fixed + args.n_dynamic,
                    n_valid_pairs=args.n_dynamic * n_fixed + args.n_dynamic * (args.n_dynamic - 1) // 2,
                    broadphase=bp, n_envs=args.n_envs,
                    compile_time=-1, fps=-1,
                ))
                continue

            for line in output.strip().splitlines():
                line = line.strip()
                if line.startswith("{"):
                    try:
                        results.append(json.loads(line))
                    except json.JSONDecodeError:
                        pass

    # --- Cancel job ---
    for jid in job_ids:
        ssh(f"scancel {jid}", check=False)
    job_ids.clear()

    # --- Write CSV ---
    local_tmp = Path.home() / "ais" / "pelican" / "tmp"
    csv_path = local_tmp / "bench_broadphase_results.csv"
    fieldnames = ["n_fixed", "n_dynamic", "n_geoms", "n_valid_pairs", "broadphase",
                  "n_envs", "compile_time", "fps", "num_steps", "elapsed"]

    with open(csv_path, "w", newline="") as f:
        writer = csv.DictWriter(f, fieldnames=fieldnames, extrasaction="ignore")
        writer.writeheader()
        for r in results:
            writer.writerow(r)
    log(f"CSV saved to {csv_path}")

    print("\n=== Results ===")
    print(f"{'n_fixed':>8} {'n_geoms':>8} {'pairs':>8} {'broadphase':>12} {'fps':>10} {'compile':>10}")
    for r in results:
        print(f"{r['n_fixed']:>8} {r['n_geoms']:>8} {r['n_valid_pairs']:>8} "
              f"{r['broadphase']:>12} {r['fps']:>10} {r['compile_time']:>10}")

    # --- Generate plot ---
    try:
        import matplotlib
        matplotlib.use("Agg")
        import matplotlib.pyplot as plt

        fig, ax = plt.subplots(figsize=(10, 6))
        for bp in broadphases:
            bp_results = [r for r in results if r["broadphase"] == bp and r["fps"] > 0]
            if bp_results:
                xs = [r["n_geoms"] for r in bp_results]
                ys = [r["fps"] for r in bp_results]
                ax.plot(xs, ys, "o-", label=bp, linewidth=2, markersize=8)

        ax.set_xlabel("Number of geometries", fontsize=13)
        ax.set_ylabel("FPS (env-steps/sec)", fontsize=13)
        ax.set_title(f"SAP vs ALL_VS_ALL broadphase\n"
                     f"({args.n_dynamic} dynamic + N fixed boxes, n_envs={args.n_envs}, "
                     f"commit={branch_commit})", fontsize=14)
        ax.legend(fontsize=12)
        ax.grid(True, alpha=0.3)

        plot_path = local_tmp / "bench_broadphase_plot.png"
        fig.tight_layout()
        fig.savefig(plot_path, dpi=150)
        plt.close(fig)
        log(f"Plot saved to {plot_path}")
    except ImportError:
        log("matplotlib not available locally, skipping plot")

    log("Done!")


if __name__ == "__main__":
    main()
```

Results:

bench_broadphase_plot

[bench_broadphase_results.csv](https://github.com/user-attachments/files/26347468/bench_broadphase_results.csv)

Screenshot 2026-03-29 at 22 40 21

Looks like, counter-intuitively, all vs all is almost always faster, and is only slower for low geom counts.

I find this suspicious tbh. So I'm happy to keep digging into this if you find this suspicious too?

hughperkins: So Opus instrumented the dispatch function:
```
  Confirmed:
  • SAP: sap_calls=11, all_vs_all_calls=0 — only _func_broad_phase_sap was called
  • ALL_VS_ALL: sap_calls=0, all_vs_all_calls=11 — only _func_broad_phase_all_vs_all was called

  The dispatch is working correctly. Each broadphase config routes to its own kernel (11 calls = 1 during build + 10 steps).
  ```

And also, thinking about it, lets say we have 100 geoms. The *maximum* possible number of valid pairs is 100^2, which is 10,000. But hte GPU has 20,000 cores. So all vs all is O(1) up to sqrt(20,000). But even when we hit 20,000 valid geom pairs, I would expect intuitively taht all vs all will keep being faster than SAP until each thread has to do at least 3-4x as much work. So plausibly up to sqrt(80,000) or so, ie about 300 geoms, all vs all will still be faster. And 300 geoms is quite a lot.

I would recommend we make the heuristic very simple:
- if it's a GPU => all vs all
- if it's a CPU => SAP

File: genesis/engine/solvers/rigid/rigid_solver.py (hunk 6 above)
Comment (duburcqa) [reply genesis/engine/solvers/rigid/rigid_solver.py:363]: > Maybe the options struct I guess?

Yes.

Comment (duburcqa) [reply genesis/engine/solvers/rigid/rigid_solver.py:363]: > added docstring to class broadphase_traversal(IntEnum):

This is fine.

File: genesis/constants.py [hunk 9]
```diff
@@ -62,6 +61,52 @@ class constraint_solver(IntEnum):
...
+    Notes
+    -----
+    ``RigidOptions.broadphase_traversal`` defaults to ``None``, which lets the
+    solver choose automatically:
+
+    - **CPU backend** → ``SAP`` (sequential sweep is efficient on CPU).
+    - **GPU backend** → ``ALL_VS_ALL`` (parallel pair checking is faster).
+    - **GPU with hibernation or heterogeneous entities** → ``SAP``
+      (``ALL_VS_ALL`` is not compatible with these features).
+    """
```
duburcqa, line 96: 120 chars linewidth.

hughperkins: wraps at 120 now

genesis/utils/array_class.py
hughperkins: Ok. note for completeness that it seems very plausible that this will cause each thread to do two sequential int reads:
```
- load collision_pairs[t_idx * 2]
- load collision_pairs[t_idx * 2 + 1]
```
In the current approach it is also sequential, but from two different arrays:
```
- load cp_a[t_idx]
- load cp_b[t_idx]
```
The difference is that in the second approach, when the warp of 32 threads reads the first load, all threads are reading from adjacent memory, and the cache read is fully coallesced.

In the first approach, when the threads read the first instruction, the therads wil be reading from strided locations, so the cache read will be less effiicent.

In practice, the geom lookup will be much slower, so likely not an issue.

hughperkins: implemented in 896b9c59

hughperkins:
Wooo 🙌  🥳

PR #2610, opened 2026-03-28
[BUG FIX] Fix motion planning crashing for short path < 3 nodes.

### LINKED ISSUE #2487: [Bug]: `shortcut_path` crash in `plan_path` with `smooth_path=True` (0.4.0 regression)
### Bug Description

`plan_path()` with `smooth_path=True` crashes about 4 out of 5 runs with a `torch.AcceleratorError`. It's a regression in 0.4.0 — same code works fine on 0.3.11.

The issue is that `shortcut_path` calls `torch.multinomial(..., 2)` which needs at least 2 valid entries to sample from. When RRTConnect happens to find a very short path (< 3 nodes), there aren't enough entries and the call fails. There's no guard for this case.

```
torch.AcceleratorError: index -2 is out of bounds: 0, range 0 to 1
```

```
File ".../genesis/utils/path_planning.py", line 282, in shortcut_path
    ind = torch.multinomial(path_mask.T, 2).sort().values.to(gs.tc_int)  # B, 2
```

### Steps to Reproduce

```python
"""Crashes ~4 out of 5 runs on Genesis 0.4.0. Works on 0.3.11."""
import genesis as gs
import numpy as np

gs.init()

franka = scene.add_entity(
    gs.morphs.MJCF(file="xml/franka_emika_panda/panda.xml")
)
scene.add_entity(gs.morphs.Cylinder(pos=(0.4, 0, 0.05), radius=0.03, height=0.02))
scene.add_entity(
    gs.morphs.Cylinder(
        pos=(0.3, 0.2, 0.05), radius=0.005, height=0.05, fixed=True
    )
)
scene.build()

franka.set_dofs_kp(
    kp=np.array([4500, 4500, 3500, 3500, 2000, 2000, 2000, 100, 100])
)
franka.set_dofs_kv(kv=np.array([450, 450, 350, 350, 200, 200, 200, 10, 10]))

ee = franka.get_link("hand")
target_pos = np.array([0.4, 0.0, 0.10])
gripper_down = np.array([0, 1, 0, 0])
seed = np.array(
    [-1.238, -0.127, 2.128, -0.699, -0.824, 0.245, -0.368, 0.04, 0.04]
)
    link=ee, pos=target_pos, quat=gripper_down, init_qpos=seed
)

# Crashes ~80% of the time on 0.4.0
path = franka.plan_path(
    qpos_goal.cpu().numpy(), num_waypoints=200, smooth_path=True
)
print(f"OK: path shape = {path.shape}")
```

The non-determinism comes from RRT's random sampling — when the trees connect in very few nodes, the resulting path is too short for `shortcut_path` to handle.

## Root Cause

In `path_planning.py`, `plan()` (line 1012) passes the RRT solution into `shortcut_path()`:

```python
sol = self.shortcut_path(
    torch.ones_like(sol[..., 0]),  # path_mask: shape [N, B], all ones
    sol,                            # path: shape [N, B, DoF]
    iterations=10,
    ...
)
```

Inside `shortcut_path()` (line 282):

```python
ind = torch.multinomial(path_mask.T, 2).sort().values.to(gs.tc_int)  # B, 2
```

`path_mask.T` has shape `[B, N]`. `torch.multinomial(input, 2)` needs each row to have at least 2 nonzero entries. When the RRT path has `N < 2` valid waypoints, this fails — there's no bounds check.

Suggested fix — just skip shortcutting for short paths:

```python
def shortcut_path(self, path_mask, path, iterations=50, ...):
    if path.shape[0] < 3:  # need at least 3 waypoints to shortcut
        return path
    for i in range(iterations):
        ind = torch.multinomial(path_mask.T, 2).sort().values.to(gs.tc_int)
        ...
```

### Expected Behavior

`plan_path(smooth_path=True)` should work regardless of how short the RRT path is. If the path is too short to shortcut, just return it as-is.

### Screenshots/Videos

_No response_

### Relevant log output

```shell

```

### Environment

- OS: macOS (Darwin 25.2.0, Apple Silicon)
- Python: 3.10.11
- PyTorch: 2.10.0
- Backend: MPS (also reproduced on CPU)

- Affected: Genesis 0.4.0
- Works on: Genesis 0.3.11

### Additional Context

_No response_

Lidang-Jiang:
## Summary
- Fix `shortcut_path()` crash when RRT path has fewer than 3 nodes
- `torch.multinomial(path_mask.T, 2)` requires at least 2 categories per row, crashes with `RuntimeError: cannot sample n_sample > prob_dist.size(-1) samples without replacement` when path has < 2 nodes
- Even with exactly 2 nodes, the sampled index gap is always 1 (`ind_mask` is always `False`), making the shortcut loop a no-op
- Add early return guard: if `path.shape[0] < 3`, return path immediately

Closes #2487

## Changes
- `genesis/utils/path_planning.py`: Add guard at the beginning of `shortcut_path()` to return early when path has fewer than 3 waypoints

## Test Plan

<details>
<summary>Before (crash with short path)</summary>

```
============================================================
BEFORE FIX: Testing shortcut_path crash condition
============================================================

Test 1: path with N=1 node (1 waypoint)
----------------------------------------
  path.shape = torch.Size([1, 1, 9])
  path_mask.shape = torch.Size([1, 1])
  path_mask.T.shape = torch.Size([1, 1])
  CRASH: RuntimeError: cannot sample n_sample > prob_dist.size(-1) samples without replacement

Test 2: path with N=2 nodes (2 waypoints)
----------------------------------------
  path.shape = torch.Size([2, 1, 9])
  path_mask.shape = torch.Size([2, 1])
  path_mask.T.shape = torch.Size([1, 2])
  multinomial result: tensor([[0, 1]])
  ind_mask (gap > 1): tensor([False])
  NOTE: With N=2, gap is always exactly 1, so ind_mask is always False
  -> Shortcut loop runs but never applies any shortcut (wasted computation)

Test 3: path with N=3 nodes (3 waypoints) - minimum for useful shortcut
----------------------------------------
  path.shape = torch.Size([3, 1, 9])
  path_mask.shape = torch.Size([3, 1])
  path_mask.T.shape = torch.Size([1, 3])
  multinomial result: tensor([[1, 2]])
  ind_mask (gap > 1): tensor([False])
  -> Can potentially shortcut if sampled indices have gap > 1

============================================================
CONCLUSION:
  N=1: CRASHES - cannot sample 2 from 1 element
  N=2: No crash but shortcut is always useless (gap always = 1)
  N>=3: Works correctly, shortcut can be applied
  Fix: Return early when path.shape[0] < 3
============================================================
```

</details>

<details>
<summary>After (no crash, short path returned as-is)</summary>

```
============================================================
AFTER FIX: Testing with early return guard
============================================================

Test 1: path with N=1 node (1 waypoint)
----------------------------------------
  path.shape = torch.Size([1, 1, 9])
  Result: returned path as-is (shape torch.Size([1, 1, 9]))
  Early return triggered: True
  OK - no crash, path returned unchanged

Test 2: path with N=2 nodes (2 waypoints)
----------------------------------------
  path.shape = torch.Size([2, 1, 9])
  Result: returned path as-is (shape torch.Size([2, 1, 9]))
  Early return triggered: True
  OK - no crash, path returned unchanged

Test 3: path with N=3 nodes (3 waypoints) - goes through normal flow
----------------------------------------
  path.shape = torch.Size([3, 1, 9])
  Result: path processed normally (shape torch.Size([3, 1, 9]))
  Early return triggered: True
  OK - normal shortcut flow executed without crash

Test 4: path with N=50 nodes (typical RRT path) - normal operation
----------------------------------------
  path.shape = torch.Size([50, 1, 9])
  Result: path processed normally (shape torch.Size([50, 1, 9]))
  OK - normal shortcut flow executed without crash

Test 5: Batched case - N=1, B=4 (multiple envs, all short paths)
----------------------------------------
  path.shape = torch.Size([1, 4, 9])
  Result: returned path as-is (shape torch.Size([1, 4, 9]))
  Early return triggered: True
  OK - no crash, all batched paths returned unchanged

============================================================
ALL TESTS PASSED
  N<3: Early return, no crash, path returned as-is
  N>=3: Normal shortcut flow, no regression
============================================================
```

</details>

<details>
<summary>Syntax check</summary>

```
Syntax OK: genesis/utils/path_planning.py
```

</details>

### DIFF
--- a/genesis/utils/path_planning.py
+++ b/genesis/utils/path_planning.py
@@ -278,6 +278,11 @@ class PathPlanner(ABC):
         iterations: int
             the number of refine iterations
         """
+        # Need at least 3 waypoints to shortcut (multinomial samples 2 indices,
+        # and the shortcut only applies when their gap > 1).
+        if path.shape[0] < 3:
+            return path
+
         for i in range(iterations):
             ind = torch.multinomial(path_mask.T, 2).sort().values.to(gs.tc_int)  # B, 2
             ind_mask = (ind[:, 1] - ind[:, 0]) > 1


duburcqa:
Usually bug fixes should come with a unit test, but it is ok for this time.

Lidang-Jiang:
Thank you for the review and the feedback\! I'll make sure to include unit tests with bug fixes in future PRs.

PR #2615, opened 2026-03-28
[MISC] Replace 2d index collision_pair_idx with table scan.

hughperkins:
Removes the O(n_geoms^2) collision_pair_idx lookup matrix and replaces it with valid_pairs_a/valid_pairs_b flat arrays + linear scan. This trades O(1) pair lookup for O(n_valid_pairs) table scan.

## Related Issue

## Checklist:

### DIFF
--- a/genesis/engine/bvh.py
+++ b/genesis/engine/bvh.py
@@ -569,11 +569,10 @@ class RigidTetLBVH(LBVH):
         """
         i_ag = self.coupler.rigid_volume_elems_geom_idx[i_a]
         i_qg = self.coupler.rigid_volume_elems_geom_idx[i_q]
-        i_ga = i_ag if i_ag < i_qg else i_qg
-        i_gb = i_qg if i_ag < i_qg else i_ag
         is_valid = False
-        for i in range(self.coupler.rigid_n_valid_pairs[None]):
-            if self.coupler.rigid_valid_pairs_a[i] == i_ga and self.coupler.rigid_valid_pairs_b[i] == i_gb:
-                is_valid = True
-                break
+        if i_ag < i_qg:
+            for i in range(self.coupler.rigid_n_valid_pairs[None]):
+                if self.coupler.rigid_valid_pairs_a[i] == i_ag and self.coupler.rigid_valid_pairs_b[i] == i_qg:
+                    is_valid = True
+                    break
         return not is_valid


genesis/engine/solvers/rigid/collider/broadphase.py
duburcqa, line 276: I don't understand why you are not just looping over valid pairs right away, since there is a continue on the very top of the for loop. this would avoid having to call `func_check_collision_valid` entirely, and `func_find_pair_idx` becomes useless.

hughperkins: Thats not how SAP works. SAP works like:
- insert the negative x edge of each geom AABB into a list
- insert the postiive x edge of each geom AABB into a list
- sort the list by x coordinate
- initialize active set as empty set
- sweep along the list
    - if hit a negative geom x:
         - you've just hit a new geom
         - check for collisions between this geom and all geoms in the active set. which means:
               - check if this is a valid collision pair
               - check AABB on other axes
               - if everythign checks out then:
                         - add pair of geoms to list of broadphase output contacts
         - add this geom to the active set
    - if fyou hit a negative geom x:
         - remove the geom frm the active set

duburcqa: Ok now I understand my mistake.

Since it is easy to guarantee the sorting order of the fields keeping track of the valid pairs, it is easy to optimize the order of pairs so that we can start checking if valid at a start idx that is monototically increasing when iterating over all geom pairs.

This would reduce the complexity from O(n_valid_pairs) to O(n_valid_pairs**0.5) <= O(n_geoms), which makes a big difference.

genesis/engine/couplers/sap_coupler.py
```diff
@@ -356,10 +356,16 @@ def _init_hydroelastic_rigid_fields_and_info(self):
         self.rigid_volume_verts_geom_idx.from_numpy(rigid_volume_verts_geom_idx_np)
         self.rigid_volume_elems_geom_idx = qd.field(gs.qd_int, shape=(self.n_rigid_volume_elems,))
         self.rigid_volume_elems_geom_idx.from_numpy(rigid_volume_elems_geom_idx_np)
-        # FIXME: Convert collision_pair_idx to field here because SAPCoupler cannot support ndarray/field switch yet
```
duburcqa, line 359: Restore this FIXME. Still applicable.

duburcqa:
Not worth the effort. I suggest to close this PR.

PR #2621, opened 2026-03-29
[MISC] Add GPU graph to decomposed solver to reduce kernel launch latency.

hughperkins:
## Description

Add GPU graph to decomposed solver to reduce kernel launch latency.
- on SM90+, the entire loop, including checking whether to continue or not, happens on the GPU
- on older CUDA GPUs, the body of the loop is launched by a single CUDA call, and the entire body runs on the GPU, with no further interaction with the host
     - checking for loop termination happens on the host side, in C++, but requires copying the loop variable from GPU to CPU
- on other GPUs, the body of the loop runs as per any other Quadrants kernel with multiple top-level loops:
     - that is, each top-level loop is launched from the C++ side, as a separate GPU kernel
     - as per pre-SM 90 CUDA GPUs, checking for loop termination happens on the C++ side, and incurs a GPU to CPU copy, for the integer counter, each iteration
- oh, actually:
     - whilst quadrants gpu graph graph can run itself on all backends
     - current decomposed hessian kernel only works with shared memory, and so doesn't work on cpu
     - in addition, decomp on main is set to only run for CUDA
         - we'll preserve this behavior on this branch for now

After this PR, decomposed *always* uses gpu graph, with one exception: if autodiff is needed.

## Related Issue

## Checklist:


hughperkins:
Running locally, wiht latest change, I get, for field:

results (25)

hughperkins:
actaully, it's curious why this pr would effect CPU time, given decomposed doesnt run except on cuda 🤔 There must be something odd happening in my benchmark script 🤔  Like, maybe it does `--backend gpu` or something? 🤔

duburcqa:
> current decomposed hessian kernel only works with shared memory, and so decomp on main is set to only run for CUDA

This is not a good reason. shared memory is already perfectly supported on Apple Metal, Vulkan and AMDGPU. What is not supported is atomics on shared memory, and non-32bits dtype.

pyproject.toml
```diff
@@ -10,7 +10,7 @@ readme = "README.md"
 requires-python = ">=3.10,<3.14"
 dependencies = [
     "psutil",
-    "quadrants==0.4.5",
+    "quadrants==0.5.0b1",
```
duburcqa: Beta.

duburcqa (dismissed):

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
```diff
@@ -215,7 +207,54 @@ def _kernel_update_search_direction(
...
+        _func_check_early_exit(constraint_state, graph_counter)
+
+
+@solver.func_solve_body.register(
+    is_compatible=lambda entities_info,
+    dofs_state,
+    constraint_state,
+    rigid_global_info,
+    static_rigid_sim_config,
+    _n_iterations: not static_rigid_sim_config.requires_grad and static_rigid_sim_config.backend == gs.cuda,
```
duburcqa: Let's see https://github.com/Genesis-Embodied-AI/Genesis/pull/2623

hughperkins:
> > current decomposed hessian kernel only works with shared memory, and so decomp on main is set to only run for CUDA
> 
> This is not a good reason. shared memory is already perfectly supported on Apple Metal, Vulkan and AMDGPU. What is not supported is atomics on shared memory, and non-32bits dtype.

Either way:
- the hessian tiled wont run on Mac or CPU
- the goal of this PR is not to turn on decomp on new paltforms, but to migrate decomp to use gpu graph
-

duburcqa:
> the hessian tiled wont run on Mac or CPU

Why? it was always supported on Mac.

duburcqa:
> the goal of this PR is not to turn on decomp on new paltforms, but to migrate decomp to use gpu graph

Sure. But then update the documentation accordingly. Instead of mentioning shared memory related issues.

hughperkins:
Note: the failures are pending a new pre-release of quadrants that addresses https://github.com/Genesis-Embodied-AI/quadrants/pulls

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
duburcqa: Merged.

hughperkins: Got it. Will take a look.

hughperkins:
Woo, +50% FPS on dex_hand field 🙌

duburcqa (approved):

PR #2639, opened 2026-04-01
[MISC] Better parallelization of add collision constraints.

hughperkins:
## Description

Split add_collision_constraints out of add_inequality_constraints into a separate kernel (add_collision_constraints_work) with a transposed flat loop: i_b = flat_idx % B, i_col = flat_idx // B. This ensures warp threads access consecutive i_b values, giving coalesced Jacobian writes.

Constraint slots are allocated deterministically (collision_con_start + i_col * 4 + i) instead of via atomic_add. A follow-up kernel (update_n_constraints_collision) bumps n_constraints per env.

## Motivation and Context

Benchmarks on field:

20260401_coll_cnostr_2b_1503

Another run. some variance:

20260401_coll_cnostr_2b_1503 (1)


hughperkins:
Woo! 🥳

PR #2655, opened 2026-04-04
[FEATURE] Add `history_length` to Sensors

### LINKED ISSUE #2646: [Feature]: `history_length` support for `ContactForceSensor`

When using `decimation > 1` (i.e. multiple physics steps per control step), `ContactForceSensor` currently returns only the instantaneous force from the final physics step of each control interval. Transient contacts that occur during earlier physics steps within the same control interval are silently dropped. A `history_length` parameter that accumulates per-physics-step force snapshots across a control interval — analogous to mjlab's [`ContactSensorCfg.history_length`](https://github.com/mujocolab/mjlab/blob/main/src/mjlab/sensor/contact_sensor.py) — would make it possible to reliably detect these transient events.

### Motivation

In RL with `decimation > 1`, I want to access per-physics-step contact force history, because transient self-collisions that occur mid-interval are silently dropped by the current instantaneous readout.

### Potential Benefit

Reliable reward function implementation for RL

A `history_length` parameter on `ContactForceSensor` that returns accumulated contact forces for each physics step within a control interval, with shape `(num_envs, num_links, history_length, 3)`.

### Additional information

_No response_

Milotrince:
## Description

- Support `history_length` for all sensors
- Detect whether to `update_ground_truth_only` instead of user having to set the option
- Add `filter_link_idx` for `Contact` sensor

## Related Issue

Resolves https://github.com/Genesis-Embodied-AI/Genesis/issues/2646

## Motivation and Context

See issue

Added `tests/test_sensors.py::test_sensor_history_length_contact_and_imu`


duburcqa:
@claude review

genesis/engine/sensors/sensor_manager.py
```diff
@@ -190,6 +227,26 @@ def draw_debug(self, context: "RasterizerContext", buffer_updates: dict[str, np.
...
+
+            hist_idx = torch.arange(sensor._history_length, device=ring.buffer.device, dtype=torch.int32)
+            blocks = []
+            for rel_slice in sensor._cache_slices:
+                abs_slice = slice(sensor._cache_idx + rel_slice.start, sensor._cache_idx + rel_slice.stop)
+                hist = ring.at(hist_idx, slice(None), abs_slice)
+                blocks.append(hist.transpose(0, 1).flatten(1, 2))
+            stacked = torch.cat(blocks, dim=1)
+            self._history_clone_cache[hist_cache_key] = stacked
+            return stacked
```
Milotrince: the current implementation of history is with minimal changes to the current system, so history per sensor is collected upon read like this. should we restructure somehow to make this more efficient or is this ok?

duburcqa:
@claude review

Milotrince:
bump!

genesis/engine/sensors/base_sensor.py
```diff
@@ -120,41 +123,66 @@ def __init__(self, sensor_options: "SensorOptions", sensor_idx: int, sensor_mana
...
+        intrinsic_shapes: tuple[tuple[int, ...], ...] = (
+            (return_format,) if isinstance(return_format[0], int) else return_format
+        )
+        self._intrinsic_return_shapes: tuple[tuple[int, ...], ...] = intrinsic_shapes
 
         self._cache_size = 0
-        for shape in self._return_shapes:
-            data_size = np.prod(shape)
+        for shape in intrinsic_shapes:
+            data_size = int(np.prod(shape))
```
duburcqa: useless.

genesis/engine/sensors/base_sensor.py
```diff
@@ -120,41 +123,66 @@ def __init__(self, sensor_options: "SensorOptions", sensor_idx: int, sensor_mana
...
             self._cache_slices.append(slice(self._cache_size, self._cache_size + data_size))
             self._cache_size += data_size
 
+        # Slices into the per-sensor tensor from get_cloned_from_cache (history stacks H frames on dim 1).
+        self._read_flat_slices: list[slice] = []
+        read_off = 0
+        for shape in intrinsic_shapes:
+            p = int(np.prod(shape))
```
duburcqa: useless

genesis/engine/sensors/base_sensor.py
```diff
@@ -120,41 +123,66 @@ def __init__(self, sensor_options: "SensorOptions", sensor_idx: int, sensor_mana
...
+            self._return_shapes = tuple((self._history_length, *s) for s in intrinsic_shapes)
+        else:
+            self._return_shapes = intrinsic_shapes
+
         self._cache_idx: int = -1  # initialized by SensorManager during build
 
     # =============================== methods to implement ===============================
 
+    def _options_require_measured_cache(self):
+        return np.any(np.abs(self._options.delay) > gs.EPS)
```
duburcqa, line 177: Why adding this helper? I hate one-liner functions.

genesis/engine/sensors/base_sensor.py
```diff
@@ -442,29 +477,41 @@ class NoisySensorMixin(Generic[NoisySensorMetadataMixinT]):
     Base sensor class for analog sensors that are attached to a RigidEntity.
     """
 
+    def _assert_measured_cache_will_update(self):
+        assert self._shared_metadata.update_ground_truth_only is False, (
+            "Tried to update noise option but update_ground_truth_only is True. "
+            "Set a noisy option to nonzero value so that the measured cache will be updated."
+        )
```
duburcqa: Could you rather define a decorator for this? à-la `gs.assert_built`

genesis/engine/sensors/base_sensor.py
```diff
@@ -496,6 +543,17 @@ def build(self):
...
+    def _options_require_measured_cache(self) -> bool:
+        return super()._options_require_measured_cache() or (
+            self._options.jitter > gs.EPS
+            or (self._options.interpolate and (self._delay_ts > 0 or self._options.jitter > gs.EPS))
+            or np.any(np.abs(self._options.bias) > gs.EPS)
+            or np.any(np.abs(self._options.noise) > gs.EPS)
+            or np.any(np.abs(self._options.random_walk) > gs.EPS)
+            or np.any(np.abs(self._options.resolution) > gs.EPS)
+            or np.any(np.array(self._options.jitter) > gs.EPS)
+        )
```
duburcqa, line 562: I'm not a huge fan of this branching. I would rather always have measured cache always enabled.

genesis/engine/sensors/sensor_manager.py
duburcqa: > Should we restructure somehow to make this more efficient or is this ok?

Yes. We should.

genesis/options/sensors/camera.py
```diff
@@ -64,6 +64,10 @@ class BaseCameraOptions(RigidSensorOptionsMixin[SensorT]):
     lights: list[dict[str, Any]] = []
     offset_T: Matrix4x4Type | None = None
 
+    def model_post_init(self, context: Any) -> None:
+        if self.history_length > 0:
+            gs.raise_exception("Camera sensors do not support history.")
```
duburcqa, line 69: Why it does not support history?

tests/test_sensors.py
```diff
@@ -520,6 +613,42 @@ def test_contact_sensors_gravity_force(n_envs, show_viewer, tol):
...
+        gs.sensors.Contact(
+            entity_idx=floor.idx,
+            filter_link_idx=(box.link_start,),
+        )
+    )
+    scene.build(n_envs=n_envs)
+    for _ in range(20):
+        scene.step()
+    assert sensor_floor.read().all()
+    assert not sensor_filtered.read().any()
```
duburcqa: I'm not convinced this is good enough. This is not validating that the filter is doing what it is supposed to do.

tests/test_sensors.py
```diff
@@ -358,6 +358,99 @@ def test_imu_sensor(show_viewer, tol, n_envs):
...
+        assert cg.shape == _expected_shape_with_history((1,))
+        ig = imu_h.read_ground_truth()
+        assert ig.lin_acc.shape == _expected_shape_with_history((3,))
+        assert ig.ang_vel.shape == _expected_shape_with_history((3,))
+        assert ig.mag.shape == _expected_shape_with_history((3,))
+
+        assert_equal(contact_h.read(), cg)
+
+        assert_allclose(
+            ig.lin_acc[0 if n_envs == 0 else (slice(None), 0)], imu_ref.read_ground_truth().lin_acc, tol=tol
```
duburcqa: Define a variable for `0 if n_envs == 0 else (slice(None), 0)`. I call it `batch_shape` usually.

tests/test_sensors.py
```diff
@@ -358,6 +358,99 @@ def test_imu_sensor(show_viewer, tol, n_envs):
...
+                prev_i.lin_acc[0 if n_envs == 0 else (slice(None), 0)],
+                tol=tol,
+            )
+            assert_allclose(
+                ig.ang_vel[1 if n_envs == 0 else (slice(None), 1)],
+                prev_i.ang_vel[0 if n_envs == 0 else (slice(None), 0)],
+                tol=tol,
+            )
+            assert_allclose(
+                ig.mag[1 if n_envs == 0 else (slice(None), 1)],
```
duburcqa: Why do you need this? `[..., 1]` is not doing the job already?

tests/test_sensors.py
```diff
@@ -358,6 +358,99 @@ def test_imu_sensor(show_viewer, tol, n_envs):
...
+        if prev_c is not None:
+            assert_equal(
+                cg[1 if n_envs == 0 else (slice(None), 1)],
+                prev_c[0 if n_envs == 0 else (slice(None), 0)],
+            )
+        if prev_i is not None:
+            assert_allclose(
+                ig.lin_acc[1 if n_envs == 0 else (slice(None), 1)],
+                prev_i.lin_acc[0 if n_envs == 0 else (slice(None), 0)],
+                tol=tol,
```
duburcqa: tol is not good enough. It should be `gs.EPS` here.

genesis/options/sensors/camera.py
Milotrince: camera sensor uses a different read()/cache system that i didnt want to modify for now

genesis/engine/sensors/base_sensor.py
Milotrince: let's check how much it affects performance

genesis/engine/sensors/base_sensor.py
Milotrince: sensor classes extend this with super()

genesis/engine/sensors/sensor_manager.py
Milotrince: what would you suggest? each sensor may have varying lengths of history

Milotrince: currently:  manager has `_ground_truth_cache` shape`(dtype_size, B)`, `_cache` shape`(dtype_size, B)`, calls `_update_shared_ground_truth_cache() `, `_update_shared_cache()` per sensor class based `update_ground_truth_only`

im thinking to do: manager has `_ground_truth_cache` and `_cache` shape `(sensor_size, max_history_len_per_sensor_class, B)`, calls `_update_shared_cache()` per sensor class which updates both caches (whether to skip some computation for measured cache based on options can be decided per sensor cls)

i want to avoid flattening the history dim into the cache because it will complicate indexing, and having max history length per dtype instead of sensor might allocate too much unused space if, say, history=10 is used for ContactForce sensors in the scene and no history used for Proximity sensor

Milotrince: Another consideration: an api to get the data tensor for all sensors of a certain type would be useful, instead of having to store the sensor objects and loop/read over each (say you have 10 contact sensors that are being used for RL)

genesis/engine/sensors/base_sensor.py
duburcqa: Updated.

genesis/engine/sensors/base_sensor.py
duburcqa: Fixed.

genesis/engine/sensors/sensor_manager.py
duburcqa: Added.

duburcqa:
@claude review

Issue #2652, opened 2026-04-03
[Bug]: 'nan' causes training to crash in the manipulation example


AvikArefin:
### Bug Description

I had previously tried to make a project, HEAVILY inspired by the code from the manipulation example, but updated to the latest version, and unfortunately, every time after a few iterations, I would start getting 'nan'. And the entire system would crash. I have heard that for there has to be something terribly wrong and I should not "ignore" it. But I had tried to sort of push it away, by resetting the entire simulation whenever there would be a 'nan', and I noticed, the frequency of the number of times the simulation would throw 'nan' would just continue to rise. 

Now, on to given example, I tried twice on to train, and after around 47 iterations (3 times in a row), it would give 'nan' and crash.
I would be grateful if you would fix issue and kindly explain why it was happening in the first place.

### Steps to Reproduce

in the `manipulation` examples
uv init --python 3.13 

pyproject.toml 

[project]
name = "manipulation"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.13"
dependencies = [
    "genesis-world>=0.4.4",
    "rsl-rl-lib>=5.0.1",
    "tensorboard>=2.20.0",
]

uv sync
uv run grasp_train.py --stage=rl

### Expected Behavior

The training should finish properly, without any 'nan' issue.

### Relevant log output

```shell
################################################################################
                           Learning iteration 47/300                             

                               Run name: grasp
                            Total steps: 2359296 
                       Steps per second: 2984 
                        Collection time: 15.818s 
                          Learning time: 0.653s 
                        Mean value loss: 0.0001
                    Mean surrogate loss: -0.0050
                      Mean entropy loss: 6.6541
                            Mean reward: 0.64
                    Mean episode length: 301.00
                        Mean action std: 0.75
             Mean episode rew_keypoints: 0.2155
--------------------------------------------------------------------------------
                         Iteration time: 16.47s
                           Time elapsed: 00:14:29
                                    ETA: 01:16:02
Traceback (most recent call last):
  File "/Users/crimson/Activity/manipulation/grasp_train.py", line 208, in <module>
    main()
    ~~~~^^
  File "/Users/crimson/Activity/manipulation/grasp_train.py", line 204, in main
    runner.learn(num_learning_iterations=args.max_iterations, init_at_random_ep_len=True)
[...]

PR #2644, opened 2026-04-02
[MISC] Update all RL examples.

duburcqa:
Resolves Genesis-Embodied-AI/Genesis#2640

PR #2657, opened 2026-04-04
[MISC] Refactor example grasp RL+BC for performance and cross-platform support.

duburcqa:
* Refactor manipulation RL+BC example
  * Camera sensors instead of scene cameras for stereo rendering, with automatic fallback for cross-platform support
  * Efficient soft-reset replacing full scene reset for faster RL training
  * NaN error detection triggering automatic env reset in both grasp and go2 environments
  * DLS inverse kinematics fix for MPS/Metal backend where batched linear algebra was ~300x slower than CPU
  * Video recording and live debug preview for camera sensors
* Faster set_pos/set_quat with zero-copy and deferred forward kinematics support
* Fix camera sensor per-env rendering with Rasterizer
* Automatically select optimal H264 codec for Video recorder
* Reject heterogeneous entities mixing fixed and non-fixed morphs
* Validate that batched rasterizer camera sensors require env_separate_rigid=True
* Fix recorder cross-thread GPU access crash.

Resolves https://github.com/Genesis-Embodied-AI/Genesis/issues/2652
Resolves https://github.com/Genesis-Embodied-AI/Genesis/pull/2644#issuecomment-4185162993

### DIFF
--- a/examples/manipulation/grasp_train.py
+++ b/examples/manipulation/grasp_train.py
@@ -176,12 +176,12 @@ def main():
     log_dir = Path("logs") / f"{args.exp_name + '_' + args.stage}"
     log_dir.mkdir(parents=True, exist_ok=True)
 
-    with open(log_dir / "cfgs.pkl", "wb") as f:
-        pickle.dump((env_cfg, reward_scales, robot_cfg, rl_train_cfg, bc_train_cfg), f)
-
     # === env ===
     # BC only needs a small number of envs, e.g., 10
     env_cfg["num_envs"] = args.num_envs if args.stage == "rl" else 10
+
+    with open(log_dir / "cfgs.pkl", "wb") as f:
+        pickle.dump((env_cfg, reward_scales, robot_cfg, rl_train_cfg, bc_train_cfg), f)
     env = GraspEnv(
         env_cfg=env_cfg,
         reward_cfg=reward_scales,


examples/manipulation/grasp_env.py
AvikArefin: I believe type should be gs.morphs.MJCF

duburcqa:
@claude review

AvikArefin:
```
[Genesis] [20:08:48] [ERROR] TypeError: RigidEntity.set_pos() got an unexpected keyword argument 'skip_forward'
```

am I using a wrong version of genesis? FYI ("genesis-world>=0.4.4", "rsl-rl-lib>=5.0.1",)

duburcqa:
> am I using a wrong version of genesis? FYI ("genesis-world>=0.4.4", "rsl-rl-lib>=5.0.1",)

Yes, you must use this specific branch. I modified the source code of Genesis in this refactoring. The scope of the modifications got significantly larger (see PR description).

examples/manipulation/grasp_train.py
AvikArefin: I believe, the cfgs.pkl file is storing the incorrect environment number.

can be checked using 
```
uv run  python -c "import pickle; from pathlib import Path; print(pickle.load(open('logs/grasp_rl/cfgs.pkl', 'rb'))[0]['num_envs'])"
```

duburcqa: Indeed you are right! I will push a fix once the CI is green.

PR #2659, opened 2026-04-05
[MISC] Add register-only tiled cholesky, and incremental H patching, for performance.

hughperkins:
## Description

- upgrade Cholesky full to use 16x16 register-only from Quadrants
    - avoids the high latency of shared memory
    - portable across all Quadrants-supported gpus (amd, metal, vulkan, cuda)
- use incremental H-patching for Hessian, instead of Givens rotations, or full Hessian
     - faster than full (because incremental)
     - avoids serial restrictions of Givens rotations
- fuse Cholesky and H-patching
    - avoid pushing to and from global memory in between

Note that we need to combine both incremental H-patching and register tiled Cholesky, because the gains for each on their own are much more modest:

For dex_hand, with field, on RTX 6000 Blackwell:

| Approach | branch name| FPS | delta FPS |
|-----|-----|----|----|
|main|main|13686| +0%|
|tile16 cholesky| hp/cholesky-tile16| 13813|+0.93%|
|H-patching| hp/incremental-hessian |13858|+1.26%|
|fused H-patching + tile16 cholesky | this PR | 15256|+11.47%|

This is because fusing both into a single kernel avoids having to make a round-trip to global memory in between them.

## Related Issue

## Checklist:


hughperkins:
~this will need a newer new version of quadrants => close for now, to avoid stressing CI.~

duburcqa:
Could you split register-only tiled cholesky and incremental H patching in 2 different PRs please? The first one is easier to review while the second one requires more effort.

hughperkins:
> Could you split register-only tiled cholesky and incremental H patching in 2 different PRs please? The first one is easier to review while the second one requires more effort.

I could, but there isn't much benefit to either of them on their own. My concern is you might not want to merge somethign that only gives ~1% or so benefit?

duburcqa:
> I could, but there isn't much benefit to either of them on their own. My concern is you might not want to merge somethign that only gives ~1% or so benefit?

I see. In that case, could you explain why together in bring sizeable benefits but not individually?

hughperkins:
> > I could, but there isn't much benefit to either of them on their own. My concern is you might not want to merge somethign that only gives ~1% or so benefit?
> 
> I see. In that case, could you explain why together in bring sizeable benefits but not individually?

added a table into the descripton above:

Screenshot 2026-04-05 at 07 02 51

hughperkins:
Also added to description "This is because fusing both into a single kernel avoids having to make a round-trip to global memory in between them."

hughperkins:
~will need a new quadrants release. closng for now.~

duburcqa:
I see. Thank you!

tests/test_rigid_physics.py
duburcqa, line 1: Why are you modifying this test? It has been working just fine for months

tests/test_rigid_physics.py
```diff
@@ -3733,13 +3733,13 @@ def rigid_solver_build(self):
...
-        scene.step()
+        for _ in range(10):
+            scene.step()
         assert not scene.rigid_solver.get_error_envs_mask().any()
         assert (scene.rigid_solver.constraint_solver.constraint_state.n_constraints.to_numpy() > 0).all()
 
-        nt_H = scene.rigid_solver.constraint_solver.constraint_state.nt_H.to_numpy()
-        assert (np.linalg.norm(nt_H.reshape((-1, 2)), axis=0) > 5.0).all()
-        values.append(nt_H)
+        qpos = gs_robot.get_qpos().cpu().numpy()
```
duburcqa: I much prefer the original test. This check is doing nothing.

tests/test_rigid_physics.py
```diff
@@ -3733,13 +3733,13 @@ def rigid_solver_build(self):
         assert scene.rigid_solver._static_rigid_sim_config.enable_tiled_cholesky_mass_matrix == enable_tiled_cholesky
         assert scene.rigid_solver._static_rigid_sim_config.enable_tiled_cholesky_hessian == enable_tiled_cholesky
 
-        scene.step()
+        for _ in range(10):
```
duburcqa: I don’t see any reason for doing 10 steps in this test.

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -1621,6 +1630,19 @@ def func_cholesky_factor_direct_batch(
             constraint_state.nt_H[i_b, j_d, i_d] = (constraint_state.nt_H[i_b, j_d, i_d] - dot) * tmp
 
 
+Tile16x16 = make_tile16x16(gs.qd_float)
+
+
+@qd.func
+def _butterfly_reduce_16(val, tid):
+    """Sum val across 16 threads using butterfly reduction via subgroup shuffles (4 rounds)."""
+    val = val + qd.simt.subgroup.shuffle(val, qd.u32(tid ^ 8))
```
duburcqa: Prefer static for loop

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -1629,77 +1651,197 @@ def func_cholesky_factor_direct_tiled(
...
+                constraint_state.nt_H[i_b, k0 : k0 + Tile16x16.SIZE, k0:n_dofs] = L_kk
+
+
+@qd.func
+def func_cholesky_and_solve_fused_tiled(
+    constraint_state: array_class.ConstraintState,
+    rigid_global_info: array_class.RigidGlobalInfo,
+    static_rigid_sim_config: qd.template(),
+):
+    """Fused blocked Cholesky factorization + forward/backward solve.
```
duburcqa: Would it be possible to this docstring a bit shorter?

tests/test_rigid_physics.py
hughperkins: in order to support H-patching, we need to preserve H. Cholesky is therefore never written to nt_H, and is not available for asserts.

tests/test_rigid_physics.py
```diff
@@ -3737,9 +3738,9 @@ def rigid_solver_build(self):
         assert not scene.rigid_solver.get_error_envs_mask().any()
         assert (scene.rigid_solver.constraint_solver.constraint_state.n_constraints.to_numpy() > 0).all()
 
-        nt_H = scene.rigid_solver.constraint_solver.constraint_state.nt_H.to_numpy()
-        assert (np.linalg.norm(nt_H.reshape((-1, 2)), axis=0) > 5.0).all()
-        values.append(nt_H)
+        Mgrad = scene.rigid_solver.constraint_solver.constraint_state.Mgrad.to_numpy()
+        assert np.linalg.norm(Mgrad) > 0.0
```
duburcqa: Could you set a threshold larger than 0.0? Or is it expected to be very small?

hughperkins: I don't think we have any expectaitons on it really. You mean, run empirically, and see what value would fit the empirical output?

duburcqa: Yes

hughperkins: (note: measured empirical value was ~1346)

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -2,6 +2,7 @@
 
 import numpy as np
 import quadrants as qd
+from quadrants.lang.simt.tile16 import make_tile16x16
 import torch
 from frozendict import frozendict
```
duburcqa, line 6: Import order (yes we should introduce auto sorting soonish)

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -1467,6 +1469,9 @@ def func_hessian_direct_tiled(
     optimization problem fits in a single block, i.e. n_constraints <= 32 and n_dofs <= 64.
 
     Note that only the lower triangular part will be updated for efficiency, because the Hessian matrix is symmetric.
+
+    When check_full_hessian is True (used with H patching), skips envs where
+    use_full_hessian == 0 (those get patched instead of rebuilt).
```
duburcqa: Only wrap lines when reaching 120 chars.

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -1629,77 +1649,194 @@ def func_cholesky_factor_direct_tiled(
...
+def func_cholesky_and_solve_fused_tiled(
+    constraint_state: array_class.ConstraintState,
+    rigid_global_info: array_class.RigidGlobalInfo,
+    static_rigid_sim_config: qd.template(),
+):
+    """Fused Cholesky factorization and triangular solve, keeping L in shared memory.
+
+    Factorizes H = L L^T using register-resident 16x16 tiles, storing completed L tiles
+    in shared memory. Then solves L L^T x = g (forward + backward substitution) in-place
+    and writes the result to Mgrad, without ever writing L to global memory.
```
duburcqa: Wrapping too early.

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -1629,77 +1649,194 @@ def func_cholesky_factor_direct_tiled(
...
+        # L is fully computed in L_sh. Load gradient into v_sh.
+        k = tid
+        while k < n_dofs:
+            v_sh[k] = constraint_state.grad[k, i_b]
+            k = k + Tile16x16.SIZE
         qd.simt.block.sync()
 
-        # Loop over all columns sequentially, which is an integral part of Cholesky-Crout algorithm and cannot be
-        # avoided.
+        # Forward substitution: solve L @ y = grad (parallel dot with 16 threads)
```
duburcqa, line 1860: This comment is loosing some important information.

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -1629,77 +1649,194 @@ def func_cholesky_factor_direct_tiled(
...
         for i_d in range(n_dofs):
-            # Compute the diagonal of the Cholesky factor L for the column i being considered, ie
-            # L_{i,i} = sqrt(A_{i,i} - sum_{j=1}^{i-1}(L_{i,j} ** 2 ))
```
duburcqa, line 1729: Why comment is still relevant, and if not completely, it should be updated. Not removed.

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -1629,77 +1649,194 @@ def func_cholesky_factor_direct_tiled(
...
-            # L_{j,i} = 1 / L_{i,i} (A_{j,i} - sum_{k=1}^{i-1}(L_{j,k} L_{i,k}), for j > i
-            inv_diag = 1.0 / H[i_d, i_d]
-            j_d = i_d + 1 + tid
-            while j_d < n_dofs:
-                dot = gs.qd_float(0.0)
-                for k_d in range(i_d):
-                    dot = dot + H[j_d, k_d] * H[i_d, k_d]
-                H[j_d, i_d] = (H[j_d, i_d] - dot) * inv_diag
-                j_d = j_d + BLOCK_DIM
+        # Backward substitution: solve L^T @ x = y (parallel dot with 16 threads)
```
duburcqa, line 1872: It feels like some information has been lost compared to the original comment.

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -1629,77 +1649,194 @@ def func_cholesky_factor_direct_tiled(
...
+        tid = i % Tile16x16.SIZE
+        i_b = i // Tile16x16.SIZE
         if i_b >= _B:
             continue
         if constraint_state.n_constraints[i_b] == 0 or not constraint_state.improved[i_b]:
             continue
 
-        # Padding +1 to avoid memory bank conflicts that would cause access serialization
-        H = qd.simt.block.SharedArray((MAX_DOFS, MAX_DOFS + 1), gs.qd_float)
+        L_sh = qd.simt.block.SharedArray((MAX_DOFS, MAX_DOFS), gs.qd_float)
```
duburcqa: Why are you removing the comment and not padding yourself? Do you have a proof that it does not bring any benefit in your case? if so, you should add a comment that clearly states this and explains why.

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -1629,77 +1649,194 @@ def func_cholesky_factor_direct_tiled(
...
+                            v_diag = L_sh[k0 + tid, j0 + t]
+                        L_ik -= qd.outer(v_own, v_diag)
+
+                L_kk.solve_triangular_(L_ik)
+
+                if i0 + tid < n_dofs:
+                    L_sh[i0 : i0 + Tile16x16.SIZE, k0:n_dofs] = L_ik
+
+            if k0 + tid < n_dofs:
+                L_sh[k0 : k0 + Tile16x16.SIZE, k0:n_dofs] = L_kk
```
duburcqa: Each of these block should have a short one-line comment on top.

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -1629,77 +1649,194 @@ def func_cholesky_factor_direct_tiled(
...
+                            v_diag = constraint_state.nt_H[i_b, k0 + tid, j0 + t]
+                        L_ik -= qd.outer(v_own, v_diag)
+
+                L_kk.solve_triangular_(L_ik)
+
+                if i0 + tid < n_dofs:
+                    constraint_state.nt_H[i_b, i0 : i0 + Tile16x16.SIZE, k0:n_dofs] = L_ik
+
+            if k0 + tid < n_dofs:
+                constraint_state.nt_H[i_b, k0 : k0 + Tile16x16.SIZE, k0:n_dofs] = L_kk
```
duburcqa: Each of these block should have a short one-line comment on top.

genesis/engine/solvers/rigid/constraint/solver.py
duburcqa: Would it be possible to share more code with `func_cholesky_factor_direct_tiled`?

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -1621,6 +1630,17 @@ def func_cholesky_factor_direct_batch(
             constraint_state.nt_H[i_b, j_d, i_d] = (constraint_state.nt_H[i_b, j_d, i_d] - dot) * tmp
 
 
+Tile16x16 = make_tile16x16(gs.qd_float)
```
duburcqa: Why is this not done by Quadrants? In some kind of singleton factory.

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
```diff
@@ -709,13 +709,118 @@ def _func_update_constraint_cost(
...
+                else:
+                    constraint_state.use_full_hessian[i_b] = 0
+
+
+@qd.func
+def _func_patch_hessian_delta(
+    constraint_state: array_class.ConstraintState,
+    rigid_global_info: array_class.RigidGlobalInfo,
+):
+    """Patch H with deltas from changed constraints for envs with use_full_hessian == 0.
```
duburcqa: The meaning of "patching" is not clear. Could you clarify that?

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
```diff
@@ -709,13 +709,118 @@ def _func_update_constraint_cost(
...
+
+    For each constraint whose active state changed, adds or subtracts its J^T D J
+    contribution to the Hessian. Uses 1 block per env (same parallelism as
+    hessian_direct_tiled) to avoid launching excessive threads.
```
duburcqa: You are wrapping lines too early.

genesis/engine/solvers/rigid/constraint/solver_breakdown.py
```diff
@@ -709,13 +709,118 @@ def _func_update_constraint_cost(
...
+    constraint_state: array_class.ConstraintState,
+    rigid_global_info: array_class.RigidGlobalInfo,
     static_rigid_sim_config: qd.template(),
 ):
-    """Step 4: Newton Hessian update (Newton only)"""
+    """Full Hessian rebuild + Cholesky for ALL improved envs (non-fused path).
+
+    Matches origin/main behavior: H is rebuilt from scratch every iteration,
+    then Cholesky overwrites nt_H with L in-place.  H patching is not used
+    because the subsequent Cholesky would destroy H anyway.
```
duburcqa: Wrapping lines too early.

tests/test_rigid_physics.py
```diff
@@ -3737,11 +3738,41 @@ def rigid_solver_build(self):
...
+    # -------------------------------------------------------
+    # tol=1e-03                         PASS            PASS
+    # tol=5e-04                         FAIL            PASS
+    # tol=2e-04                         FAIL            PASS
+    # tol=1e-04                         FAIL            FAIL
+    # tol=5e-05                         FAIL            FAIL
+    # tol=1e-05                         FAIL            FAIL
+
+    # ========================================================================
+    assert_allclose(*values, tol=5e-4)
```
duburcqa: Keep this comment in the PR, with a link referencing your post in code comment.

Apart from that, it seems that you are saying that your branch improves accuracy, yet you are manually setting the tolerance to 5e-4, which is worst that the original value (5e-5).

hughperkins: Main fails with tolerance < 1e-3

hughperkins: Rememeber: on main, the test checks np_H, but here we are checking Mgrad, which needs higher tolerance, as evidenced by the experiments on main.

hughperkins: Ok, so:
- on main:
    - tolerance on nt_H can be <5e-5
    - tolerance on Mgrad needs to be >1e-3
- on this branch:
     - tolerance on Mgrad can be as small as 2e-4

genesis/engine/solvers/rigid/constraint/solver.py
hughperkins: what code do you feel is in common?

all this code is operating on shuffle tiles.

AFAIK none of this code is usable in the direct tiled

hughperkins: oh wait, maybe I'm thinking of a differnet function 🤔

hughperkins: oh I see what you mean. Looking...

tests/test_rigid_physics.py
duburcqa: I see! Thank you for the details. Ok so in the end better accuracy for lower cost, which is expected but great.

genesis/engine/solvers/rigid/constraint/solver.py
hughperkins: this is what Opus says:

  Then the remaining options are limited:
  • Template flag (fused: qd.template()) with qd.static(fused) branches inside the loop. But the non-fused version would still need to allocate an unused SharedArray to pass as
    a parameter, wasting shared memory.
  • Keep as-is. The duplication is ~25 lines. The factorization loops look similar but the storage source/dest (the thing that matters most for performance) is different at
    every read/write point. Factoring it out would require scattering if qd.static(fused) branches throughout, making the code harder to follow than the current two clean
    self-contained functions.

tests/test_rigid_physics.py
hughperkins: # Analysis for choice of tol.
    # 'main' here represents 'Cholesky shared memory implementation', which was what was in place at commit 7bf5f6a7.
    # 'Branch' here represents the shuffle based fused Cholesky and patched-H implementation.
    #
    # Created using https://github.com/Genesis-Embodied-AI/perso_hugh/blob/main/demos/test_cholesky_tiled.py
    #
    # ========================================================================
    # Cholesky Tiled vs Non-Tiled: Numerical Accuracy Comparison
    # ========================================================================
    # Main commit:   7bf5f6a7
    # Branch commit: 6f90d8a8 (hp/incremental-hessian-fuse-chol-tiles-api)
    # Quadrants:     38f16c77b (hp/tile16-tiles-only)

    # Metric                               Main          Branch
    # ---------------------------------------------------------
    # max_abs_diff               2.825928e-02    1.220703e-03
    # max_rel_diff               2.477033e-02    1.056026e-02
    # norm_tiled                 1.236271e+03    1.346164e+03
    # norm_notiled               1.236246e+03    1.346164e+03

    # Tolerance                          Main          Branch
    # -------------------------------------------------------
    # tol=1e-03                         PASS            PASS
    # tol=5e-04                         FAIL            PASS
    # tol=2e-04                         FAIL            PASS
    # tol=1e-04                         FAIL            FAIL
    # tol=5e-05                         FAIL            FAIL
    # tol=1e-05                         FAIL            FAIL

    # ========================================================================

genesis/engine/solvers/rigid/constraint/solver.py
duburcqa: Let's keep what you have know if it is too complicated.

genesis/engine/solvers/rigid/constraint/solver.py
hughperkins: How are you imaginging this usage woud look?

genesis/engine/solvers/rigid/constraint/solver.py
hughperkins: is some related comment back now? (hard to tell from github review)

genesis/engine/solvers/rigid/constraint/solver.py
hughperkins: is some related comment back now? (hard to tell from github review)

genesis/engine/solvers/rigid/constraint/solver.py
duburcqa: Similar to "annotations"?
```python
qd.types.Tile16x16(dtype=gs.qd_float)
```

hughperkins:
Note: updated the vector bits to use a VectorProxy too now, similar to the TileProxy:
```
                        v_own = L_sh[i0:n_dofs, j0 + t]
                        v_diag = L_sh[k0:n_dofs, j0 + t]
                        L_ik -= qd.outer(v_own, v_diag)
```

hughperkins:
added some commentshere:

```
        # --- Scalar triangular solve using L from shared memory ---
        # No longer using 16x16 tiles; the 16 threads parallelize each row's
        # dot product by striping across columns, then butterfly-reduce to
        # sum the partial products. Thread 0 writes each solved element.
```

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -1621,6 +1629,17 @@ def func_cholesky_factor_direct_batch(
             constraint_state.nt_H[i_b, j_d, i_d] = (constraint_state.nt_H[i_b, j_d, i_d] - dot) * tmp
 
 
+Tile16x16 = qd.types.Tile16x16(dtype=gs.qd_float)
+
+
+@qd.func
+def _butterfly_reduce_16(val, tid):
```
hughperkins: I wonder if this should be moved into quadrants somewhere?

hughperkins: Probalby right?

hughperkins: (also, I kind of lean towards the name descrinbg the effect and behavior, rather than the detailed implementation?)

duburcqa: Yes that would be much better in Genesis where lower expertise from maintainers can be expected. Nice to mention the implementation in doc string though.

duburcqa: Indeed you should move it in Quadrants i think. My previous comment still hold since it would be public API.

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -1681,77 +1697,212 @@ def func_cholesky_factor_direct_tiled(
...
+        # are processed sequentially (they only depend on the diagonal, but each tile uses all threads).
+        for kb in range(N_BLOCKS):
+            k0 = kb * 16
+
+            # Load diagonal tile H[k,k], padding out-of-bounds rows with identity
+            L_kk = qd.simt.Tile16x16.zeros(dtype=gs.qd_float)
+            if k0 + tid < n_dofs:
+                L_kk[:] = constraint_state.nt_H[i_b, k0 : k0 + 16, k0:n_dofs]
+            else:
+                L_kk.eye_()
```
duburcqa: I don't know if we care, but on paper, having empty init instead of zero-init would be more efficient.

hughperkins: updated. ths was old code. thanks for reminding me :)

PR #2688, opened 2026-04-10
[BUG FIX] Fix contact overflow causing unbounded memory access.

duburcqa:
* Handling unsupported 'pygel3d' dependency on Linux ARM.
* Fix CUDA illegal memory access from contact counter overflow.
* Workaround for atomic or not working on Apple Metal.


duburcqa:
@claude review

PR #2690, opened 2026-04-10
[FEATURE] Add history_length support for ContactForceSensor

### LINKED ISSUE #2646: [Feature]: `history_length` support for `ContactForceSensor`

vlordier:
## Summary
- Adds `history_length` parameter to `ContactForce` sensor options to store and return historical force readings
- When `history_length > 1`, returns a tensor with past force readings from the ring buffer
- Useful for force filtering, smoothing, and computing force rates/derivatives

## Changes
- `genesis/options/sensors/options.py`: Added `history_length: PositiveInt = 1` field to `ContactForce` options
- `genesis/engine/sensors/contact_force.py`: Modified `read()` to return history from ring buffer, updated `_get_return_format()`
- `genesis/engine/sensors/sensor_manager.py`: Extended ring buffer size to accommodate history length

## Usage
```python
sensor = scene.add_sensor(
    gs.sensors.ContactForce(
        entity_idx=box.idx,
        history_length=5,
    )
)
forces = sensor.read()
# Shape: (history_length, 3) for n_envs=1
# Shape: (n_envs, history_length, 3) for batched
# Index 0 = most recent, Index N-1 = oldest
```

## Backward Compatible
- Default `history_length=1` returns same shape as before: `(n_envs, 3)`

Closes #2646


duburcqa:
PR already under review that this exact feature.

PR #2691, opened 2026-04-10
[FEATURE] Add get_height_at() and get_normal_at() for terrain entities

### LINKED ISSUE #2094: [Feature]: Get_Height interface when using terrain

when using terrain, we set the horizontal scale as a minimum size of block in terrain, to get terrain height we need to use height_field

### Motivation

When studying locomotion problems, somtimes terrain infos may help the learning process, e.g. reward robot feet fitting the terrain surface, and this requires we can query terrain normal by giving the x,y axis corrdinate, and another example is we can reward feet swing height base on the terrain height.

when using terrain, we set the horizontal scale as a minimum size of block in terrain,e.g. in height field[[1,2],[5,4]] with horizontal scale 0.25, and vertical scale 1.0, we get a 0.25*0.25 square and in simulation it seems to divide into two triangle like (
1----2
|  \    |
|    \  |
5----4
)  and to get height of position in this area need to calculate based on which triangle this position falls in.
That's to say to get terrain height we need to use height_field and we only get the 'points' height, and the height of positions between 'points' are complex to get.

I wonder if we can add feature that we can get height of terrain by giving (x,y), if convenient, get normal by given (x,y) will also be very nice.

### Potential Benefit

more precise terrain info get when using terrain

vlordier:
## Summary
- Adds `get_height_at(x, y)` method to query interpolated terrain height at world position
- Adds `get_normal_at(x, y)` method to compute surface normal at world position
- Both methods handle boundary conditions (return 0 for height outside terrain, [0,0,1] for normal)
- Use bilinear interpolation for height from height field
- Compute normal from height field gradient using central differences

## Changes
- `genesis/engine/entities/rigid_entity/rigid_entity.py`: Added two methods to RigidEntity class:
  - `get_height_at(x, y)` - Returns interpolated height using bilinear sampling from terrain height field
  - `get_normal_at(x, y)` - Returns unit normal vector computed from height field gradient
- Both methods use `@gs.assert_built` decorator and include detailed docstrings

## Usage
```python
# Create terrain entity
terrain = scene.add_entity(
    gs.morphs.Terrain(
        height_field=height_field_array,
        horizontal_scale=0.25,
        vertical_scale=0.1
    )
)
scene.build()

# Query terrain properties at world position
height = terrain.get_height_at(3.0, 4.5)  # Returns interpolated height
normal = terrain.get_normal_at(3.0, 4.5)  # Returns unit normal vector [nx, ny, nz]

# Useful for:
# - Locomotion rewards (feet fitting terrain surface)
# - Terrain-following control
# - Contact detection and response tuning
```

## Backward Compatible
- No changes to existing APIs
- Methods only available on entities with terrain morphology

Closes #2094


duburcqa:
@claude review

duburcqa (changes requested):
This PR contains tons of unrelated changes.

duburcqa:
This PR is stale. Closing.

PR #2692, opened 2026-04-11
[MISC] Enable parallel linesearch on all GPU backends.

duburcqa:


tests/test_rigid_physics.py
```diff
@@ -2184,6 +2184,8 @@ def test_contact_forces(show_viewer, tol):
         pos=np.array([0.65, 0.0, 0.3]),
         quat=np.array([0, 1, 0, 0]),
     )
+    franka.set_dofs_kp(5000.0, dofs_idx_local=motors_dof)
+    franka.set_dofs_kv(500.0, dofs_idx_local=motors_dof)
```
hughperkins: out of personal curiosity, what is the purpose of these two new lines?

hughperkins: out of personal curiosity, what is the purpose of these two new lines?

duburcqa: My objective was to improve tracking accuracy. Without increasing these gains, the final configuration of the robot will be slightly shifted from the target `qpos` computed by inverse kinematics to keep the gripper perfectly horizontal won't be accurate. 

It all started with this unit test failing on MacOS after enabling graph on Apple Metal. I wanted to improve tracking accuracy to reduce the tolerance of the final assertion checking that the total contact force applied my the grippers on the box matches the gravity. It worked in practice, but it turns out it should not be necessary because `get_links_net_contact_force` returns the force in world frame instead of local frame, so it should always match the gravity.

hughperkins: I see. Context is, this is he one test that was failing for me when I implemented a faster, more parallel, iterative linesearch (without doing a Newton step at the end etc). Opus told me that the gripper was gradually moving sideways, as a result of tiny changes. The solution to this turned out to be, I believe, always taking the mid point out of 3 points tried, rather than one of the end-points, on the final iteration.

I'm not sue exactly what set dofs kp and set dofs kv do, but they *sound* like kp is a spring, and kv adds damping?

So, wouldnt this force the gripper to remain in place, rather than slowly drifting?

And thus , this might disguise issues that might occur in practice, in the absence of having a damped spring keeping the gripper in place?

hughperkins: (On the whole, I'd prefer to test my linesearch algorithms *without* this spring in place, I feel. I'd prefer stuff to fail during testing, than once I've merged stuff into Genesis, and people start using it)

duburcqa: Sure. Having to increase the gains has no sense here. It just worked out of luck, hiding the real issue.

duburcqa: Updated the unit test. I would recommend that you have a look.

duburcqa: Note that I won't move on before understanding what is going on here. Having to increase the tolerance is often hiding more serious bugs.

hughperkins: oh wow, two orders of magnitude difference...

duburcqa: I'm working on this issue. I will let you know if I need help. I will at least request your review if you have time.

duburcqa: Ok I improved the unit test. It is quite robust now. The good news is that it is not Apple Metal related. The bad news is that parallel linesearch is very inaccurate compared to iterative (x1000).

duburcqa: I will merge this PR is the CI pass, since it is unrelated to Apple Metal in the end. That does not mean I'm giving up with this issue.

PR #2695, opened 2026-04-11
[FEATURE] Enable/disable collision and visualization for rigid geometries.

vlordier:
## Summary

Added public API methods to enable/disable collision and visualization for rigid geometries.

This feature allows users to toggle collision detection and visualization on/off for individual rigid geometries after they have been added to the scene. This is useful for:
- Temporarily disabling collision for specific objects during simulation
- Controlling visualization without removing/re-adding entities
- Performance optimization by disabling unnecessary collision checks

## Changes

Modified `genesis/engine/entities/rigid_entity/rigid_geom.py` to add:
- `set_collision_enabled(bool)` method
- `set_visualization_enabled(bool)` method
- Internal tracking of enabled states
- Early return in collision/visualization methods when disabled

## Usage

```python
# Get a rigid entity
rigid_entity = scene.add_entity(...)

# Disable collision for a specific geom
rigid_entity.geoms[0].set_collision_enabled(False)

# Disable visualization for a specific geom  
rigid_entity.geoms[0].set_visualization_enabled(False)

# Re-enable when needed
rigid_entity.geoms[0].set_collision_enabled(True)
rigid_entity.geoms[0].set_visualization_enabled(True)
```

## Testing

- Verified that disabling collision prevents collision responses
- Verified that disabling visualization hides the geom from rendering
- Verified that re-enabling restores normal behavior
- Existing tests continue to pass
- No breaking changes to existing API

PR #2696, opened 2026-04-11
[MISC] Fix PBD entities to properly add visual vertices to solver for BatchRendererCamera compatibility

vlordier:
## Summary

Fixed PBD entities to properly add visual vertices to solver for BatchRendererCamera compatibility. 

Previously, PBD entities were not calling `_add_vverts_to_solver()` in their `_add_to_solver()` method, which meant that visual vertices were not being added to the solver. This caused the BatchRendererCamera to fail when trying to render PBD entities because it couldn't find the visual vertex data.

This fix ensures that PBD entities properly add their visual vertices to the solver, making them compatible with the BatchRendererCamera.

## Testing

- Verified that PBD entities now work correctly with BatchRendererCamera
- Ran existing tests to ensure no regression
- The fix follows the same pattern used by other entity types (like rigid entities)


duburcqa:
@claude review

vlordier:
## Analysis

After investigation, this PR is currently a **net-zero change**.

- Commit `44d22c5` adds a `_add_vverts_to_solver()` call inside `PBDBaseEntity._add_to_solver()`.
- Commit `57eee9d` removes those exact same 4 lines.

The reason for the revert (documented in `57eee9d`) is correct: `ParticleEntity._add_to_solver()` (the parent class, `particle_entity.py:187–188`) already calls `_add_vverts_to_solver()` when `_need_skinning=True`, which is the case for all PBD tet entities. The duplicate call would have caused redundant precomputation with no functional benefit.

**If the original issue (BatchRendererCamera not rendering PBD entities) is still open**, the fix should be investigated at a different layer — the `batch_renderer.py` `GenesisGeomRetriever` does not include PBD visual geometry today. A more complete fix is included in PR #2695 which adds `retrieve_pbd_meshes_static()` to `GenesisGeomRetriever`.

Closing this PR as the net change is zero and the fix belongs elsewhere.

PR #2710, opened 2026-04-14
[MISC] Unify linesearch refinement between decomposed and monolith solver paths.

duburcqa:
## Motivation and Context

Falling back to newton alpha if no point on the grid is showing improvement is a real fix. This is necessary to correct the error distribution compared to iterative. Without it, we are just stopping too early compared to the desired tolerance, so it runs faster because it does less iterations but does not satisfy the required tolerance. This was validated by [@Mingrui](https://genesis-ai-company.slack.com/team/U09EYPPHWHW) earlier.
Once fixed (one-liner), I can confirm that parallel linesearch does not bring any performance improvement. Parallel vs Iterative are almost perfectly matching. it was just faster because of the wrong gating I just mentioned
Asking for 1e-6 tolerance using float32 precision is not realistic. This means that the gradient is scaled by 1e-8. If you use float32 in the first place, you do not expect machine precision in the first place.

## Description

Replace the cooperative grid search + bisection refinement with a simpler architecture: the P0 kernel computes a Newton step estimate, and the eval kernel refines it serially via a shared `func_linesearch_refine` (bracketing walk + 3-alpha dual-bracket). This function is used by both the decomposed and monolith solver paths, eliminating ~120 lines of duplicated refinement logic.

The `candidates[12, B]` array is replaced with 4 named scalar fields (`ls_alpha`, `ls_p0_cost`, `ls_alpha_newton`, `ls_gtol`). The user-facing option `prefer_parallel_linesearch` is removed; an internal `prefer_decomposed_solver` flag auto-selects based on backend/context.

Lowering the tolerance from 1e-6 to 1e-5. I did a small analysis and this is the sweet spot. Typically, I get 40k FPS on box pyramid 6 with 1e-5, but 30k FPS with 5e-6, which is only twice larger. This should preserve (and ever increase the performance compared to the current decomposed version on main), while satisfying the tolerance requirement and ensuring consistency between monolith and decomposed.

### BEFORE

image

### AFTER

image

For completeness, benchmark script is available [here](https://github.com/user-attachments/files/26691038/benchmark_linesearch_accuracy.py).

### DIFF
--- a/genesis/engine/solvers/rigid/constraint/solver.py
+++ b/genesis/engine/solvers/rigid/constraint/solver.py
@@ -115,7 +115,10 @@ class ConstraintSolver:
         self.jv = cs.jv
         self.quad_gauss = cs.quad_gauss
 
-        self.candidates = cs.candidates
+        self.ls_alpha = cs.ls_alpha
+        self.ls_p0_cost = cs.ls_p0_cost
+        self.ls_alpha_newton = cs.ls_alpha_newton
+        self.ls_gtol = cs.ls_gtol
         self.ls_it = cs.ls_it
         self.ls_result = cs.ls_result
         if self._solver_type == gs.constraint_solver.CG:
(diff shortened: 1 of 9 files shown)


duburcqa:
@claude review

duburcqa:
@claude review

duburcqa:
@claude review

PR #2713, opened 2026-04-15
[BREAKING][MISC] Set default constraint solver tolerance based on floating-point precision.

duburcqa:
Related to https://github.com/Genesis-Embodied-AI/Genesis/pull/2710

### DIFF
--- a/genesis/engine/solvers/rigid/rigid_solver.py
+++ b/genesis/engine/solvers/rigid/rigid_solver.py
@@ -337,6 +337,10 @@ class RigidSolver(KinematicSolver):
         self.n_fixed_verts_ = max(1, self.n_fixed_verts)
         self.n_candidate_equalities_ = max(1, self.n_equalities + self._options.max_dynamic_constraints)
 
+        # Resolve precision-dependent tolerance default
+        if self._options.tolerance is None:
+            self._options.tolerance = 1e-5 if gs.qd_float == qd.f32 else 1e-8
+
         super().build()
 
         self._init_mass_mat()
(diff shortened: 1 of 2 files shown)


genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -337,6 +337,10 @@ def build(self):
         self.n_fixed_verts_ = max(1, self.n_fixed_verts)
         self.n_candidate_equalities_ = max(1, self.n_equalities + self._options.max_dynamic_constraints)
 
+        # Resolve precision-dependent tolerance default
+        if self._options.tolerance is None:
+            self._options.tolerance = 1e-5 if gs.qd_float == qd.f32 else 1e-8
```
hughperkins, line 342: Nice!

duburcqa, line 342: Actually options are supposed to be instantiable without having to initialise Genesis first. This will have to be delegated to rigid solver build, which sadly obfuscate a bit the behaviour.

genesis/options/solvers.py
```diff
@@ -500,7 +500,7 @@ class RigidOptions(Options):
     # constraint solver
     constraint_solver: gs.constraint_solver = gs.constraint_solver.Newton
     iterations: PositiveInt = 50
-    tolerance: PositiveFloat = 1e-6
+    tolerance: PositiveFloat | None = None
```
duburcqa, line 503: The documentation should be updated to reflect the new default.

PR #2714, opened 2026-04-15
[BUG FIX] Fix float modulo buggy on AMDGPU.

duburcqa:


tests/conftest.py
duburcqa, line 265: Add:
`else: return -1`

duburcqa:
@claude review

PR #2717, opened 2026-04-16
[MISC] Fix safe GJK fallback.

duburcqa:
## Problem

PR https://github.com/Genesis-Embodied-AI/Genesis/pull/2714 fixed a correctness bug in `_func_count_supports_mesh` where `dot_max` was never updated, causing the support count to always return 1. The fix was reverted in https://github.com/Genesis-Embodied-AI/Genesis/pull/2716 because it caused an ~20% FPS regression on benchmarks.

## Root cause of the regression

In `func_safe_gjk_support`, the perturbation loop (i=0..8) was calling `func_count_support` for **all 9 directions** -- the original plus 8 EPS-perturbed variants. With the corrected count function, meshes with flat faces (very common: cube-like links on Franka) return count > 1 for the original direction. The loop then tries 8 perturbed directions, but since the EPS perturbation is far smaller than the support-field grid resolution, it maps to the **same grid cells** and yields the **same count** every time. The result: 8 wasted count evaluations that always return > 1, the loop falls through to i=8, and the function returns the exact same i=0 baseline result.

## Fix

Compute `func_count_support` only once at i=0. For perturbed directions (i=1..8), skip the count entirely and go straight to computing support points + checking simplex validity via `func_is_new_simplex_vertex_valid`. This way:

- **count == 1** (unique support): break immediately, same cost as before.
- **count > 1** (ambiguous, e.g. flat face): perturbed directions now actually compute support points and check if the resulting simplex vertex is non-degenerate, instead of fruitlessly re-checking the same count.

### DIFF
--- a/genesis/engine/solvers/rigid/collider/gjk.py
+++ b/genesis/engine/solvers/rigid/collider/gjk.py
@@ -1770,6 +1770,14 @@ def func_safe_gjk_support(
     function perturbs the support direction to find the best support points that guarantee non-degenerate simplex
     in the GJK algorithm.
 
+    The support count (number of candidate support vertices tied for the maximum dot product) is only evaluated once
+    for the original direction (i=0). For perturbed directions (i=1..8), we skip the count and go straight to
+    computing support points and checking simplex validity. This is critical for performance because the count check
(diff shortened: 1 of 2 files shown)


hughperkins:
Interesting.

PR #2740, opened 2026-04-25
[Test] Skipping test_collision_edge_cases which fail on amdgpus with fast-math

jamesETsmith:
# Summary

These changes skip tests which pass when fast math is turned off. Since we want to leave fast math enabled in the JIT by default, we're skipping these tests to make the CI easier to work with.


jamesETsmith:
Apologies for the noise, I meant to open this against the ROCm fork of genesis.

duburcqa:
> Apologies for the noise, I meant to open this against the ROCm fork of genesis.

@jamesETsmith  What is that? We are currently trying to improve support of ROCm in Genesis. I'm not aware of any specific bugs. All the unit tests are now passing on main.

PR #2741, opened 2026-04-26
[MISC] Tensor 2: Migration of V to qd.Tensor.

hughperkins:


hughperkins:
@claude review

genesis/utils/misc.py
```diff
@@ -690,6 +696,9 @@ def qd_to_numpy(
         copy (bool, optional): Wether to enforce returning a copy no matter what. None to avoid copy if possible
         without raising an exception if not.
     """
+    if isinstance(value, qd.Tensor):
+        value = value._unwrap()
```
duburcqa: This should be moved in `qd_to_python`.

genesis/utils/misc.py
```diff
@@ -637,6 +640,9 @@ def qd_to_torch(
...
```
duburcqa: This should be moved in `qd_to_python`.

duburcqa: Actually not moved but duplicated. Since `qd_to_python` is called directly in a few cases.

duburcqa: Hum... actually it is no longer the case. Forget about this.

genesis/utils/misc.py
duburcqa: Forget about this.

duburcqa: Remove this one. It is useless.

genesis/utils/misc.py
duburcqa: You should be this under `if gs.use_zerocopy:` branch. Not for performance but for clarity.

genesis/utils/array_class.py
```diff
@@ -1,86 +1,58 @@
...
-
-def maybe_shape(shape, is_on):
-    return shape if is_on else ()
-
-
-@dataclass_transform(eq_default=True, order_default=True, kw_only_default=False, frozen_default=True)
-class AutoInitMeta(type):
     def __new__(cls, name, bases, namespace):
-        names = tuple(namespace["__annotations__"].keys())
+        names = tuple(namespace.get("__annotations__", {}).keys())
```
duburcqa: Why are you doing this? Not having annotations should fail hard.

genesis/utils/array_class.py
```diff
@@ -1,86 +1,58 @@
...
 
-
-PLACEHOLDER = V(dtype=gs.qd_float, shape=())
-
-
-def maybe_shape(shape, is_on):
-    return shape if is_on else ()
-
-
-@dataclass_transform(eq_default=True, order_default=True, kw_only_default=False, frozen_default=True)
```
duburcqa, line 41: Why are you removing this?

genesis/utils/array_class.py
```diff
@@ -1,86 +1,58 @@
...
             for key in names:
                 if key not in assigned:
                     raise TypeError(f"{name}() missing required argument: '{key}'")
-
-            # Set attributes
             for key, value in assigned.items():
                 setattr(self, key, value)
 
         namespace["__init__"] = __init__
-
```
duburcqa, line 75: Why are you stripping all the comments?

genesis/utils/array_class.py
```diff
@@ -156,26 +128,26 @@ def get_rigid_global_info(solver, kinematic_only):
...
-            mass_mat_L_bw=V(dtype=gs.qd_float, shape=()),
-            mass_mat_D_inv=V(dtype=gs.qd_float, shape=()),
-            mass_mat_mask=V(dtype=gs.qd_bool, shape=()),
-            mass_parent_mask=V(dtype=gs.qd_float, shape=()),
+            envs_offset=qd.Vector.tensor(3, gs.qd_float, (_B,), backend=_TENSOR_BACKEND),
+            gravity=qd.Vector.tensor(3, gs.qd_float, (), backend=_TENSOR_BACKEND),
+            meaninertia=qd.tensor(gs.qd_float, (), backend=_TENSOR_BACKEND),
+            n_awake_dofs=qd.tensor(gs.qd_int, (_B,), backend=_TENSOR_BACKEND),
+            n_awake_entities=qd.tensor(gs.qd_int, (_B,), backend=_TENSOR_BACKEND),
+            n_awake_links=qd.tensor(gs.qd_int, (_B,), backend=_TENSOR_BACKEND),
```
duburcqa: I don't like it. Could you keep some `V`...etc that include the backend? Having to specify the backend everywhere is adding a lot of noise without any advantage.

genesis/utils/misc.py
hughperkins: because the _T_tc is on the unwrapped value, so we need to unwrap it first? but we dont need to do that if we are just going to call qd_to_python immediately?

genesis/utils/array_class.py
hughperkins: put them back now, I think.

genesis/utils/array_class.py
hughperkins: reverted

genesis/utils/array_class.py
hughperkins: reverted

genesis/utils/misc.py
hughperkins: (these are questoins t oyou just to be clear. just clarifying this is why you want it moved under the if zerocopy)

genesis/utils/misc.py
hughperkins: done

genesis/utils/array_class.py
hughperkins: you mean, `V = partial(qd.tensor, backend=_TENSOR_BACKEND)`?

genesis/utils/misc.py
hughperkins: done

genesis/engine/solvers/fem_solver.py
```diff
@@ -376,7 +376,7 @@ def build(self):
         # Overwrite gravity because only field is supported for now
         if self._gravity is not None:
             gravity = self._gravity.to_numpy()
-            self._gravity = qd.field(dtype=gs.qd_vec3, shape=(self._B,))
+            self._gravity = array_class.V(gs.qd_vec3, (self._B,))
```
duburcqa: Are you sure this is supported? The kernels involving gravity are not assuming field? 

I don't think this would work.

genesis/engine/solvers/mpm_solver.py
```diff
@@ -237,7 +237,7 @@ def build(self):
         # Overwrite gravity because only field is supported for now
         if self._gravity is not None:
             gravity = self._gravity.to_numpy()
-            self._gravity = qd.field(dtype=gs.qd_vec3, shape=(self._B,))
+            self._gravity = array_class.V(gs.qd_vec3, (self._B,))
```
duburcqa: Same. I doubt this is valid.

genesis/engine/solvers/pbd_solver.py
```diff
@@ -232,7 +233,7 @@ def build(self):
         # Overwrite gravity because only field is supported for now
         if self._gravity is not None:
             gravity = self._gravity.to_numpy()
-            self._gravity = qd.field(dtype=gs.qd_vec3, shape=(self._B,))
+            self._gravity = array_class.V(gs.qd_vec3, (self._B,))
```
duburcqa: Same. I don't think this would work.

genesis/__init__.py
```diff
@@ -417,10 +417,13 @@ def destroy():
...
 
     # Clear global state
-    global _theme, device, backend, EPS
+    global _theme, device, backend, use_ndarray, use_fastcache, use_zerocopy, EPS
     _theme = None
     device = None
     backend = None
+    use_ndarray = None
+    use_fastcache = None
+    use_zerocopy = None
```
duburcqa: This is not valid. You cannot reset at least fastcache.

genesis/engine/solvers/fem_solver.py
hughperkins: Why do you feel this would not work? What do you think might break?

hughperkins: Opus says:

  These are @qd.kernel methods on a @qd.data_oriented class. self._gravity is accessed via self (a template struct). The _gravity attribute goes through the build_Attribute path
   (not the flattened name path) and gets unwrapped/promoted properly. It's used with simple indexing self._gravity[i_b] — standard subscript access.

  Let me check the _kernel_set_gravity in base_solver.py which uses qd.Tensor annotation:

   genesis/engine/solvers/base_solver.py lines 171-174
```
  @qd.kernel
  def _kernel_set_gravity(tensor: qd.types.ndarray(), envs_idx: qd.types.ndarray(), gravity: qd.Tensor):
          for j in qd.static(range(3)):
```
  That uses qd.Tensor which handles both backends. And the @qd.data_oriented class kernels access self._gravity as a struct field — which also works for both backends (that's
  what the build_Attribute / _promote_ndarray_if_declared path handles).

  So there's nothing FEM-specific that requires field. The "only field is supported" comment was likely written before ndarray support existed. The code should work fine with
  V() in ndarray mode — it's just subscript access _gravity[i_b], same pattern used everywhere else.

  The reviewer may be going by the old comment rather than the actual kernel code. I don't see any field-only API usage (no SNode operations, no _loop_range() calls, no
  field-specific methods).

hughperkins: ### does anything else use gravity?

  Every usage is the same pattern — self._gravity[i_b] inside @qd.kernel methods on @qd.data_oriented classes, plus _kernel_set_gravity which already uses qd.Tensor annotation.
  All of these work with both backends.

  The "Overwrite gravity because only field is supported for now" comment is copy-pasted across all 5 subclass solvers (fem, pbd, sf, sph, mpm). It's stale — that comment
  predates ndarray/Tensor support. The overwrite blocks themselves are now redundant since base_solver.build() (line 42) already creates _gravity via V(), and the subclasses
  re-create it with... the same V() call. They're overwriting it with an identical value.

  Actually, those overwrite blocks could just be deleted entirely now — base_solver.build() already does the right thing and the subclasses are just repeating it.

genesis/engine/solvers/mpm_solver.py
hughperkins: see above

genesis/engine/solvers/pbd_solver.py
hughperkins: see above

genesis/__init__.py
hughperkins: reverted

genesis/utils/array_class.py
hughperkins: done

genesis/utils/array_class.py
hughperkins: (so, addressed)

genesis/engine/solvers/base_solver.py
```diff
@@ -171,14 +168,7 @@ def _repr_brief(self):
...
 @qd.kernel
-def _kernel_set_gravity_field(tensor: qd.types.ndarray(), envs_idx: qd.types.ndarray(), gravity: qd.template()):
-    for i_b_ in range(envs_idx.shape[0]):
-        for j in qd.static(range(3)):
-            gravity[envs_idx[i_b_]][j] = tensor[i_b_, j]
-
-
-@qd.kernel
-def _kernel_set_gravity_ndarray(tensor: qd.types.ndarray(), envs_idx: qd.types.ndarray(), gravity: qd.types.ndarray()):
+def _kernel_set_gravity(tensor: qd.types.ndarray(), envs_idx: qd.types.ndarray(), gravity: qd.Tensor):
```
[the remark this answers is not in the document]
hughperkins: addressed

hughperkins:
@claude review

hughperkins:
(note: we need a new version of quadrants, to address a one-line bug fix needed for the new many-reset test to pass 😅 )

genesis/engine/solvers/sf_solver.py
```diff
@@ -1,7 +1,6 @@
 import quadrants as qd
 
 import genesis as gs
-
```
duburcqa, line 4: Why this change?

genesis/__init__.py
```diff
@@ -133,19 +133,9 @@ def init(
...
-            raise_exception("Genesis previous initialized. Quadrants dynamic array mode cannot be updated anymore.")
+    _use_ndarray = not (is_ndarray_disabled or performance_mode)
     is_fastcache_disabled = os.environ.get("GS_ENABLE_FASTCACHE", "1") == "0"
-    if use_fastcache is None:
-        _use_fastcache = not is_fastcache_disabled and _use_ndarray
-    else:
-        _use_fastcache = use_fastcache
-        if use_fastcache and is_fastcache_disabled:
-            raise_exception("Genesis previous initialized. Quadrants fast cache mode cannot be disabled anymore.")
+    _use_fastcache = not is_fastcache_disabled and _use_ndarray
```
duburcqa: You cannot do this for fastcache.

tests/test_backend_switching.py
```diff
@@ -0,0 +1,68 @@
...
+import os
+
+import numpy as np
+import pytest
+import quadrants as qd
+
+import genesis as gs
+
+
+def _run_cycle(use_nd, cycle_idx):
```
duburcqa: Why not just inlining this function inside 'test_backend_switching' since it is only used once?

genesis/utils/array_class.py
```diff
@@ -156,26 +152,26 @@ def get_rigid_global_info(solver, kinematic_only):
...
+            qpos=V(gs.qd_float, (solver.n_qs_, _B)),
+            qpos_next=V(gs.qd_float, (solver.n_qs_, _B)),
+            links_T=V_MAT(4, 4, gs.qd_float, (solver.n_links_,)),
+            geoms_init_AABB=V_VEC(3, gs.qd_float, ()),
+            mass_mat=V(gs.qd_float, ()),
+            mass_mat_L=V(gs.qd_float, ()),
+            mass_mat_L_bw=V(gs.qd_float, ()),
+            mass_mat_D_inv=V(gs.qd_float, ()),
+            mass_mat_mask=V(gs.qd_bool, ()),
+            mass_parent_mask=V(gs.qd_float, ()),
```
duburcqa: I'm not sure to see the point of this change.

genesis/utils/array_class.py
```diff
@@ -191,26 +187,26 @@ def get_rigid_global_info(solver, kinematic_only):
...
+        qpos=V(gs.qd_float, (solver.n_qs_, _B), needs_grad=requires_grad),
+        qpos_next=V(gs.qd_float, (solver.n_qs_, _B), needs_grad=requires_grad),
+        links_T=V_MAT(4, 4, gs.qd_float, (solver.n_links_,)),
+        geoms_init_AABB=V_VEC(3, gs.qd_float, (solver.n_geoms_, 8)),
+        mass_mat=V(gs.qd_float, mass_mat_shape, needs_grad=requires_grad),
+        mass_mat_L=V(gs.qd_float, mass_mat_shape, needs_grad=requires_grad),
+        mass_mat_L_bw=V(gs.qd_float, mass_mat_shape_bw, needs_grad=requires_grad),
+        mass_mat_D_inv=V(gs.qd_float, (solver.n_dofs_, _B), needs_grad=requires_grad),
+        mass_mat_mask=V(gs.qd_bool, (solver.n_entities_, _B)),
+        mass_parent_mask=V(gs.qd_float, (solver.n_dofs_, solver.n_dofs_)),
```
duburcqa: Same.

genesis/utils/array_class.py
hughperkins: you want it to use kwargs? It's not an intnetional change to not use kwargs, but, dont you normally prefer args?

genesis/utils/array_class.py
```diff
@@ -1,45 +1,45 @@
...
+PLACEHOLDER = V(gs.qd_float, ())
 
 
 def maybe_shape(shape, is_on):
     return shape if is_on else ()
 
 
 @dataclass_transform(eq_default=True, order_default=True, kw_only_default=False, frozen_default=True)
-class AutoInitMeta(type):
+class _AutoInitMeta(type):
```
duburcqa: I'm not sure to understand why adding '_' here. This module is not user facing and if a user need it, then this user will likely need this convenience metaclass.

genesis/utils/array_class.py
```diff
@@ -72,15 +72,11 @@ def __init__(self, *args, **kwargs):
                 setattr(self, key, value)
 
         namespace["__init__"] = __init__
-
```
duburcqa, line 75: Still one line skip deleted for no reason.

genesis/utils/array_class.py
```diff
@@ -100,42 +96,42 @@ class ErrorCode(IntEnum):
 # =========================================== RigidGlobalInfo ===========================================
 
 
-@DATA_ORIENTED
-class StructRigidGlobalInfo(metaclass=BASE_METACLASS):
+@dataclasses.dataclass(frozen=True)
+class StructRigidGlobalInfo:
```
duburcqa: Why are you no longer using our metaclass?

duburcqa: Hum ok I see. If you use dataclass everything, our meta class can be deleted entirely!

duburcqa: So all you need to do is pass `eq_default=True, order_default=True, kw_only_default=False, frozen_default=True` directly to `@dataclasses.dataclass` I guess, to preserve these nice properties.

genesis/engine/solvers/fem_solver.py
hughperkins: Ok, this was wrong.

- Updated to keep the other solvers using fiedl for gravity
- Added a unit test for calling set gravity for one of the other solvers
- added FIXMEs explaining the options for fixing this at some point.

genesis/utils/array_class.py
hughperkins: addressed

genesis/utils/array_class.py
hughperkins: addressed

genesis/utils/array_class.py
hughperkins: I agree that this PR should not be changing the visibliyt of existing things.

But I disagree that this module is not user-facing. It is clearly public. The module has no `_` prefix at any point in its path.

genesis/utils/array_class.py
```diff
@@ -1,34 +1,32 @@
...
 import quadrants as qd
+from typing_extensions import dataclass_transform  # Made it into standard lib from Python 3.12
 import numpy as np
 import torch
-from typing_extensions import dataclass_transform  # Made it into standard lib from Python 3.12
 
 import genesis as gs
 
 if not gs._initialized:
     gs.raise_exception("Genesis hasn't been initialized. Did you call `gs.init()`?")
```
duburcqa, line 13: You can finally remove this \o/

hughperkins: I was expecting one of the test cases to hit these. But they don't. Some test missing? This is just completely superfluous dead code?

duburcqa: It is orthogonal. I'm not asking you to add a unit test for this, but it would be great to have one indeed.
```python
import genesis as gs
import genesis.utils.array_class

genesis.init([...])

<run an extremely basic simulation, everything runs fine>
```
This should be done in a subprocess, since genesis is already imported and init/destroyed multiple times when hitting this test.

For me you test checking that you can toggle on/off performance mode between init/destroy is sufficient to prove that it works. Not the same high-level issue, but related to the same root cause.

genesis/utils/array_class.py
hughperkins: for dataclasses, per Opus:

  | @dataclass_transform param | Effect                                      | @dataclasses.dataclass equivalent |
  |----------------------------|---------------------------------------------|-----------------------------------|
  | eq_default=True            | Generate __eq__ and __ne__                  | eq=True (already the default)     |
  | order_default=True         | Generate __lt__, __le__, __gt__, __ge__     | order=True (would break: tensors don't support <) |
  | kw_only_default=False      | Allow positional args in __init__           | kw_only=False (already the default) |
  | frozen_default=True        | Make instances immutable (no setattr)       | frozen=True (already explicit)    |

summary: Opus says that these are already the case, except for order_default=True, which Opus says isn't currenlty possible.

Thoughts?

hughperkins: As far as AutoInitMeta, Opus says:  We can't delete AutoInitMeta entirely because 3 @qd.data_oriented classes still use it (StructColliderStaticConfig, StructGJKStaticConfig, StructRigidSimStaticConfig) — those
  hold plain Python types, not tensors.

genesis/utils/array_class.py
hughperkins: updated

tests/test_backend_switching.py
hughperkins: updated

genesis/utils/array_class.py
hughperkins: updated (per Opus)

genesis/utils/array_class.py
hughperkins: added test. guard removed.

(PS I thought this was a different exception actually. the one about ndarray vs field).

genesis/utils/array_class.py
hughperkins: added in https://github.com/Genesis-Embodied-AI/Genesis/pull/2741/changes/15b8104e1bb370e668cde8da904b28aadec12d0c

genesis/__init__.py
hughperkins: Opus says:

  The reviewer says "You cannot do this for fastcache." His concern is that we removed the guard that prevented use_fastcache from being changed between init() calls. On main,
  once use_fastcache was set, subsequent init() calls couldn't change it.

  The issue is: fastcache compiles kernels into a cached format tied to the ndarray/fastcache mode. If you switch fastcache on/off between init()/destroy() cycles, previously
  cached kernels may be invalid.

  However, our test_backend_switching test calls gs.destroy() between cycles, which calls qd.reset(). That clears the kernel caches. So switching fastcache between full
  init/destroy cycles should be safe — the kernel state is fully reset.

  The original guards were there for the case where init() was called without destroy() first. But gs.init() already checks _initialized and raises if you double-init. So the
  guard was redundant.

  My take: the change is safe because gs.destroy() → qd.reset() clears all compiled kernel state, making it fine to recompute use_fastcache on the next init(). The old guard was
   protecting against a scenario (re-init without destroy) that was already blocked by the _initialized check. But if the reviewer feels strongly, we could restore just the
  fastcache guard. Want me to reply with this reasoning, or would you rather just restore the guard to unblock?

hughperkins: Thoughts?

genesis/utils/array_class.py
duburcqa: Ok let's drop 'order_default'. Use explicit options rather than implicit. And keep 'AutoInitMeta' where it is still needed.

genesis/engine/solvers/sf_solver.py
hughperkins: addressed (per Opus)

genesis/__init__.py
duburcqa: Yes. You cannot do this for fastcache.

> The issue is: fastcache compiles kernels into a cached format tied to the ndarray/fastcache mode.

This is not the issue. The issue is just that fastcache is hard-coded at import time.

Now that I am thinking, it also means that 
```python
if not gs._initialized:
    gs.raise_exception("Genesis hasn't been initialized. Did you call `gs.init()`?")
```
cannot be safely removed yet...

genesis/utils/array_class.py
duburcqa: > you want it to use kwargs? It's not an intnetional change to not use kwargs, but, dont you normally prefer args?

yes, I much prefer explicit keywords in this context. Because the order here does not follow torch or numpy, so I would rather make it order independent.

hughperkins: Ok. And I suppose also because shape is anonymous?

genesis/utils/array_class.py
hughperkins: cool. question : do you mean drop order_default also from the metaclass?

duburcqa: yes.

genesis/utils/array_class.py
duburcqa: > And I suppose also because shape is anonymous?

yes.

genesis/utils/array_class.py
hughperkins: addressed in https://github.com/Genesis-Embodied-AI/Genesis/pull/2741/commits/68ec7a5815ee46f1e8521a31d3036f3710a3a90d

genesis/utils/array_class.py
hughperkins: (addressed)

hughperkins:
Note: the failing unit test is beacsue need a new Quadratns release, for the bakend siwtching test to pass.

genesis/utils/array_class.py
```diff
@@ -1,34 +1,29 @@
...
 import quadrants as qd
+from typing_extensions import dataclass_transform  # Made it into standard lib from Python 3.12
 import numpy as np
 import torch
-from typing_extensions import dataclass_transform  # Made it into standard lib from Python 3.12
 
 import genesis as gs
 
-if not gs._initialized:
-    gs.raise_exception("Genesis hasn't been initialized. Did you call `gs.init()`?")
```
duburcqa, line 15: Please add back this guard.

hughperkins: restored in https://github.com/Genesis-Embodied-AI/Genesis/pull/2741/changes/fa7a29c175d9d3b826d402daf451b192c341c72e

duburcqa (dismissed):

hughperkins:
Note: I verified on the 'always-fastcache' branch that fastcache was still working, on both ubuntu and mac. Before we merge this branch, should probably do a similar check (or else merge the 'always-fastcache' branch directly)

hughperkins:
also checked that fastcache is still working on this branch, on both ubuntu and mac

PR #2751, opened 2026-04-30
[MISC] Migrate to unified Quadrant's tensor abstraction.

hughperkins:
## Related Issue

## Checklist:


hughperkins:
At this point:
- benchmarks run
- mostly look ok
- except for anymal_uniform_kinematic
2026-04-30-1826-always-fastcache

hughperkins:
Updated benchmarks, with quadrnats main at this point (actually tested on a branch, but that has now been merged to main):

20260501-1517-non-field-plan-shortcut

hughperkins:
Note: I manually verified that fastcache was working, on both ubuntu, and on Mac.

But probably good that at least one other person does this.

genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -232,7 +232,7 @@ def add_inequality_constraints(self):
             self._solver._static_rigid_sim_config,
         )
 
-    def resolve(self):
+    def resolve(self, entities_info=None, rigid_global_info=None):
```
duburcqa, line 235: What is this?

genesis/utils/array_class.py
```diff
@@ -2134,37 +2127,37 @@ def get_viewer_raycast_result():
...
+MPRInfo = StructMPRInfo
+SupportFieldInfo = StructSupportFieldInfo
+ConstraintState = StructConstraintState
+GJKState = StructGJKState
+GJKInfo = StructGJKInfo
+SDFInfo = StructSDFInfo
+ContactIslandState = StructContactIslandState
+DiffContactInput = StructDiffContactInput
+RigidAdjointCache = StructRigidAdjointCache
+RaycastResult = StructViewerRaycastResult
```
duburcqa: What about renaming all the 'Struct' so that we can get rid of this useless mapping?

genesis/engine/solvers/rigid/constraint/solver.py
hughperkins: When we enable fastcache=True, these were violating purity.

genesis/utils/array_class.py
hughperkins: makes sense. was pondering that too. Which direction do you want to sync the names? (keep Struct, or remove Struct?)

hughperkins: (from your phrasing, sounds like: remove'Struct'?)

duburcqa: Remove struct indeed.

genesis/engine/solvers/rigid/constraint/solver.py
duburcqa: I don't understand how this related to violating purity. Here you are just changing the signature of a pure Python function.

hughperkins: Opus says:

  Now I see. The caller rigid_solver.py:1070 passes entities_info and rigid_global_info, but this solver.py resolve method doesn't use them — it accesses everything through
  self._solver. The =None defaults were added to make the signature compatible with the island solver's resolve(self, entities_info, rigid_global_info) being called
  polymorphically from the same call site.

  Let me check how the caller decides which solver to use:

Read ...is/genesis/engine/solvers/rigid/rigid_solver.py lines 1065-1074

  Right — self.constraint_solver is either ConstraintSolver (from solver.py) or ConstraintSolverIsland (from solver_island.py). The caller passes the args unconditionally, so
  solver.py's resolve needs to accept them to match the interface.

  The =None defaults are there purely for interface compatibility — this solver doesn't need them because it accesses everything via self._solver. It's a bit ugly but
  functional.

  My take: The comment is reasonable — it looks confusing. A cleaner approach would be either:

  1. Use `kwargs** to swallow unused args: def resolve(self, **kwargs):` — less noise but hides the interface
  2. Accept the params explicitly and just don't use them: def resolve(self, entities_info, rigid_global_info): — makes the shared interface obvious, no =None weirdness
  3. Add a comment explaining it's for interface parity with ConstraintSolverIsland.resolve

  Option 2 is cleanest — accept the params positionally (no defaults), making it clear both solvers share the same calling convention. Want me to fix it?

hughperkins: Note that (this is Hugh here) island solver resolve is a kernel:

Screenshot 2026-05-02 at 14 59 16

genesis/utils/array_class.py
hughperkins: addressed

genesis/engine/solvers/rigid/constraint/solver.py
duburcqa: Not a fan of modifying the resolve to accommodate legacy half-broken island solver.

duburcqa (approved):

PR #2761, opened 2026-05-04
[BUG FIX] Make Surface shortcut resolution idempotent

Kashu7100:
## Summary

The `model_validator(mode="after")` resolvers on `Surface` (`_resolve_shortcuts`) and `Glass` (`_post_init`) route shortcut fields (`color`, `opacity`, `roughness`, `metallic`, `emissive`, `thickness`) into their corresponding `*_texture` fields and then raise if both are populated. They never clear the shortcut after applying it, so the instance ends up with both fields set.

This works when constructing a surface in isolation, but raises `GenesisException: 'color' and 'diffuse_texture' cannot both be set.` the moment the surface is nested inside another Pydantic model (downstream frameworks that wrap Genesis options for declarative scene configs commonly hit this). Pydantic 2 re-runs after-mode validators on nested instances even with `revalidate_instances="never"` — the second pass sees both `color` and `diffuse_texture` populated and trips the conflict check.

## Repro

```python
from pydantic import BaseModel
import genesis as gs
from genesis.options.surfaces import Surface

class Wrapper(BaseModel):
    surface: Surface

s = gs.surfaces.Rough(color=(0.4, 0.4, 0.4))   # OK
Wrapper(surface=s)                              # GenesisException — 'color' and 'diffuse_texture' cannot both be set.
```

## Fix

Clear each shortcut after routing it into the texture field, so re-running the validator is a no-op. The `default_roughness` sync moves to the top of `_resolve_shortcuts` because the loop that consumes `self.roughness` now nulls it.

A repo-wide grep confirmed no other code reads `surface.<shortcut>` after construction — these fields have `Field(exclude=True, repr=False)` and are documented as input-only shortcut parameters.

## Test plan

- [x] `gs.surfaces.Rough(color=...)`, `Glass(color=..., thickness=...)`, `BSDF(color=..., roughness=..., metallic=...)`, `Emission(color=...)` round-trip safely through nested Pydantic models.
- [x] Asserts the resolved texture values are correct (not just non-None).
- [x] `default_roughness` sync still mirrors the `roughness` shortcut, and an explicit `default_roughness` still wins over the shortcut.
- [x] First-construction conflicts still raise (e.g., `Rough(color=..., diffuse_texture=...)`, `Glass(thickness=..., thickness_texture=...)`).
- [x] Re-using an already-resolved surface in multiple wrappers does not mutate it further.

### DIFF
--- a/genesis/options/surfaces.py
+++ b/genesis/options/surfaces.py
@@ -3,7 +3,7 @@ from typing import ClassVar, Literal
 from typing_extensions import Self
 
 import numpy as np
-from pydantic import Field, StrictBool, model_validator
+from pydantic import Field, PrivateAttr, StrictBool, model_validator
 
 import genesis as gs
 from genesis.typing import FArrayType, UnitInterval, ValidFloat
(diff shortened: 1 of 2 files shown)


Kashu7100:
tests are failing due to fetch issue - not related to this PR.

genesis/options/surfaces.py
duburcqa, line 1: I'm not a fan of this design. I will try to find a better way.

Kashu7100: SGTM

duburcqa: Reworked: validator is now `model_validator(mode="before")` classmethod, matching the `morphs.py:_resolve_orientation` pattern for input-only shortcuts. Drops the `PrivateAttr` state flag entirely - `mode="before"` does not fire on re-validation of an already-built instance, so re-entry is naturally a no-op.

PR #2721, opened 2026-04-20
[FEATURE] Support visual mesh deformation of KinematicEntity and rendering of KinematicEntity in raycast sensor.

Kashu7100:
## Description

- Multi-solver BVH raycasting. RaycasterSharedMetadata now carries an extra_visual_bvhs
  list. Per-solver BVHs are built and traversed independently; per-ray results are merged
  via a new kernel_merge_ray_hits that keeps the closest hit. A single depth camera can
  now see rigid + kinematic geometry simultaneously.
- Cross-solver sensor attachment. Replaced the single-solver links_idx lookup with
  per-sensor (solver, link_idx) pairs. _gather_sensor_link_poses() groups sensors by
  solver, does one bulk get_links_pos/quat per solver, and scatters into a (B, n_sensors,
  3/4) tensor. Static sensors (entity_idx=-1) get identity transforms with properly padded
   per-sensor arrays.
- Material-level opt-in. Moved use_visual_raycasting from an entity property to
  gs.materials.Kinematic, so visibility to sensors is configured at entity creation.
  Non-opt-in entities have their vverts moved to (1e10, 1e10, 1e10) after FK so their
  AABBs fall outside any ray's max_range — no kernel-signature changes needed.
- Robustness fixes. Raise on cross-solver metadata mismatch; force-run FK before the
  first BVH build so non-root kinematic links have valid poses; validate no_hit_value >=
  max_range when multi-solver merge is active; setter for use_visual_raycasting raises
  post-build instead of silently no-oping; set _is_forward_pos_updated=True after internal
   FK to avoid redundant per-frame recomputation.
- Examples. examples/sensors/depth_camera_custom_vverts.py exercises the full pipeline:
  Go2 quadruped (rigid, articulated) + ground plane + a deforming kinematic sphere (opted
  in) + a static kinematic box (opted out, visible in viewer but absent from depth).
  examples/rendering/custom_visual_mesh.py updated to the new material API.

## Related Issue

Resolves Genesis-Embodied-AI/Genesis#<your-issue-number>

## Checklist:

- [ ] I read the **CONTRIBUTING** document.
- [ ] I followed the `Submitting Code Changes` section of **CONTRIBUTING** document.
- [ ] I tagged the title correctly (including BUG FIX/FEATURE/MISC/BREAKING)
- [ ] I updated the [documentation](https://github.com/Genesis-Embodied-AI/genesis-doc) accordingly or no change is needed.
- [ ] I tested my changes and added instructions on how to test it for reviewers.

[...]

PR #2768, opened 2026-05-09
[FEATURE] Per-frame visual vertex deformation for KinematicEntity

Kashu7100:
## Summary

Adds `KinematicEntity.set_vverts(vverts, envs_idx=None)` so externally-skinned meshes (e.g. SMPL body models, deforming objects) can drive visual vertex positions per frame without going through the per-vgeom transform path.

The rasterizer grows a per-env "skinned" node path: when an entity has `has_custom_vverts` set, each rendered env gets its own pyrender node and `_update_rigid_custom_vverts` blits the user vertex buffer into the node's GL buffers (positions and recomputed normals) per frame. The instanced `rigid_node` is lazily migrated on first use, so paying the per-env cost is opt-in via `set_vverts`.

Mesh topology is unchanged; only vertex positions move per frame, so the rasterizer's vertex-reorder cache hit path stays valid.

## Files changed

- `genesis/engine/entities/rigid_entity/rigid_entity.py` — `_custom_vverts` buffer, `has_custom_vverts` property, `set_vverts(vverts, envs_idx=None)` API.
- `genesis/vis/rasterizer_context.py` — per-env `skinned_nodes` dict, `add_skinned_node`, `_update_rigid_custom_vverts`, `_seg_key_for_geom` helper extracted for reuse, lazy migration from instanced rigid node.
- `examples/rendering/custom_visual_mesh.py` — wave-deforming subdivided box (no external deps) and SMPL body model demo (requires `smplx`).

## Relationship to the previous PR

This PR is the rendering half of #2721. The other half — visual-mesh raycasting (depth camera / lidar against `KinematicEntity`) — is split into a separate dependent PR.

## Test plan

- [ ] `python examples/rendering/custom_visual_mesh.py -v` renders a wave-deforming box that visibly deforms each frame
- [ ] `python examples/rendering/custom_visual_mesh.py -v -B 4` renders 4 batched envs each with their own deformation phase
- [ ] `python examples/rendering/custom_visual_mesh.py -v --smpl --model_path <path>` renders an animated SMPL body
- [ ] Existing rendering paths unaffected for entities without `set_vverts`

### DIFF
--- a/examples/rendering/custom_visual_mesh.py
+++ b/examples/rendering/custom_visual_mesh.py
@@ -1,0 +1,173 @@
+"""
+Custom Visual Mesh in Genesis
+==============================
+
+Demonstrates how to use ``set_vverts()`` on a ``KinematicEntity`` to update
+visual vertex positions at runtime.
+
+``set_vverts()`` is a fast path where only vertex positions change while the
+mesh topology (faces) stays constant. This is ideal for SMPL-style body
+models and other externally-skinned meshes.
+
+Requirements (only for SMPL demo)
+----------------------------------
+- ``smplx`` package  (``pip install smplx``)
+- SMPL model files   (download from https://smpl.is.tue.mpg.de)
+
+Usage
+-----
+    # Wave-deforming box (no external dependencies)
+    python custom_visual_mesh.py -v
+
+    # SMPL body mesh (requires smplx)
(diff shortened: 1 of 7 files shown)


duburcqa:
it turns out the split is a bad idea. Let's close this and merge with https://github.com/Genesis-Embodied-AI/genesis-world/pull/2769.

PR #2770, opened 2026-05-11
[MISC] Refactor sensor update.

Milotrince:
## Description
- Include ground truth update to `_update_shared_cache()` and remove `_update_shared_ground_truth_cache()`. 

## Related Issue
Sensor infra update for future new sensor features.

## Motivation and Context
For noise that affects intermediate calculations (as opposed to imperfections that postprocess the final output value, e.g. a noise on raycast direction) this would allow for one kernel call to handle both the noised and unnoised outputs, without having to store/copy values in shared_metadata.

All sensor tests pass

## Checklist:

### DIFF
--- a/genesis/engine/sensors/base_sensor.py
+++ b/genesis/engine/sensors/base_sensor.py
@@ -1,6 +1,6 @@
 from dataclasses import dataclass, field
 from functools import partial
-from typing import TYPE_CHECKING, ClassVar, Generic, NamedTuple, TypeVar, get_args, get_origin
+from typing import TYPE_CHECKING, ClassVar, Generic, NamedTuple, Sequence, TypeVar, get_args, get_origin
 
 import numpy as np
 import torch
@@ -50,8 +50,8 @@ class SharedSensorMetadata:
     history_lengths: list[int] = field(default_factory=list)
     # Per-sensor: interpolate delayed reads (aligned with cache_sizes). Appended in Sensor.build.
     interpolate: list[bool] = field(default_factory=list)
-    # True iff at least one sensor in the class has a nonzero read delay. Precomputed at build (and latched True
-    # by `set_delay`) so the per-step fast path in SensorManager._apply_delay_to_shared_cache can avoid a GPU-syncing reduction.
+    # True iff at least one sensor in the class has a nonzero read delay. Precomputed at build (and latched True by
+    # `set_delay`) so the per-step fast path in `_apply_delay_to_shared_cache` can avoid a GPU-syncing reduction.
     has_any_delay: bool = False
     # True iff at least one sensor in the class has a nonzero jitter option. Stays False for non-noisy sensor classes
     # (Contact, Raycaster) since they never sample jitter. Same precompute-and-latch contract as `has_any_delay`.
@@ -275,6 +275,73 @@ class Sensor(RBC, Generic[OptionsT, SharedSensorMetadataT, DataT]):
         return self._is_built
 
     # =============================== private shared methods ===============================
+
+    @classmethod
+    def _apply_delay_to_shared_cache(
+        cls,
+        shared_metadata: SharedSensorMetadataT,
+        shared_cache: torch.Tensor,
+        measured_data_timeline: "TensorRingBuffer",
+        interpolate: Sequence[bool] | None = None,
+    ):
+        """
+        Applies the read delay (and jitter, if any) to the shared cache tensor by copying the buffered data at the
+        appropriate index. When the class has neither delay nor jitter, takes a fast path that writes the most
+        recent ring frame to the cache class-wide.
+
+        Parameters
+        ----------
+        shared_metadata : SharedSensorMetadata
+            The shared metadata for the sensor.
+        shared_cache : torch.Tensor
+            The shared cache tensor.
+        measured_data_timeline : TensorRingBuffer
+            Pre-delay measured timeline ring for this sensor class slice (current step already written by
+            ``_update_shared_cache``).
+        interpolate : Sequence[bool] | None
+            Whether to interpolate the sensor data for the read delay + jitter. Defaults to False.
+        """
+        # Fast path: when no sensor in the class has any delay or jitter, this is equivalent to copying the most
+        # recent ring frame to the measured cache class-wide. Avoids a Python loop over every sensor. Both flags
+        # are precomputed Python bools (no GPU sync) latched at build / by setters.
+        if not shared_metadata.has_any_delay and not shared_metadata.has_any_jitter:
+            shared_cache.copy_(measured_data_timeline.at(0))
+            return
+
+        # Sample uniform jitter in [0, jitter_ts) per env per sensor. One-sided so samples are non-negative;
+        # combined with the `jitter < dt` option constraint, the effective per-step shift is strictly bounded
+        # within [0, 1) steps and cannot wrap the ring.
+        if shared_metadata.has_any_jitter:
+            shared_metadata.cur_jitter_ts.uniform_(0.0, 1.0).mul_(shared_metadata.jitter_ts)
+            cur_jitter_ts = shared_metadata.cur_jitter_ts
+        else:
+            cur_jitter_ts = None
+
+        if interpolate is None:
+            interpolate = [False for _ in shared_metadata.cache_sizes]
+
+        tensor_start = 0
+        for sensor_idx, (tensor_size, interp) in enumerate(zip(shared_metadata.cache_sizes, interpolate)):
+            # Compute the current delay of the sensor, taking into account jitter if any
+            cur_delay_ts = shared_metadata.delays_ts[:, sensor_idx]
+            if cur_jitter_ts is not None:
+                cur_delay_ts = cur_delay_ts + cur_jitter_ts[:, sensor_idx]
+
+            # Get int for indexing into ring buffer (0 = most recent, 1 = delayed by one timestep, etc.)
+            cur_delay_ts_int = cur_delay_ts.to(dtype=torch.int64)
+
+            # Update shared cached with left data (Zero Order Hold) or linearly interpolated data (First Order)
+            tensor_slice = slice(tensor_start, tensor_start + tensor_size)
+            sensor_cache = shared_cache[:, tensor_slice]
+            data_left = measured_data_timeline.at(cur_delay_ts_int, tensor_slice, per_row=True)
+            if interp:
+                ratio = torch.frac(cur_delay_ts)
+                data_right = measured_data_timeline.at(cur_delay_ts_int + 1, tensor_slice, per_row=True)
+                torch.lerp(data_left, data_right, ratio[:, None], out=sensor_cache)
+            else:
+                sensor_cache.copy_(data_left)
+
+            tensor_start += tensor_size
 
     def _get_formatted_data(self, tensor: torch.Tensor, envs_idx=None) -> torch.Tensor:
         """
--- a/genesis/engine/sensors/base_sensor.py
+++ b/genesis/engine/sensors/base_sensor.py
@@ -1,6 +1,6 @@
 from dataclasses import dataclass, field
 from functools import partial
-from typing import TYPE_CHECKING, ClassVar, Generic, NamedTuple, Sequence, TypeVar, get_args, get_origin
+from typing import TYPE_CHECKING, ClassVar, Generic, NamedTuple, TypeVar, get_args, get_origin
 
 import numpy as np
 import torch
@@ -48,8 +48,10 @@ class SharedSensorMetadata:
     cache_sizes: list[int] = field(default_factory=list)
     delays_ts: torch.Tensor = make_tensor_field((0, 0), dtype_factory=lambda: gs.tc_int)
     history_lengths: list[int] = field(default_factory=list)
+    # Per-sensor: interpolate delayed reads (aligned with cache_sizes). Appended in Sensor.build.
+    interpolate: list[bool] = field(default_factory=list)
     # True iff at least one sensor in the class has a nonzero read delay. Precomputed at build (and latched True
-    # by `set_delay`) so the per-step fast path in `_apply_delay_to_shared_cache` can avoid a GPU-syncing reduction.
+    # by `set_delay`) so the per-step fast path in SensorManager._apply_delay_to_shared_cache can avoid a GPU-syncing reduction.
     has_any_delay: bool = False
     # True iff at least one sensor in the class has a nonzero jitter option. Stays False for non-noisy sensor classes
     # (Contact, Raycaster) since they never sample jitter. Same precompute-and-latch contract as `has_any_delay`.
@@ -174,6 +176,7 @@ class Sensor(RBC, Generic[OptionsT, SharedSensorMetadataT, DataT]):
         )
         self._shared_metadata.cache_sizes.append(self._cache_size)
         self._shared_metadata.history_lengths.append(self._options.history_length)
+        self._shared_metadata.interpolate.append(getattr(self._options, "interpolate", False))
         if self._delay_ts > 0:
             self._shared_metadata.has_any_delay = True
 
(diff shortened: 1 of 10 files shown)


duburcqa:
### What the PR currently does

1. Folds `_update_shared_ground_truth_cache` into `_update_shared_cache`. Each sensor now writes both ground truth and pre-delay measured values in one classmethod.
2. New `_update_shared_cache` signature: `(cls, shared_metadata, current_ground_truth_data_T, measured_data_timeline)`.
3. Moves `_apply_delay_to_shared_cache` from a `Sensor` classmethod to a `SensorManager` instance method.
4. Adds `_measured_timeline_ring` on `SensorManager`, sharing its `_idx` counter with `_gt_timeline_ring`.
5. Migrates `interpolate: list[bool]` from `ImperfectSensorMetadataMixin` up into the base `SharedSensorMetadata`.
6. Rebalances `test_contact_sensors_gravity_force` step counts (1+10+90=101 → 3+20+80=103) and introduces `DT = 1e-2`, `DELAY_STEPS = 2` constants.

### Audit findings (issues with the PR as submitted)

| # | Location | Issue | Severity |
|---|---|---|---|
| 1 | `base_sensor.py:178` | `getattr(self._options, "interpolate", False)` — explicit CLAUDE.md violation | blocker |
| 2 | `tests/test_sensors.py` | Dead mock method `_update_current_ground_truth_data_T` overrides nothing | blocker |
| 3 | `temperature.py` | `# use torch for fft` noise comment (function name already says it) | blocker |
| 4 | `sensor_manager.py` | `gt_slice` / `_gt_timeline_ring` — `gt` abbreviation forbidden by user | blocker |
| 5 | `base_sensor.py` | `interpolate` moved from `ImperfectSensorMetadataMixin` to base `SharedSensorMetadata` — wrong layer; root cause of issue #1 | blocker |
| 6 | `*sensor.py` (×6) | `measured = measured_data_timeline.at(0, copy=False); measured.copy_(current_ground_truth_data_T.T)` repeated 6× — symptom of misplaced responsibility | design smell |
| 7 | `sensor_manager.py` | `_apply_delay_to_shared_cache` placed on manager — sensor-class polymorphism doesn't belong on the manager | design smell |
| 8 | `tests/test_sensors.py` | `test_contact_sensors_gravity_force` step rebalance lacks clear justification | verify |

### User priorities for this PR

- Be very picky; make sure everything is done right.
- A perfect sensor does not have delay.
- Derived sensors should not be responsible for boilerplate that has only one correct implementation.
- Masking is **not** an imperfection — it's a sensor-physics constraint.

## Design evolution (back-and-forth that led to v12)

The plan went through twelve revisions; recording the discarded paths matters because future-you may otherwise re-propose them.

### v1 — Each sensor calls `_seed_measured` helper on `Sensor` base
**Rejected.** "It is not the responsibility of derived sensors to deal with this."

### v2 — `SensorManager` seeds after the sensor populates ground truth
**Rejected.** "It is not ideal." The manager shouldn't own per-sensor compute; that's where the smell in audit finding #7 comes from.

### v3 — Template method: `Sensor._update_shared_cache` is concrete on the base, calls abstract `_update_ground_truth` then auto-seeds; `ImperfectSensorMetadataMixin._update_shared_cache` overrides via `super()` and appends `_apply_imperfections`
**Accepted in shape, refined.** This is the spine of the final design.

### v4 — Added `_apply_intermediate` hook between delay and imperfections
**Rejected.** "This could just be implemented as part of `cls._apply_imperfections`." Folded `_apply_intermediate` into `_apply_imperfections`. (Later partially reverted — see v10 — and ultimately re-rejected in v11 in favor of `_transform_measurement`.)

### v5 — `_apply_delay_to_shared_cache` as a staticmethod helper on `ImperfectSensorMetadataMixin`
**Rejected.** "Then just remove `_apply_delay_to_shared_cache` entirely and inline it in `_make_measurement`." It was only called from a non-overrideable method, so the indirection earned nothing.

### v6 — `delay` stays on base `SensorOptions`; `DelayedSensor*` mixins as a follow-up
**Rejected (misunderstanding).** User clarified the follow-up was about *removing* delay from basic sensors, not introducing the mixin. The mixin discussion belonged in this PR.

### v7 — Full orthogonalization: `DelayedSensorOptionsMixin` + `DelayedSensorMetadataMixin` independent of `Imperfect*`
**Rejected.** No real sensor mixes in `Delayed` without `Imperfect` or vice versa. The orthogonalization is theoretical with zero users. YAGNI. User: "if all sensors are already always combining `DelayedSensorOptionsMixin + ImperfectSensorOptionsMixin` or using neither, why not just settle to what we said and move delay in `ImperfectSensorOptionsMixin` directly?"

### v8 — Single `ImperfectSensorOptionsMixin` owns `delay` + `jitter` + `interpolate` + `noise` + `bias` + `quantize`
**Accepted.** Contact and Raycaster reject those options at construction (Pydantic `extra="forbid"`). This is the intended API break ("a perfect sensor does not have delay").

### v9 — Removed `_seed_measurement` from base `Sensor`
**Accepted.** Basic sensors never touch a timeline ring, so `_seed_measurement` lives only on `ImperfectSensorMetadataMixin`.

### v10 — Split `_apply_imperfections` into a non-overrideable core flanked by `_apply_pre_imperfections` / `_apply_post_imperfections` hooks
**Rejected.** Naming was poor ("`_apply_pre_imperfections` is only relevant for mixing measurement and GT — should be stated in the name"). Also: "clamp should probably go in `_apply_imperfections`, asking the user to overload it. It seems its natural place."

### v11 — `_transform_measurement` (pre, no `super()` discipline) + `_apply_imperfections` (overrideable, `super()` discipline). Tried folding ContactForce's `mask` into the `_apply_imperfections` override
**Rejected.** "Masking is NOT an imperfection. I guess it is `_make_measurement` that should be overloaded to add a post-processing step of masking."

### v12 — Locked
- `_seed_measurement` on imperfect mixin, overrideable (Temperature RC filter).
- `_transform_measurement` on imperfect mixin, overrideable, no-`super()` discipline (IMU rotation).
- `_apply_imperfections` on imperfect mixin, overrideable, `super()` discipline (ContactForce: `super()` + clamp).
- `_make_measurement` overrideable as a last-resort hook for sensor-physics post-processing (ContactForce: `super()` + mask).
- Delay sampling inlined inside `_make_measurement` (not a separate method).
- `_apply_delay_to_shared_cache` deleted entirely.

## Final design (v12)

### Options layer

```python
# genesis/options/sensors/options.py
class SensorOptions:                                # base — no delay, no imperfections
    name: str | None
    history_length: NonNegativeInt = 0
    ...

class ImperfectSensorOptionsMixin(SensorOptions):   # owns delay + imperfections (single mixin)
    delay: NonNegativeFloat = 0.0
    jitter: NonNegativeFloat = 0.0
    interpolate: StrictBool = False
    noise: ...
    bias: ...
    quantize: ...

    def validate_scene(self, scene):
        super().validate_scene(scene)
        # delay-vs-dt check moves here from SensorOptions
```

**API break:** Contact and Raycaster lose `delay`, `jitter`, `interpolate`. Pydantic `extra="forbid"` rejects callers at construction.

### Metadata layer

```python
# genesis/engine/sensors/base_sensor.py
@dataclass
class SharedSensorMetadata:                              # base — no delay state, no imperfection state
    cache_sizes: list[int] = field(default_factory=list)
    history_lengths: list[int] = field(default_factory=list)
    solver: ...
    ...
    # Note: `interpolate` was moved to base in the PR — that move is REVERTED.

@dataclass
class ImperfectSensorMetadataMixin(SharedSensorMetadata):
    delays_ts: torch.Tensor = ...
    jitter_ts: torch.Tensor = ...
    cur_jitter_ts: torch.Tensor = ...
    interpolate: list[bool] = field(default_factory=list)
    has_any_delay: bool = False
    has_any_jitter: bool = False
    noise_std: torch.Tensor = ...
    bias: torch.Tensor = ...
    quantize_step: torch.Tensor = ...
```

### Sensor base

```python
class Sensor:
    @classmethod
    def _update_shared_cache(cls, metadata, current_ground_truth_data_T, measured_data_timeline, shared_cache):
        # Standard orchestrator. Override-able as an escape hatch for "internal noise" sensors
        # (e.g. noise on raycast direction) where a single kernel writes both GT and measured
        # in one pass. Current sensors all fit the standard chain and do NOT override this.
        cls._update_ground_truth(metadata, current_ground_truth_data_T)
        cls._make_measurement(metadata, current_ground_truth_data_T, measured_data_timeline, shared_cache)

    @classmethod
    def _update_ground_truth(cls, metadata, current_ground_truth_data_T):
        raise NotImplementedError

    @classmethod
    def _make_measurement(cls, metadata, current_ground_truth_data_T, measured_data_timeline, shared_cache):
        # Basic (perfect, untimed) sensor: measured == ground truth. `measured_data_timeline` is None.
        shared_cache.copy_(current_ground_truth_data_T.T)
```

### Imperfect mixin chain

```python
class ImperfectSensorMetadataMixin:
    @classmethod
    def _make_measurement(cls, metadata, current_ground_truth_data_T, measured_data_timeline, shared_cache):
        cls._seed_measurement(metadata, current_ground_truth_data_T, measured_data_timeline)
        # ---- inlined delay sampling: writes shared_cache from measured_data_timeline ----
        if not metadata.has_any_delay and not metadata.has_any_jitter:
            shared_cache.copy_(measured_data_timeline.at(0))
        else:
            if metadata.has_any_jitter:
                metadata.cur_jitter_ts.uniform_(0.0, 1.0).mul_(metadata.jitter_ts)
                cur_jitter_ts = metadata.cur_jitter_ts
            else:
                cur_jitter_ts = None
            tensor_start = 0
            envs_idx = metadata.solver.scene._envs_idx
            for sensor_idx, (tensor_size, interp) in enumerate(zip(metadata.cache_sizes, metadata.interpolate)):
                cur_delay_ts = metadata.delays_ts[:, sensor_idx]
                if cur_jitter_ts is not None:
                    cur_delay_ts = cur_delay_ts + cur_jitter_ts[:, sensor_idx]
                cur_delay_ts_int = cur_delay_ts.to(dtype=torch.int64)
                tensor_slice = slice(tensor_start, tensor_start + tensor_size)
                sensor_cache = shared_cache[:, tensor_slice]
                data_left = measured_data_timeline.at(cur_delay_ts_int, envs_idx, tensor_slice)
                if interp:
                    ratio = torch.frac(cur_delay_ts)
                    data_right = measured_data_timeline.at(cur_delay_ts_int + 1, envs_idx, tensor_slice)
                    torch.lerp(data_left, data_right, ratio[:, None], out=sensor_cache)
                else:
                    sensor_cache.copy_(data_left)
                tensor_start += tensor_size
        # ----------------------------------------------------------------------------------
        cls._transform_measurement(metadata, shared_cache)
        cls._apply_imperfections(metadata, shared_cache)

    @classmethod
    def _seed_measurement(cls, metadata, current_ground_truth_data_T, measured_data_timeline):
        measured_data_timeline.at(0, copy=False).copy_(current_ground_truth_data_T.T)

    @classmethod
    def _transform_measurement(cls, metadata, shared_cache):
        pass  # no-op default — only IMU overrides for frame rotation

    @classmethod
    def _apply_imperfections(cls, metadata, shared_cache):
        # generic noise / bias / quantize
        ...
```

### Override matrix

| Sensor       | `_update_ground_truth` | `_seed_measurement`              | `_transform_measurement` | `_apply_imperfections`   | `_make_measurement`     |
| ------------ | ---------------------- | -------------------------------- | ------------------------ | ------------------------ | ----------------------- |
| Contact (bool) | yes                  | —                                | —                        | —                        | —                       |
| Raycaster    | yes                    | —                                | —                        | —                        | —                       |
| Camera       | unchanged              | —                                | —                        | —                        | —                       |
| ContactForce | yes                    | —                                | —                        | `super()` + clamp        | `super()` + mask        |
| IMU          | yes                    | —                                | frame rotation           | —                        | —                       |
| Proximity    | yes                    | —                                | —                        | —                        | —                       |
| KinematicTactile (contact) | yes      | —                                | —                        | —                        | —                       |
| KinematicTactile (elastomer) | yes    | —                                | —                        | —                        | —                       |
| Temperature  | yes                    | RC filter (replaces default)     | —                        | —                        | —                       |

### Conceptual split between override points

- **`_seed_measurement`** — stateful pre-delay seeding. Default: copy ground truth into timeline slot 0. Override when the measured signal at slot 0 depends on prior measured slots (Temperature's RC filter).
- **`_transform_measurement`** — deterministic transform of the post-delay value before generic noise. Default: no-op. Override when the sensor's output frame differs from ground truth's frame (IMU's body-frame rotation).
- **`_apply_imperfections`** — generic noise/bias/quantize, plus sensor-specific output-range constraints. Override with `super()` + extra when the sensor's output has additional range limits (ContactForce's clamp).
- **`_make_measurement`** — last-resort sensor-physics post-processing that doesn't fit any leaf. Override with `super()` + extra. (ContactForce's mask: zero-contact bins read exactly zero regardless of noise — a sensor-detection characteristic, not an imperfection.)

### Compatibility with Trinity's original motivation

Trinity opened the PR to enable two things:
1. Combine GT and measured writes into one method (`_update_shared_cache`).
2. Allow sensors whose noise lives *inside* the GT kernel (e.g. noised raycast direction) to write both noised and unnoised outputs in a single kernel call, without copying intermediate state through `shared_metadata`.

v12 supports both:
- (1) directly — `Sensor._update_shared_cache` is the single orchestrator that produces both buffers.
- (2) via the escape hatch — a future "internal noise" sensor overrides `_update_shared_cache` itself. The signature includes both `current_ground_truth_data_T` and `measured_data_timeline`, so a custom kernel can write both in one pass and skip the standard `_make_measurement` chain entirely.

The standard chain (seed → delay → transform → imperfections) is the clean path for 100% of current sensors. The escape hatch covers Trinity's future motivation without polluting the common case.

### Why mask cannot be folded elsewhere

- Pre-noise (`_transform_measurement`): zero-contact bins would receive noise additively, becoming ghost contacts.
- Inside `_apply_imperfections`: conceptually wrong — mask is a sensor-physics gate, not a perturbation. The user explicitly rejected this in v11.
- Post-noise via a separate post hook: rejected in v10 in favor of folding into `_apply_imperfections`. Going back to a post hook just to host mask would be excessive.
- Solution: override `_make_measurement` directly with `super()` + mask. Rare; ContactForce is the only current user.

### `SensorManager`

- Delete `SensorManager._apply_delay_to_shared_cache` (logic inlined inside `ImperfectSensorMetadataMixin._make_measurement`).
- Rename `_gt_timeline_ring` → `_ground_truth_timeline_ring`. Rename `gt_slice` → `ground_truth_slice`. Grep-audit for stragglers.
- Allocate `_measured_timeline_ring[dtype]` only when at least one sensor in that dtype is imperfect. Basic-sensor dtypes get `None`. The base `Sensor._make_measurement` never references `measured_data_timeline`, so `None` is harmless without a guard.
- `step()` becomes:
  ```python
  for ring in self._ground_truth_timeline_ring.values():
      ring.rotate()
  for ring in self._measured_timeline_ring.values():
      ring.rotate()
  for sensor_cls, group in self._sensor_groups.items():
      metadata = group.metadata
      ground_truth_slice = self._ground_truth_timeline_ring[group.dtype].at(0, copy=False)
      measured_timeline = self._measured_timeline_ring.get(group.dtype)
      shared_cache = group.shared_cache
      sensor_cls._update_shared_cache(metadata, ground_truth_slice, measured_timeline, shared_cache)
      # write into history ring if history_length > 0
  ```
- Index-sharing between the two ring dicts is preserved (existing canonical pattern in `TensorRingBuffer.__getitem__(idx=self._idx)`).
- The `interpolate.append(...)` build step lives on the imperfect-sensor build path (where `self._options.interpolate` is statically present). The `getattr(self._options, "interpolate", False)` line in `base_sensor.py:178` is deleted.

### Behavior changes worth flagging in the commit message

- Contact and Raycaster reject `delay=`, `jitter=`, `interpolate=` at construction (Pydantic `extra="forbid"`).
- Temperature: imperfections previously applied *inside* the seed/filter step (noise propagated through the ring). After this refactor, imperfections apply to `shared_cache` *after* delay — fresh noise per output, not baked into the timeline. Affects history reads only; the current measurement at each step is unchanged in distribution.
- `metadata.interpolate` no longer exists on basic sensors' metadata. Any external code reading it on a basic sensor would have always been meaningless; now it fails loudly.

Milotrince:
1. i think all sensors should have delay and history, but by moving the delay handling to `ImperfectSensorMixin` sensors which dont inherit `ImperfectSensorMixin` won't have delay.
  - interpolation doesn't make sense for the boolean `Contact` sensor or `Raycaster` 
  - white noise, etc. doesn't make sense for the boolean `Contact` sensor
2. whats the motivation for separating `_apply_imperfections` from `_transform_measurement`?

so I would keep delay on base sensor but other than that i think the plan is good

duburcqa:
Interpolation do make sense for Contact. Why not? And all sensor that have delay to expose some kind of interpolation, even if it is a zero-order-hold. Similarly, even Contact sensor are affected by white noise in reality. Why not? But the noise must happen at the right place, ie before it is casted as a Boolean.

Delay is a form of imperfection, why should it be part of all base (perfect) sensors? We can have a dedicated mixin for delay if you prefer.

Apply transform and imperfections are 2 different things. Transform is what gives you to “real” signal from the raw data, and apply imperfection is what applies at noise etc. Actually we should have pre and post transform to handle all types of sensors properly, typically for Boolean sensors casting should move that the every end of the process physically.

duburcqa:
Actually the issue comes from the fact we are mixing the notion of ground truth and raw data but they are not the same. "raw data" is what is being observed in the physical word, and ground truth is the hardware processing that comes with it, whether it is perfect on not. On top of that we have imperfection that happens at physics-level, and some others that happens at the hardware-level. Both should be dealt with separately, in different part of the pipeline, eventually by different mixin.

Basically, when you said: "if we are always updating gt and measured cache, why not combine into one _update_shared_cache() (and remove _update_shared_ground_truth_cache())? for noise that affects intermediate calculations (as opposed to imperfections that postprocess the final output value, e.g. a noise on raycast direction) this would allow for one kernel call to handle both the noised and unnoised outputs, without having to store/copy values in shared_metadata"

I think it is possible in most cases and we should make the API flexible enough to allow users to do this, but it is not possible in general, where you have:
```
raw data ---> transform (eg IMU frame rotation) ---> casting (incl. clamping, casting to bool) ---> ground truth
          |
          |-> physics imperfections -> transform -> hardware imperfections (incl. delay, noise...etc) --> casting -> measured signal
```

duburcqa:
Updated plan v16: [pr2770-sensor-update-restructure.md](https://github.com/user-attachments/files/27772006/pr2770-sensor-update-restructure.md)

duburcqa (dismissed):

PR #2785, opened 2026-05-15
[FEATURE] Restructure sensor base hierarchy + eager _post_process + SensorDataSpec API.

duburcqa:
## Summary

Builds on @Milotrince's #2770 (consolidate `_update_shared_cache`) to introduce the v17 sensor base hierarchy. This PR is **stacked on #2770** and should land after it.

Key changes:

- **`Sensor` / `SimpleSensor` split.** `Sensor` is the minimal customization contract (4 entry-points: `_update_shared_cache`, `_post_process`, `return_spec`/`intermediate_spec`, `uses_measured_pipeline`). `SimpleSensor` is the standard pipeline that most Genesis sensors derive from, with four hooks (`_update_raw_data`, `_update_current_timestep_data`, `_apply_transform`, `_apply_hardware_imperfections`).

- **No `Imperfect*` classes.** Options-side: time-related (`delay`, `jitter`, `interpolate`, `history_length`) on `SensorOptions` base; imperfection parameters (`noise`, `bias`, `random_walk`, `resolution`) on `SimpleSensorOptions(SensorOptions)`. Metadata-side: time-related on `SharedSensorMetadata`; imperfection-parameter state on `SimpleSensorMetadata(SharedSensorMetadata)`. Behavior: `ImperfectSensorMixin` folded into `SimpleSensor`.

- **`SensorDataSpec` API.** Self-contained `return_spec` and `intermediate_spec` cached_properties (public, no underscore prefix), each a `SensorDataSpec(shape, dtype)` NamedTuple. Replaces the old `_get_return_format` / `_get_cache_dtype` method pair. Overriding `_post_process` REQUIRES overriding `intermediate_spec` — `Sensor.__init_subclass__` enforces this at class-definition time.

- **Snapshot-time imperfections.** Hardware imperfections (noise, bias, random_walk, resolution) apply at the embedded-sampler snapshot (PRE-delay), inside the timeline ring's slot 0 — not at read time. This matches the abstraction: `read()` is a query into shared memory, idempotent within a sampling period; delayed reads return stale snapshots with their original imperfection state. v16's "fresh noise per `read()`" framing implicitly modeled an analog amplifier on the wire — wrong abstraction once you cross into the digital embedded layer.

- **Eager `_post_process`.** Applied once per step by the orchestrator into a per-class return cache; `read()` is then a pure memory-lookup view. Per-class `_return_cache[sensor_cls]` alias-views the per-dtype `_intermediate_cache[dtype]` when `_post_process` is identity, separate buffer when overridden. `read_sensors(copy=False)` is honest for every sensor class.

- **Per-sensor migrations.** All Genesis sensors except Camera derive from `SimpleSensor` and inherit the standard pipeline. IMU's body-frame rotation moves to `_apply_transform`. Temperature's RC filter moves to `_apply_transform` with `if timeline is not None` gating. ContactSensor refactored to float intermediate + bool return; ContactForceSensor deadband + saturation moved to `_post_process`. Camera derives from `Sensor` directly and inherits `uses_measured_pipeline = False`.

- **Single measured ring.** v17 merges the v16 `_measured_timeline_ring` and `_measured_history_ring` (imperfections were AFTER delay sampling, requiring a separate post-noise ring for history). Now the timeline ring stores post-noise values directly; `read(history_length=N)` reads from the per-class linearized history snapshots of the return cache.

## User-observable behavior changes

- `Contact` options gains a new `threshold` field (default 0.0 preserves current behavior).
- `ContactSensor`'s internal intermediate cache is now `float` (kernel writes a contact count); read paths still return `bool` via `_post_process`.
- `ContactForceSensor`'s deadband + saturation runs at read time (eagerly into the return cache) instead of at write time. No user-facing read behavior change; internal cache may transiently contain values outside the deadband.
- Temperature's hardware imperfections move from inside the seed/filter step to snapshot time (PRE-delay) — delayed reads correctly return stale snapshots with their original noise state, and `read(history_length=N)` is consistent with prior `read()` values.
- `SensorManager.read_sensors(copy=False)` (the default) returns honest views for every sensor class, including `ContactSensor` / `ContactForceSensor` which now have real per-class return-cache storage. No three-valued `copy=None` semantics.
- `return_spec` / `intermediate_spec` are public properties — users may inspect the spec of any sensor instance.

## Test plan
- [ ] `pytest tests/test_sensors.py` — 34/34 passing (33 pre-existing + 1 new test for the `intermediate_spec` strict-override rule).
- [ ] `pytest tests/test_viewer.py` — 6/6 passing.
- [ ] `pytest tests/test_rigid_physics.py` — no sensor-specific tests; verified no spillover.

### DIFF
--- a/genesis/engine/sensors/base_sensor.py
+++ b/genesis/engine/sensors/base_sensor.py
@@ -1,6 +1,6 @@
 from dataclasses import dataclass, field
 from functools import partial
-from typing import TYPE_CHECKING, ClassVar, Generic, NamedTuple, TypeVar, get_args, get_origin
+from typing import TYPE_CHECKING, ClassVar, Generic, NamedTuple, Sequence, TypeVar, get_args, get_origin
 
 import numpy as np
 import torch
@@ -50,8 +50,8 @@ class SharedSensorMetadata:
     history_lengths: list[int] = field(default_factory=list)
     # Per-sensor: interpolate delayed reads (aligned with cache_sizes). Appended in Sensor.build.
     interpolate: list[bool] = field(default_factory=list)
-    # True iff at least one sensor in the class has a nonzero read delay. Precomputed at build (and latched True
-    # by `set_delay`) so the per-step fast path in SensorManager._apply_delay_to_shared_cache can avoid a GPU-syncing reduction.
+    # True iff at least one sensor in the class has a nonzero read delay. Precomputed at build (and latched True by
+    # `set_delay`) so the per-step fast path in `_apply_delay_to_shared_cache` can avoid a GPU-syncing reduction.
     has_any_delay: bool = False
     # True iff at least one sensor in the class has a nonzero jitter option. Stays False for non-noisy sensor classes
     # (Contact, Raycaster) since they never sample jitter. Same precompute-and-latch contract as `has_any_delay`.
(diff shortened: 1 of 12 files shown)


duburcqa:
Reopening with base set to Trinity's PR branch (#2770) for clean stacking diff.

PR #2786, opened 2026-05-15
[FEATURE] Refactor sensor pipeline (cont'd).

duburcqa:
## Summary

Follow-up to #2770 (Trinity's `_update_shared_cache` consolidation). Splits the sensor base hierarchy and reshapes the per-step pipeline so that:

- `Sensor` carries only the time-related plumbing (`delay`, `jitter`, `interpolate`, `history_length`); `SimpleSensor(Sensor)` adds the standard pipeline and the imperfection parameters (`noise`, `bias`, `random_walk`, `resolution`). `SimpleSensorOptions` / `SimpleSensorMetadata` mirror the split on the options and metadata layers. Camera derives from `Sensor` directly and sees only the time-related fields. No `Imperfect*` classes remain.
- Sensor-class spec is declared via four accessors: `_get_return_format(self)` and `_get_intermediate_format(self)` for shape (instance methods, since sensor options may affect shape — Raycaster pattern, Camera resolution, Proximity probes, etc.); `_get_cache_dtype(cls)` and `_get_intermediate_dtype(cls)` for dtype (classmethods, since dtype is class-uniform — the per-class slice into the per-dtype intermediate buffer must be contiguous). The intermediate methods default to the return ones. `__init_subclass__` enforces that overriding `_post_process` also requires overriding `_get_intermediate_format` and/or `_get_intermediate_dtype` (no silent guessing about the intermediate buffer space; a no-op override is acceptable when shape and dtype coincide with return).
- Hardware imperfections (`_apply_hardware_imperfections`) are applied **PRE-delay** at snapshot time inside the SimpleSensor orchestrator. This matches the embedded-sampler abstraction (`read()` is a memory lookup against a snapshot the firmware wrote earlier) and lets a single measured-timeline ring serve both delay sampling and `read(history_length=N)`.
- `_post_process` becomes **eager** (applied once per step into a per-class `_return_cache` in the return space). `read()` returns a pure view into the return cache for non-history reads. Stateful `_post_process` is well-defined under eager placement, opening the door to in-sensor software signal processing (complementary / Mahony / Kalman filters, fused-orientation estimates). For sensors with overridden `_post_process` AND `history_length > 0`, the manager allocates a per-class return-space ring so history reads pull post-processed snapshots directly (correct for stateful projections).
- `_apply_transform`'s signature gains a `timeline` keyword so coordinate transforms and stateful filters live in the same override (transform on both branches; filter portion gated by `if timeline is not None:`). Filter-then-transform is not expressible by design — GT has no hardware response.
- `uses_measured_pipeline: ClassVar[bool] = True` is the class-level capability flag declaring whether the class reads or writes the measured-timeline ring inside `_update_shared_cache`. Default `True` (every `SimpleSensor`-derived class needs the ring); `BaseCameraSensor` explicitly sets it to `False` and raises in `__init__` if any of `delay > 0`, `jitter > 0`, `history_length > 0` is requested (those features depend on the ring it bypasses). Ring allocation is therefore a binary class-level decision, decoupled from runtime imperfection state — setters like `set_noise` / `set_bias` / `set_random_walk` / `set_jitter` always work because the ring is already there. Per-step gating of actual imperfection work is done by the `has_any_*` flags inside `_apply_hardware_imperfections`.
- `read_sensors(envs_idx=..., is_ground_truth=...)` always returns a fresh tensor per class (env axis is sanitized to a 1D tensor, forcing fancy indexing). The `copy` parameter is removed at all three levels (manager / scene / entity); the mutation-safe contract is uniform.
- Per-sensor migrations: ContactSensor moves the bool projection from kernel to `_post_process` (float intermediate, bool return, per-instance `threshold` option); ContactForceSensor's existing stateless deadband + saturation moves from the write path into `_post_process`; IMU gains an `_apply_transform` override (body-frame rotation); TemperatureGrid gains an `_apply_transform` override (RC filter on the measured branch). Other sensors only need `_update_raw_data`.

A complete walkthrough of the runtime pipeline and a custom-sensors how-to are in the user guide under Advanced Topics: [`sensor_pipeline.md`](https://genesis-world.readthedocs.io/en/latest/user_guide/advanced_topics/sensor_pipeline.html) and [`custom_sensors.md`](https://genesis-world.readthedocs.io/en/latest/user_guide/advanced_topics/custom_sensors.html) (merged via Genesis-Embodied-AI/genesis-doc#110).

### DIFF
--- a/genesis/engine/entities/base_entity.py
+++ b/genesis/engine/entities/base_entity.py
@@ -111,17 +111,17 @@ class Entity(RBC):
         return self._sim._sensor_manager.get_sensors_by_entity(self._idx)
 
     @gs.assert_built
-    def read_sensors(self, envs_idx=None, copy: bool = False) -> "dict[type[Sensor], torch.Tensor]":
+    def read_sensors(self, envs_idx=None) -> "dict[type[Sensor], torch.Tensor]":
         """
         Read every sensor attached to this entity as a tensor per sensor class.
+
+        Always returns a fresh tensor independent of the internal sensor storage; the caller is free to mutate the
+        result.
 
         Parameters
         ----------
         envs_idx : array-like | int | slice | None
-            Environment selection. Integer/slice indexing returns a view along the batch axis; list/tensor
-            (fancy) indexing returns a copy.
-        copy : bool
-            When True, returned tensors are cloned. When False (default), they are views into the shared cache.
+            Environment selection. Defaults to all environments.
 
         Returns
         -------
@@ -129,7 +129,7 @@ class Entity(RBC):
             For each sensor class with at least one sensor on this entity, a tensor of shape
             (B, [history,] entity_cache_size_for_class).
         """
-        return self._sim._sensor_manager.read_sensors(entity_idx=self._idx, envs_idx=envs_idx, copy=copy)
+        return self._sim._sensor_manager.read_sensors(entity_idx=self._idx, envs_idx=envs_idx)
 
     # ------------------------------------------------------------------------------------
     # --------------------------------- naming methods -----------------------------------
--- a/genesis/engine/scene.py
+++ b/genesis/engine/scene.py
@@ -636,24 +636,24 @@ class Scene(RBC):
         return self._sim._sensor_manager.create_sensor(sensor_options)
 
     @gs.assert_built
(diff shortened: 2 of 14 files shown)


duburcqa:
@claude review

PR #2790, opened 2026-05-16
[MISC] Reduce indentation in narrowphase.py and save 89 lines

hughperkins:
For every `if cond: <large body>` that wraps the tail of a `for` loop and where the body contains further nested control flow, invert the condition and use `continue` to skip ahead. This collapses one level of nesting on the body and removes no other code.

17 sites flattened, picked from a survey of all `if` statements whose body covers >=20 lines and contains at least one nested `if`/`for`/`while`. Sites where the body was a flat sequence (no nested control flow) were left alone -- inverting+continuing there only renames the guard without reducing depth.

Functions whose max AST nesting depth dropped:
  func_narrow_phase_nonconvex_vs_nonterrain   12 -> 9
  func_contact_mpr_terrain                    10 -> 8
  func_contact_convex_convex_sdf               7 -> 5
  _func_multicontact_mpr                       6 -> 4
  func_contact_edge_sdf                        6 -> 4

Lines at >=8 levels of nesting drop ~38% (381 -> 237); lines at >=11 levels drop ~61% (98 -> 38). File line count: 2491 -> 2482.

`func_convex_convex_contact`'s max depth (14) is unchanged because its outermost wrap `if not (PLANE and BOX): <whole body>` is a function-level guard, not a loop tail -- flattening it requires return-inside-if support in qd.func, which is a separate work item.

Two multi-line conditions (sites at lines 442 and 2045) were negated by hand via De Morgan, preserving their multi-line layout. All other sites are single-line condition flips. File was syntax-checked after every mutation; semantics are preserved by construction for the `continue` substitution and by inspection for the negations.

## Related Issue

## Checklist:


duburcqa:
Going to close this one. Dug into it and the problem is most of the line savings come from the early-`continue` flattening, and Quadrants autodiff flat-out rejects `continue` (and `break`/`while`) — `IdentifyIndependentBlocks` throws `"ContinueStmt is not supported in AutoDiff."`. Since we're pushing the rigid pipeline (incl. the monolithic narrowphase convex path) toward full differentiability — see #2842 — dropping `continue` into those kernels goes the wrong way and we'd just have to rip it back out.

The parts that *are* autodiff-safe (merging nested `if`s, `else: if` -> `elif`) net ~0 lines on their own, so not really worth it. Happy to take just the `if`-merges as a small follow-up if you want, but no worries either way.

hughperkins:
> Quadrants autodiff flat-out rejects continue (and break/while) — IdentifyIndependentBlocks throws "ContinueStmt is not supported in AutoDiff."

Interesting.

I'm kind of a fan of reducing indentation levels by using things like continue, break, return, so might take a look at some point. Not in the foreseeable future though.

duburcqa:
> I'm kind of a fan of reducing indentation levels by using things like continue, break, return, so might take a look at some point. Not in the foreseeable future though.

Me too... That code is horrible to read because of this.

PR #2804, opened 2026-05-18
[MISC] Fail loudly when failing to upload benchmark stats on main branch.

hughperkins:


.github/workflows/production.yml
```diff
@@ -183,7 +183,7 @@ jobs:
                    --print -m "benchmarks" ./tests
 
             cat speed_test*.txt > "/mnt/data/artifacts/speed_test_${SLURM_JOB_NAME}.txt"
-            if [ -n "${WANDB_API_KEY}" ] ; then (
+            if [ "${{ github.ref }}" = "refs/heads/main" ] ; then (
```
duburcqa: It would fail of the some external contributor opens a PR using its main branch.

PR #2827, opened 2026-05-23
[MISC] Speed up rigid constraint solver by adjusting cholesky tile size.

hughperkins:
## Description

Interestingly, even though this only targets dex_hand with 32x32 (dofs 64 or so), this PR also modifies shared memory usage:

  For g1_fall (n_dofs=35):
  • Upstream rounds up to multiple of 32: ceil(35/32)*32 = 64
  • This branch rounds up to multiple of cholesky_tile_size=16: ceil(35/16)*16 = 48

This reduction in shared memory usage then plausibly lead to increasd occupancy.

In any case, we see +7% FPS on g1_fall, even though g1_fall continues to use 16x16 tiles.


duburcqa (approved):

hughperkins:
Thanks! 🙌

PR #2843, opened 2026-05-27
[MISC] Add unit tests for TensorRingBuffer.

yeezhouyi:
## Summary
- Add 20 unit tests for `TensorRingBuffer` in `genesis/utils/ring_buffer.py`, covering initialization, data flow, rotation, relative indexing, copy semantics, slicing, cloning, wraparound, and multidimensional access.
- All tests are marked `@pytest.mark.required`.

## Test plan
- [x] `pytest tests/test_utils.py -k ring_buffer -v` passes.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -15,6 +15,7 @@ from genesis.utils.misc import tensor_to_array
 from genesis.utils import warnings as warnings_mod
 from genesis.utils.warnings import warn_once
 from genesis.utils.urdf import compose_inertial_properties
+from genesis.utils.ring_buffer import TensorRingBuffer
 
 from .utils import assert_allclose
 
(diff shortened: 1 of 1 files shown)


duburcqa:
We do not have unit test in Genesis, only integration test. Why do you think it is relevant to add unit test for the TensorRingBuffer specifically?

yeezhouyi:
Thanks for the feedback! I understand that Genesis uses integration tests, and I respect that convention. The reason I added dedicated tests for TensorRingBuffer is that it's a core data structure shared across the entire sensor pipeline, with subtle semantics around memory sharing, view vs clone behavior, and index wraparound that integration tests can't easily catch. For example, the __getitem__ slice behavior can silently produce wrong views for multi-dimensional tensors — a kind of bug that wouldn't surface as a wrong sensor reading but could cause hard-to-debug memory corruption in long-running simulations. That said, I'm happy to adapt the tests to match the project style if you prefer. Let me know what you think!

duburcqa:
> a kind of bug that wouldn't surface as a wrong sensor reading but could cause hard-to-debug memory corruption in long-running simulations.

I don't think a bug that wouldn't surface as a wrong sensor reading is worth fixing. Since it is basically our only requirement today for this class.

PR #2853, opened 2026-05-27
[MISC] Switch dedupe contact sort to use Quadrants bitonic sort

hughperkins:
Replaces the inlined 15-stage bitonic compare-exchange schedule in ``func_clamp_prune_and_sort_contacts_coop`` (phase 1a) with a one-line call to the new Quadrants subgroup primitive:

    my_key, my_idx = qd.simt.subgroup.bitonic_sort_kv_tiled(my_key, my_idx, 5)

The primitive (added in quadrants hp/bitonic-sort-kv) is a @qd.func that inlines at compile time and unrolls the same 15 compare-exchange stages this code used to write inline, so the generated kernel IR is bit-identical to today on CUDA. Net change: ~30 lines of hand-rolled bitonic code removed, the sentinel load + write-back wrapper is unchanged, and the rest of the kernel (clamp + key init + bucket walk + phase 2 + phase 3) is untouched.

``log2_size = 5`` pins the sort to a 32-lane tile, matching the kernel's hard-coded ``block_dim = _K = 32``. Using the tiled form rather than the bare ``bitonic_sort_kv(...)`` wrapper keeps the sort width fixed at 32 even on AMDGPU wave64, where the bare wrapper would otherwise sort across all 64 lanes and mix in garbage from the inactive upper half.

Requires the matching quadrants change to be installed (the public symbol ``qd.simt.subgroup.bitonic_sort_kv_tiled`` is added by that PR).

## Related Issue

## Checklist:


genesis/engine/solvers/rigid/collider/contact.py
```diff
@@ -991,26 +991,7 @@ def func_clamp_prune_and_sort_contacts_coop(
...
-                        their_lt_mine = (their_key < my_key) or (their_key == my_key and their_idx < my_idx)
-                        if take_min:
-                            if their_lt_mine:
-                                my_key = their_key
-                                my_idx = their_idx
-                        else:
-                            if not their_lt_mine and (their_key != my_key or their_idx != my_idx):
-                                my_key = their_key
-                                my_idx = their_idx
+                my_key, my_idx = qd.simt.subgroup.bitonic_sort_kv_tiled(my_key, my_idx, _LOG2_K)
```
duburcqa, line 1004: Could we have an API that does not require specifying `_LOG2_K`? Like this could be the default value, because I guess `block_dim` can be queried from context no? if not, then we should probably expose it. It should not be that hard, since it is already possible to set it from python.

duburcqa (dismissed):

genesis/engine/solvers/rigid/collider/contact.py
hughperkins: again, this is not controllable frm genesis side. We already reviewed and merged the quadrants-side code.

hughperkins: I doubt we can query block_dim from context.

hughperkins: Oh, hmm, expose it 🤔

hughperkins: Note that we already have an untiled versoin of bitonic sort, which uses the subgroup size by default (maybe hti is what you meant by the 'block size'? In any case, this is already they default, for the untiled version.

Here we are using the tiled version, because the subgroup size might not be 32.

hughperkins:
🙌

PR #2860, opened 2026-05-28
[MISC] Move register-tile Cholesky into quadrants.

hughperkins:
Delete `genesis/utils/_tile16.py` and `_tile32.py`, whose contents (the inlined column-setter cascades + qd.static-unrolled cholesky_ body) now live in `quadrants/lang/simt/_tile16.py` and `_tile32.py` upstream. The constraint solver switches from `Tile16x16Cholesky` / `Tile32x32Cholesky` to `qd.simt.Tile16x16` / `qd.simt.Tile32x32`, and the blocked-Cholesky bodies go back to the readable slice syntax (`tile[:] = arr[r:r2, c:c2]`, `tile -= qd.outer(v, v)`) instead of the explicit `_load3d` / `_store3d` / `_resolve_vec3d` / `_ger_sub` private calls that were required while two distinct tile classes co-existed in one kernel.

Pairs with quadrants PR hp/import-tiles-from-genesis.

## Related Issue

## Checklist:


genesis/engine/solvers/rigid/constraint/solver.py
```diff
@@ -2130,11 +2125,11 @@ def func_cholesky_and_solve_fused_tiled(
...
     if qd.static(static_rigid_sim_config.cholesky_tile_size == 32):
         _cholesky_and_solve_fused_tiled_impl(
-            constraint_state, rigid_global_info, static_rigid_sim_config, Tile32x32Cholesky, write_L_to_nt_H
+            constraint_state, rigid_global_info, static_rigid_sim_config, qd.simt.Tile32x32, write_L_to_nt_H
         )
     else:
         _cholesky_and_solve_fused_tiled_impl(
-            constraint_state, rigid_global_info, static_rigid_sim_config, Tile16x16Cholesky, write_L_to_nt_H
+            constraint_state, rigid_global_info, static_rigid_sim_config, qd.simt.Tile16x16, write_L_to_nt_H
         )
```
duburcqa, line 2292: Would it be possible to have `qd.simt.Tile` that accept static "size" argument, so that ce can pass size as input argument instead of entire class? This would be more idiomatic.

duburcqa: And even more natural and cleaner, because you would just have to pass "static_rigid_sim_config.cholesky_tile_size" directly inside `_cholesky_and_solve_fused_tiled_impl` instead of adding one more indirection level as it is currently the case.

duburcqa (dismissed):

genesis/engine/solvers/rigid/constraint/solver.py
hughperkins: would be better to comment these on the quadrants-side PR. I've already merged that though.

hughperkins: (No way to modify such parameters from genesis side)

hughperkins:
🙌

PR #2868, opened 2026-05-30
[MISC] Speed up heterogeneous-env raycasts with a ray-coherent thread mapping

Kashu7100:
## Description

Follow-up to #2867 — **stacked on it**. The first two commits here are #2867 (skip static BVH rebuild; share one BVH across envs when geometry is env-identical); this PR adds the third commit. Please review/merge after #2867 lands — the diff will narrow to just the new commit once #2867 is merged into `main`.

**New change: route heterogeneous-env raycasts to a ray-coherent thread mapping.** #2867's shared-BVH optimization maps consecutive GPU threads to *one ray across consecutive envs*. That is optimal when all envs share a single BVH copy (the warp broadcasts a single batch-0 node), but pathological when envs have *distinct* geometry (domain-randomized terrain, per-env obstacles): a warp then straddles `n_env` different trees, maximizing execution divergence and scattering node loads.

This commit selects the thread→`(ray, env)` decomposition at compile time from the **existing** `shared_bvh` flag, in a single flat parallel loop:
- **shared (homogeneous)** → env is the fastest-varying index → coalesced batch-0 broadcast (unchanged).
- **not shared (heterogeneous)** → ray is the fastest-varying index → each warp stays within one env's tree and rides ray coherence instead of diverging across `n_env` trees.

Routing is **automatic and requires no API change**: `shared_bvh` is already set per BVH entry from whether the solver's link poses match across all envs, so any scene with per-env-different geometry takes the ray-coherent path on its own.

## Related Issue

No tracking issue filed yet — same as #2867.

## Motivation and Context

After #2867, the homogeneous (shared-terrain) case reaches parity with an NVIDIA Warp `mesh_query_ray` baseline, but the heterogeneous case (per-env geometry) still pays the full per-env cost. The penalty turned out to be almost entirely the thread mapping, not the BVH itself.

Controlled measurement (RTX 3080, 2048 envs × 64×36 rays, identical scene, only the mapping varied):

| Path | Mapping | ms/cast |
|---|---|--:|
| Heterogeneous | env-innermost (#2867) | 27.1 |
| Heterogeneous | ray-innermost (this PR) | **12.1** (2.2×) |
| Homogeneous | env-innermost broadcast | 4.76 (unchanged) |

On *truly* divergent trees the env-innermost baseline only degrades further, so 2.2× is a lower bound.

The change is behavior-preserving: the mapping is a pure permutation of which thread handles which `(ray, env)` pair, and each pair writes its own output slot with no cross-thread interaction — results are identical.

Environment: single RTX 3080, CUDA backend.

- `pytest tests/test_sensors.py -k "raycaster or lidar"` — all 8 pass. In particular `test_lidar_bvh_parallel_env` builds a scene with **genuinely per-env-different** obstacle positions and asserts the *exact* expected per-env distances, exercising the new ray-innermost (`shared_bvh=False`) path end-to-end.
- Perf: the controlled 27.1 → 12.1 ms comparison above (same geometry, only the loop mapping swapped), with the homogeneous path re-confirmed at 4.76 ms (no regression).

## Checklist:

### DIFF
--- a/genesis/engine/sensors/raycaster.py
+++ b/genesis/engine/sensors/raycaster.py
@@ -30,7 +30,8 @@ if TYPE_CHECKING:
     from .sensor_manager import SensorManager
 
 
-class _SolverBVH(NamedTuple):
+@dataclass
+class _SolverBVH:
     """
     One BVH built against a solver's mesh.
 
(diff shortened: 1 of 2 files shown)


Kashu7100:
Aggregated into #2867 — the ray-coherent heterogeneous mapping is now the third commit there. Closing this stacked PR to keep all three raycaster perf changes in one place.

PR #2869, opened 2026-05-31
[MISC] Speedup rigid solver for large dofs count (> ~128).

Kashu7100:
## Problem

The GPU rigid constraint solver is dramatically slow for environments with many DOFs (multi-humanoid, dexterous hands, SMPL-X). On an RTX 3090, two SMPL-X humanoids (318 DOFs) run at ~2.1 s per `scene.step()`, and step time is roughly **flat in `n_envs`** (occupancy-bound) and **independent of solver `iterations`** — i.e. the cost is once-per-step dense factorizations, not the iterative solve.

Two factorizations are gated by the GPU shared-memory size and silently fall back to **serial one-GPU-thread-per-env** dense factorization (`O(n_dofs³)`) above the cap:

1. **Constraint Hessian** `H = M + JᵀDJ` — gated by `enable_tiled_cholesky_hessian = 16 <= n_dofs <= max_n_threads` (`max_n_threads` ~96–128, bounded by holding the full `L` in shared memory).
2. **Mass matrix** `M` (per entity) — gated by `enable_tiled_cholesky_mass_matrix = max_n_dofs_per_entity <= max_n_threads`; factored every substep for the unconstrained acceleration.

A single SMPL-X humanoid is 159 DOFs/entity, so **both** hit the serial path. A Franka micro-benchmark (stack *k* Frankas) shows the Hessian cliff exactly at the cap: 126 DOFs → 77 ms, 135 DOFs → 252 ms.

## Fixes

**1. Hessian (commit 1) — re-route, no new kernel.** The standalone `func_cholesky_factor_direct_tiled` is a left-looking blocked Cholesky that streams `T×T` tiles through registers (*"no shared memory… no inherent DOF limit"*) — only the *fused* factor+solve and the *tiled triangular solve* need the full `L` resident. Add `enable_tiled_cholesky_hessian_large` (`n_dofs > max_n_threads`) and route the above-cap factor to the existing register-streaming kernel; the triangular solve stays on the scalar `O(n_dofs²)` batch path and per-iteration incremental rank-1 updates remain scalar.

**2. Mass matrix (commit 2) — new cooperative LDLᵀ.** The mass matrix has no register-streaming factor, and its tiled path stages the whole entity submatrix in shared memory. Add `enable_tiled_cholesky_mass_matrix_large` and a cooperative LDLᵀ that factors the entity submatrix **in-place in global memory** over a block of threads. To be both uncapped and race-free, each elimination step snapshots the pivot row into a small shared **vector** (`O(n_dofs)`, not `O(n_dofs²)`) before updating the trailing submatrix, so the parallel per-row updates only ever *read* the pivot row — no read/write aliasing on the eliminated row regardless of scheduling. Otherwise identical to the scalar branch.

Both keep the existing shared-memory fast paths for `n_dofs ≤ cap` (no low-DOF regression) and leave CPU / `sparse_solve` / `n_dofs < 16` on their current paths.

## Results (RTX 3090, Newton, iterations=10/20, num_envs=256)

Hessian fix — Franka micro-benchmark (small entities, so only the Hessian crosses the cap):

| env DOFs | before | after | speedup |
|--:|--:|--:|--:|
| 126 (below cap, control) | 36.6 ms | 37.1 ms | 1.0× |
| 135 | 118.6 ms | 31.8 ms | **3.7×** |
| 180 | 273.4 ms | 51.4 ms | **5.3×** |
| 270 | 731.2 ms | 114.8 ms | **6.4×** |

Both fixes — SMPL-X humanoids (159 DOFs/entity, so the mass matrix is also above the cap):

| scene | before | after | speedup |
|--|--:|--:|--:|
| 1 humanoid (159 DOFs) | 1162 ms | **96 ms** | **12.1×** |
| 2 humanoids (318 DOFs) | 2136 ms | **208 ms** | **10.3×** |

(The mass-matrix fix unlocks the bulk of the humanoid gain: with only the Hessian fix the dual-humanoid was ~1115 ms; the per-entity mass factor was the dominant remaining serial cost.)

## Correctness

Both factorizations match their scalar counterparts to floating-point precision and produce stable dynamics:
- Hessian: Franka link-position checksum after 160 substeps 35613.3506 (scalar) vs 35613.3532 (tiled).
- Mass: SMPL-X checksums 72459.9002 vs 72459.8918 (1 humanoid), 161006.689 vs 161006.681 (2 humanoids), all finite.
- Below-cap and small-entity controls (Franka) are unchanged.

## Notes / follow-ups

- The Hessian *triangular solve* and *fused* kernel remain shared-memory-capped; above the cap the solve uses the scalar `O(n_dofs²)` batch path. A tile-streaming triangular solve would uncap it fully (the `O(n_dofs³)` factor was the dominant cost, so this is minor).
- Validated by applying the change to an installed build and running the benchmarks above; I have not run the full upstream CI/test suite locally. Happy to add unit tests (factor a known SPD/mass system above the cap, compare to the scalar factor) and split into two PRs if preferred.

### DIFF
--- a/genesis/engine/solvers/rigid/abd/forward_dynamics.py
+++ b/genesis/engine/solvers/rigid/abd/forward_dynamics.py
@@ -492,7 +492,87 @@ def func_factor_mass(
         n_entities = entities_info.n_links.shape[0]
         _B = dofs_state.ctrl_mode.shape[1]
 
-        if qd.static(
+        if qd.static(static_rigid_sim_config.enable_tiled_cholesky_mass_matrix_large):
+            # Uncapped cooperative per-entity LDL^T (max_n_dofs_per_entity > max_n_threads). Factors the entity
+            # mass submatrix in-place in global memory (mass_mat_L), cooperatively over a block of BLOCK_DIM
+            # threads. Each elimination step snapshots the pivot row into a small shared vector (O(n_dofs), not
+            # O(n_dofs^2)) BEFORE updating the trailing submatrix, so the parallel row updates only ever READ
+            # the pivot row (from shared) -- race-free regardless of thread scheduling. Numerically identical to
+            # the scalar branch below; only the parallelization (and lack of an O(n_dofs^2) shared array) differs.
+            BLOCK_DIM = qd.static(32)
+            MAX_DOFS_PER_ENTITY = qd.static(static_rigid_sim_config.tiled_n_dofs_per_entity_large)
+
+            qd.loop_config(name="factor_mass", block_dim=BLOCK_DIM)
+            for i in range(n_entities * _B * BLOCK_DIM):
+                tid = i % BLOCK_DIM
+                i_e = (i // BLOCK_DIM) % n_entities
+                i_b = i // (BLOCK_DIM * n_entities)
+                if i_b >= _B:
+                    continue
+
+                if rigid_global_info.mass_mat_mask[i_e, i_b]:
+                    entity_dof_start = entities_info.dof_start[i_e]
+                    entity_dof_end = entities_info.dof_end[i_e]
+                    n_dofs = entities_info.n_dofs[i_e]
+
+                    pivot_row = qd.simt.block.SharedArray((MAX_DOFS_PER_ENTITY,), gs.qd_float)
+
+                    # Copy the lower triangle of M into mass_mat_L (+ implicit damping on the diagonal), cooperatively.
+                    i_d_ = tid
+                    while i_d_ < n_dofs:
+                        i_d = entity_dof_start + i_d_
+                        for j_d in range(entity_dof_start, i_d + 1):
+                            rigid_global_info.mass_mat_L[i_d, j_d, i_b] = rigid_global_info.mass_mat[i_d, j_d, i_b]
+                        if qd.static(implicit_damping):
+                            I_d = [i_d, i_b] if qd.static(static_rigid_sim_config.batch_dofs_info) else i_d
+                            rigid_global_info.mass_mat_L[i_d, i_d, i_b] = (
+                                rigid_global_info.mass_mat_L[i_d, i_d, i_b]
+                                + dofs_info.damping[I_d] * rigid_global_info.substep_dt[None]
+                            )
+                            if qd.static(static_rigid_sim_config.integrator == gs.integrator.implicitfast):
+                                if dofs_state.ctrl_mode[i_d, i_b] <= gs.CTRL_MODE.VELOCITY:
+                                    rigid_global_info.mass_mat_L[i_d, i_d, i_b] = (
+                                        rigid_global_info.mass_mat_L[i_d, i_d, i_b]
+                                        - dofs_info.act_bias[I_d][2] * rigid_global_info.substep_dt[None]
+                                    )
+                        i_d_ = i_d_ + BLOCK_DIM
+                    qd.simt.block.sync()
+
+                    # In-place LDL^T, eliminating dofs from last to first (matches the scalar branch).
+                    for j in range(n_dofs):
+                        i_d_ = n_dofs - j - 1
+                        i_d = entity_dof_end - j - 1
+                        D_inv = 1.0 / rigid_global_info.mass_mat_L[i_d, i_d, i_b]
+                        if tid == 0:
+                            rigid_global_info.mass_mat_D_inv[i_d, i_b] = D_inv
+
+                        # Phase A: snapshot the (Schur-updated) pivot-row entries below the diagonal into shared.
+                        j_d_ = tid
+                        while j_d_ < i_d_:
+                            pivot_row[j_d_] = rigid_global_info.mass_mat_L[i_d, entity_dof_start + j_d_, i_b]
+                            j_d_ = j_d_ + BLOCK_DIM
+                        qd.simt.block.sync()
+
+                        # Phase B: each lane eliminates one column j_d, updating its own row j_d of the trailing
+                        # submatrix from the read-only snapshot. Distinct rows per lane => no write conflicts, and
+                        # the pivot row is only read (from shared) => no read/write race on row i_d.
+                        j_d_ = tid
+                        while j_d_ < i_d_:
+                            a = pivot_row[j_d_] * D_inv
+                            j_d = entity_dof_start + j_d_
+                            for k_d_ in range(j_d_ + 1):
+                                rigid_global_info.mass_mat_L[j_d, entity_dof_start + k_d_, i_b] = (
+                                    rigid_global_info.mass_mat_L[j_d, entity_dof_start + k_d_, i_b]
+                                    - a * pivot_row[k_d_]
+                                )
+                            rigid_global_info.mass_mat_L[i_d, j_d, i_b] = a
+                            j_d_ = j_d_ + BLOCK_DIM
+                        qd.simt.block.sync()
+
+                        # Diagonal coeffs of L are ignored downstream (see scalar branch) but set to 1.0 to match.
+                        if tid == 0:
+                            rigid_global_info.mass_mat_L[i_d, i_d, i_b] = 1.0
+        elif qd.static(
             not static_rigid_sim_config.enable_tiled_cholesky_mass_matrix or static_rigid_sim_config.backend == gs.cpu
         ):
             qd.loop_config(name="factor_mass", serialize=static_rigid_sim_config.para_level < gs.PARA_LEVEL.PARTIAL)
--- a/genesis/engine/solvers/rigid/abd/forward_dynamics.py
+++ b/genesis/engine/solvers/rigid/abd/forward_dynamics.py
@@ -492,14 +492,17 @@ def func_factor_mass(
         n_entities = entities_info.n_links.shape[0]
         _B = dofs_state.ctrl_mode.shape[1]
 
-        if qd.static(static_rigid_sim_config.enable_tiled_cholesky_mass_matrix_large):
-            # Uncapped cooperative per-entity LDL^T (max_n_dofs_per_entity > max_n_threads): factors the entity mass
-            # submatrix in-place in global memory (mass_mat_L) over a block of BLOCK_DIM threads. Each elimination step
-            # snapshots the pivot row into a small shared vector (O(n_dofs), not O(n_dofs^2)) before updating the
+        if qd.static(
+            static_rigid_sim_config.enable_tiled_cholesky_mass_matrix
+            and not static_rigid_sim_config.mass_matrix_fits_shared
+        ):
+            # Uncapped cooperative per-entity LDL^T (entity submatrix does not fit shared memory): factors the entity
+            # mass submatrix in-place in global memory (mass_mat_L) over a block of BLOCK_DIM threads. Each elimination
+            # step snapshots the pivot row into a small shared vector (O(n_dofs), not O(n_dofs^2)) before updating the
             # trailing submatrix, so the parallel per-row updates only READ the pivot row (from shared) -- race-free
             # regardless of scheduling. Numerically identical to the scalar branch below; only parallelization differs.
             BLOCK_DIM = qd.static(32)
-            MAX_DOFS_PER_ENTITY = qd.static(static_rigid_sim_config.tiled_n_dofs_per_entity_large)
+            MAX_DOFS_PER_ENTITY = qd.static(static_rigid_sim_config.tiled_n_dofs_per_entity)
 
             qd.loop_config(name="factor_mass", block_dim=BLOCK_DIM)
             for i in range(n_entities * _B * BLOCK_DIM):
--- a/genesis/engine/solvers/rigid/abd/forward_dynamics.py
+++ b/genesis/engine/solvers/rigid/abd/forward_dynamics.py
@@ -493,12 +493,11 @@ def func_factor_mass(
         _B = dofs_state.ctrl_mode.shape[1]
 
         if qd.static(static_rigid_sim_config.enable_tiled_cholesky_mass_matrix_large):
-            # Uncapped cooperative per-entity LDL^T (max_n_dofs_per_entity > max_n_threads). Factors the entity
-            # mass submatrix in-place in global memory (mass_mat_L), cooperatively over a block of BLOCK_DIM
-            # threads. Each elimination step snapshots the pivot row into a small shared vector (O(n_dofs), not
-            # O(n_dofs^2)) BEFORE updating the trailing submatrix, so the parallel row updates only ever READ
-            # the pivot row (from shared) -- race-free regardless of thread scheduling. Numerically identical to
-            # the scalar branch below; only the parallelization (and lack of an O(n_dofs^2) shared array) differs.
+            # Uncapped cooperative per-entity LDL^T (max_n_dofs_per_entity > max_n_threads): factors the entity mass
+            # submatrix in-place in global memory (mass_mat_L) over a block of BLOCK_DIM threads. Each elimination step
+            # snapshots the pivot row into a small shared vector (O(n_dofs), not O(n_dofs^2)) before updating the
+            # trailing submatrix, so the parallel per-row updates only READ the pivot row (from shared) -- race-free
+            # regardless of scheduling. Numerically identical to the scalar branch below; only parallelization differs.
             BLOCK_DIM = qd.static(32)
             MAX_DOFS_PER_ENTITY = qd.static(static_rigid_sim_config.tiled_n_dofs_per_entity_large)
 
--- a/genesis/engine/solvers/rigid/rigid_solver.py
+++ b/genesis/engine/solvers/rigid/rigid_solver.py
@@ -470,6 +470,13 @@ class RigidSolver(KinematicSolver):
 
                 enable_tiled_cholesky_mass_matrix = 8 <= max_n_dofs_per_entity <= max_n_threads and self.n_envs <= 16384
                 enable_tiled_cholesky_hessian = 16 <= self.n_dofs <= max_n_threads and self.n_envs <= 16384
+                # Above the shared-memory cap, the standalone register-streaming tiled factor
+                # (func_cholesky_factor_direct_tiled) has no DOF limit and replaces the scalar
+                # one-thread-per-env Cholesky fallback. Not used for sparse_solve (its incremental
+                # update assumes a different jac layout) -- that path keeps the scalar direct rebuild.
+                enable_tiled_cholesky_hessian_large = (
+                    self.n_dofs > max_n_threads and self.n_envs <= 16384 and not self._options.sparse_solve
+                )
 
                 # n_dofs-based dispatch between Tile16x16 and Tile32x32 Cholesky kernels (Hessian only).
                 # Derived from a padded-volume + sub-warp utilization model:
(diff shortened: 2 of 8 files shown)


genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -470,6 +470,13 @@ def _build_static_config(self):
 
                 enable_tiled_cholesky_mass_matrix = 8 <= max_n_dofs_per_entity <= max_n_threads and self.n_envs <= 16384
                 enable_tiled_cholesky_hessian = 16 <= self.n_dofs <= max_n_threads and self.n_envs <= 16384
+                # Above the shared-memory cap, the standalone register-streaming tiled factor
+                # (func_cholesky_factor_direct_tiled) has no DOF limit and replaces the scalar
+                # one-thread-per-env Cholesky fallback. Not used for sparse_solve (its incremental
+                # update assumes a different jac layout) -- that path keeps the scalar direct rebuild.
+                enable_tiled_cholesky_hessian_large = (
+                    self.n_dofs > max_n_threads and self.n_envs <= 16384 and not self._options.sparse_solve
```
hughperkins: why hard-code 16384? What is the basis for 16384? Is it the number of cores on your 30x0? Perhaps we can compare with the actual number of cores on the gpu instead? (there are existing formulas that compare with the number of gpu cores. eg look into narrowphase grid size formulae)

hughperkins: (Note: I suspect it should be `<` rather than `<=`, because if the number of envs exactly matches the available gpu cores, single thread per env might be faster perhaps?)

genesis/engine/solvers/rigid/abd/forward_dynamics.py
```diff
@@ -492,7 +492,87 @@ def func_factor_mass(
...
         _B = dofs_state.ctrl_mode.shape[1]
 
-        if qd.static(
+        if qd.static(static_rigid_sim_config.enable_tiled_cholesky_mass_matrix_large):
+            # Uncapped cooperative per-entity LDL^T (max_n_dofs_per_entity > max_n_threads). Factors the entity
+            # mass submatrix in-place in global memory (mass_mat_L), cooperatively over a block of BLOCK_DIM
+            # threads. Each elimination step snapshots the pivot row into a small shared vector (O(n_dofs), not
+            # O(n_dofs^2)) BEFORE updating the trailing submatrix, so the parallel row updates only ever READ
+            # the pivot row (from shared) -- race-free regardless of thread scheduling. Numerically identical to
+            # the scalar branch below; only the parallelization (and lack of an O(n_dofs^2) shared array) differs.
```
duburcqa: Maximize linewidth in code comments.

genesis/engine/solvers/rigid/abd/forward_dynamics.py
```diff
@@ -492,7 +492,87 @@ def func_factor_mass(
         n_entities = entities_info.n_links.shape[0]
         _B = dofs_state.ctrl_mode.shape[1]
 
-        if qd.static(
+        if qd.static(static_rigid_sim_config.enable_tiled_cholesky_mass_matrix_large):
```
duburcqa: Why do we need yet another completely independent block? It cannot be merged with 'enable_tiled_cholesky_mass_matrix'? This makes the code harder to test and harder to maintain.

genesis/engine/solvers/rigid/rigid_solver.py
duburcqa: See comment line 460. It is just a pre-existing heuristics that has been extended.

genesis/utils/array_class.py
```diff
@@ -2142,6 +2155,9 @@ class RigidSimStaticConfig(metaclass=AutoInitMeta):
     # subgroup-cooperative refinement in the linesearch and contiguous per-thread access.
     constraint_layout_transposed: bool = False
     tiled_n_dofs_per_entity: int = -1
+    # Shared-vector length (>= max_n_dofs_per_entity, multiple of 32) for the pivot-row snapshot used by the
+    # uncapped cooperative mass factor selected by ``enable_tiled_cholesky_mass_matrix_large``.
```
duburcqa: Remove all these code comments. I don't think they are helpful here. Moreover, keep all your code comment generic and avoid method stuffs like 'g1_fall'.

duburcqa: If these comments are helpful, move them close to where they are used, not where they are defined.

genesis/engine/solvers/rigid/rigid_solver.py
hughperkins: That there is a pre-existing heuristic which hard codes the number of gpu cores is not a good reason to continue to use that approach in new code I feel.

We have the technology to find out the actual number of gpu cores.

genesis/utils/array_class.py
```diff
@@ -2127,7 +2127,20 @@ class RigidSimStaticConfig(metaclass=AutoInitMeta):
     parallel_init: bool = False  # parallelize init over (constraints, envs) when GPU is not saturated by envs alone
     broadphase_traversal: int = 0
     enable_tiled_cholesky_mass_matrix: bool = False
+    # The per-entity tiled mass-matrix factor keeps the whole entity mass submatrix in shared memory, so
+    # ``enable_tiled_cholesky_mass_matrix`` is capped at ``max_n_dofs_per_entity <= max_n_threads``. Above that
+    # cap this flag selects a cooperative LDL^T that factors the entity submatrix in-place in global memory,
+    # snapshotting only the current pivot row into a small shared vector (O(n_dofs) shared, not O(n_dofs^2)),
+    # so it has no DOF cap. Replaces the scalar one-thread-per-(entity,env) fallback for high-DOF entities.
+    enable_tiled_cholesky_mass_matrix_large: bool = False
     enable_tiled_cholesky_hessian: bool = False
```
duburcqa: I'm not convinced we need to add all these variables with '*_large' suffix. Initially, it was introduced to because it was not possible to enable tiling systematically because of limited shared memory. If this is no longer the case, maybe the logics could be simplified.

genesis/engine/solvers/rigid/rigid_solver.py
duburcqa: > That there is a pre-existing heuristic which hard codes the number of gpu cores is not a good reason to continue to use that approach in new code I feel.

Sure. I'm just answering your question: "why hard-code 16384?".

> We have the technology to find out the actual number of gpu cores.

Sure, but this heuristics is not related to the number of gpu cores initially.

hughperkins: I'm not sure I see how the comment at line 460 explains 16834? It says:

```
            # TODO: These alternative tiled algorithms are designed to reduce the impact of latency. However, naive
            # implementation scales slightly better asymptotically than shared memory-based implementation because the
            # scheduler of modern GPUs is able to hides latency by swapping warps if the workload is sufficient. The
            # crossover threshold is both hardware and kernel-dependent. As a result, the optimal implementation should
            # be selected based on dynamic timer-based profiling instead of hard-coded heuristic.
```

This is basically hinting at the fact that it relates to the number of gpu cores, without making that connection. Here is the connection:

When the number of envs is at least equal to the nubmer of gpu cores, we can utilize all cores by running each env single threaded.

When the number of envs is much less than the number of gpus cores (eg 25% or fewer), then using cooperative threads, eg 32 threads per env, will enable fully utilizing the gpu cores, where using one thread per env would under-utilize the cores.

There is some crossover point between num envs == 25% of gpu cores, and num envs == 100%, where cooperative and single threaded swap over in which is fastest.

personally I've been using num_envs <= 0.5 * num gpu cores. We could also use num_envs < num gpu cores. They are approximately equivalent, and both are near enough I feel. Both of these are more accureate, than assuming that the gpu has 16384 cores I feel, and better expresses the reasoning behind the heuristic.

Kashu7100: Good point — I've removed the `n_envs <= 16384` guard from both new uncapped paths (`enable_tiled_cholesky_*_large`) in the latest push. Those branches only ever replace the scalar one-thread-per-env factor, which is O(n_dofs^3) serial and never preferable for high-DOF problems, so the env-saturation heuristic that motivates the cap on the existing flags doesn't apply here. I left the pre-existing flags' `16384`/`<=` untouched to keep this PR scoped, but happy to follow up with a core-count-based threshold for all of them in a separate change if you'd like.

genesis/utils/array_class.py
Kashu7100: The cap is still real for the shared-memory paths: the *fused factor+solve* and the *tiled triangular solve* both keep the full L in shared memory, so `enable_tiled_cholesky_hessian` (which gates those) must stay capped at `n_dofs <= max_n_threads`. The `_large` flag only selects the *standalone* factor (`func_cholesky_factor_direct_tiled`), which streams tiles through registers and has no such limit — so these are genuinely two regimes (capped solve vs uncapped factor), not redundant. If you'd prefer one consistent naming, I can instead split `enable_tiled_cholesky_hessian` into a factor flag (uncapped) and a solve/fused flag (capped) rather than adding a `_large` variant — let me know which you'd rather review.

genesis/engine/solvers/rigid/abd/forward_dynamics.py
Kashu7100: The existing tiled mass branch stages the whole entity submatrix in a shared array sized `tiled_n_dofs_per_entity` (capped), so it can't handle entities above the cap — hence a separate branch rather than extending it. I kept the shared-array path because it's faster for the common small-entity case (Franka/ANYmal/Go2); the cooperative in-place LDL^T trades shared-memory reuse for having no DOF cap and is only a win above the cap. I can fold them into one branch that picks shared-array vs streaming internally if you'd prefer a single code path — slightly more logic in one block, no behavior change.

genesis/utils/array_class.py
Kashu7100: Done — removed the field comments from the struct in the latest push; the rationale now lives at the usage sites (`rigid_solver` where the flags are computed, and the kernels where they're used). Also widened the kernel comment block to fill the line width.

Kashu7100:
### On the `anymal_uniform_kinematic` regression (−12.3%, ndarray/cuda/30000)

I believe this is measurement jitter rather than a real regression. ANYmal is 18 DOFs, and at `n_envs=30000` **all four** tiled-Cholesky flags evaluate to `False`, so the executed factorization path is byte-identical to `main`:

| flag | evaluation | result |
|---|---|---|
| `enable_tiled_cholesky_mass_matrix` | `8 ≤ 18 ≤ 128 and 30000 ≤ 16384` | **False** (n_envs cap) |
| `enable_tiled_cholesky_mass_matrix_large` | `18 > 128` | **False** |
| `enable_tiled_cholesky_hessian` | `16 ≤ 18 ≤ 128 and 30000 ≤ 16384` | **False** (n_envs cap) |
| `enable_tiled_cholesky_hessian_large` | `18 > 128` | **False** |

Both the mass and Hessian factorizations therefore take the same scalar branches as before (the new `qd.static(...)` branches are compile-time eliminated). Corroborating that it's not a code effect: only the `ndarray` variant moved while `field` for the same benchmark is +0.0%/within noise, and `run_benchmark`'s own docstring documents this exact failure mode ("intermittent 50%+ apparent runtime_fps regressions ... even though steady-state step() time was unchanged" from CUDA cold-start/compile timing landing in the wall-clock window). Could a maintainer re-run the benchmark job to confirm? Happy to dig further if it persists across re-runs.

### New benchmark: `double_smplx`

Per the suggestion to add a benchmark representing the target, I added `double_smplx` (two 159-DOF SMPL-X humanoids dropped on a plane with random torques), adapted from @hughperkins' `hp/double-smpl-benchmark` — thanks! I switched it from SMPL (75 DOF) to **SMPL-X (159 DOF)** so the *per-entity* mass factor also crosses the cap (the 318-DOF environment crosses the Hessian cap either way), making it the only suite entry that exercises **both** new paths; everything else is ≤62 DOFs/entity. Self-collision is disabled (the SMPL-X hands carry many finger geoms) to keep the cost solver-dominated. Asset comes from the `Kashu7100/eden_assets` HF dataset. Validated locally (builds, runs, finite; both `*_large` paths engage).

genesis/engine/solvers/rigid/rigid_solver.py
duburcqa: That’s make no sense

genesis/engine/solvers/rigid/abd/forward_dynamics.py
duburcqa: Do not let your AI answer PR reviews

genesis/utils/array_class.py
duburcqa: Do not let your AI answer PR reviews

duburcqa:
> I believe this is measurement jitter rather than a real regression.

that makes no sense.

Edit: ok now I see what you mean. That’s not jitter but may indeed be related to kernel dispatching peaking the wrong implementation.

duburcqa:
> Self-collision is disabled (the SMPL-X hands carry many finger geoms) to keep the cost solver-dominated.

that’s not how our benchmark are built.

duburcqa:
> Asset comes from the Kashu7100/eden_assets HF dataset.

that is not ok.

genesis/engine/solvers/rigid/rigid_solver.py
duburcqa: > This is basically hinting at the fact that it relates to the number of gpu cores, without making that connection.

Yes, I was not aware of this connection buck in November.

tests/test_rigid_benchmarks.py
```diff
@@ -550,6 +550,82 @@ def step():
...
+    """Two SMPL-X humanoids (159 DOFs each) dropped onto a plane with random per-DOF torques.
+
+    A high-DOF, multi-humanoid stress test for the rigid constraint solver: each entity has 159 DOFs
+    (> the per-entity tiled-Cholesky shared-memory cap), and the two-entity environment has 318 DOFs
+    (> the constraint-Hessian cap), so this exercises both the mass-matrix and Hessian factorizations on
+    high-DOF problems. Mirrors :func:`make_g1_fall`; adapted from @hughperkins' ``hp/double-smpl-benchmark``.
+    Self-collision is disabled (the SMPL-X hands carry many finger geoms) to keep the cost dominated by the
+    solver rather than self-collision broad-phase. The MJCF is fetched from the ``Kashu7100/eden_assets``
+    HuggingFace dataset (the SMPL-X humanoid tracked by the Eden project, https://github.com/Kashu7100/Eden).
+    """
```
duburcqa: Remove docstring. We don't do this in Genesis.

tests/test_rigid_benchmarks.py
```diff
@@ -550,6 +550,82 @@ def step():
...
+    step_dt = 0.005
+
+    scene = gs.Scene(
+        rigid_options=gs.options.RigidOptions(
+            dt=step_dt,
+            iterations=10,
+            tolerance=1e-5,
+            ls_iterations=20,
```
duburcqa: Let's keep default unless you have a strong reason not to do this that is well documented.

tests/test_rigid_benchmarks.py
```diff
@@ -550,6 +550,82 @@ def step():
...
+
+    for robot, y in zip(robots, y_offsets):
+        init_qpos = torch.zeros((robot.n_qs,), dtype=gs.tc_float, device=gs.device)
+        init_qpos[1] = y  # y position
+        init_qpos[2] = init_z  # z position
+        init_qpos[3] = 1.0  # quaternion w component
+        robot.set_qpos(init_qpos)
+
+    random_forces = [torch.zeros((n_envs, robot.n_dofs), dtype=gs.tc_float, device=gs.device) for robot in robots]
+    max_force = 50.0
```
duburcqa: Move hardcoded constants on top of the function, in uppercase.

tests/test_rigid_benchmarks.py
```diff
@@ -550,6 +550,82 @@ def step():
...
+
+    A high-DOF, multi-humanoid stress test for the rigid constraint solver: each entity has 159 DOFs
+    (> the per-entity tiled-Cholesky shared-memory cap), and the two-entity environment has 318 DOFs
+    (> the constraint-Hessian cap), so this exercises both the mass-matrix and Hessian factorizations on
+    high-DOF problems. Mirrors :func:`make_g1_fall`; adapted from @hughperkins' ``hp/double-smpl-benchmark``.
+    Self-collision is disabled (the SMPL-X hands carry many finger geoms) to keep the cost dominated by the
+    solver rather than self-collision broad-phase. The MJCF is fetched from the ``Kashu7100/eden_assets``
+    HuggingFace dataset (the SMPL-X humanoid tracked by the Eden project, https://github.com/Kashu7100/Eden).
+    """
+    step_dt = 0.005
```
duburcqa: Move hardcoded constants on top of the function, in uppercase.

tests/test_rigid_benchmarks.py
```diff
@@ -550,6 +550,82 @@ def step():
...
+    time_start = time.time()
+    scene.build(n_envs=n_envs)
+    compile_time = time.time() - time_start
+
+    for robot, y in zip(robots, y_offsets):
+        init_qpos = torch.zeros((robot.n_qs,), dtype=gs.tc_float, device=gs.device)
+        init_qpos[1] = y  # y position
+        init_qpos[2] = init_z  # z position
+        init_qpos[3] = 1.0  # quaternion w component
+        robot.set_qpos(init_qpos)
```
duburcqa: You are duplicating the init pos logics.

tests/utils.py
duburcqa, line 1: We are not going to do remove. Revert this entire file.

duburcqa (approved):

duburcqa:
For the record, here are the benchmarks data when running `GS_ENABLE_NDARRAY=0 pytest -m benchmarks -n 0 "tests/test_rigid_benchmarks.py::test_speed[double_smplx-Newton-None-4096-gpu]"`

```
### main

compile_time=64.2 	| runtime_fps=2842.0 	| realtime_factor=14.2

### This PR

compile_time=84.2 	| runtime_fps=6912.0 	| realtime_factor=34.6
```

Beyond this specific benchmark, not performance difference is observed, without surprise. Moreover, dropping the shared-memory code path to only keep the newly introduced register-based implementation is causing 15% performance drop.

|env                     |batch_size|constraint_solver|gjk_collision|PR_vs_main_pct|PR-coop_only_vs_main_pct|
|------------------------|----------|-----------------|-------------|--------------|------------------------|
|anymal_random           |30000     |                 |             |-1.1          |-1.2                    |
|anymal_uniform          |30000     |                 |             |-0.4          |-0.4                    |
|anymal_uniform_kinematic|30000     |                 |             |1.8           |2.4                     |
|anymal_zero             |30000     |                 |             |-0.8          |-0.6                    |
|box_pyramid_3           |4096      |                 |             |0.1           |-1.5                    |
|box_pyramid_4           |4096      |                 |             |2.9           |3.1                     |
|box_pyramid_5           |4096      |                 |             |0.2           |0.8                     |
|box_pyramid_6           |4096      |                 |False        |-0.2          |-0.1                    |
|box_pyramid_6           |4096      |                 |True         |1.9           |1.0                     |
|dex_hand                |4096      |                 |             |-9.4          |0.7                     |
|double_smplx            |4096      |Newton           |             |              |                        |
|duck_in_box_easy        |30000     |                 |False        |0.1           |1.8                     |
|duck_in_box_easy        |30000     |                 |True         |1.8           |0.9                     |
|duck_in_box_hard        |30000     |                 |False        |-1.7          |-0.6                    |
|duck_in_box_hard        |30000     |                 |True         |0.9           |-1.4                    |
|franka                  |30000     |                 |             |-3.6          |2.1                     |
|franka_accessors        |30000     |                 |             |2.5           |1.7                     |
|franka_free             |30000     |                 |             |1.7           |1.7                     |
|franka_random           |30000     |                 |False        |1.7           |-0.6                    |
|franka_random           |30000     |                 |True         |-2.0          |-1.3                    |
|franka_random           |30000     |CG               |             |-0.9          |-0.4                    |
|franka_random           |30000     |Newton           |             |-1.5          |-1.5                    |
|g1_fall                 |4096      |Newton           |             |-17.8         |0.1                     |
|go2                     |4096      |                 |True         |-10.5         |-1.2                    |
|go2                     |4096      |CG               |False        |-11.7         |-0.0                    |
|go2                     |4096      |Newton           |False        |-12.7         |-1.1                    |

PR #2872, opened 2026-06-01
[FEATURE] Add complementarity-free constraint solver

Kashu7100:
## Summary

- Adds the ComFree (complementarity-free) constraint solver from arXiv:2603.12185 as a third `constraint_solver` option alongside CG and Newton: every constraint force is computed analytically in a single pass instead of iterating a complementarity solve to convergence.
- `ComFreeSolver` subclasses `ConstraintSolver` and overrides only `resolve()`, sharing the whole assembly pipeline (contacts, equalities, joint limits, frictionloss), reset/clear, and force publication. The resolution is one kernel: analytical per-row force, `J^T @ force` gather, pre-factored mass solve.
- Per-row forces are projected onto each row's admissible set: contact and joint-limit rows push only (realizing the friction pyramid facet-by-facet), frictionloss rows are box-bounded, equality rows stay two-sided.
- Two new options, `comfree_stiffness` and `comfree_damping`, tune the analytical force. Incompatible options are rejected (elliptic cone, noslip, differentiable mode) or resolved automatically (contact islands).
- Zero cost for CG/Newton users: the two extra assembly outputs (`efc_dist`, `efc_mass`) are statically compiled out and their buffers are empty unless ComFree is selected.

## Motivation and Context

RL workloads step thousands of environments where the iterative constraint solve dominates the step budget. A single-pass analytical solver trades contact accuracy (compliant, tunable spring-damper behavior) for large throughput gains on contact-dense batched scenes.

Measured whole-step throughput vs Newton (interleaved A/B/A runs, Apple M-series):
- 16-box pile, CPU, 1 env: ~1.6x
- 16-box pile, Metal, 512 envs: ~2.5x
- 16-box pile, Metal, 4096 envs: ~4.0x
- Franka + cube, Metal, 4096 envs: ~1.5x

- New `tests/rigid/test_constraints.py::test_comfree_solver` (parametrized over `n_envs=[0, 2]`): one packed scene asserting analytic free fall, compliant contact settling and rest, two-sided connect equality holding a hung body, symmetric frictionloss braking of opposite spins, joint-limit containment (transient and steady-state), and identical batched trajectories; plus `test_comfree_incompatible_options` for option validation. Passing locally on CPU and Metal.
- Full `tests/rigid` required suite passing on CPU (no regression for CG/Newton: their compiled kernels are unchanged).

## Checklist:


duburcqa:
The implementation was audited in depth, rebased onto current `main`, and extended on this branch (torsional/rolling friction support, a phantom-force fix on zeroed friction rows, equality impedance aligned with the iterative solvers, and mutation-tested coverage of every constraint-row type). The solver is a faithful port of the reference `comfree_warp` implementation, and its resting/sliding contact behavior, friction dissipation, equalities, and joint limits are all correct.

Unfortunately the audit surfaced a disqualifying limitation that is inherent to the published method, and this PR cannot be merged as is.

## ComFree has no static friction

The analytical facet force (paper Eq. 8) is a pure function of the predicted velocity: opposing friction facets can only produce a net tangential force through a velocity asymmetry. Resisting a sustained tangential load therefore requires steady slip — friction is viscous by construction. Exact complementarity solvers find the facet forces that null the acceleration (true stiction); a single closed-form pass cannot.

Minimal quantification — a 1kg cube pressed against a vertical wall at 20N (Coulomb capacity 20N, weight 9.8N, mu=1.0; any solver with static friction holds it):

| Solver | substeps (dt=0.01) | vertical creep |
|---|---|---|
| Newton | 1 | 0.0045 m/s |
| Newton | 4 | 0.0026 m/s |
| ComFree | 1 | **2.54 m/s** (free fall) |
| ComFree | 4 | **0.55 m/s** |

Consequences: `test_pick_and_place` fails with ComFree at the 1N grip that Newton holds with — the cube slides out of the gripper during the lift (video below: identical scene, gains, and grip force; only the constraint solver differs). Anything relying on stiction (grasps, stacks under shear, feet) is affected. The escape hatches do not work either: the creep rate is independent of the normal force, and stiffening the contact impedance (the reference repo's own fingertip `solimp="0.97 0.995"`) crosses the stability boundary of the explicit update in this configuration and ejects the object instead of sticking. The paper never tests a sustained tangential load (its friction experiments are all decay-to-rest), and its own Eq. 10 exact dual-cone impedance — which would fix this — is explicitly discarded as too expensive, in favor of the under-stiffened heuristic that causes the creep.

<details>
<summary>Standalone reproduction (run on this branch)</summary>

```python
import genesis as gs

for solver in (gs.constraint_solver.Newton, gs.constraint_solver.ComFree):
    for substeps in (1, 4):
        scene = gs.Scene(
            sim_options=gs.options.SimOptions(dt=0.01, substeps=substeps),
            rigid_options=gs.options.RigidOptions(constraint_solver=solver, box_box_detection=True),
            show_viewer=False,
        )
        scene.add_entity(gs.morphs.Box(size=(0.1, 0.5, 0.5), pos=(-0.05, 0.0, 0.5), fixed=True))
        cube = scene.add_entity(gs.morphs.Box(size=(0.1, 0.1, 0.1), pos=(0.0501, 0.0, 0.5)))
        scene.build()
        cube.control_dofs_force([-20.0, 0.0, 0.0, 0.0, 0.0, 0.0])
        for _ in range(50):
            scene.step()
        z1 = float(tensor_to_array(cube.get_pos())[2])
        for _ in range(100):
            scene.step()
        z2 = float(tensor_to_array(cube.get_pos())[2])
        print(f"{solver.name:8s} substeps={substeps}: vertical creep = {z1 - z2:9.6f} m/s")
        gs.destroy()
```

</details>

## What would change the verdict

A mechanism providing genuine stiction, e.g. the exact dual-cone impedance of Eq. 10 made stable, or an elastic tangential-anchor state — both of which are research on the method itself rather than integration work. Until then the solver would silently break the manipulation and locomotion use cases Genesis users rely on, and a solver that must not be used for grasping cannot ship as a general `constraint_solver` option.

Closing for now — happy to revisit if the stiction problem is addressed upstream. The audited branch is preserved should that happen.

duburcqa:
The reproduction video referenced above — same scene, same control gains, same 1N grip force; only the constraint solver differs. Newton (left) lifts and holds the cube; ComFree (right) lets it creep out of the closed gripper during the lift. A second symptom of the same under-stiffened analytical law follows: once the cube is lost, the sustained 1N closing command drives the fingers through their compliant joint limits.

Newton vs ComFree Franka grasp at identical 1N grip

PR #2874, opened 2026-06-01
[MISC] Add solver state-change subscriber mechanism. Fix raycast sensor for heterogeneous entities.

duburcqa:
## Description

Two independent, self-contained changes, bundled so the raycaster-perf PR (#2867) can rebase on top and consume the mechanism instead of its per-step `torch.equal` pose diff.

### 1. Solver state-change subscriber mechanism (`[MISC]`)

A generic, solver-agnostic way for a solver to notify interested parties when its state is mutated by a setter, so caches (e.g. sensor BVHs) can be invalidated without polling.

- `StateChange` enum — categories only (`GEOMETRY` / `DYNAMICS`), never an index space (links/dofs/particles differ per solver).
- `Subscriber` — a unique handle a consumer constructs and registers via `solver.subscribe(subscriber)`; the solver keeps a `set` of them. Mode is fixed at construction:
  - eager (a callback is given): each matching change calls `callback(change, envs_idx)` immediately;
  - lazy (no callback): changes accumulate in `subscriber.pending` until the owner calls `clear()` — letting a sensor refresh on its next update rather than on every `set_pos`.
- `@mutates(StateChange.GEOMETRY)` tags the geometry-mutating setters (`set_base_links_pos/quat`, `set_qpos`, `set_dofs_position`) in the rigid and kinematic solvers. Reads and physics integration never notify, and there is zero overhead when a solver has no subscribers.

### 2. Heterogeneous-env collision raycasting fix (`[BUG FIX]`)

The raycaster collision BVH treated heterogeneous entities (`add_entity(morph=(variant_0, variant_1, ...))`) as the **union** of all variants in every env: all envs share one vertex buffer holding every variant, and `update_aabbs` built every face's AABB unconditionally. A ray then hit whichever variant was closer, identically in all envs.

`update_aabbs` now includes a face in env `i_b` only when its geom lies in that env's active range (`links_info.geom_start/geom_end`, indexed via the existing `batch_links_info` idiom). Inactive faces keep an inverted AABB and are skipped by ray queries — exactly like the visual `face_mask`. For a homogeneous solver every geom is always in range, so this is a no-op.

## Testing

`pytest tests/test_solver_subscribers.py tests/test_sensors.py -k "raycaster or lidar or subscriber"` (CPU backend):

- `test_solver_state_change_subscribers` — eager vs lazy modes, category filter, `envs_idx` forwarding, and that reads / physics steps never notify.
- `test_raycaster_heterogeneous_object` — box variant in env 0, sphere variant in env 1 at the same pose; asserts per-env distances `[0.9, 0.8]` (the union bug returned `[0.8, 0.8]`).
- Existing raycaster/lidar tests still pass (the masking is a verified no-op for homogeneous scenes).

### DIFF
--- a/genesis/engine/sensors/raycaster.py
+++ b/genesis/engine/sensors/raycaster.py
@@ -115,6 +115,7 @@ class RaycasterSensor(KinematicSensorMixin, SimpleSensor[RaycasterOptions, Rayca
                     faces_info=entry.solver.faces_info,
                     free_verts_state=entry.solver.free_verts_state,
                     fixed_verts_state=entry.solver.fixed_verts_state,
+                    links_info=entry.solver.links_info,
                     static_rigid_sim_config=entry.solver._static_rigid_sim_config,
                     aabb_state=entry.aabb,
                 )
--- a/genesis/engine/solvers/base_solver.py
+++ b/genesis/engine/solvers/base_solver.py
@@ -1,4 +1,7 @@
-from typing import TYPE_CHECKING
+import enum
+import functools
+import inspect
+from typing import TYPE_CHECKING, Callable
 
 import quadrants as qd
 import numpy as np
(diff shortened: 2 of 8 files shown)


duburcqa:
Split into two orthogonal PRs: #2875 (solver state-change subscriber mechanism) and #2876 (heterogeneous-env raycast fix).

PR #2908, opened 2026-06-07
[FEATURE] Add distances-only mode to raycaster.

Kashu7100:
## Summary

A `Raycaster`/`DepthCamera` always stores **both** the per-ray xyz hit point and the hit distance in its output cache (`H*W*(3+1)` floats per env), even though depth-image consumers read only `.distances`. For a depth camera that means 3/4 of the cache (and the per-ray write bandwidth) is spent on points nobody reads.

This PR adds **`return_points: bool = True`** to the raycaster options. With `return_points=False` the sensor skips computing and storing the hit points: the output cache shrinks ~4x (just the `H*W` distances), and the cast kernel stops doing the per-ray point transform + 3 writes. `read().points` is `None` when disabled; `.distances` is **bit-identical** to the points-on result.

## Implementation

- `return_points` flag on `Raycaster` options (inherited by `DepthCamera`); `read()` keeps the `(points, distances)` NamedTuple contract with `points=None`.
- `_get_return_format` drops the `(*shape, 3)` points field when disabled, so the shared cache is allocated 4x smaller automatically.
- The cast kernels gate point writes per-sensor via `sensor_return_points` and locate the distance block accordingly, so a points sensor and a distances-only sensor share one BVH cast.
- Per-sensor cache offsets are now a running cumulative sum (was `cache_size*(idx+1)`, which assumed a uniform per-sensor cache size) so sensors of differing cache sizes pack correctly. The imperfection setters (`set_bias`/`set_noise`/`set_resolution`/`set_random_walk`) had the same uniform-size assumption in their flat-field indexing; they now write at the sensor's actual cache offset.
- Debug drawing reconstructs local hit points as `distance * unit ray_dir` for distances-only sensors, with missed rays collapsing onto the ray start exactly as when points are stored.

## Benchmark

Static depth scene (RTX 3080, plane + fixed boxes), perception cost as the marginal `scene.step()` time; `return_points=True` -> `False`:

| envs | res | VRAM | step time |
|-----:|-----|------|-----------|
| 4096 | 64x36 | 1153 -> 289 MB (**4.0x**) | 9.3 -> 5.8 ms (1.6x) |
| 1024 | 128x96 | 1538 -> 385 MB (**4.0x**) | 19.9 -> 7.7 ms (**2.6x**) |
| 4096 | 128x96 | 6147 -> 1538 MB (**4.0x**) | 77.5 -> 30.1 ms (**2.6x**) |

The speed win grows with rays x envs because the cast was bandwidth-bound writing the unused xyz.

## Tests

`tests/sensors/test_raycaster.py::test_hits` gains a distances-only twin of its depth camera (`n_envs in {0, 2}`): distances bit-identical to the points-on twin sharing the cast, point/distance norm consistency at `gs.EPS` while heterogeneous cache blocks are packed together, and a `set_bias` regression asserting imperfections land only on the sensor they are set on (fails with the previous uniform-stride indexing).


Milotrince:
Yeah thats a lot of MB so the motivation is fair. I suspect draw_debug would fail with this modification. Should write a warning that draw_debug is disabled when `return_points=False` and `draw_debug=True`.

Kashu7100:
sounds good!

Milotrince:
Since you're busy I'll update this PR with the debug draw fix and some minor things.

duburcqa (approved):

PR #2911, opened 2026-06-08
[diagnosis] Nyx headless renderer crash on H20-3e (SM90): root cause + repro + workaround

SueMarsR:
## What

Adds a **root-cause diagnosis, reproduction, and a temporary LD_PRELOAD workaround** for the Nyx path-traced renderer (`gs-nyx`) crashing during `scene.build()` on a headless datacenter GPU. No engine code is changed — this is documentation + diagnostic tooling under `docker-genesis/`.

> ⚠️ The actual fix must be in the closed-source **`gs-nyx`** wheel, not in this repo. This PR documents the bug precisely (with backtraces and Vulkan call traces) so the Nyx team can fix it; I'm also filing an issue. Maintainers may prefer to take only `NYX_BUG_REPORT.md` — happy to trim.

## Root cause (the "EGL segfault" was a red herring)

Earlier this looked like an NVIDIA EGL incompatibility (`libnvidia-eglcore` in the backtrace). It is **not**. Step-by-step debugging shows two unchecked degenerate-scene bugs in the gs-nyx native renderer:

1. **Unchecked zero-size external-memory allocation.** During build, gs-nyx calls `vkAllocateMemory(allocationSize = 0)` for its CUDA↔Vulkan interop buffer (invalid per the Vulkan spec). The NVIDIA driver returns `VK_ERROR_OUT_OF_DEVICE_MEMORY (-2)` and a **NULL** `VkDeviceMemory`; gs-nyx does not check the result and binds the NULL handle via `vkBindBufferMemory` → SIGSEGV **inside the driver** (this is why an `libnvidia-eglcore` frame appears — it's the driver dereferencing NULL, not an EGL-context problem).
2. **Divide-by-zero in the same path.** Forcing the allocation to succeed advances past the NULL bind, then hits a hardware **SIGFPE** (`div %rsi`, `rsi = 0`). Both stem from one element count being `0`.

A version-skew blocker was also found: **`gs-nyx-plugin` 0.1.2 targets the pre-1.0 sensor API** and fails at `add_sensor` on `genesis-world >= 1.0` with `TypeError: NyxCameraData.__new__() missing 1 required positional argument: 'rgb'`. **`gs-nyx-plugin` 0.1.3 + `gs-nyx` 0.1.2** are required to reach `build()`.

## Proof it is not driver / EGL / headless

- A ~20-line **pure-Vulkan** program on the same node does `vkCreateInstance` / `vkEnumeratePhysicalDevices` (2× H20-3e) / `vkCreateDevice` / `vkDestroy*` **cleanly**.
- `vulkaninfo --summary` enumerates both GPUs (`apiVersion 1.4.303`).
- Forcing GLFW's **Null** platform and stripping all WSI instance/device extensions (`VK_KHR_swapchain`, surface extensions) **does not change the crash**.

## Device / environment

| | |
|---|---|
| GPU | NVIDIA **H20-3e** (Hopper, **SM 90** / compute 9.0, 143 GB), **headless** |
| Driver | **570.133.20**, CUDA 12.8, Vulkan 1.4.303 |
| Host OS | Alibaba Cloud Linux 3 (glibc 2.32) |
| Container | `pytorch/pytorch:2.11.0-cuda12.8-cudnn9-devel` (Ubuntu 24.04, glibc 2.39) |
| Python / torch | 3.12 / 2.11.0+cu128 |
| genesis-world / quadrants | 1.0.0 / 1.0.0 |
| gs-nyx / gs-nyx-plugin | **0.1.2 (cp312-abi3)** / **0.1.3** |

## Reproduction parameters

```python
scene = gs.Scene(sim_options=gs.options.SimOptions(dt=0.01), show_viewer=False)
scene.add_entity(morph=gs.morphs.Plane(plane_size=(10.0, 10.0)))
scene.add_entity(morph=gs.morphs.Sphere(radius=0.3, pos=(0, 0, 0.5)))
cam = scene.add_sensor(NyxCameraOptions(
    res=(320, 240), pos=(-1, 1, 1.2), lookat=(0, 0, 0.1),
    fov=20.0, spp=4, render_mode=npr.ERenderMode.FastPathTracer,
))
scene.build(n_envs=1)   # SIGSEGV
```

Scene richness (extra geometry, env map) does not change the outcome. `examples/01_hello_nyx.py` crashes identically.

## Suggested upstream fix (in `gs-nyx`)

1. Never call `vkAllocateMemory` with `allocationSize == 0`; clamp or skip when the count is 0.
2. **Check `vkAllocateMemory`'s result** before use; never bind a NULL `VkDeviceMemory`.
3. Guard the divisor against 0 in the same resource-setup routine.

## Files

- `docker-genesis/NYX_BUG_REPORT.md` — full report (v1 EGL report kept as `_v1_egl_superseded.md` to document the ruled-out theory).
- `docker-genesis/nyx_headless_shim.c` — LD_PRELOAD diagnostic/workaround (GLFW Null platform; logs+guards `vkAllocateMemory`/`vkBindBufferMemory`; converts the SIGSEGV into a clean error).
- `docker-genesis/nyx_probe_counts.py`, `nyx_upgrade_test.sh`, `nyx_env_matrix.sh` — probes.
- `docker-genesis/nyx_upgrade/*.txt` — gdb backtraces (SIGSEGV + SIGFPE). The large `gs-nyx` wheel is gitignored (fetched from PyPI).


SueMarsR:
Tracking issue: Genesis-Embodied-AI/genesis-nyx#12

duburcqa:
Wrong repo. Please reopen a PR and/or an issue in https://github.com/Genesis-Embodied-AI/genesis-nyx.

PR #2914, opened 2026-06-08
[MISC] Reduce memory footprint of ray casting-based sensors via cross-env memory sharing.

Kashu7100:
## Description

The raycast BVHs (`Raycaster` / `DepthCamera` / `Lidar`, also traversed by the tactile probes) allocate one tree per env: nodes, AABBs, morton codes and the radix-sort scratch are all `(n_envs, ...)`. For a high-poly static terrain this replication dominates GPU memory and OOMs at a few thousand envs, even though the per-env trees are identical or fall into a handful of variants.

Building on the static/dynamic split of #3078, this PR makes every static BVH automatically allocate one tree per distinct per-env geometry instead of one per env: identical envs share a single tree, N heterogeneous variants get N trees. It applies to the static collision subset of a mixed robot-on-terrain solver, to fully static solvers, and to the static visual BVH.

Envs are grouped by exact equality of every per-env input of the tree build - active geom ranges and free-vert positions for collision, vgeom poses and custom vverts for visual, each restricted to the entry's own subset - so a wrong merge is impossible and the grouping is blind to the movable geometry diverging under physics. An `is_env_uniform` entry (#3078) collapses to one tree with the signature left unread. A per-env `set_pos` on fixed geometry regroups on rebuild (up to one tree per env) and an identical reset merges back; tree allocations are pooled per tree count so diverge/reset cycles reuse them. Two routing arrays drive the cast: `env_bvh_idx` (tree slot per env) and `env_cast_order` (envs sorted by tree, keeping the env-major thread mapping coalesced even when group membership interleaves). Tactile probes traverse both folded trees through the same routing.

## Related Issue

None.

## Motivation and Context

Massively parallel RL on static terrain: static-BVH memory drops by the ratio of env count to distinct-geometry count, on top of #3078's rebuild-cost win.

Measured on the #3078 benchmark scene (default Terrain, 41.7k static faces + Franka, 2.3k rays/env, distances-only), interleaved A/B/A/B against main:

- CUDA, 1024 envs: total GPU memory 7755 -> 2051 MB (3.8x; the static subset collapses from 1024 trees to 1), stepping 11.0 -> 11.0 ms/step and per-step configuration teleports 16.0 -> 16.0 ms/step (parity), per-step `scene.reset()` 93.8 -> 29.5 ms (3.2x: one static tree rebuilt instead of 1024, and the identical-envs short-circuit replaces the per-env AABB comparison).
- CPU, single env: parity within run noise on every arm (~6 ms/step).

On a fully static 4.6k-face terrain scene against per-env trees, 1284 -> 497 MB; per-env trees OOM at 4096 envs while the shared tree runs.

- `tests/sensors/test_raycaster.py::test_shared_static_bvh_regroup` (new, n_envs 0 and 4, mixed solver): collision static subset and kinematic visual BVH collapse to one tree each with analytic distances; a non-monotonic per-env `set_pos` splits the static subset to per-env trees with per-env distances; an interleaved per-env `set_pos` on the visual entity splits the visual BVH into two trees; a per-env divergence of the movable entity leaves the restored static subset on one shared tree; reset merges everything back.
- `tests/sensors/test_raycaster.py::test_static_dynamic_bvh_split_merge` asserts the static subset holds a single tree while the dynamic subset stays per env, through every merge/rebuild phase of #3078 (tactile fold included).
- `tests/sensors/test_raycaster.py::test_heterogeneous_object` now runs 6 envs and asserts the collapse to 3 trees with per-variant distances.
- `tests/sensors/test_tactile.py::test_raycast_probe_on_fully_fixed_solver` (new): tactile probes traverse the shared tree through the env routing.
- Full `tests/sensors/test_raycaster.py` + `test_tactile.py` + `test_api.py` + `tests/core/test_misc.py` pass on CPU, CUDA and Metal backends.

## Checklist:


Milotrince:
I removed the `shared_static_raycast_bvh` flag in favor of it being automatic!

duburcqa (approved):

PR #2937, opened 2026-06-12
[BUG FIX] offscreen camera renders at viewer resolution when show_viewer=True

### LINKED ISSUE #2927: [Bug]: Offscreen camera renders at viewer resolution instead of configured resolution when show_viewer=True
### Bug Description

When `show_viewer=True`, offscreen camera renders (via `camera.render()`) return images at the interactive viewer's resolution rather than the resolution explicitly configured via `scene.add_camera(res=...)`. This causes `camera.intrinsics` to be inconsistent with the actual rendered image dimensions, which in turn makes `camera.render_pointcloud()` fail because the intrinsic matrix is computed for the configured resolution but the depth image has a different shape.
When `show_viewer=False`, everything works correctly.

This bug seems to be introduced in 0.4.7 and still occurs in latest release (1.1.1 at the time of this issue).

### Steps to Reproduce

```python
import genesis as gs
import numpy as np

gs.init(backend=gs.gpu)

RES = (360, 240)  # (width, height)


def test_resolution(show_viewer):
    scene = gs.Scene(
        show_viewer=show_viewer,
        vis_options=gs.options.VisOptions(env_separate_rigid=True),
    )

    scene.add_entity(morph=gs.morphs.Sphere(pos=(0.5, 0.0, 0.5), radius=0.1))

    camera = scene.add_camera(
        res=RES, pos=(1.0, 0.0, 0.5), lookat=(0.5, 0.0, 0.0), fov=35.0,
    )

    scene.build(n_envs=1)
    scene.step()

    _, depth, _, _ = camera.render(depth=True, force_render=True)

    actual_h, actual_w = depth.shape
    expected_w, expected_h = RES

    print(f"\nshow_viewer={show_viewer}")
    print(f"  Configured resolution (W, H): {expected_w} x {expected_h}")
    print(f"  Actual depth shape (H, W):    {actual_h} x {actual_w}")

    if (actual_w, actual_h) == RES:
        print(f"  PASS: Resolution matches")
    else:
        print(f"  FAIL: Resolution mismatch!")
        print(f"    camera.intrinsics uses configured resolution"
              f" but depth has different shape.")

    # Downstream impact: point cloud unprojection fails
    try:
        camera.render_pointcloud()
        print(f"  Point cloud render succeeded")
    except Exception as e:
        print(f"  Point cloud render FAILED: {type(e).__name__}: {e}")

    return (actual_w, actual_h) == RES


pass_with_viewer = test_resolution(show_viewer=True)
pass_without_viewer = test_resolution(show_viewer=False)

print("\n" + "=" * 60)
if pass_without_viewer and pass_with_viewer:
    print("All tests passed!")
elif pass_without_viewer and not pass_with_viewer:
    print("Bug reproduced: resolution mismatch only when show_viewer=True")
else:
    print("Unexpected behavior")
print("=" * 60)
```

### Expected Behavior

`camera.render()` should always return images at the resolution specified in `scene.add_camera(res=...)`, regardless of whether the viewer is enabled and regardless of the viewer's resolution. `camera.render_pointcloud()` should work correctly in both cases.

### Screenshots/Videos

_No response_

### Relevant log output

```shell
[Genesis] [10:14:32] [INFO] ╭─────────────────────────────────────────────────────────────────────────────────────╮
[Genesis] [10:14:32] [INFO] │┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉ Genesis ┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉┈┉│
[Genesis] [10:14:32] [INFO] ╰─────────────────────────────────────────────────────────────────────────────────────╯
[Genesis] [10:14:33] [INFO] Running on [NVIDIA GeForce RTX 3080] with backend gs.cuda. Device memory: 9.64 GB.
[Genesis] [10:14:33] [INFO] 🚀 Genesis initialized. 🔖 version: 1.1.1, 🎨 theme: dark, 🌱 seed: None, 🐛 debug: False, 📏 precision: 32, 🔥 performance: False, 💬 verbose:INFO
[Genesis] [10:14:34] [INFO] Scene <b616561> created.
[Genesis] [10:14:34] [INFO] Adding <gs.engine.entities.RigidEntity>. idx: 0, uid: <7daca5b>, morph: <gs.morphs.Sphere>, material: <gs.materials.Rigid>.
[Genesis] [10:14:34] [INFO] Building scene <b616561>...
[Genesis] [10:14:39] [INFO] Compiling simulation kernels...
[Genesis] [10:14:45] [INFO] Building visualizer...
[Genesis] [10:15:02] [INFO] Viewer created. Resolution: 720×540, refresh_rate: 60, dt: 0.01s, realtime_factor: 1.0 (real-time target: 100.0 FPS).

show_viewer=True
  Configured resolution (W, H): 360 x 240
  Actual depth shape (H, W):    540 x 720
  ✗ FAIL: Resolution mismatch!
    camera.intrinsics uses configured resolution but depth has different shape.
  Point cloud render FAILED: ValueError: cannot reshape array of size 259200 into shape (540,720,3)
[Genesis] [10:15:03] [INFO] Scene <7852769> created.
[Genesis] [10:15:03] [INFO] Adding <gs.engine.entities.RigidEntity>. idx: 0, uid: <65c023a>, morph: <gs.morphs.Sphere>, material: <gs.materials.Rigid>.
[Genesis] [10:15:03] [INFO] Building scene <7852769>...
[Genesis] [10:15:03] [INFO] Compiling simulation kernels...
[Genesis] [10:15:04] [INFO] Building visualizer...

show_viewer=False
  Configured resolution (W, H): 360 x 240
  Actual depth shape (H, W):    240 x 360
  ✓ PASS: Resolution matches
  Point cloud render succeeded

============================================================
Bug reproduced: resolution mismatch only when show_viewer=True
============================================================
[Genesis] [10:15:04] [INFO] 💤 Exiting Genesis and caching compiled kernels...
```

### Environment

- OS: Ubuntu 24.04
- GPU/CPU: NVIDIA GeForce RTX 3080

genesis-world==1.1.0

### Additional Context

The downstream impact is that `camera.render_pointcloud()` (and any manual point cloud unprojection using `camera.intrinsics`) produces incorrect or crashed results because the intrinsic matrix is computed for the configured resolution while the rendered depth image has the viewer's actual resolution and the two don't match. Even if `render_pointcloud()` doesn't crash, the resulting point cloud would be geometrically wrong.

HaozheZhang6:
## What

Fixes #2927. An offscreen `camera.render()` returned images at the **interactive viewer's** resolution instead of the camera's configured `res` when `show_viewer=True`, so `camera.intrinsics` no longer matched the rendered image shape and `camera.render_pointcloud()` failed.

## Root cause

In `Viewer._event_loop_step_offscreen` (`genesis/ext/pyrender/viewer.py`), the camera's offscreen render target was forced to `self._offscreen_viewport_size` (the interactive window's size) before rendering:

```python
saved_viewport = (target.viewport_width, target.viewport_height)
target.viewport_width, target.viewport_height = self._offscreen_viewport_size
```

`target` is the camera's own offscreen FBO, already sized to the camera's resolution and independent of the OS window — so the window-clamping workaround that override was added for doesn't apply to it. The override just made every camera render come out at the viewer's size whenever a viewer existed. The fix renders at the target's own viewport. (`show_viewer=False` doesn't go through the viewer, which is why it was already correct.)

## Verification

Using the issue's repro (rasterizer, camera `res=(360, 240)`):

| | before | after |
|---|---|---|
| `show_viewer=True` depth shape (H, W) | `(512, 682)` ❌ | `(240, 360)` ✅ |
| `show_viewer=True` `render_pointcloud()` | fails | works ✅ |
| `show_viewer=False` depth shape | `(240, 360)` ✅ | `(240, 360)` ✅ (no regression) |

Verified on Linux + GPU backend under a virtual display (Xvfb).

I drafted a regression test in `tests/test_render.py` (assert `depth.shape == configured res`, parametrized like `test_render_api`), but the bug path only triggers when an interactive viewer is actually available — happy to add it in whatever form fits your CI's display/EGL setup.

### DIFF
--- a/genesis/ext/pyrender/viewer.py
+++ b/genesis/ext/pyrender/viewer.py
@@ -798,19 +798,17 @@ class Viewer(pyglet.window.Window):
                 # Update context, just in case is not already done before
                 self.gs_context.update()
 
-                # Render current frame from camera viewpoint. Force the renderer back to the originally
-                # requested viewport size so the offscreen FBO honors what the caller asked for even when the
-                # window has since been clamped to a smaller content area by the OS (e.g. macOS CI runners).
+                # Render current frame from camera viewpoint. ``target`` is the camera's own
+                # offscreen FBO, already sized to the camera's configured resolution, so render at
+                # that size. Forcing it to ``self._offscreen_viewport_size`` (the interactive
+                # window's size) made offscreen camera renders come out at the viewer resolution
+                # whenever ``show_viewer=True`` (#2927); the camera FBO is independent of the window
+                # so it is never subject to the OS window clamping that override was guarding against.
                 self._offscreen_results = []
                 self.render_flags["offscreen"] = True
                 self.render_flags["skip_markers"] = skip_markers
-                saved_viewport = (target.viewport_width, target.viewport_height)
-                target.viewport_width, target.viewport_height = self._offscreen_viewport_size
-                try:
-                    self.clear()
-                    retval = self._render(camera, target, normal)
-                finally:
-                    target.viewport_width, target.viewport_height = saved_viewport
+                self.clear()
+                retval = self._render(camera, target, normal)
                 self._offscreen_result = retval if retval else (None, None)
                 self.render_flags["offscreen"] = False
                 self.render_flags["skip_markers"] = False
--- a/genesis/ext/pyrender/viewer.py
+++ b/genesis/ext/pyrender/viewer.py
@@ -798,11 +798,10 @@ class Viewer(pyglet.window.Window):
                 # Update context, just in case is not already done before
                 self.gs_context.update()
 
-                # ``target`` is the camera's own offscreen FBO, already sized to the camera's configured resolution,
-                # so render at that size and never resize it to ``self._offscreen_viewport_size`` (the interactive
-                # window size). The camera FBO is independent of the window: forcing the window size onto it renders
-                # the camera at the wrong resolution, and unlike the window the FBO is never subject to the OS
-                # content-area clamping that such a resize would otherwise be guarding against.
+                # ``target`` is the camera's own offscreen FBO, already sized to the camera's configured resolution;
+                # render at that size and do not resize it to ``self._offscreen_viewport_size``. That viewport tracks
+                # the interactive window, a surface distinct from this FBO; forcing it onto the FBO would render the
+                # camera at the window's resolution instead of its own.
                 self._offscreen_results = []
                 self.render_flags["offscreen"] = True
                 self.render_flags["skip_markers"] = skip_markers
--- a/genesis/ext/pyrender/viewer.py
+++ b/genesis/ext/pyrender/viewer.py
@@ -798,12 +798,11 @@ class Viewer(pyglet.window.Window):
                 # Update context, just in case is not already done before
                 self.gs_context.update()
 
-                # Render current frame from camera viewpoint. ``target`` is the camera's own
-                # offscreen FBO, already sized to the camera's configured resolution, so render at
-                # that size. Forcing it to ``self._offscreen_viewport_size`` (the interactive
-                # window's size) made offscreen camera renders come out at the viewer resolution
-                # whenever ``show_viewer=True`` (#2927); the camera FBO is independent of the window
-                # so it is never subject to the OS window clamping that override was guarding against.
+                # ``target`` is the camera's own offscreen FBO, already sized to the camera's configured resolution,
+                # so render at that size and never resize it to ``self._offscreen_viewport_size`` (the interactive
+                # window size). The camera FBO is independent of the window: forcing the window size onto it renders
+                # the camera at the wrong resolution, and unlike the window the FBO is never subject to the OS
+                # content-area clamping that such a resize would otherwise be guarding against.
                 self._offscreen_results = []
                 self.render_flags["offscreen"] = True
                 self.render_flags["skip_markers"] = skip_markers


genesis/ext/pyrender/viewer.py
```diff
@@ -798,19 +798,17 @@ def _event_loop_step_offscreen(self):
...
 
-                # Render current frame from camera viewpoint. Force the renderer back to the originally
-                # requested viewport size so the offscreen FBO honors what the caller asked for even when the
-                # window has since been clamped to a smaller content area by the OS (e.g. macOS CI runners).
+                # Render current frame from camera viewpoint. ``target`` is the camera's own
+                # offscreen FBO, already sized to the camera's configured resolution, so render at
+                # that size. Forcing it to ``self._offscreen_viewport_size`` (the interactive
+                # window's size) made offscreen camera renders come out at the viewer resolution
+                # whenever ``show_viewer=True`` (#2927); the camera FBO is independent of the window
+                # so it is never subject to the OS window clamping that override was guarding against.
```
duburcqa: Please avoid wrapping too early and maximise linewidth in code comments (120chars).

duburcqa: Moreover, do not reference issues and PR in code comments.

duburcqa: Do not document the past, only the current behaviour, and why it must be implemented the way it is, to avoid re-introducing bugs in potential later refactoring by mistake.

genesis/ext/pyrender/viewer.py
```diff
@@ -798,19 +798,17 @@ def _event_loop_step_offscreen(self):
...
                 self.render_flags["skip_markers"] = skip_markers
-                saved_viewport = (target.viewport_width, target.viewport_height)
-                target.viewport_width, target.viewport_height = self._offscreen_viewport_size
-                try:
-                    self.clear()
-                    retval = self._render(camera, target, normal)
-                finally:
-                    target.viewport_width, target.viewport_height = saved_viewport
+                self.clear()
+                retval = self._render(camera, target, normal)
```
duburcqa, line 809: Remove a try-finally block that sounds perfectly fine without code comment to explain why it matters is not ok.

HaozheZhang6:
Done — reworded to describe only the current behaviour and why the FBO must keep its own size (so the override is not reintroduced), widened the lines, and dropped the issue reference.

HaozheZhang6:
Done — widened to 120 and reworded to current behaviour + why only, no issue/past refs.

duburcqa:
Thanks for the thorough investigation @HaozheZhang6 - the root-cause diagnosis (per-camera FBO already sized to `camera.res`, override clobbering it with the window size) is correct and was the key insight.

The one issue with this PR as written: `render_offscreen` is also called with the viewer's own renderer (`self._renderer`) -- see the `@required` test `test_render_offscreen_oversized_resolution`, which passes `pyrender_viewer._renderer`. That renderer's viewport follows `on_resize` and gets clamped below the requested size on macOS runners, so it genuinely needs the override. Removing it unconditionally fixes the camera case but regresses that test on clamped displays (this is what the Codex P2 note was pointing at).

I've opened #2954, which scopes the override to `target is self._renderer` (window renderer keeps the clamp workaround, per-camera FBOs render at their own resolution) and adds a regression test for the camera path.

HaozheZhang6:
Ah, you're right — the viewer's own renderer (`test_render_offscreen_oversized_resolution`) genuinely needs the clamp override; I missed that path. #2954's `target is self._renderer` scoping is the correct fix — thanks for picking it up.

duburcqa:
No problem! And remember: no fix without accompanying unit test :)

PR #2938, opened 2026-06-12
[BUG FIX] Restore the carried object's pose when plan_path fails

### LINKED ISSUE #2715: [Bug]: Calling plan_path with with_entity causes unintended object motion
### Bug Description

Calling `plan_path` with the `with_entity` argument modifies the simulation state, even though no trajectory is executed.

Specifically, after calling `plan_path`, objects in the scene (e.g., a grasped cube) unexpectedly change their state (position / velocity), despite no control commands being applied to the robot.

This suggests that the planner internally performs simulation (e.g., for collision checking or trajectory validation) but does not properly restore the original scene state.

As a result, the scene becomes inconsistent after planning, which can also lead to downstream issues such as invalid paths or planner instability.

### Steps to Reproduce

Run this python script.
Observe that:
- The cube position changes after calling plan_path
- The cube may gain velocity or move unexpectedly
- This happens even though no trajectory is executed
```python
import warnings

import genesis as gs
import numpy as np

START_QPOS = np.array(
    [
        1.6877899169921875, -1.7629612684249878, -1.1454275846481323, -2.7444674968719482, -1.2441562414169312,
        1.2233363389968872, 0.5208253860473633, 0.03999991714954376, 0.040000006556510925
    ],
    dtype=np.float32,
)


def _to_numpy(tensor):
    if hasattr(tensor, "detach"):
        return tensor.detach().cpu().numpy()
    return np.asarray(tensor)

def main():

    gs.init(backend=gs.gpu)

    scene = gs.Scene(
        sim_options=gs.options.SimOptions(dt=0.01),
        rigid_options=gs.options.RigidOptions(box_box_detection=True),
        show_viewer=True
    )

    scene.add_entity(
        gs.morphs.Box(
            size=(0.5, 0.8, 0.2),
            pos=(0.5, 0.0, 0.11),
        ),
        surface=gs.surfaces.Plastic(color=(0.35, 0.55, 0.85)),
    )
    cube = scene.add_entity(
        gs.morphs.Box(
            size=(0.05, 0.05, 0.05),
            pos=(0.30, 0.10, 0.35),
        ),
        surface=gs.surfaces.Plastic(color=(0.95, 0.45, 0.15)),
    )
    franka = scene.add_entity(gs.morphs.MJCF(file="xml/franka_emika_panda/panda.xml"))

    scene.build()

    franka.set_dofs_kp(np.array([4500, 4500, 3500, 3500, 2000, 2000, 2000, 100, 100]))
    franka.set_dofs_kv(np.array([450, 450, 350, 350, 200, 200, 200, 10, 10]))
    franka.set_dofs_force_range(
        np.array([-87, -87, -87, -87, -12, -12, -12, -100, -100]),
        np.array([87, 87, 87, 87, 12, 12, 12, 100, 100]),
    )
    hand = franka.get_link("hand")

    for _ in range(10):
        scene.step()

    franka.set_dofs_position(START_QPOS)
    franka.control_dofs_position(START_QPOS)
    for _ in range(20):
        scene.step()
    fingers_dof = np.arange(7, 9)
    franka.control_dofs_position(np.array([0.0, 0.0], dtype=np.float32), fingers_dof)
    for _ in range(100):
        scene.step()

        link= hand,
        pos=np.asarray([0.2999962 , 0.1000018 , 0.34340014], dtype=np.float32),
        quat=np.asarray([1.32832755e-05 ,7.07106803e-01 ,7.07106759e-01 ,4.74470964e-06], dtype=np.float64),
    )
    qpos_goal[-2:] = 0.0
    print(
        f"{qpos_goal=}"
    )
    for _ in range(20):
        scene.step()
    try:
        path, valid = franka.plan_path(
            qpos_goal=qpos_goal,
            num_waypoints=120,
            smooth_path=True,
            max_retry=10,
            ee_link_name="hand",
            with_entity=cube,
            return_valid_mask=True,
        )
        print(f"plan result: shape={tuple(path.shape)}, valid={_to_numpy(valid)}")
    except Exception as exc:
        warnings.warn(f"plan raised: {exc}")

    for _ in range(1000):
        scene.step()


if __name__ == "__main__":
    main()

```

### Expected Behavior

Calling `plan_path` should not modify the simulation state.

Specifically:
- No entity (including `with_entity`) should change position, velocity, or any physical state
- The scene should remain identical before and after planning
- Path planning should be a purely computational process (e.g., using internal copies or temporary states)

In other words, `plan_path` should have no side effects on the simulation unless the resulting trajectory is explicitly executed.

### Screenshots/Videos

https://github.com/user-attachments/assets/9ebfb23b-62b0-4df8-b10f-e0683923a7b8

### Relevant log output

```shell
[Genesis] [13:28:26] [INFO] Running at 58.93 FPS.
[Genesis] [13:28:26] [INFO] Running at 58.90 FPS.
/Users/shiinarikka/anaconda3/envs/TestGenesis/lib/python3.12/site-packages/quadrants/_test_tools/warnings_helper.py:11: UserWarning:

cannot create weak reference to 'tuple' object. Template mapper caching disabled.

[Genesis] [13:28:30] [INFO] RRTConnect planning failed in 1 environments
[Genesis] [13:28:30] [INFO] Planning failed. Retrying for 1 environments...
[Genesis] [13:28:59] [INFO] RRTConnect planning exceeded maximum number of nodes (2000).
[Genesis] [13:29:00] [INFO] Planning failed. Retrying for 1 environments...
plan result: shape=(120, 9), valid=True
2026-04-16 13:29:03.160 python[73996:44786352] error messaging the mach port for IMKCFRunLoopWakeUpReliable
[Genesis] [13:29:03] [INFO] Running at 0.06 FPS.
[Genesis] [13:29:03] [INFO] Running at 48.35 FPS.
```

### Environment

- OS: macos sequoia 15.7.4
- GPU/CPU Apple M1 Pro
- GPU-driver N/A
- CUDA / CUDA-toolkit version N/A

**SHA/hash**:e807698b8aa773fad3a6dfb4556889b251c30924

### Additional Context

_No response_

HaozheZhang6:
## Description

`plan_path(..., with_entity=obj)` mutates the scene whenever a planning attempt fails. Both `RRT.plan` and `RRTConnect.plan` restore the robot's `qpos` in the failure early-return (`if is_invalid.all():`), but only the *success* path also calls `update_object(...)` to put the carried object back. So a failed attempt leaves the object at whatever configuration the planner last explored. Because planning is retried (`max_retry`), a failed attempt also corrupts the relative grasp offset that the next retry captures, so even an eventually-successful plan can return with the object displaced.

The fix mirrors the success path: restore the object via `update_object(...)` in the failure branch of both planners (+2 lines each).

## Related Issue

Resolves #2715

## Motivation and Context

`plan_path` is a query and should not change the scene. Today, calling it with `with_entity` (e.g. a grasped cube) and hitting a failed attempt teleports the object, which corrupts downstream control and makes retried planning unsound.

Issue repro (Franka + cube), headless, recording the cube pose immediately before/after `plan_path(with_entity=cube)` with no trajectory executed:

| | cube pose after `plan_path` | \|Δ\|max |
|---|---|---|
| before fix | `[0.399, -0.021, 1.078]` | 0.844 m |
| after fix  | `[0.300, 0.132, 0.233]` | 1.5e-8 |

Isolated single failed attempt (`max_retry=0`, in-collision goal so the attempt fails deterministically): before → the cube is dragged 1.12 m; after → preserved (~3e-8).

Added a regression test that reproduces this without depending on the planner's RNG (an in-collision goal makes the single attempt fail deterministically), then asserts the attached object's pose is unchanged:

```
pytest tests/test_rigid_physics.py -k plan_path_with_entity_preserves_state_on_failure
```

It passes with the fix and fails without it (the object moves ~1.16 m). `ruff check` / `ruff format` are clean on both files. I ran the new test on the CPU backend; I was not able to run the full suite on my box (its CUDA driver is too old for the installed torch), so I'm relying on CI for the rest.

## Checklist:


HaozheZhang6:
Gentle nudge -- this has been green for ~10 days. It restores the carried object's pose when placing (currently it gets corrupted). Would a maintainer be able to take a look when there's time?

duburcqa:
> Gentle nudge -- this has been green for ~10 days. It restores the carried object's pose when placing (currently it gets corrupted). Would a maintainer be able to take a look when there's time?

I'm postponing the merge of this because I'm considering rewriting 'plan_path' entirely. The current implementation is so bad, it is barely usable. So I'm not convinced it is worth fixing it.

duburcqa:
Superseded by https://github.com/Genesis-Embodied-AI/genesis-world/pull/3109

PR #2944, opened 2026-06-15
[MISC] Use warn_once for legacy URDF parser fallback warning

Kunal-Somani:
## Description
Replace `gs.logger.warning()` with `warn_once()` for the legacy URDF parser fallback warning in `rigid_entity.py`.

The warning at line 768 was firing on every scene build that triggers the fallback path meaning if the same broken asset is loaded across multiple `scene.build()` calls, test loops, or parallel environments, the same message would spam the log repeatedly.

`warn_once()` already exists in `genesis/utils/warnings.py` and is used in `scene.py` for exactly this pattern (deprecated API warnings). This change applies it consistently to the URDF parser fallback.

## Related Issue
No related issue, self-contained improvement spotted during codebase review.

## Motivation and Context
The URDF parser fallback warning is triggered whenever Mujoco's unified parser fails to parse a URDF file and genesis falls back to the legacy parser. In practice this happens for assets with unsupported mesh formats (e.g. `.glb` collision meshes). Loading such an asset repeatedly in one session is common in tests, loops, or multi-scene setups which causes the same warning to print every time, adding noise with no additional information.

The fix is one line: `gs.logger.warning(` to `warn_once(`.

```bash
# Run URDF/MJCF/mesh parsing tests
uv run pytest tests/test_rigid_physics.py -k "urdf or mjcf or parse or mesh" -q
# Result: 44 passed, 1 skipped
# (1 pre-existing failure in test_urdf_align unrelated to this change —
#  physics numerical drift on CPU, present before this PR)

# Run misc, mesh, utils tests
uv run pytest tests/test_misc.py tests/test_mesh.py tests/test_utils.py -q
# Result: 69 passed
```

**To manually verify the warning fires only once:** 
load the same URDF asset that triggers the fallback in two separate `scene.add_entity()` calls within the same Python session and the warning appears once instead of twice.

## Checklist:

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -22,6 +22,7 @@ from genesis.utils import mjcf as mju
 from genesis.utils import terrain as tu
 from genesis.utils import urdf as uu
 from genesis.utils.misc import DeprecationError, broadcast_tensor, qd_to_numpy, qd_to_torch
+from genesis.utils.warnings import warn_once
 from genesis.engine.states.entities import RigidEntityState
 
 from ..base_entity import Entity
@@ -764,7 +765,7 @@ class KinematicEntity(Entity):
                         if is_col:
                             link_g_infos.append(g_info)
             except (ValueError, AssertionError) as e:
-                gs.logger.warning(
+                warn_once(
                     "Falling back to legacy URDF parser. Default values of physics properties may be off:\n"
                     + str(e).replace("\n", " - ")
                 )


duburcqa:
That's fundamentally wrong. We must warn for every URDF falling into this category, otherwise there is no way for a user to terminate whether multiple URDF files are affected. This warning may be painful, I can understand that, but I don't have any good solution to offer at the moment I'm afraid.

Kunal-Somani:
> That's fundamentally wrong. We must warn for every URDF falling into this category, otherwise there is no way for a user to terminate whether multiple URDF files are affected. This warning may be painful, I can understand that, but I don't have any good solution to offer at the moment I'm afraid.

Makes sense. The per-file signal is important for diagnosing which assets are affected. Thanks for the clarification.

PR #2945, opened 2026-06-15
[MISC] Clarify deprecated 'gs view' help text.

Kunal-Somani:
## Description
The `gs view` subparser help text read `[DEPRECATED] Alias of 'launch'.` which tells the user the command is deprecated but not what to use instead.

The runtime warning message already says `"Use 'gs launch' instead"` because this aligns the help text to say the same thing so users get the migration path from both `gs --help` and `gs view --help`.

## Related Issue
No related issue; self-contained UX improvement.

## Motivation and Context
Before:
```bash
view    [DEPRECATED] Alias of 'launch'.
```
After:
```bash
view    [DEPRECATED] Use 'gs launch' instead.
```

```bash
uv run python -m genesis._main --help
# view line now reads: [DEPRECATED] Use 'gs launch' instead.
```

## Checklist:

### DIFF
--- a/genesis/_main.py
+++ b/genesis/_main.py
@@ -172,7 +172,7 @@ def main():
     launch_args.add_argument("-l", "--link_frame", action="store_true", default=False, help="Show link frame")
 
     subparsers.add_parser("launch", parents=[launch_args], help="Visualize a given asset (Mesh/URDF/MJCF/USD)")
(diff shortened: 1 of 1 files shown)


duburcqa (dismissed):

Kunal-Somani:
ah right, should i just remove the migration hint entirely and keep it as [DEPRECATED] with no replacement reference, or using the launch command would instead work better

PR #2946, opened 2026-06-15
[BUG FIX] Fix non-deterministic rasterizer rendering on Nvidia GPU

Kunal-Somani:
## Description

`test_deterministic[RASTERIZER]` was failing on Nvidia GPUs with `assert np.float32(1.3333333e-06) < 1e-15` identical simulation steps producing slightly different render outputs.

The root cause is documented in `jit_render.py:533` with an existing FIXME comment: `GL_BLEND` yields non-deterministic results on Nvidia GPU when alpha-blended fragments are processed in non-deterministic order by the driver.

`GL_BLEND` is enabled for any entity with transparency (alpha < 1). The test scene includes a transparent sphere (`color=(1.0, 1.0, 1.0, 0.5)`), which triggers `GL_BLEND` on every render call, producing ~1e-6 pixel variance between otherwise identical frames on Nvidia hardware.

The test already handles `darwin + Apple Software Renderer` with a relaxed tolerance. This PR extends that pattern to `linux + Nvidia` by:

1. Detecting Nvidia GPU in `OffscreenRenderer` via `glGetString(GL_RENDERER)`  same location where software renderer detection already lives.
2. Exposing `is_nvidia` property on `Visualizer`  mirrors the existing `is_software`.
3. Applying a tolerance of `2e-6` for Nvidia rasterizer in `test_deterministic`  just above the observed `1.33e-6` variance, consistent with how `darwin + software` is already handled.

## Related Issue

No related issue; discovered via `uv run pytest tests/ -m required` failure on Linux + RTX 4060 Laptop GPU.

## Motivation and Context

The `GL_BLEND` non-determinism on Nvidia is a known limitation documented by the existing FIXME comment in `jit_render.py`. The correct long-term fix is to disable `GL_BLEND` for offscreen deterministic renders or implement order-independent transparency. This PR takes the minimal safe approach of relaxing the tolerance to match observed hardware behaviour, consistent with how the macOS software renderer case is already handled in the same test.

Before this PR, on Linux + Nvidia GPU:

```text
uv run pytest tests/test_render.py::test_deterministic -v
FAILED — AssertionError: Per-step renders do not match
assert np.float32(1.3333333e-06) < 1e-15
```

After this PR:

```text
uv run pytest tests/test_render.py::test_deterministic -v
PASSED

uv run pytest tests/test_render.py -q
34 passed, 32 skipped
```

Tested on: Linux, Python 3.11.15, NVIDIA GeForce RTX 4060 Laptop GPU.

## Checklist:

### DIFF
--- a/genesis/ext/pyrender/offscreen.py
+++ b/genesis/ext/pyrender/offscreen.py
@@ -33,6 +33,7 @@ class OffscreenRenderer(object):
         self.point_size = point_size
         self._platform = None
         self._is_software = False
+        self._is_nvidia = False
         self._has_valid_context = False
         self._create(pyopengl_platform)
         self._seg_node_map = seg_node_map
@@ -280,6 +281,7 @@ class OffscreenRenderer(object):
             renderer = glGetString(GL_RENDERER).decode()
             gs.logger.debug(f"Using offscreen rendering OpenGL device: {renderer}")
             self._is_software = any(e in renderer for e in ("llvmpipe", "Apple Software Renderer"))
+            self._is_nvidia = "NVIDIA" in renderer
         except Exception:
             pass
         if self._is_software:
--- a/genesis/vis/visualizer.py
+++ b/genesis/vis/visualizer.py
@@ -276,6 +276,17 @@ class Visualizer(RBC):
         return self._rasterizer._renderer._is_software
 
     @property
+    @gs.assert_built
+    def is_nvidia(self):
+        if self._batch_renderer is not None or self._raytracer is not None:
+            return False
+        if self._viewer is not None:
+            assert self._viewer._pyrender_viewer is not None
+            return self._viewer._pyrender_viewer._is_nvidia
+        assert self._rasterizer is not None and self._rasterizer._renderer is not None
+        return self._rasterizer._renderer._is_nvidia
+
+    @property
     def batch_renderer(self):
         return self._batch_renderer
 
--- a/tests/test_render.py
+++ b/tests/test_render.py
@@ -317,9 +317,14 @@ def test_deterministic(tmp_path, renderer_type, renderer, show_viewer, tol):
     )
     scene.build(n_envs=3, env_spacing=(2.0, 2.0))
 
-    # Apple Software Renderer is fully deterministic: successive captures do not match exactly
+    # Apple Software Renderer is not fully deterministic between successive captures.
     if sys.platform == "darwin" and scene.visualizer.is_software:
         tol_env = tol_step = 1.0
+    elif scene.visualizer.is_nvidia:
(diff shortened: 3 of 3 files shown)


duburcqa:
This unit test is passing just fine on our production GPU. I'm sorry but we are not going to relax the test globally. it must be strict on platforms that supports it. Closing for now. If you manage to find a more targeted gating, let me know and I will reopen.

Kunal-Somani:
> This unit test is passing just fine on our production GPU. I'm sorry but we are not going to relax the test globally. it must be strict on platforms that supports it. Closing for now. If you manage to find a more targeted gating, let me know and I will reopen.

Thanks for the classification. I looked at the existing pattern for apple software renderer at line 371 and was thinking something similar could work here — detect the failure at the assertion point and call pytest.xfail only when the error falls in the gl_blend noise range (like < 1e-5) and on linux only. machines where the test already passes would be untouched

something roughly like

if rgb_step_err >= tol_step and sys.platform == "linux" and rgb_step_err < 1e-5:
    pytest.xfail(f"gl_blend non-determinism on this driver, see jit_render.py fixme")

is this the kind of targeted gating you had in mind or would you prefer a different approach ??

duburcqa:
I don’t think this is a good approach. You need to determine what is causing the renderer to not act deterministically on some platform, so that we can relax the tolerance for these machines specifically. With what you suggest, unit tests that are failing on machines where it should be deterministic will be indiscriminately flagged as xfail instead of hard failing as they are supposed to. This is weakening the guarantees that the production CI currently offers.

Kunal-Somani:
> I don’t think this is a good approach. You need to determine what is causing the renderer to not act deterministically on some platform, so that we can relax the tolerance for these machines specifically. With what you suggest, unit tests that are failing on machines where it should be deterministic will be indiscriminately flagged as xfail instead of hard failing as they are supposed to. This is weakening the guarantees that the production CI currently offers.

got it, so the goal is to identify what exactly is different about my driver that causes gl_blend to be non-deterministic, would querying the opengl driver version or checking for a specific gl extension be the right approach, or is there something else in the rendering pipeline i should be looking at first

duburcqa:
> got it, so the goal is to identify what exactly is different about my driver that causes gl_blend to be non-deterministic, would querying the opengl driver version or checking for a specific gl extension be the right approach, or is there something else in the rendering pipeline i should be looking at first

I'm afraid this quest is VERY challenging and could take weeks of effort... like it is finding a needle in haystack.

Kunal-Somani:
> I'm afraid this quest is VERY challenging and could take weeks of effort... like it is finding a needle in haystack.

fair enough, i'll park this one and look for something more tractable. Thanks for being upfront about the complexity.

PR #2952, opened 2026-06-16
[BUG FIX] Raise on `.morph` for heterogeneous entities instead of returning the primary

### LINKED ISSUE #2750: [Bug]: Size of heterogeneous morphs
### Bug Description

What is the expected way of getting "size" information from a rigid entity with heterogeneous morphs?
Currently setting a rigid body to have two different morphs, a `gs.morphs.Box` and `gs.morphs.Cylinder` for example makes the code order dependent and leads to errors as well.

Let's assume that I define the morphs so that the first is a `Cylinder` and the second is a `Box`, then `obj.morph.size` fails. 
Now if the morph order is reversed, I only get the size of the box.

Even if we define both morphs to be of type `Box`, but with different sizes, we only get the size of the first `Box`.

Question: Is there an example or an expected way of doing this?

### Steps to Reproduce

```python
# code that defined size parameters for box and cylinder

# Define the morphs
morphs_heterogeneous = []
for r, h in zip(cylinder_radii, heights):
    morphs_heterogeneous.append(
        gs.morphs.Cylinder(radius=r, height=h)
    )
for w, h in zip(widths, heights):
    morphs_heterogeneous.append(
        gs.morphs.Box(size=(l, w, h))
    )

object = self.scene.add_entity(
    morph=morphs_heterogeneous,
)
# Then much later in the code
object.morph.size <--- raises AttributeError: 'Cylinder' object has no attribute 'size'
```
But:
```python
# code that defined size parameters for box and cylinder

# Define the morphs
morphs_heterogeneous = []
for w, h in zip(widths, heights):
    morphs_heterogeneous.append(
        gs.morphs.Box(size=(l, w, h))
    )
for r, h in zip(cylinder_radii, heights):
    morphs_heterogeneous.append(
        gs.morphs.Cylinder(radius=r, height=h)
    )

object = self.scene.add_entity(
    morph=morphs_heterogeneous,
)
# Then much later in the code
object.morph.size <--- Returns the specified size of the first box
```

### Expected Behavior

I think the expected behavior here is to get a list of box sizes when all are Box type. And then each primitive should have a similar function let's call it `size` for now that can be used even with all primitives without raising an `AttributeError`.

### Screenshots/Videos

_No response_

### Relevant log output

```shell

```

### Environment

- OS:  Ubuntu 22.04
- GPU/CPU NVIDIA RTX A5500
- GPU-driver version (N/A if no GPU) 582
- CUDA / CUDA-toolkit version (N/A if non-Nvidia) CUDA 12.8

Genesis version: 0.4.5

### Additional Context

_No response_

HaozheZhang6:
## What

`RigidEntity.morph` returned only the first variant for a heterogeneous entity (one built from a list/tuple of morphs), silently hiding the rest — so `type(entity.morph)` just depended on morph order and the entity looked homogeneous. Per #2750, `.morph` now raises for heterogeneous entities and points callers to `.morphs` (all variants) or `.main_morph` (primary). Homogeneous entities are unchanged.

## Root cause

`.morph` (inherited from `base_entity`) returns `self._morph`, the primary variant, with no awareness of `self._morph_heterogeneous`. `RigidEntity` already exposes `.morphs` and `.main_morph` for the heterogeneous case, so `.morph` was the only ambiguous accessor — matching the note on the issue that it should raise here.

## Test

Added `test_morph_accessor_rejects_heterogeneous_entity` (CPU): a homogeneous entity's `.morph` still returns its morph; a heterogeneous entity's `.morph` raises `GenesisException` while `.morphs` / `.main_morph` keep working. Fails before the change, passes after.

Resolves Genesis-Embodied-AI/Genesis#2750

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -1982,6 +1982,17 @@ class KinematicEntity(Entity):
         return len(self._links)
 
     @property
+    def morph(self):
+        """The morph of the entity.
+
+        Raises for heterogeneous entities, where a single morph is ambiguous; use `morphs` for all
+        variants or `main_morph` for the primary one.
+        """
+        if self._enable_heterogeneous:
+            gs.raise_exception("`morph` is ambiguous for heterogeneous entities; use `morphs` or `main_morph`.")
+        return self._morph
+
+    @property
     def main_morph(self):
         """The main morph of the entity (first morph for heterogeneous entities)."""
         return self._morph
--- a/tests/test_rigid_physics.py
+++ b/tests/test_rigid_physics.py
@@ -6989,6 +6989,26 @@ def test_heterogeneous_invalid_material_raises():
         )
 
 
+@pytest.mark.required
+def test_morph_accessor_rejects_heterogeneous_entity():
+    """`.morph` is ambiguous once an entity has heterogeneous variants (#2750): it must raise and steer
+    callers to `.morphs` / `.main_morph` instead of silently returning only the primary morph."""
+    scene = gs.Scene(show_viewer=False)
+    homogeneous = scene.add_entity(gs.morphs.Box(size=(0.04, 0.04, 0.04), pos=(-0.3, 0.0, 0.2)))
+    heterogeneous = scene.add_entity(
+        morph=(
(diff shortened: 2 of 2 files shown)


duburcqa:
Duplicated

PR #2960, opened 2026-06-17
[FEATURE] Add support of heterogeneous entities in batch renderer.

Kashu7100:
## Summary

Heterogeneous entities — objects whose geometry variant only exists in a *subset* of environments — were drawn in **every** environment by the **Madrona batch renderer**. This is the Madrona-path counterpart to #2958, which fixed the same class of bug in the pyrender rasterizer.

### Root cause (and why it differs from #2958)
The pyrender fix in #2958 had two parts: (a) stop *compacting* per-env poses, and (b) add a per-`(primitive, env)` `active_envs` visibility mask.

The Madrona retriever (`GenesisGeomRetriever.retrieve_rigid_state_torch`) already feeds **dense** per-env poses — `vgeoms_state.pos` transposed to `[num_worlds, num_vgeoms]` — and never compacts them, so part (a) is already correct here. The only missing piece is part (b): per-environment **visibility**. Without it, a variant vgeom that exists only in envs *X, Y* is still drawn in all worlds (the Madrona renderer draws every geom in every world).

### Change
`GenesisGeomRetriever.retrieve_rigid_meshes_static` now emits an optional `geom_env_mask` of shape `[num_worlds, num_vgeoms]` (`1` = visible, `0` = hidden), built from each vgeom's existing `active_envs_mask` (`None` ⇒ active in all envs). This is the direct Madrona analog of the `active_envs` mask added to the pyrender rasterizer in #2958.

- Omitted entirely for homogeneous scenes → preserves legacy behavior **and** stays back-compatible with Madrona builds that don't yet accept the kwarg.
- No pose changes needed (the dense per-env poses are already aligned); inactive `(world, geom)` entries are simply masked out and never drawn / never enter the BVH.

```python
if any(vgeom.active_envs_mask is not None for vgeom in vgeoms):
    num_worlds = max(self.rigid_solver.n_envs, 1)
    geom_env_mask = np.ones((num_worlds, self.n_vgeoms), dtype=np.int32)
    for vgeom in vgeoms:
        if vgeom.active_envs_mask is not None:
            geom_env_mask[:, vgeom.idx] = vgeom.active_envs_mask.detach().cpu().numpy()
    args["geom_env_mask"] = geom_env_mask
```

### Dependency
Requires the matching renderer-side change in gs-madrona that consumes `geom_env_mask`: **Genesis-Embodied-AI/gs-madrona#53** (adds the optional `geom_env_mask` kwarg → per-`(world, geom)` renderable disable). With an older gs-madrona, the mask is only emitted for heterogeneous scenes, so homogeneous workloads are unaffected; heterogeneous scenes need both PRs.

### Testing / follow-up
- `py_compile` clean; pre-commit (ruff) passes.
- Functional validation needs a CUDA + Madrona environment. Suggested follow-up: a `test_render`-style test mirroring `test_rasterizer_renders_heterogeneous_entities` from #2958 but driving the batch renderer (assert a variant only covers pixels in its active envs). Happy to add it.

> [!NOTE]
> Draft — opened for review of the approach alongside gs-madrona#53.


Kashu7100:
comparison_old_vs_new

Kashu7100:
https://github.com/user-attachments/assets/5887d08a-f6b4-435e-9e5a-17b98040c033


https://github.com/user-attachments/assets/6a5f06bc-e0c7-45d3-98f9-f318990569d0

Kashu7100:
Now https://github.com/Genesis-Embodied-AI/gs-madrona/pull/53 is landed.

Kashu7100:
gs-madrona 0.0.8 has been released.

duburcqa (changes requested):
Need a snapshot-based unit test. You should also remove 'test_heterogeneous_geom_env_mask_for_batch_renderer', we do not provide any guarantee regarding these internals.

PR #2963, opened 2026-06-17
[FEATURE] Add analytical sphere-sphere collision detection.

Kashu7100:
## Description
Sphere-sphere collision pairs fall through to the iterative MPR narrow-phase path (and GJK+EPA on a cold contact cache), whose normals wander by a per-mille angle on grazing contacts at single precision. A sphere-sphere contact is a closed form, so this routes it through a dedicated analytic function, mirroring the existing sphere-capsule, sphere-box and capsule-capsule specializations: `func_sphere_sphere_contact` computes the contact directly from the two centers and radii, with an arbitrary-direction fallback for coincident centers, wired into all three narrow-phase dispatch sites.

The closed form is also exactly differentiable, where EPA fails to converge on the smoothly-curved sphere-sphere Minkowski boundary: sphere-sphere pairs are now supported with `requires_grad=True` through `func_differentiable_sphere_contact`, registered by the analytic fill kernel alongside the plane-convex contacts, and the build-time rejection of smooth pairs keeps covering sphere-ellipsoid and ellipsoid-ellipsoid only.

```bash
pytest tests/rigid/test_narrowphase.py tests/grad/test_rigid_collision.py
```
`test_sphere_sphere_vs_gjk` compares the analytic contact against GJK on both backends over overlapping, grazing, separated and concentric configurations, keeping the previous EPA smooth-surface regression coverage on its GJK arm. `test_contact_detection_jacobian_matches_fd` gains an overlapping sphere pair, validating the analytic gradients against finite differences, plus a coincident-centers pass asserting finite gradients at the degenerate cusp.

## Checklist:
- [x] I tagged the title correctly (including BUG FIX/FEATURE/BREAKING)


hughperkins:
Nice that you updated tests/test_rigid_physics_analytical_vs_gjk.py 🙌 

Skimming the PR, looks plausible to me. (skimmed very quickly, like 30 seconds, to be clear).

Kashu7100:
Addressed the Codex P2 in 38c9a78 — sphere-sphere is now fully supported under `requires_grad=True`, not just exempt from the guard.

Codex correctly noted the guard made the analytic path unreachable. Just relaxing the guard wasn't enough, though: the differentiable backward (`func_narrow_phase_diff_convex_vs_convex`) reconstructs every contact from a Minkowski triangle, which is degenerate for two spheres — it produced NaN gradients. So the fix is end-to-end:

- **Guard** (`collider.py`): exempt sphere-sphere only; sphere-ellipsoid / ellipsoid-ellipsoid still raise (no analytic path, still tunnel through EPA).
- **Backward** (`diff_gjk.py`): new `func_differentiable_sphere_contact` reconstructs the contact in closed form from the geom centres + radii.
- **Backward kernel** (`narrowphase.py`): both contact loops dispatch sphere-sphere pairs to the analytic reconstruction.
- **Forward** (`narrowphase.py` + `contact.py`): `func_add_diff_contact_input_sphere` registers a self-referential diff-input entry so the backward processes the analytic contact.

Gradients w.r.t. geom positions match finite differences to **~1e-11** (64-bit). `test_diff_smooth_pair_raises` now covers a sphere-ellipsoid pair (still unsupported), and a new `test_diff_sphere_sphere_contact` checks the exact forward contact + FD-validates the gradients. Full `test_grad.py` + `test_rigid_physics_analytical_vs_gjk.py` pass (25 passed).

@codex review

duburcqa:
image

This table is false advertising. sphere-sphere collision detection is NOT drifting but exact (and there is neither penetration depth nor normal error). MPR is biased, but the post-processing step `func_apply_smooth_refinement` that could right after fixes this before it gets stored in the contact state. We have a unit test validating exactly this: `test_no_drift`.

Kashu7100:
that's fair. but the speed ups are real.

duburcqa:
@codex review

duburcqa (approved):

PR #2987, opened 2026-06-24
[FEATURE] Add joint torque sensor (closes #2983)

### LINKED ISSUE #2983: [Feature]: Add joint torque sensor

Some robots have joint torque sensors placed in serie with their actuator. These sensor measure the total torque going through the joint: this can be viewed as being either:
 - the motor's input torque minus all actuator loss (friction, armature inertia, motor dynamics...)
 - the sum of all external forces acting on the joint: gravity, inertia and virtual forces, interaction forces...
 
Currently, this quantity is not readily available: Genesis gives access to
- `get_dofs_control_force`: motor torque output
- `get_dofs_force`: this looks like what I need, but when I compute it for a robot at rest, this quantity is close to zero - so it's at least missing the gravity term.

### Motivation

In order to implement a compliant controller on a collaborative robot (e.g. impedance control), some robot are built with torque sensors at the joint output. Feedback is then performed on this sensor to regulate the robot's torque. To simulate this controller, I need the same output available in simulation.

### Potential Benefit

Ability to simulate robots with joint torque sensors

- [  ] A way to obtain the desired force- through a helper function  on a `RigidEntity` or a Sensor
- [  ] Unit test showing this is the quantity we want: maybe a robotic arm with and without interaction force to check both cases?

### Additional information

_No response_

Kashu7100:
## Summary

Adds a **joint torque sensor** — the torque transmitted through each joint — closing #2983. Real collaborative robots place a torque sensor in series with the actuator and close an impedance/compliance loop on it; this exposes the same signal in simulation.

The issue reporter found `get_dofs_force` reads ~0 for a stationary arm. That's expected: it returns the *net* generalized force (`≈ M·q̈`), which is zero at rest and so omits the gravity-support torque. The transmitted torque is the **actuation-side** force `qf_applied + qf_passive` (actuator/applied command + passive damping/friction & spring), which retains the gravity term when the arm holds against gravity.

## What's added

- **`RigidSolver.get_dofs_actuation_force`** / **`RigidEntity.get_dofs_actuation_force`** — expose `qf_applied + qf_passive`. Docstrings contrast it with `get_dofs_force` (net force) and `get_dofs_control_force` (pre-loss motor command), and note the gravity-compensation caveat.
- **`gs.sensors.JointTorque`** options + **`JointTorqueSensor`** — plug into the existing sensor framework (`scene.add_sensor`, recorders), measuring a selected set of DOFs (defaults to all of the entity's DOFs).
- **Tests** (`tests/test_sensors.py`, `n_envs ∈ {0, 2}`) on a gravity-loaded Franka:
  - at rest the reading equals the helper and is non-zero, while `get_dofs_force` stays ~0;
  - feeding the reading back open-loop in force mode holds the arm (gravity-compensation feed-forward);
  - an externally commanded joint torque is read back.

## Notes / scope

- This is the **closed-form** definition (`qf_applied + qf_passive`). A Cartesian contact wrench enters via the constraint force, so it's reflected only once the controller reacts. An exact-fidelity variant (RNE-post-constraint projection of the link interaction force, à la MuJoCo's joint torque sensor / `cfrc_int`) is a natural follow-up.
- Genesis gravity compensation removes gravity from the entity's dynamics, so with that flag on the reading carries no gravity term (the simulated robot has no gravity load) — documented in the docstrings.

Closes #2983

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -3741,6 +3741,44 @@ class RigidEntity(KinematicEntity):
         dofs_idx = self._get_global_idx(dofs_idx_local, self.n_dofs, self._dof_start, unsafe=True)
         return self._solver.get_dofs_force(dofs_idx, envs_idx)
 
+    @gs.assert_built
+    def get_dofs_actuation_force(self, dofs_idx_local=None, envs_idx=None):
+        """
+        Get the torque transmitted through the entity's joints, i.e. the joint torque sensor measurement.
+
+        This is the actuation-side force ``qf_applied + qf_passive``: the actuator/applied generalized
+        force (motor torque from the controller, plus any `set_dofs_force` contribution) combined with the
+        passive joint force (damping/friction and spring). It is the total torque a sensor mounted in
+        series with the actuator would measure.
+
+        Note
+        ----
+        This differs from the two other force getters:
+
+        - `get_dofs_force` returns the *net* force on each dof (mass times acceleration), which converges
+          to zero for a robot at rest and therefore does not capture the gravity-support torque.
+        - `get_dofs_control_force` returns the raw motor command before passive/actuator losses.
+
+        At rest with a controller holding position, this getter reports the gravity-support torque on each
+        joint. If gravity compensation is enabled for the entity (which removes gravity from its dynamics),
+        no gravity term appears here, since the simulated robot carries no gravity load; keep gravity
+        enabled and let the controller compensate to read a real gravity load.
+
+        Parameters
+        ----------
+        dofs_idx_local : None | array_like, optional
+            The indices of the dofs to get. If None, all dofs will be returned. Note that here this uses the local `q_idx`, not the scene-level one. Defaults to None.
+        envs_idx : None | array_like, optional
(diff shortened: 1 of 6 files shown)


matthieuvigne:
Thank  you for looking at `https://github.com/Genesis-Embodied-AI/genesis-world/issues/2983`. However, the equation you propose does not yield the quantity I'm looking for ; maybe my issue wasn't clear.

I think the simplest way to compute a torque sensor's output is to view it from the motor side. Between the motor and the sensor, there is a gearbox - and this gearbox introduces torque loss in two fashion. Through its armature inertia, and through friction. So the torque sensor output can be computed as (control force  - friction - armature * ddq). See for instance eq. 1 of 
https://hal.science/hal-02889421v1/file/Actuator_Model_Identification_and_Differential_Dynamic_Programming_for_a_TALOS_Humanoid_Robot.pdf

I actually already started working on it when I open the issue; I've opened another PR https://github.com/Genesis-Embodied-AI/genesis-world/pull/2989, and I think this one can be closed. I compared it with this implementation and the results are indeed different.

Kashu7100:
Thank you for the comment, and yes my PR does not take into account the loss in gearbox. I think yours is better.

PR #3002, opened 2026-07-02
Perf/fix decomposed path v100 reenable

zhihuidu-amd:
## Summary

Re-enable the 7-kernel decomposed solver path that was disabled in v1.0.0
due to a one-line positional-arg mismatch.

## Root Cause

`_decomposed_amdgpu_is_compatible` used `args[5]` for `static_rigid_sim_config`,
but v1.0.0 inserted `dofs_info` as the 2nd positional arg, shifting it to `args[6]`.

## Fix

Update positional fallback from `args[5]` to `args[6]` and remove early `return False`.
Sub-kernels need no changes — they already pass all solver device functions by keyword.

## Expected Impact

J^T@efc_force dispatched as 2D ndrange(n_dofs, n_envs) producing ~5504 WGs
vs the tiled_wc inline loop. Fully saturates MI325X SIMDs for the dominant
constraint-update work.

## Validation

Awaiting CI benchmark results on MI325X.


zhihuidu-amd:
wrong pr

PR #3022, opened 2026-07-10
[AMDGPU] Perf: cache repeated Jaref lookup in func_update_constraint_batch

gayathiri-venkataraman:
## Summary

`func_update_constraint_batch` reads `constraint_state.Jaref[i_c, i_b]` several times
within the same iteration. This hoists the element into a local variable so it is loaded
once and reused, removing redundant global-memory reads in a hot constraint-solver loop.
This is pure common-subexpression elimination: the computed values are identical and only
the number of memory loads is reduced. No correctness change. Backend-agnostic (benefits
all backends), motivated by AMD/ROCm performance profiling on MI300X.

Verified with the rigid-physics `required` test set, run with and without this change:
- AMD MI300X (VF slice, `-n 8 --backend gpu`): identical results — 281 passed / 5 xfailed,
  a byte-identical pass/fail set versus baseline (zero behavioral delta).
- AMD MI325X (baremetal, Docker/theRock, `-n 8`, gpu-only): 0 GPU failures.

Ported from AMD ROCm/Genesis integration work.

### DIFF
--- a/genesis/engine/solvers/rigid/constraint/solver.py
+++ b/genesis/engine/solvers/rigid/constraint/solver.py
@@ -4416,8 +4416,12 @@ def func_update_constraint_batch(
     cost_i = gs.qd_float(0.0)
     gauss_i = gs.qd_float(0.0)
 
-    # Beware 'active' does not refer to whether a constraint is active, but rather whether its quadratic cost is active
+    # Beware 'active' does not refer to whether a constraint is active, but rather whether its quadratic cost is active.
+    # `Jaref[i_c, i_b]` is read up to 6 times in the original loop body; the DSL cannot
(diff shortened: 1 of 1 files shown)


duburcqa:
Any reproducible benchmark with accompanying statistics to assess the benefit of this PR? We do not merge on principle, it has to be grounded with real numbers.

gayathiri-venkataraman:
Withdrawing this PR. On the AMD RL benchmark (MI325X, 8192 envs) the change shows no measurable throughput improvement — the cached lookup sits in a code path the AMDGPU cooperative solver kernel bypasses below n_envs=16384, so it can't be justified as a standalone performance change. The optimization is being folded into the coordinated solver-stack upstreaming instead. The change itself is a safe CSE (no numeric change) and passed the rigid-physics test set with no regression.

PR #3036, opened 2026-07-14
[BUG FIX] Fix MJCF parsing of 2D textures.

### LINKED ISSUE #3027: [Bug]: MJCF importer does not match MuJoCo 2D texture mapping
### Bug Description

The MJCF importer does not reproduce MuJoCo's texture-coordinate behavior for textures with `type="2d"`.

Genesis currently assigns normalized UVs to planes and boxes and then applies:

```python
uv *= mj_mat.texrepeat
```

This differs from MuJoCo in several ways:

* `texuniform` is not used.
* Generated coordinates do not include MuJoCo's object-linear phase.
* Texture repetition does not correctly account for geometry render size or Genesis morph scale.
* Meshes without authored texture coordinates do not receive equivalent generated coordinates.
* Authored mesh texture coordinates are incorrectly multiplied by `texrepeat`.
* The vertical-coordinate conversion modifies a view of the shared `mj.mesh_texcoord` array.

MuJoCo uses generated object-linear coordinates when explicit mesh texture coordinates are unavailable:

```text
s =  0.5 * effective_repeat_x * x - 0.5
t = -0.5 * effective_repeat_y * y - 0.5
```

The effective repetition depends on `texrepeat`, `texuniform`, geometry render size, and whether the geometry is already pre-scaled.

When authored mesh texture coordinates are available, MuJoCo gives them priority and does not apply the generated-coordinate path.

Genesis currently reads authored coordinates using:

```python
uv = mj.mesh_texcoord[tex_vert_start:tex_vert_end]
uv[:, 1] = 1 - uv[:, 1]
```

Because this slice is a NumPy view, the conversion modifies the shared MuJoCo model. Referencing the same mesh more than once can therefore produce inconsistent coordinates.

### Steps to Reproduce

Save the following as `texture_mapping.xml`:

```xml
<mujoco>
  <asset>
    <texture
      name="checker"
      type="2d"
      builtin="checker"
      width="256"
      height="256"
      rgb1="0.1 0.2 0.8"
      rgb2="0.9 0.8 0.1"/>

    <material
      name="floor_material"
      texture="checker"
      texrepeat="3 3"
      texuniform="true"/>
  </asset>

  <worldbody>
    <geom
      name="floor"
      type="plane"
      size="3 3 0.1"
      material="floor_material"/>
  </worldbody>
</mujoco>
```

Load the same file in MuJoCo 3.6.0 and Genesis and compare the floor texture.

Genesis generates object-wide UVs in the `[0, 3]` range. MuJoCo's generated mapping for this finite plane produces a coordinate span of 9 in each direction, so the texture density differs.

The phase difference can be made more visible with asymmetric or non-integer repetition, for example:

```xml
texrepeat="2.5 3.5"
```

### Expected Behavior

Genesis should reproduce MuJoCo's two 2D texture-coordinate paths:

* Generated coordinates should use MuJoCo-equivalent object-linear phase and scaling.
* `texrepeat` and `texuniform` should follow MuJoCo semantics.
* Geometry render size and Genesis morph scale should be handled consistently.
* Finite and infinite planes should preserve equivalent texture alignment.
* Meshes without authored coordinates should receive generated coordinates.
* Authored mesh coordinates should take priority and should not be multiplied by `texrepeat`.
* Coordinate-convention conversion should not modify the shared `mujoco.MjModel`.

### Environment

* OS: Linux
* MuJoCo version: 3.6.0
* Affected component: MJCF importer
* Affected path: `genesis/utils/mjcf.py`
* The generated-coordinate mismatch occurs during import and is not specific to one Genesis renderer.

### Release Version or Commit ID

```text
958fe1ec96e59f1c428b41dffab05137a9434527
```

The relevant implementation is also present on `main`:

```text
https://github.com/Genesis-Embodied-AI/genesis-world/blob/main/genesis/utils/mjcf.py
```

### Additional Context

The mismatch changes image observations on floors, walls, primitives, and meshes.

References:

* https://mujoco.readthedocs.io/en/stable/XMLreference.html#asset-mesh
* https://mujoco.readthedocs.io/en/stable/XMLreference.html#asset-material
* https://github.com/google-deepmind/mujoco/blob/main/src/render/classic/render_gl3.c

thanyu-hub:
## Description

This PR makes the MJCF importer's 2D texture-coordinate behavior consistent with MuJoCo.

- Generates MuJoCo-compatible object-linear coordinates for primitives and meshes without authored texture coordinates, respecting `texrepeat`, `texuniform`, compiled render dimensions, finite and infinite plane alignment, and Genesis morph scale when `texuniform="true"`.
- Preserves authored mesh coordinates as a separate path: they take priority, are not multiplied by material `texrepeat`, and are copied before vertical-coordinate conversion to avoid modifying the shared `mujoco.MjModel`.
- Adds regression coverage for generated, mesh-fitted, and authored-coordinate paths, including repeated references to the same mesh.

## Related Issue

Resolves https://github.com/Genesis-Embodied-AI/genesis-world/issues/3027

## Motivation and Context

Genesis previously generated normalized UVs for some MJCF geometry types and multiplied all available coordinates by the material `texrepeat`.

This differs from MuJoCo's generated object-linear mapping and its separate authored-coordinate path. As a result, the same MJCF model and texture assets could produce different texture density, phase, and orientation in MuJoCo and Genesis.

The shared `mj.mesh_texcoord` array could also be modified in place while converting coordinate conventions, making repeated references to an authored-UV mesh order-dependent.

These differences affect visual compatibility and image observations used by vision-based policies and simulation benchmarks.

The following checks passed:

```bash
.venv/bin/pre-commit run --files genesis/utils/mjcf.py tests/test_mesh.py
.venv/bin/pytest -n 0 tests/test_mesh.py
```

The mesh test suite completed with `26 passed`.

The required suite was also run:

```bash
.venv/bin/pytest -v --forked -m required ./tests
```

It reported six failures, none specific to this change. Five also reproduce on the base commit `958fe1ec96e59f1c428b41dffab05137a9434527`; the remaining parallel BatchRenderer process crash passes in isolation on both branches.

## Screenshots

MuJoCo 3.6.0 is shown on the left, Genesis before this fix in the center, and Genesis after this fix on the right.

genesis-issue-3027-comparison

## Checklist


genesis/utils/mjcf.py
```diff
@@ -598,8 +602,41 @@ def parse_geom(mj, i_g, scale, surface, xml_path):
...
+                mujoco.mjtGeom.mjGEOM_CAPSULE,
+                mujoco.mjtGeom.mjGEOM_CYLINDER,
+            ):
+                render_size[1] = render_size[0]
+            is_size_finite = render_size > 0
+
+            object_xy = mesh_params["vertices"][:, :2].copy()
+            if mj_geom.type[0] != mujoco.mjtGeom.mjGEOM_MESH:
+                # Finite primitives use unit coordinates, while infinite plane dimensions use spatial coordinates
+                np.divide(object_xy, render_size, out=object_xy, where=is_size_finite)
```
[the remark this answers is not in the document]
thanyu-hub: For a mesh without authored texture coordinates and with `texuniform="false"`, the `dataid >= 0` branch immediately below already divides `repeat` by the compiled `render_size`, matching [MuJoCo's pre-scaled-geom path](https://github.com/google-deepmind/mujoco/blob/3.6.0/src/render/classic/render_gl3.c#L121-L149). 
For a mesh spanning `[-2, 2]` with a size of `2`, this gives an effective repeat of `1 / 2` and a UV span of exactly `1`. Normalizing `object_xy` as well would apply the size correction twice, so no change is needed.

genesis/utils/mjcf.py
```diff
@@ -573,8 +576,9 @@ def parse_geom(mj, i_g, scale, surface, xml_path):
         index_faces = [faces.ravel(), norm_faces.ravel()]
         if tex_vert_start != -1:  # -1 means no texcoord
             tex_faces = mj.mesh_facetexcoord[face_start:face_end]
-            uv = mj.mesh_texcoord[tex_vert_start:tex_vert_end]
+            uv = mj.mesh_texcoord[tex_vert_start:tex_vert_end].copy()
```
duburcqa: Avoid doing a copy and rather stop mutating uv data in place. Add a comment to explain that we must be careful not to mutate mujoco data.

tests/test_mesh.py
```diff
@@ -942,6 +943,233 @@ def test_mjcf_parse_mesh_normals(normals_mjcf):
...
+    )
+    ET.SubElement(
+        worldbody,
+        "geom",
+        name="mesh_explicit_b",
+        type="mesh",
+        mesh="explicit_mesh",
+        material="uniform",
+    )
+
```
duburcqa: Keep arguments on 1 line whenever it fits.

tests/test_mesh.py
```diff
@@ -942,6 +943,233 @@ def test_mjcf_parse_mesh_normals(normals_mjcf):
...
+    texcoord_end = texcoord_start + model.mesh_texcoordnum[explicit_mesh.id]
+    explicit_uvs = model.mesh_texcoord[texcoord_start:texcoord_end]
+
+    for vgeom in entity.vgeoms:
+        geom = model.geom(vgeom.metadata["name"])
+        if vgeom.metadata["name"].startswith("mesh_explicit"):
+            assert_allclose(
+                np.unique(vgeom.uvs, axis=0),
+                np.unique(explicit_uvs, axis=0),
+                tol=1e-6,
```
duburcqa: why 1e-6 and not `gs.EPS` ?

tests/test_mesh.py
```diff
@@ -942,6 +943,233 @@ def test_mjcf_parse_mesh_normals(normals_mjcf):
...
+            )
```
duburcqa: Keep all arguments on 1 line.

tests/test_mesh.py
```diff
@@ -942,6 +943,233 @@ def test_mjcf_parse_mesh_normals(normals_mjcf):
...
+        repeat = material.texrepeat.copy()
+        if geom.dataid[0] >= 0:
+            np.divide(repeat, render_size, out=repeat, where=is_size_finite)
+        if material.texuniform[0]:
+            repeat *= np.where(is_size_finite, render_size, 1.0) * SCALE
+
+        expected_uvs = np.empty_like(object_xy)
+        expected_uvs[:, 0] = 0.5 * repeat[0] * object_xy[:, 0] - 0.5
+        expected_uvs[:, 1] = -0.5 * repeat[1] * object_xy[:, 1] - 0.5
+        assert_allclose(vgeom.uvs, expected_uvs, tol=1e-6)
```
duburcqa: Same. Why `1e-6`?

tests/test_mesh.py
```diff
@@ -942,6 +943,233 @@ def test_mjcf_parse_mesh_normals(normals_mjcf):
...
+    scene = gs.Scene()
+    entity = scene.add_entity(
+        gs.morphs.MJCF(
+            file=textured_mjcf,
+            scale=SCALE,
+        ),
+    )
+    scene.build()
+
+    model = mujoco.MjModel.from_xml_string(textured_mjcf)
```
duburcqa: You are not supposed to call 'mujoco.MjModel' manually. You should rather use `gs_sim, mj_sim` fixtures.

genesis/utils/mjcf.py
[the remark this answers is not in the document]
duburcqa: Add a comment in the code to explain this. Please keep it compact (max 3 lines should be fine). If codex got confused, it is apparently not clear enough in the code.

duburcqa:
Sorry about the annoying rebase conflict. I merged a complete refactoring of the unit test infrastructure.

tests/test_mesh.py
thanyu-hub: Thanks for the review :) Before migrating this test to `gs_sim` and `mj_sim`, I’d like to confirm how to preserve its `scale=2.0` coverage. The current `gs_sim` path constructs `gs.morphs.MJCF` without passing `scale` ([`tests/utils.py:677-685`](https://github.com/Genesis-Embodied-AI/genesis-world/blob/763e43e529b53ca0f3f40e233455079f9e570d14/tests/utils.py#L677-L685)), so it uses the inherited default of `1.0` ([`genesis/options/morphs.py:608`](https://github.com/Genesis-Embodied-AI/genesis-world/blob/763e43e529b53ca0f3f40e233455079f9e570d14/genesis/options/morphs.py#L608)). Migrating as-is would therefore lose the check that `texuniform` incorporates the Genesis morph scale. Should I thread the existing `scale` fixture through `gs_sim` and `build_genesis_sim`, or is there another fixture pattern I should follow?

duburcqa: I see. Ok let’s forget about it then. But I think it is unnecessary to use mujoco as a baseline at all. If you MJCF is careful crafted, you should know what you expect without ever needing to rely on mujoco parser to validate it.

genesis/utils/mjcf.py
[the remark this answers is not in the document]
thanyu-hub: Added compact comments clarifying that true meshes keep spatial object coordinates and apply the compiled mesh size correction through `repeat / render_size`. The test now covers scales 1 and 3.

genesis/utils/mjcf.py
thanyu-hub: The MuJoCo's view is left unchanged, and UV transformations create new arrays. I also added a compact ownership comment.

tests/test_mesh.py
thanyu-hub: The test now uses analytical UV expectations for a carefully crafted MJCF, without constructing a separate MuJoCo model as an oracle.

duburcqa (dismissed):

PR #3037, opened 2026-07-14
[MISC] Restructure test suite into per-component folders with per-capability files.

duburcqa:
## Description

The test suite (820+ tests) was packed into single files per component, some of which had grown far beyond maintainability (`test_rigid_physics.py` alone was 8500+ lines). This PR restructures `tests/` into per-component folders holding one file per capability, with shared model fixtures in folder-level `conftest.py`:

- `rigid/` - dynamics, mujoco_parity, control, kinematics, collision, collision_nonconvex, narrowphase, terrain, friction, constraints, asset_loading, api, heterogeneous, islands, sparse
- `deformable/` - fem, sph, pbd, deformable (cross-solver)
- `ipc/` - api, rigid, deformable (+ shared scene-inspection helpers in `utils.py`)
- `coupling/` - hybrid (classical couplers: SAP, hydroelastic, legacy)
- `sensors/` - api, imu, joint_torque, contact, raycaster, temperature, tactile, camera, recorders
- `rendering/` - offscreen, debug_draw, dynamic_meshes, madrona, viewer, imgui_overlay (+ shared renderer fixtures in `conftest.py`)
- `loaders/` - mesh, usd
- `core/` - quadrants, backend_switching, grad, utils, misc, bvh
- `integration/` - end-to-end scenarios
- `benchmarks/` - performance benchmarks (`-m benchmarks`)

`tests/test_examples.py` stays flat at the root since it is a runner for `examples/`, not a component test. Suite-wide fixtures and helpers remain in `tests/conftest.py` / `tests/utils.py`.

The snapshots dataset repository now mirrors the tests tree (`<component>/__snapshots__/<module>/`), so snapshot lookup is derived from the snapshot directory path relative to the tests root and no longer requires suite-wide unique module basenames. The pinned revision is bumped accordingly.

## Verification

- Collection parity: 918/918 test IDs (names and parametrizations) identical to `main` across all three marker passes (default, `-m benchmarks`, `-m examples`).
- Smoke runs across the new folders pass, including the cross-package fixture chain (`model_name` -> `rigid/conftest.py` -> `gs_sim`/`mj_sim`) and the subprocess-spawning tests whose module paths changed.
- Snapshot download validated end-to-end against the reorganized dataset repository (local snapshot deleted, re-fetched at the new pinned revision, test passes).


Milotrince (dismissed):
this was much needed convention change, thank you!! It does mean all PRs will have to rebased, but that needs to happen anyway, so no problem

PR #3052, opened 2026-07-16
[FEATURE] Add MJCF skip_root_plane option for worldbody floors

### LINKED ISSUE #2782: [Feature]: Add MJCF option to skip loading ground plane.
### Bug Description

in the latest main when spawning using MJCF morphs, i think this issue was fixed previously.
Because of this some robots start with their links penetrating the ground if not dealt with properly this will causes NaNs in the simulation

### Steps to Reproduce

spawn the 'ant' xml asset from teh genesis assets 

### Expected Behavior

the ground plane be removed

### Screenshots/Videos

Image

latest main

### Additional Context

_No response_

Fstarnb:
## Description
Adds an opt-in `skip_root_plane` flag on `gs.morphs.MJCF` that skips plane geoms attached to the MJCF worldbody (body id 0) when loading an entity.

## Related Issue
Resolves Genesis-Embodied-AI/Genesis#2782

## Motivation and Context
Many MJCF robot models (e.g. `xml/ant.xml`) embed a ground plane in the worldbody. Loading them as a single entity into a scene that already has a floor, or placing the robot such that links penetrate that embedded plane, is awkward. Maintainer guidance on the issue was to keep MJCF planes by default and offer an explicit morph option rather than silently stripping them.

Default remains `False` so existing behavior is unchanged.

- Unit test `tests/core/test_misc.py::test_mjcf_skip_root_plane` loads `xml/ant.xml` with and without `skip_root_plane=True` and asserts exactly one fewer plane geom when the flag is set.
- Filter logic was also checked against MuJoCo's geom body/type arrays for ant.xml (one root plane at body id 0).

Example:

```python
import genesis as gs

# drop ant's embedded floor so only the scene plane is used
ant = scene.add_entity(gs.morphs.MJCF(file="xml/ant.xml", skip_root_plane=True))
scene.build()
```

## Checklist:

### DIFF
--- a/genesis/options/morphs.py
+++ b/genesis/options/morphs.py
@@ -991,12 +991,18 @@ class MJCF(FileMorph):
     default_armature : float, optional
         Default rotor inertia of the actuators. In practice it is applied to all joints regardless of whether they are
         actuated. None to disable. Default to 0.1.
+    skip_root_plane : bool, optional
+        If True, skip plane geoms attached to the MJCF worldbody (body id 0) when loading the entity. Many MJCF robot
+        models ship with an embedded ground plane that is inconvenient when the scene already has its own floor, or
+        when the robot is placed so that its links would otherwise penetrate that plane. This is opt-in: planes that
+        are part of the MJCF file are kept by default. Defaults to False.
     """
 
     pos: Vec3FType | None = None
     quat: UnitVec4FType | None = None
     requires_jac_and_IK: StrictBool = True
     default_armature: float | None = Field(default=0.1, ge=0)
+    skip_root_plane: StrictBool = False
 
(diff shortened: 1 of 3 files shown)


duburcqa:
Thanks for this, and sorry for the duplicate effort! This overlaps with #3050, which resolves the same issue (#2782) and is a bit further along, so we're consolidating on that one. Closing this in its favor.

For reference, the direction we settled on there: a positive `is_ground_plane_included` flag (matching our boolean naming convention), the worldbody planes stripped from the source XML before compilation (so planes authored under child bodies survive even with `fusestatic`), and the test living in `tests/rigid/test_asset_loading.py`.

Really appreciate the contribution, and hope to see more PRs from you!

PR #3057, opened 2026-07-16
[MISC] Stop requiring 64bit precision in tactile sensors.

Milotrince:
## Summary

Follow-up to #3058 (which fixed the tactile sensors on the Metal backend by running the bonded-layer transfer's float64 mode solve on CPU and moving the result to the device). This replaces that ill-conditioned per-mode 4x4 linear solve with its exact closed form, evaluated natively in the simulation dtype:

```
S(q) = 2 q^2 / (sinh(2q) - 2q)
```

Since the closed form needs no float64, the CPU round-trip #3058 added is no longer necessary: the wavenumbers and the transfer are built directly on-device in `tc_float`.

`sinh(2q) - 2q` cancels catastrophically at small `q`, so a short Taylor series (5 terms, by recurrence) is used below the crossover and the direct form above it.

## Why this is better

Validated against a 50-digit (mpmath) ground truth:

| q | exact | old float64 solve | this closed form (float32) |
|---|---|---|---|
| 1e-3 | 1500.0 | 9e-9 | 1.3e-7 |
| 1.0 | 1.229 | 5e-16 | 3e-9 |
| 15.0 | 8.4e-11 | 2.3e-6 | 5e-8 |
| 17.0 | 2.0e-12 | 1.3e-5 | 3e-8 |

- Uniformly ~1e-7 in float32, and **more accurate than the float64 solve at large q**, where the solve's matrix entries span a huge dynamic range.
- Drops the CPU round-trip and the batched `torch.linalg.solve` from the kernel build.
- Keeps `@torch.jit.script` (the caller `_precompute_hydroshear_dilate_kernel_fft` is scripted, so the transfer must be too).

## Testing

`tests/sensors/test_tactile.py` -- 19 passed (CPU backend), including the four `elastomer` tests that exercise this bonded-layer path.

### DIFF
--- a/genesis/engine/sensors/point_cloud_tactile.py
+++ b/genesis/engine/sensors/point_cloud_tactile.py
@@ -959,54 +959,35 @@ def _collect_collision_geom_idx(solver, track_link_idx: np.ndarray) -> tuple[tor
     return torch.tensor(geom_idx, dtype=gs.tc_int, device=gs.device), torch.stack(active_masks, dim=0)
 
 
-# [numerical] Peak of the Gaussian first-moment kernel r * exp(-lambda r^2): value e^{-1/2} / sqrt(2 lambda)
-# at r = 1/sqrt(2 lambda). Peak-normalizes the local kernel in the compressibility blend.
+# Peak of r * exp(-lambda r^2) is this / sqrt(2 lambda); peak-normalizes the local kernel in the blend.
 _INV_SQRT_E = math.exp(-0.5)
 
-# [numerical] Clamp range for q = |k| * h, the dimensionless wavenumber fed to _bonded_layer_transfer's S(q).
-# Q_MIN is set by the float64 conditioning of that 4x4 mode solve: cond(M) ~ 4.5/q^3 (lubrication limit), so
-# q = 1e-3 gives cond ~ 4.5e9 -- the smallest q still solved to ~6 digits in double. Clamping there costs no
-# accuracy: S has already reached its 1.5/q asymptote, and real FFT grids never get this low anyway (smallest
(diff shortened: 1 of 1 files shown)


duburcqa (dismissed):

tests/sensors/test_tactile.py
```diff
@@ -903,6 +903,19 @@ def test_gaussian_crosstalk_kernel_helper():
...
+    # Closed-form S(q) = 2q^2/(sinh(2q)-2q) evaluated in float32 (Taylor series near q->0 where the direct form
+    # cancels, direct otherwise) must match the analytic value across the wavenumber range, into the 1.5/q asymptote.
+    # Local import: point_cloud_tactile reads gs.qd_float at module load, so it cannot be imported before gs.init.
+    from genesis.engine.sensors.point_cloud_tactile import _bonded_layer_transfer
+
+    q = torch.logspace(-3, np.log10(5.0), 200, dtype=torch.float32)
+    s = _bonded_layer_transfer(q)
+    q64 = q.to(torch.float64)
+    s_ref = 2.0 * q64 * q64 / (torch.sinh(2.0 * q64) - 2.0 * q64)
+    assert_allclose(s, s_ref, rtol=1e-6)
```
duburcqa: What is this tolerance? You should probably use `tol` fixture?

Milotrince: it's checking specifically for tol w.r.t float32's comfortable range. but i dont even think this test is necessary

duburcqa: Yeah, so that’s basically what our `tol` fixture do then!

duburcqa (approved):

PR #3061, opened 2026-07-17
[BUG FIX] Fix fastcache support (cont'd).

hughperkins:
The earlier purity PR fixed the host-side copy of the 1/sqrt(e) constant but not the kernel-side one, and left BVH_STACK_SIZE captured in a fastcache kernel. Both are [PURE.VIOLATION] warnings that Genesis CI escalates to errors.

- _INV_SQRT_E in _func_elastomer_direct_dilate_contribution (reached by the fastcache kernel _kernel_elastomer_dilate_accumulate): inline math.exp(-0.5) instead of capturing the global. The math module is an allowed fastcache capture, so this is exempt and cache-safe. This is the site the tactile sensor tests were failing on.

- BVH_STACK_SIZE in the fastcache kernel _kernel_elastomer_surface_state_bvh: pass it as a qd.template() param (bvh_stack_size), matching the existing qd.template() usage in these sensor kernels. Template primitives are baked and keyed into the fastcache key, so distinct stack sizes recompile correctly. BVH_STACK_SIZE stays the uppercase source of truth, captured only in host code at the call site. This one was latent: the failing tests errored in the dilate path before reaching this kernel.

### DIFF
--- a/genesis/engine/sensors/point_cloud_tactile.py
+++ b/genesis/engine/sensors/point_cloud_tactile.py
@@ -916,7 +916,9 @@ def _func_elastomer_direct_dilate_contribution(
         if compressibility <= 0.0:  # pure incompressible r_hat / r
             w = depth * inv
         else:  # blend, each kernel peak-normalized (see the FFT builder for the closed-form peaks)
-            norm_g = gs.qd_float(qd.static(_INV_SQRT_E)) / qd.sqrt(gs.qd_float(2.0) * lam)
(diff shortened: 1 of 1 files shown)


hughperkins:
Running Genesis unit tests against quadrants branch that removes the qd.static loophole.

On main:

Screenshot 2026-07-17 at 11 08 43

On this branch:

Screenshot 2026-07-17 at 11 08 34

duburcqa (approved):

hughperkins:
🙌

PR #2904, opened 2026-06-05
[FEATURE] Decouple visual from physics geometry for FEM entities.

ACMLCZH:
## Description

FEM entities now expose their morph sub-meshes as visual geoms (`fem_entity.vgeoms`, mirroring the rigid entity API): each `FEMVisGeom` pairs a render mesh carrying its own surface and UVs with the map from its vertices to the simulated vertices. The rasterizer and raytracer draw the authored input surface per visual geom and update its vertices from the simulation state through that map, while the FEM solver drops its render-only topology and UV fields. Sub-meshes are welded into a single simulation mesh before tetrahedralization, which now runs before the morph position is applied so the tetrahedralization and its on-disk cache are shared across placements of the same geometry. The normal render pass supports double-sided primitives in both offscreen and interactive viewer renders through a shared shader cache keyed on the primitive defines.

## Motivation and Context

Multi-material assets lost their appearance when simulated as FEM entities: rendering used the tetrahedral boundary with a single entity-wide surface and zero-padded UVs. Decoupling render geometry from simulation geometry preserves the authored look (per-sub-mesh materials, textures, UVs) and enables sub-mesh level segmentation for FEM entities.

## How Has This Been / Can This Be Tested?

- `tests/deformable/test_fem.py`: the new `test_multi_submesh_render_decoupling` asserts exact per-vertex tracking between visual geoms and simulated vertices on a two-material GLB; `test_interior_tetrahedralized_vertex` checks the same invariant on a tetrahedralized box. The full file passes, including the implicit falling sphere/box rest-state checks.
- `tests/rendering/test_offscreen.py`: the new `test_multi_submesh_fem_render` snapshot test covers rgb/depth/segmentation/normal on the rasterizer and rgb on the raytracer; `test_segmentation_map` covers the per-sub-mesh segmentation keys at "geom" level.
- `Genesis-Intelligence/snapshots` updated to revision `11a81c48367a18aedf0a3bed460b69c7ba7ccd4e` (5 new PNGs), pinned in `tests/utils.py`.

## Screenshots (if appropriate):

**Pair 1 - Ducks (Left Rigid, Right FEM)**

| Before | After (This PR) |
|---|---|
|  sbs_upstream |sbs_ours|

**Pair 2 - Multi-sub-mesh FEM rendering (trashbag_rope.glb)**

| Before (uniform single surface) | After (per-sub-mesh GLB materials) |
|---|---|
| trashbag_before| trashbag_after|

## Checklist:

[...]

PR #3079, opened 2026-07-20
[FEATURE] Decouple FEM render geometry from simulation geometry.

duburcqa:
## Description

FEM entities now expose their morph sub-meshes as visual geoms (`fem_entity.vgeoms`, mirroring the rigid entity API): each `FEMVisGeom` pairs a render mesh carrying its own surface and UVs with the map from its vertices to the simulated vertices. The rasterizer and raytracer draw the authored input surface per visual geom and update its vertices from the simulation state through that map, while the FEM solver drops its render-only topology and UV fields. Sub-meshes are welded into a single simulation mesh before tetrahedralization, which now runs before the morph position is applied so the tetrahedralization and its on-disk cache are shared across placements of the same geometry. The normal render pass supports double-sided primitives in both offscreen and interactive viewer renders through a shared shader cache keyed on the primitive defines.

## Related Issue

Supersedes #2904.

## Motivation and Context

Multi-material assets lost their appearance when simulated as FEM entities: rendering used the tetrahedral boundary with a single entity-wide surface and zero-padded UVs. Decoupling render geometry from simulation geometry preserves the authored look (per-sub-mesh materials, textures, UVs) and enables sub-mesh level segmentation for FEM entities.

- `tests/deformable/test_fem.py`: the new `test_multi_submesh_render_decoupling` asserts exact per-vertex tracking between visual geoms and simulated vertices on a two-material GLB; `test_interior_tetrahedralized_vertex` checks the same invariant on a tetrahedralized box. The full file passes, including the implicit falling sphere/box rest-state checks.
- `tests/rendering/test_offscreen.py`: the new `test_multi_submesh_fem_render` snapshot test covers rgb/depth/segmentation/normal on the rasterizer and rgb on the raytracer; `test_segmentation_map` covers the per-sub-mesh segmentation keys at "geom" level.
- `Genesis-Intelligence/snapshots` updated to revision `11a81c48367a18aedf0a3bed460b69c7ba7ccd4e` (5 new PNGs), pinned in `tests/utils.py`.

**Pair 1 - Ducks (Left Rigid, Right FEM)**

| Before | After (This PR) |
|---|---|
|  sbs_upstream |sbs_ours|

**Pair 2 - Multi-sub-mesh FEM rendering (trashbag_rope.glb)**

| Before (uniform single surface) | After (per-sub-mesh GLB materials) |
|---|---|
| trashbag_before| trashbag_after|

## Checklist:


duburcqa:
Closing in favor of #2904, which keeps the original authorship: the consolidated changes are convention and API-surface alignment on top of an unchanged design. Both branches carry the same tree.

PR #3085, opened 2026-07-21
[FEATURE] Add torque visualization for tactile sensor examples.

Milotrince:
## Description
- Cleanup comments from the prev tactile PRs.
- Fix a bug in torque computation for ProximityTaxel sensor
- Visualize torque alongside force
- Bump docs with updated tactile sensor videos (it was wrong before) https://github.com/Genesis-Embodied-AI/genesis-doc/pull/136

https://github.com/user-attachments/assets/1fb316ba-fb63-4002-87ca-6b3c64500670

## Checklist:

### DIFF
--- a/genesis/engine/sensors/point_cloud_tactile.py
+++ b/genesis/engine/sensors/point_cloud_tactile.py
@@ -363,6 +363,7 @@ def _kernel_point_cloud_proximity_taxel_bvh(
     probe_gains: qd.types.ndarray(),
     stiffness: qd.types.ndarray(),
     shear_coupling: qd.types.ndarray(),
+    twist_scalar: qd.types.ndarray(),
     proximity_density_scale: qd.types.ndarray(),
     output_gt: qd.types.ndarray(),
     output_measured: qd.types.ndarray(),
@@ -381,6 +382,7 @@ def _kernel_point_cloud_proximity_taxel_bvh(
 
         k_stiff = stiffness[i_s]
         k_shear = shear_coupling[i_s]
+        k_twist = twist_scalar[i_s]
         dens = proximity_density_scale[i_s, i_b]
         n_probes = n_probes_per_sensor[i_s]
         cache_start = sensor_cache_start[i_s]
@@ -415,10 +417,12 @@ def _kernel_point_cloud_proximity_taxel_bvh(
 
         sum_p_gt = gs.qd_float(0.0)
         fv_gt = qd.Vector.zero(gs.qd_float, 3)
-        tau_w_gt = qd.Vector.zero(gs.qd_float, 3)
+        # Penetration-weighted relative twist-rate (about the normal); divided by sum_p after the loop to get the
+        # contact-averaged omega_n consumed by the spin torque term.
+        omega_w_gt = gs.qd_float(0.0)
         sum_p_m = gs.qd_float(0.0)
         fv_m = qd.Vector.zero(gs.qd_float, 3)
-        tau_w_m = qd.Vector.zero(gs.qd_float, 3)
+        omega_w_m = gs.qd_float(0.0)
 
         chunk_start = bvh.sensor_chunk_start[i_s]
         n_chunks = bvh.sensor_chunk_count[i_s]
@@ -430,6 +434,9 @@ def _kernel_point_cloud_proximity_taxel_bvh(
             rcom_o = dyn_state.links.root_COM[track_link_idx, i_b]
             cdv_o = dyn_state.links.cd_vel[track_link_idx, i_b]
             cda_o = dyn_state.links.cd_ang[track_link_idx, i_b]
+            # Relative angular velocity of this tracked link about the normal (constant across its samples), the
+            # scalar spin rate feeding the twist torque term.
+            omega_c = (cda_o - s_ang).dot(a_w)
             # BVH nodes live in tracked-link local frame: bring the probe sphere center over.
             probe_link = gu.qd_inv_transform_by_trans_quat(probe_world, track_pos, track_quat)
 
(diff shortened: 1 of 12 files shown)


genesis/engine/sensors/point_cloud_tactile.py
```diff
@@ -533,12 +553,12 @@ def _kernel_point_cloud_proximity_taxel_bvh(
             for j in qd.static(range(3)):
                 f_w_m[j] = f_w_m[j] + k_shear * dens * fv_m[j]
 
-        tau_scaled_m = qd.Vector.zero(gs.qd_float, 3)
+        t_w_m = lever_world.cross(f_w_m)
```
duburcqa: I don’t like these variable names. They are violating the guidelines.

genesis/recorders/plotters.py
```diff
@@ -689,6 +697,12 @@ def _project_to_plane(normal: np.ndarray, *arrays: np.ndarray) -> tuple[np.ndarr
     return tuple(data @ uv for data in arrays)
 
 
+def _as_vector_field(data: "np.ndarray | torch.Tensor") -> np.ndarray:
```
duburcqa: I hate this kind of utility. Are you sure it is necessary ? You would sanitize your input at input once and for all, then never care about it again. We should already have some misc utility for this no? This kind of converters tends to come back every now and then, with a new adhoc implementation every time.

duburcqa:
```
The two open review threads — both valid, here's the specific fix

Thread 1 — point_cloud_tactile.py:556, "variable names violate the guidelines" — Confirmed. The PR-introduced names are cryptic against the _idx/descriptive-name conventions. The worst offenders are the new ones, not the pre-existing f_w_gt/f_l_m convention:
- omega_c, omega_w_gt, omega_w_m, omega_n_gt, omega_n_m — single-letter suffixes carrying real meaning (_c=per-chunk?, _w=penetration-weighted-accumulator, _n=contact-averaged). These read as noise.
- t_w_gt/t_w_m (renamed from tau_scaled_*) and k_twist at least mirror the existing f_w_*/k_stiff/k_shear pattern, and lever_world is fine.
- Suggest spelling out the omega family, e.g. rel_spin_rate (omega_c), spin_rate_pen_weighted_gt/_m, spin_rate_avg_gt/_m. There's real tension with the surrounding kernel's terse _gt/_m/_w/_l house style, so confirm how far you want the rename to go — but at minimum the omega_* accumulators should stop hiding their averaging semantics behind _w/_n.

Thread 2 — plotters.py:700, _as_vector_field ad-hoc converter — Confirmed, and the misc utility you're thinking of already exists. _as_vector_field's coercion line
field = tensor_to_array(data) if isinstance(data, torch.Tensor) else np.asarray(data, dtype=float)
is exactly what tensor_to_array already does (tensor_to_array(x) == np.asarray(tensor_to_cpu(x)), and tensor_to_cpu is a no-op on non-tensors). So the isinstance branch is dead duplication — tensor_to_array(data, dtype=float) covers both inputs. The only remaining logic is the (N,3)→(1,N,3) promotion, and per your point that shape is fixed at build time by subplot_titles, so it doesn't need per-frame re-sniffing. Minimal fix: delete the branch and use tensor_to_array; better fix: validate/promote once at build and drop the helper entirely.

Correctness (library)

- Torque rework is physically correct. t_w = lever_world x f_w - a_w * k_twist * omega_n gives the per-taxel moment about the sensor origin plus a torsional-friction spin term opposing relative twist (correct sign). The gain-cancels-in-the-ratio comment for omega_n is right, and the sum_p > eps gating avoids div-by-zero. No issues found.
- The old torque was indeed wrong (a d x a_w moment with no consistent reference point and ignoring shear-force moment), so "bug fix" is accurate.

Conventions / style nits

- Title & tag undersell the change. The title is about example visualization, but the PR changes the default ProximityTaxel.torque output (formula rewrite + new twist_scalar=1.0 adding a spin term by default) and adds a public option + recorder feature. That's a default-contract change — it should carry [BUG FIX]/[CHANGING] framing, and the title should state the user-facing benefit (corrected tactile torque), not the example plot. Per your own [CHANGING]-tag semantics this is the tag that matters for the changelog.
- plotters.py — _cache_backgrounds now caches a single background (self._background), so the plural method name is now a misnomer. Rename to _cache_background for accuracy (and the s_hand throwaway local in build() could be inlined; self._twist_sign already carries the meaning).
- tactile_sandbox.py — comment removal deletes load-bearing rationale. The "cleanup" dropped comments that explained non-obvious behavior, e.g. the crosstalk comment was truncated to # Spatial crosstalk needs a regular grid layout, losing "the 3x3 kernel sums to 1, so it conserves total force" and "only the plane grid, not the dome"; the Schmitt-trigger threshold explanation was removed entirely. Per the guideline to preserve comments that justify non-obvious behavior, keep the why even if trimming prose. (Boilerplate/obvious ones are fine to drop.)
- shear_coupling=0.0,  # isolate the spin term ... in the test is a trailing inline comment; guideline prefers it on its own line above.
- twist_scalar option docstring uses *about* markdown emphasis — harmless but not the sibling style; the formula and tradeoff ("set to 0.0 to report tilting moments alone") are good and follow the shear_coupling sibling.

Tests

- test_proximity_taxel_twist_torque is strong: parametrized over n_envs=[0,2], no docstring, feature-named, 4-step budget, and physics-based. With shear_coupling=0.0 the normal force contributes zero moment about the normal, so torque·n = -k_twist·omega_n exactly — which makes the linearity (2x), zero (twist0), and residual-tilting-moment assertions analytically exact rather than tolerance-fudged. Nicely done.
- Kernel param order: twist_scalar inserted after shear_coupling in the kernel signature, build() concat, and call site — all three consistent. Good.
- Camera viewpoint: the guideline asks every scene.step() test to set ViewerOptions with a tuned camera; this test doesn't — but no sibling in test_tactile.py does either, so it matches the established file pattern. Flag for the file as a whole, not this PR specifically.
- Regression coverage: the test exercises the new twist term thoroughly; the last assertion (twist0 still has torque) is ever-arm moment bug fix itself. Adequate, but if you want the "fixed the torque formula" claim pinned, an assertion tying
twist0's tilting moment to lever x f would nail it.
```

genesis/recorders/plotters.py
Milotrince: moved to BaseRecorder

genesis/recorders/file_writers.py
Milotrince, line 1: The extra "blast radius" in changes is due to removing conversion to np array redundancy, resolved by enforcing np array at `base_recorder` level.

Milotrince (commented):
self-reviewed! I added ruff import style fix too, but if you don't like this it can be removed. I have auto-organize imports on in my IDE so it's quite helpful.

genesis/recorders/plotters.py
```diff
@@ -6,26 +6,25 @@
...
     MPLImagePlot as MPLImagePlotterOptions,
+    MPLLinePlot as MPLLinePlotterOptions,
     MPLVectorFieldPlot as MPLVectorFieldPlotterOptions,
+    PyQtLinePlot as PyQtLinePlotterOptions,
 )
-from genesis.utils import has_display, tensor_to_array
+from genesis.utils import has_display
 
-from .base_recorder import Recorder
+from .base_recorder import Recorder, _to_numpy
```
duburcqa: If you need to import a private helper '_to_numpy', there is a design issue.

genesis/recorders/plotters.py
```diff
@@ -59,8 +58,6 @@
 
 
 def _data_to_array(data: Sequence) -> np.ndarray:
```
duburcqa: No one-line helper '_data_to_array'. This is violating the guidelines I think no?

genesis/recorders/file_writers.py
duburcqa: Yes, and this is great!

genesis/recorders/plotters.py
duburcqa: What is '_to_numpy' in the first place? Looks like something we should get rid of by properly pre-processing data right at the entrypoint.

PR #3106, opened 2026-07-26
  [BUG FIX] Record one camera frame per simulation step.

### LINKED ISSUE #3105: [Bug]: Camera recording stores duplicate frames because the timestamp is never updated
### Bug Description

I was reading through `genesis/vis/camera.py` and hit what looks like a typo in `Camera.render()`.

After a frame is stored for recording, line 495 does this:

```python
self._recorded_t_prev == self._visualizer.scene._t
```

That's a comparison, so the result is discarded. It looks like it was meant to be an assignment (`=` rather than `==`).

The effect is that `_recorded_t_prev` stays at `-1` for the whole recording. The only other places it gets written are `__init__` and `stop_recording()`, and both set it to `-1`.

Two things depend on that value, and I've confirmed neither works:

- Line 433: `is_recording = self._in_recording and self._recorded_t_prev != self._visualizer.scene._t`. With `-1` this is always true, so calling `render()` twice in the same step stores the frame twice.
- Line 491: the missing-frame check starts with `self._recorded_t_prev < 0`, which is always true, so `gs.raise_exception("Missing frames in recording...")` can never fire.

So both the "one frame per timestamp" behaviour and the missing-frame check are silently inactive, and have been since the field was introduced.

### Steps to Reproduce

```python
import genesis as gs

gs.init(backend=gs.cpu)

scene.add_entity(gs.morphs.MJCF(file="xml/franka_emika_panda/panda.xml"))
cam = scene.add_camera(res=(320, 240), pos=(2.5, 0.0, 1.5), lookat=(0, 0, 0.5), GUI=False)
scene.build()

cam.start_recording()

scene.step()
cam.render()
cam.render()   # same scene timestamp as the previous call

print(len(cam._recorded_imgs))   # 2, but only 1 frame should be stored
```

The missing-frame check is also unreachable. This records happily even though a step is skipped every iteration:

```python
cam.start_recording()
for _ in range(5):
    scene.step()
    scene.step()   # deliberately skipped
    cam.render()
# no exception raised
```

Observed on 1.2.3, CPU backend:

- first snippet stores **2** frames instead of 1, and `cam._recorded_t_prev` is still `-1` afterwards
- second snippet stores 3 frames and raises nothing, despite a step being skipped every iteration

### Expected Behavior

Rendering twice at the same scene timestamp should only store one frame, and skipping a step while recording should trigger the existing "Missing frames in recording" error.

One thing worth flagging: changing `==` to `=` on its own isn't quite enough. `pause_recording()` sets `_in_recording = False` but leaves `_recorded_t_prev` as it was, and `start_recording()` doesn't reset it either. So pausing, stepping a few times, then calling `start_recording()` again would trip the missing-frame check on a workflow that's meant to work (the `pause_recording` docstring says recording can be resumed with `start_recording()`).

Resetting `_recorded_t_prev` in `start_recording()` would cover both the fresh-start and the resume case.

### Screenshots/Videos

_No response_

### Relevant log output

```shell
TEST 1 - two render() calls at the SAME scene timestamp
  scene._t            = 1
  _recorded_t_prev    = -1   (should track scene._t)
  frames stored       = 2   (expected 1)

TEST 2 - skipping a step should raise 'Missing frames'
  _recorded_t_prev    = -1
  frames stored       = 3
  (no exception raised)
```

### Environment

- OS: macOS 15.7.3
- GPU/CPU: Apple M4 (CPU backend, `gs.init(backend=gs.cpu)`)
- GPU-driver version: N/A

1.2.3 (commit 2a40c87ad4e89a0f115cc43fb45195a8260aad0e)

### Additional Context

_No response_

jeetrex17:
## Description

  `Camera.render()` did not update `_recorded_t_prev` after storing a video frame because it used `==`
  instead of `=`.

  This caused repeated calls to `render()` at the same simulation timestamp to store duplicate frames.
  It also kept the existing missing-frame check inactive because `_recorded_t_prev` remained at `-1`.

  This change updates the timestamp after storing a frame. It also resets timestamp tracking when
  recording transitions from inactive to active, so simulation steps elapsed while recording is paused
  are not treated as missing frames. Calling `start_recording()` while already recording preserves the
  current timestamp.

  ## Related Issue

  Resolves #3105

  ## Motivation and Context

  Camera recording is intended to store at most one frame per simulation timestamp and report skipped
  timestamps during continuous recording. Both behaviors were inactive because the previous timestamp
  was never updated.

  This restores the existing intended behavior without changing the public API.

  Tested on macOS 15.7.3 with an Apple M4, the CPU backend, and Python 3.13.11.

  A regression test was added for duplicate rendering, consecutive timestamps, repeated
  `start_recording()` calls, pause/resume, and skipped-frame detection.

  ```bash
  pytest tests/rendering/test_offscreen.py::test_recording_stores_one_frame_per_timestamp -v -n 0 --forked
  pytest tests/rendering/test_offscreen.py::test_render_api_advanced -v -n 0
  pytest tests/rendering/test_offscreen.py -v -m required -n 0 --forked
  pre-commit run --files genesis/vis/camera.py tests/rendering/test_offscreen.py
```

  The regression test fails on main because two calls to render() at the same timestamp store two
  frames. It passes with this change.

  All locally supported required offscreen-rendering cases passed, including 162 snapshots. Tests
  requiring CUDA, LuisaRender, or an interactive viewer were skipped because those backends are not
  available on my machine. Both pre-commit hooks passed.

  Not applicable.

  ## Checklist:

  - [x] I read the CONTRIBUTING document.
  - [x] I followed the Submitting Code Changes section of CONTRIBUTING document.

### DIFF
--- a/genesis/engine/scene.py
+++ b/genesis/engine/scene.py
@@ -1017,6 +1017,9 @@ class Scene(RBC):
         # TODO: Could be optimized to only clear cache associated the the environments being reset.
         if self._visualizer.is_built:
             self._visualizer.reset()
+            # Restart camera recording sequences because the scene clock was reset above
+            for camera in self._visualizer.cameras:
+                camera._recorded_t_prev = -1
 
         # TODO: sets _next_particle = 0; not sure this is env isolation safe
         for emitter in self._emitters:
--- a/genesis/vis/camera.py
+++ b/genesis/vis/camera.py
@@ -492,7 +492,7 @@ class Camera(RBC):
                 gs.raise_exception(
                     "Missing frames in recording. Please call 'camera.render()' after 'every scene.step()'."
                 )
-            self._recorded_t_prev == self._visualizer.scene._t
+            self._recorded_t_prev = self._visualizer.scene._t
             rgb_frame = tensor_to_array(rgb_arr)
             self._recorded_imgs.append(rgb_frame)
 
@@ -684,6 +684,9 @@ class Camera(RBC):
         Start recording on the camera. After recording is started, all the rgb images rendered by `camera.render()`
         will be stored, and saved to a video file when `camera.stop_recording()` is called.
         """
+        # Restart timestamp tracking when recording becomes active, excluding simulation steps elapsed while paused
+        if not self._in_recording:
+            self._recorded_t_prev = -1
         self._in_recording = True
 
     @gs.assert_built
--- a/tests/rendering/test_offscreen.py
+++ b/tests/rendering/test_offscreen.py
@@ -1532,3 +1532,60 @@ def test_rasterizer_sensor_env_spacing_invariance(renderer, context_mode):
     # Per-env sensor images must be invariant to env_spacing — the offset is purely for
     # visual separation in the interactive viewer and must be transparent to sensors.
     assert np.abs(img_ref.astype(np.float32) - img_test.astype(np.float32)).mean() < 1.0
+
+
+@pytest.mark.required
+@pytest.mark.parametrize("renderer_type", [RENDERER_TYPE.RASTERIZER])
+def test_recording_stores_one_frame_per_timestamp(renderer, show_viewer):
+    scene = gs.Scene(
+        viewer_options=gs.options.ViewerOptions(
+            camera_pos=(2.5, 0.0, 1.5),
+            camera_lookat=(0.0, 0.0, 0.5),
+        ),
+        renderer=renderer,
+        show_viewer=show_viewer,
+    )
+    scene.add_entity(
+        gs.morphs.Plane(),
+    )
+    scene.add_entity(
+        gs.morphs.Box(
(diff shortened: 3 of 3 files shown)


duburcqa:
Thanks for tracking this down - the `==` typo was the real cause, and the diagnosis in #3105 was spot on.

Closing in favour of #3107, which takes a different route to the same fix. Activating the missing-frame check turned out to break recording at a framerate below the physics rate, which `examples/rigid/hibernation.py`, `examples/fem_hard_and_soft_constraint.py` and `examples/coupling/fem_cube_linked_with_arm.py` all rely on deliberately. Rather than dropping the check, #3107 makes the camera capture its own frames at a requested framerate, which removes the duplicate frames without a timestamp comparison at all, and streams them to file so a recording is no longer capped by available memory.

Issue #1608, opened 2025-08-20
[Bug]: Wrong IK result for mobile manipulator


HeegerGao:
### Bug Description

Hi, I am using Genesis for mobile manipulation. I implemented a simple mobile manipulator (a base with a Franka arm). It works normally with joint space control. 

However, when I want to use the end effector Cartesian space PD control (e.g., I want the end effector to move forward for 0.01m and keep the base unmoved), the IK seems to return a wrong qpos. 

Specifically, in the following script, I have a mobile_franka robot, where the base has 3 dofs: moving left, right, and rotating, and the arm is a Franka Panda arm. I want to control the robot arm to move forward for `0.01m` every step and keep the base unmoved. I set the `arm_delta_action = torch.tensor([0.01, 0, 0, 0, 0, 0], device="cuda")`, and solve the IK to get the `target_qpos`, where I have set the `dofs_idx_local` to only count for the robot arm dofs (that is `[3,4,5,6,7,8,9]`). 

```
from copy import deepcopy
import sys

import torch
sys.path.append(".")
sys.path.append("..")
import numpy as np

import genesis as gs

from transform_utils import (
    euler_angles_to_matrix,
    matrix_to_quaternion,
    quaternion_multiply,
    quaternion_invert,
    Pose,
)

def get_pos(ee_link):
    return ee_link.get_pos()

def get_quat(ee_link):
    return ee_link.get_quat()

def get_pose(ee_link):
    return Pose.create_from_pq(ee_link.get_pos(), ee_link.get_quat())

def get_ee_link(entity, ee_link_name="hand"):
    return entity.get_link(ee_link_name)

gs.init(backend=gs.gpu)

scene = gs.Scene(
    viewer_options = gs.options.ViewerOptions(
        camera_pos    = (0, -3.5, 2.5),
        camera_lookat = (0.0, 0.0, 0.5),
        camera_fov    = 30,
        res           = (960, 640),
        max_FPS       = 60,
    ),
    sim_options = gs.options.SimOptions(
        dt = 0.01,
    ),
    show_viewer = True,
    show_FPS=True,
)

cam = scene.add_camera(
    res    = (960, 640),
    pos    = (0, -3.5, 2.5),
    lookat = (0.0, 0.0, 0.5),
    fov    = 30,
    GUI    = False
)

plane = scene.add_entity(
    gs.morphs.Plane(),
)
mobile_franka = scene.add_entity(
    gs.morphs.MJCF(
[...]

Issue #726, opened 2025-02-13
[Feature]:  Motion planning for rigid entities with free joints.


Bingkun-Huang:
### What feature or enhancement are you proposing?

The plan_path() is not yet supported for robots with free joints (like Tiago with moving based and manipulator) 

### Motivation

The plan_path() is not yet supported for robots with free joints (like Tiago with moving based and manipulator) 

### Potential Benefit

So that in Genesis could simulate with more diverse robots 

### What is the expected outcome of the implementation work?

Use existed libraries for planning

PR #3109, opened 2026-07-27
[BREAKING][FEATURE] Add more robust and faster motion planning algorithm.

### LINKED ISSUE #2715: [Bug]: Calling plan_path with with_entity causes unintended object motion

duburcqa:
## Description

Replaces the RRT/RRTConnect path planner with a batched one under `genesis/engine/solvers/rigid/planner/`.

- One code path for a single environment on CPU and a batch on GPU: every environment of a batch is planned in one call, rather than looping.
- Joint-space and Cartesian goals, both resolved in-kernel. Held entities are attached and certified along with the arm.
- Planning never mutates solver state - the planner owns its scratch buffers.
- Tree storage is pooled across a fixed number of concurrent searches and reused in waves, so a batch's memory stops scaling with its environment count.
- A returned path is certified collision-free between its waypoints, not merely at them.
- Certification reads exact geometry wherever it can - the collider's GJK against a convex obstacle, a geom's own vertices against a half-space - so the sphere proxy's conservatism cannot wall off a corridor that is really clear.

API break on `plan_path`: `max_nodes`, `resolution`, `timeout`, `smooth_path`, `planner` and `return_valid_mask` are gone, `num_waypoints` now defaults to scene-dt spacing, and `goal_link` / `goal_pos` / `goal_quat`, `safety_margin`, `attach_held_entities` and `seed` are new. The return value is a `PlannerPath` carrying `qpos`, `dofs_vel`, `dofs_acc`, `dt` and `is_valid`.

## How it works

```mermaid
flowchart TB
    subgraph once["set up, once per call"]
        SNAP["world snapshot<br/>freeze obstacle poses, bounds, shape class"]
        EXCL["boundary exclusions<br/>excuse the contacts the start and goal already have"]
    end

    subgraph ladder["attempt ladder, repeated until every environment certifies or the budget runs out"]
        GOAL["goal resolution<br/>multi-restart damped-least-squares IK"]
        RRT["RRT-Connect escalation<br/>two trees per environment, every edge certified"]
        SEED["seeding<br/>straight line, or a certified RRT polyline"]
        MPPI["MPPI warm start<br/>annealed smooth-noise particles, softmax update"]
        LBFGS["L-BFGS refinement<br/>two-loop direction, parallel line search"]
        VAL["validator<br/>densified swept certification at the requested margin"]
        FOLD["fold<br/>mark certified environments, end the ladder"]
    end

    subgraph finish["finish, once the ladder ends"]
        STR["straightening<br/>chord splices at decreasing radii, each re-certified"]
        POL["smoothness polish<br/>L-BFGS with raised acceleration and jerk weights"]
        RT["retiming<br/>speed profile along the certified polyline"]
    end

    CORE["collision core<br/>proxy spheres against obstacles, robot self pairs,<br/>exact re-check: collider GJK, or geom vertices for a half-space"]

    SNAP --> EXCL --> GOAL --> RRT --> SEED --> MPPI --> LBFGS --> VAL --> FOLD
    FOLD -->|"environment still unsolved"| GOAL
    FOLD -->|"all certified"| STR
    STR --> POL --> RT --> OUT["PlannerPath<br/>qpos, dofs_vel, dofs_acc, dt, is_valid"]

    GOAL -.->|exact| CORE
    RRT -.->|exact| CORE
    VAL -.->|exact| CORE
    MPPI -.->|proxy| CORE
    LBFGS -.->|proxy| CORE
```

| block | what it is responsible for |
| --- | --- |
| world snapshot | Freezes every obstacle's pose, bounds and shape class at the moment of the call, so nothing the planner reads can move under it and planning never touches solver state. |
| boundary exclusions | Records the contacts the start and goal configurations already have, so a grasp pose that touches its target is not read as a collision. Each excused pair is anchored to where it was excused and its allowance ramps away from that anchor, so the excuse cannot travel with the arm. |
| goal resolution | Cartesian goals only. Turns a pose into a joint-space goal: many independent damped-least-squares solves from low-discrepancy restarts, each certified as a hold at that pose, and the closest-to-start acceptable one becomes the goal. It is re-run every pass, so a bad branch is never permanent. |
| seeding | Gives every candidate a starting trajectory - a straight line on the first pass, or the certified polyline a tree found on a later one. |
| MPPI warm start | Gets a candidate out of the basin a straight line starts in: samples annealed smooth-noise trajectories around it and moves it along a softmax-weighted mean. |
| L-BFGS refinement | Drives the candidate to a local optimum of the trajectory cost - obstacle and self clearance, joint limits, acceleration and jerk, plus pose pinning for a Cartesian goal - along a two-loop recursion direction, with the line-search trials scored in parallel. It shares one scoring loop with the warm start. |
| RRT-Connect escalation | The fallback for what optimisation cannot reach: two trees per environment grow toward each other with every edge certified along its continuum, a shortcut pass straightens the result, and the polyline is transplanted back as a seed. It runs only for environments an earlier pass left unsolved, and continues the trees it already grew rather than replanting them. |
| validator | The only thing that decides validity. It densifies a candidate between its knots and certifies each sample with a Lipschitz sweep allowance, so a passing path is collision-free along its continuum and not merely at its waypoints. |
| fold | Marks the environments whose candidate certified, and ends the ladder as soon as none is left unsolved. |
| straightening and polish | Improve the path without being allowed to break it: chord splices at decreasing deviation radii, then an L-BFGS round with raised smoothness weights, each kept only where it still certifies. |
| retiming | Turns the certified polyline into a timed trajectory: a trapezoidal speed profile along the path under per-DOF velocity and acceleration limits, with parabolic corner blends, sampled at scene `dt`. |
| collision core | Shared by everything above that needs geometry: proxy spheres against obstacle distance fields, the robot's self pairs, and an exact re-check of borderline pairs - the collider's GJK against a convex obstacle, the geom's own vertices against a half-space. The optimiser reads the cheap proxy value; every certification path asks for the exact one. |

## Related Issue

Resolves Genesis-Embodied-AI/Genesis#2715 - planning no longer touches the scene. The planner owns its scratch state, and `tests/rigid/test_path_planning.py::test_plan_path_with_attached_entity` asserts that a carried entity's pose is identical before and after a `plan_path` call.

Related to Genesis-Embodied-AI/Genesis#1608 - not fixed here. The inverse-kinematics path is reorganised by this PR, which now shares a scratch-based forward-kinematics and Jacobian core with the planner, but its solving behaviour is unchanged, so the reported discontinuity stands.

Related to Genesis-Embodied-AI/Genesis#726 - free-joint robots are still unplannable, but the failure is now explicit: free and spherical joints raise rather than being silently mishandled.

## Motivation and Context

motion_planning_clutter_16tasks

Sixteen problems from one batch, each drawn independently, as one clip: in every tile the solid arm follows what the planner returned and the dark ghost follows the reference motion the clutter was built against. They are rarely alike, which is the point - the reference only proves a route exists.

The old planner did not survive contact with batched scenes: plain RRT failed to compile, RRTConnect crashed for `n_envs > 0`, and a successful plan displaced grasped entities. What it did return was disappointing in clutter, and there was no benchmark honest enough to say why - randomly drawn scenes are either trivial or unsolvable, so any failure could be blamed on an impossible problem.

`examples/rigid/motion_planning_clutter.py` removes that excuse, and is how the rewrite is judged. It generates the solution first and the obstacles afterwards: a reference motion is drawn in an empty world, then 80 boxes are placed against it, each pushed until its closest approach over the whole motion is 2 cm. Every scene is therefore solvable by construction - the reference is a witness - and no box is decorative, since each one comes within 2 cm of the robot at some instant. Half of them are seeded on the straight start-to-goal line, the route a planner tries first, which has to be shut off for clutter to constrain anything.

Its parameters were tuned to minimise the solve rate, over 192 problems per setting:

| parameter | chosen | alternatives measured |
| --- | --- | --- |
| spline control points | 3 | 2 -> 96% solved, 4 -> 75%, 5 -> 77% |
| end-effector path floor | 1.75 m | 2.50 m -> 63.0%, 3.25 m -> 64.1% |
| boxes per scene | 80 | 16 -> 87.5%, 40 -> 60.9%, 64 -> 46.9%, 96 -> 40.6% |

Against that generator the planner solves **256 of 256 problems** - sixteen seeds of sixteen environments each, at the shipped default retry budget, eight of those seeds held out and never used while tuning.

`python examples/rigid/motion_planning_clutter.py -v` poses one problem, plans it, and replays both the plan and the reference motion (as a translucent ghost) so the two can be compared. `-n` poses several independent problems at once. `tests/rigid/test_path_planning.py` covers the planner itself.

Scene generation costs about 13 s per batch of 16 environments at 80 boxes on CPU.

Beyond the suite, the planner is held to a local matrix on a hand-authored 16-environment pre-grasp scene, over six seeds: full validity at both precisions and on the Metal backend, zero real penetration along returned paths, bit-identical plans across runs, and peak acceleration within the retimer's bound. Every row holds. The same validity and penetration rows were run on CUDA, on both the GPU and the CPU backend, and hold there too. An accepted Cartesian goal is held to 2 mm of real penetration, measured against the collider rather than against the planner's own proxy geometry, and both backends draw the same restarts when resolving one.

## Performance

Full profile, with the stack it was measured on and the spread behind every figure: [PROFILING_REPORT.md](https://github.com/user-attachments/files/30466849/PROFILING_REPORT.md). It covers per-problem distributions, batch scaling on CPU, Metal and CUDA, per-kernel phase shares, an in-kernel `clock_counter` breakdown, the optimisations it justified, and the ideas measurement rejected.

| backend | batch | ms per plan | validity |
| --- | --- | --- | --- |
| CPU | 1 | 134 * | 128/128 |
| CPU | 16 | **76** | 128/128 |
| Metal | 16 / 64 / 256 | 114 / 44 / **19** | 2688/2688 |
| CUDA | 16 / 64 / 256 / 1024 | 133 / 49 / **19** / **14** ** | 4080/4080 |

\* Mean over 128 problems (16 grasp spots x 8 seeds), measured before the last three optimisations. The spread comes from the problem far more than from the seed: per-spot means run from 84 to 239 ms, with an overall standard deviation of 53 ms. The batched row is lower than the single-problem one because planning one problem per call also pays about 50 ms of fixed overhead that a batch amortises.

\*\* The CUDA rows predate the last two optimisations, which are worth a further 5-6% on the GPU arm and 9% on CPU; they were not re-measured there. Every figure is an interleaved A/B/A over 8 seeds, and each optimisation was gated on full validity at both precisions and on both GPU backends before being believed.

The cost is dominated by the certified collision check of a single configuration. At small batch that check is latency-bound - a batch's wall time barely grows with its environment count, so per-plan cost falls almost linearly with it - while at 256 environments the device is oversubscribed twelve-fold and the wall becomes throughput-bound, so only doing less work per check can help there. On CPU the arm is single-threaded, so batching buys nothing beyond amortising the per-call overhead.

The two arms are bound by different resources, which is why two of the optimisations are split between them: the GPU arm is bandwidth- and occupancy-bound and rejects anything that widens a load or grows the hottest function body, while the CPU arm is flop-bound and barely notices bytes.

## Blockers

- **This needs a quadrants carrying the fastcache parameter-pruning fix** (Genesis-Embodied-AI/quadrants#809). Without it the graph kernels compile with an unbound parameter and every environment silently fails to resolve a Cartesian goal, on CPU as well as GPU, so the `quadrants==1.1.3` pin has to move before this can land. Note also that the offline kernel cache is not keyed on the quadrants build, so a cache written before that fix keeps being served after it.

## Checklist:

Resolves SIM-184


genesis/engine/solvers/rigid/abd/inverse_kinematics.py
```diff
@@ -276,15 +635,98 @@ def kernel_rigid_entity_inverse_kinematics(
...
                                 q_start = dyn_info.joints.q_start[I_j]
-                                rigid_info.qpos[q_start, i_b] = dof_limit[0] + qd.random() * (
-                                    dof_limit[1] - dof_limit[0]
-                                )
-                else:
-                    pass  # When respect_joint_limit=False, we can simply continue from the last solution
+                                # Counter-based draw keyed on (env, restart, DOF): a qd.random() per-thread stream
+                                # would order by the parallel schedule, making the resampled - and returned -
+                                # configuration nondeterministic across runs.
+                                fk.qpos[q_start - q_offset, i_b] = dof_limit[0] + gu.qd_hash01(
```
YilingQiao, line 641: can we remove `env` from the keys? One use case is that we want the result to be reproducible no matter which env a trajectory is in.

duburcqa: He is issue addressed on main? I just noticed your comment...

PR #3113, opened 2026-07-28
[BUG FIX] Correct environment indexing for link information.

### LINKED ISSUE #3112: [Bug]: Link info environment indexing is inverted and breaks energy getters
### Bug Description

`RigidSolver.get_links_inertial_mass()` and `get_links_invweight()` handle `envs_idx` with the `batch_links_info` condition reversed.

  When `batch_links_info=False`, link information is shared between environments and has no environment dimension. However, `envs_idx` is accepted and passed to the underlying tensor conversion. It is then interpreted as a link index instead of an environment index. With multiple links, this silently returns the wrong link information.

  When `batch_links_info=True`, link information does have an environment dimension, but providing `envs_idx` raises:

  `envs_idx` cannot be specified for non-batched links info.

  The same inconsistency affects energy calculations:

  - `RigidSolver.get_total_energy(envs_idx=...)` can use the wrong link masses in non-batched mode.
  - `RigidEntity.get_potential_energy(envs_idx=...)` does not select the corresponding mass row in batched mode and can return values for every environment instead of only the requested one.
  - `RigidEntity.get_total_energy()` inherits the potential-energy behavior.

### Steps to Reproduce

The following example reproduces both sides of the issue using only the public API:

  ```python
  import genesis as gs

  gs.init(backend=gs.cpu)


  def make_scene(batch_links_info):
      scene = gs.Scene(
              batch_links_info=batch_links_info,
          ),
          show_viewer=False,
      )
      entities = (
          scene.add_entity(
              gs.morphs.Box(
                  size=(0.1, 0.1, 0.1),
                  pos=(0.0, 0.0, 1.0),
              )
          ),
          scene.add_entity(
              gs.morphs.Box(
                  size=(0.2, 0.2, 0.2),
                  pos=(0.0, 0.0, 2.0),
              )
          ),
      )
      scene.build(n_envs=2)
      return scene, entities


  # Non-batched link information
  scene, _ = make_scene(batch_links_info=False)

  print("Non-batched link masses:")
  print("all:", scene.rigid_solver.get_links_inertial_mass())
  print(
      "envs_idx=[1]:",
      scene.rigid_solver.get_links_inertial_mass(envs_idx=[1]),
  )

  print("\nNon-batched solver energy:")
  print("all:", scene.rigid_solver.get_total_energy())
  print(
      "envs_idx=[1]:",
      scene.rigid_solver.get_total_energy(envs_idx=[1]),
  )


  # Batched link information
  scene, entities = make_scene(batch_links_info=True)

  entities[0].links[0].set_mass((1.0, 2.0))
  entities[1].links[0].set_mass((3.0, 4.0))

  print("\nBatched link masses:")
  print("all:", scene.rigid_solver.get_links_inertial_mass())

  try:
      print(
          "envs_idx=[1]:",
          scene.rigid_solver.get_links_inertial_mass(envs_idx=[1]),
      )
  except gs.GenesisException as error:
      print("envs_idx=[1] error:", error)

  print("\nBatched entity potential energy:")
  print("all:", entities[0].get_potential_energy())
  print("envs_idx=[1]:", entities[0].get_potential_energy(envs_idx=[1]))
```

### Expected Behavior

When `batch_links_info=False`, the direct link-information getters should reject `envs_idx` because the stored link information has no environment dimension.

Energy getters should still support `envs_idx` in this mode by selecting the dynamic environment data while using the shared link masses.

When `batch_links_info=True`, the direct link-information getters should accept `envs_idx` and return only the requested environment rows.

For both batching modes, selecting environment 1 from a solver or entity energy getter should produce the same result as selecting row 1 from the getter's complete result.

### Screenshots/Videos

  Not applicable. The console output demonstrates the incorrect indexing directly.

### Relevant log output

```shell
Non-batched link masses:
  all: tensor([0.6000, 4.8000])
  envs_idx=[1]: tensor([4.8000])

  Non-batched solver energy:
  all: tensor([100.0620, 100.0620])
  envs_idx=[1]: tensor([141.2640])

  Batched link masses:
  all: tensor([[1., 3.],
               [2., 4.]])
  envs_idx=[1] error: `envs_idx` cannot be specified for non-batched links info.

  Batched entity potential energy:
  all: tensor([ 9.8100, 19.6200])
  envs_idx=[1]: tensor([ 9.8100, 19.6200])
```

### Environment

 - OS: macOS 15.7.3 (24G419)
  - GPU/CPU: Apple M4
  - GPU-driver version: N/A (Apple Metal)
  - Python: 3.11.14
  - Genesis: 1.2.3
  - Backend used for reproduction: CPU

17ea5a4cd3aae348f603230a31c40811384ae741

### Additional Context

The behavior was reproduced on an unchanged checkout of the commit above.

When `batch_links_info=False`, link mass and inverse-weight storage is one-dimensional over links. Consequently, passing `envs_idx=[1]` can silently select link 1 rather than environment 1.

When `batch_links_info=True`, the storage includes an environment dimension, so environment selection should be valid.

The neighboring batched-information getters reject `envs_idx` when their corresponding batching option is disabled, which is the opposite of the current link-information guard.

jeetrex17:
## Description

I fixed environment indexing for batched link information.

The link mass and inverse-weight getters checked `batch_links_info` in reverse. This allowed `envs_idx` for shared link information and rejected it for batched link information. I corrected those checks.

I also updated the solver and entity energy calculations so they select per-environment masses only when link information is batched.

I added a regression test for shared and batched link information, combined link and environment selection, and solver and entity energy values.

## Related Issue

Resolves #3112

## Motivation and Context

The previous behavior could interpret an environment index as a link index in shared mode. It could also reject valid environment selection in batched mode or return energy values for the wrong set of environments.

I ran:

```bash
python -m pytest tests/rigid/test_api.py::test_link_info_environment_indexing -v
python -m pytest tests/rigid/test_api.py -m "required and not slow" -v
python -m pytest tests/rigid/test_dynamics.py::test_energy_analytical_and_conservation -v
uvx --from ruff==0.14.11 ruff check genesis/engine/entities/rigid_entity/rigid_entity.py genesis/engine/solvers/rigid/rigid_solver.py tests/rigid/test_api.py
uvx --from ruff==0.14.11 ruff format --check genesis/engine/entities/rigid_entity/rigid_entity.py genesis/engine/solvers/rigid/rigid_solver.py tests/rigid/test_api.py
uv build --wheel
```

Results:

- Link information regression test: 2 passed
- Required non-slow rigid API tests: 8 passed
- Analytical energy tests: 2 passed
- Ruff checks passed
- Wheel build passed

I also started the full required test suite. It reached 21% before I stopped it because of its runtime. Two unrelated macOS rendering and plotting tests failed during that run. The rigid API and batching tests reached during the run passed.

Not applicable.

## Checklist:

### DIFF
--- a/tests/rigid/test_api.py
+++ b/tests/rigid/test_api.py
@@ -456,6 +456,55 @@ def test_batched_info(batch_links_info, batch_joints_info, batch_dofs_info):
     assert act_gain.shape == (9, 2) if batch_dofs_info else (9,)
 
 
+@pytest.mark.required
+@pytest.mark.parametrize("batch_links_info", [False, True])
+def test_link_info_environment_indexing(batch_links_info, tol):
+    scene = gs.Scene(
+        rigid_options=gs.options.RigidOptions(
+            batch_links_info=batch_links_info,
+        ),
+        show_viewer=False,
+    )
+    entities = (
+        scene.add_entity(gs.morphs.Box(size=(0.1, 0.1, 0.1), pos=(0.0, 0.0, 1.0))),
+        scene.add_entity(gs.morphs.Box(size=(0.2, 0.2, 0.2), pos=(0.0, 0.0, 2.0))),
+    )
+    scene.build(n_envs=3)
+
+    getters = (scene.rigid_solver.get_links_inertial_mass, scene.rigid_solver.get_links_invweight)
+    if not batch_links_info:
+        for getter in getters:
+            with pytest.raises(gs.GenesisException, match="envs_idx.*non-batched links info"):
+                getter(envs_idx=[1])
+    else:
+        entities[0].links[0].set_mass((1.0, 2.0, 3.0))
+        entities[1].links[0].set_mass((4.0, 5.0, 6.0))
+        assert_allclose(
+            scene.rigid_solver.get_links_inertial_mass(),
+            ((1.0, 4.0), (2.0, 5.0), (3.0, 6.0)),
+            tol=tol,
+        )
+        for getter in getters:
+            values = getter()
+            assert_allclose(getter(envs_idx=[1]), values[[1]], tol=tol)
+            assert_allclose(getter(links_idx=[1], envs_idx=[2, 0]), values[[2, 0]][:, [1]], tol=tol)
+        assert_allclose(
+            scene.rigid_solver.get_links_inertial_mass(links_idx=[1], envs_idx=[2, 0]),
+            ((6.0,), (4.0,)),
+            tol=tol,
+        )
+        for getter in (entities[0].get_links_inertial_mass, entities[0].get_links_invweight):
+            values = getter()
+            assert_allclose(getter(links_idx_local=[0], envs_idx=[2, 0]), values[[2, 0]][:, [0]], tol=tol)
+
+    total_energy = scene.rigid_solver.get_total_energy()
+    assert_allclose(scene.rigid_solver.get_total_energy(envs_idx=[1]), total_energy[[1]], tol=tol)
+    for entity in entities:
+        for getter in (entity.get_potential_energy, entity.get_total_energy):
+            values = getter()
+            assert_allclose(getter(envs_idx=[1]), values[[1]], tol=tol)
+
+
 @pytest.mark.slow  # ~200s
 @pytest.mark.required
 def test_info_batching(tol):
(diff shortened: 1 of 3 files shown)


tests/rigid/test_api.py
```diff
@@ -456,6 +456,55 @@ def test_batched_info(batch_links_info, batch_joints_info, batch_dofs_info):
     assert act_gain.shape == (9, 2) if batch_dofs_info else (9,)
 
 
+@pytest.mark.required
+@pytest.mark.parametrize("batch_links_info", [False, True])
+def test_link_info_environment_indexing(batch_links_info, tol):
```
duburcqa, line 461: Note that extending existing unit tests is strongly preferred over adding new unit tests, to keep all specifications / contract tightly packed.

I'm adding this to CODING_GUIDELINES.md.

PR #3117, opened 2026-07-28
[BUG FIX] Fix handling of negative index for rigid accessor filter 'envs_idx'.

### LINKED ISSUE #3116: [Bug]: envs_idx=-1 silently selects no environment
### Bug Description

`Scene._sanitize_envs_idx()` converts a scalar `envs_idx` into a one-element tensor by slicing:

  ```python
  if isinstance(envs_idx, (int, np.integer)):
      return self._envs_idx[envs_idx : envs_idx + 1]
  ```

  For `envs_idx=-1`, this becomes `self._envs_idx[-1:0]`, which is empty. As a result, no environment is selected and the operation silently does nothing.

Other valid negative scalar indices produce nonempty slices. `-1` is the special case where `envs_idx + 1` becomes `0`.

Out-of-range positive scalar indices have a related problem because their slices are also empty. They are silently ignored instead of raising an error.

`_sanitize_envs_idx()` is shared by APIs including scene resets, sensor reads, camera pose updates, and rigid solver accessors, so this can affect multiple per-environment operations.

### Steps to Reproduce

The following example uses the public rigid-entity API:

  ```python
  import genesis as gs

  gs.init(backend=gs.cpu)

  box = scene.add_entity(
      gs.morphs.Box(size=(0.1, 0.1, 0.1)),
  )
  scene.build(n_envs=4)

  print("Before:")
  print(box.get_pos())

  box.set_pos((9.0, 9.0, 9.0), envs_idx=-1)

  print("After envs_idx=-1:")
  print(box.get_pos())

  box.set_pos((8.0, 8.0, 8.0), envs_idx=3)

  print("After envs_idx=3:")
  print(box.get_pos())
  ```

  Output:

  ```text
  Before:
  tensor([[0., 0., 0.],
          [0., 0., 0.],
          [0., 0., 0.],
          [0., 0., 0.]])

  After envs_idx=-1:
  tensor([[0., 0., 0.],
          [0., 0., 0.],
          [0., 0., 0.],
          [0., 0., 0.]])

  After envs_idx=3:
  tensor([[0., 0., 0.],
          [0., 0., 0.],
          [0., 0., 0.],
          [8., 8., 8.]])
  ```

  The negative index is silently ignored, while the equivalent positive index updates the last environment.

### Expected Behavior

`envs_idx=-1` should select the last environment, matching normal Python indexing and the behavior of other valid negative scalar indices.

Out-of-range scalar indices should raise an exception rather than silently selecting no environments.

### Screenshots/Videos

_No response_

### Relevant log output

```shell

```

### Environment

 - OS: macOS 15.7.3
  - GPU/CPU: Apple M4
  - Backend: CPU
  - GPU-driver version: N/A
  - Python: 3.11.14
  - Genesis: 1.2.3

78801c6c

### Additional Context

 The scalar slice branch was introduced in #1310.

I am preparing a small fix with a regression test and plan to open a pull request linked to this issue. The fix is scoped to scalar environment indices. The separate `slice`, `range`, and collection paths are left unchanged.

jeetrex17:
## Description

  `Scene._sanitize_envs_idx()` used a slice to preserve the dimension of scalar environment indices. For `envs_idx=-1`,
  this produced the empty slice `[-1:0]`, so the operation silently affected no environment.

  This changes the scalar branch to index the requested environment and then restore the one-element dimension. Valid
  negative scalar indices now follow normal Python indexing, and out-of-range scalar indices raise instead of being
  silently ignored.

  The change is limited to scalar integer indices. The `slice`, `range`, and collection paths are unchanged.

  ## Related Issue

  Resolves #3116

  ## Motivation and Context

  Using `envs_idx=-1` should select the last environment, but it previously produced no warning and made no change.
  Since `_sanitize_envs_idx()` is shared by several per-environment APIs, the same silent behavior could affect scene,
  sensor, camera, and rigid-body operations.

  The existing gravity test now uses `envs_idx=-1` to update the last of three environments. Its existing assertion
  verifies that only the last environment receives the new gravity.

  The regression test failed before the fix because the last environment remained unchanged, and passed after the fix.

  Tested on macOS 15.7.3 with an Apple M4, Python 3.11.14, and the CPU backend:

  ```bash
  pytest tests/rigid/test_dynamics.py::test_gravity -n 0 -v
  pytest 'tests/rigid/test_api.py::test_data_accessor[3-False-cpu]' -v
  pre-commit run --files genesis/engine/scene.py tests/rigid/test_dynamics.py
  ```

  Results:

  ```text
  test_gravity: 1 passed
  test_data_accessor[3-False-cpu]: 1 passed
  ruff check: Passed
  ruff format: Passed
  ```

  Not applicable.

  ## Checklist:

  - [x] I updated the [documentation](https://github.com/Genesis-Embodied-AI/genesis-doc) accordingly or no change is
  needed.

### DIFF
--- a/genesis/engine/scene.py
+++ b/genesis/engine/scene.py
@@ -1606,7 +1606,7 @@ class Scene(RBC):
         if isinstance(envs_idx, (slice, range)):
             return self._envs_idx[envs_idx]
         if isinstance(envs_idx, (int, np.integer)):
-            return self._envs_idx[envs_idx : envs_idx + 1]
+            return self._envs_idx[envs_idx][None]
 
         return sanitize_index(envs_idx, -1, self.n_envs, 0, "envs_idx")
 
--- a/genesis/engine/scene.py
+++ b/genesis/engine/scene.py
@@ -1610,7 +1610,7 @@ class Scene(RBC):
             if not -n_envs <= envs_idx < n_envs:
                 gs.raise_exception(f"`envs_idx` out of range: {envs_idx} not in [{-n_envs}, {n_envs}).")
             if envs_idx < 0:
-                envs_idx += n_envs
+                envs_idx = envs_idx + n_envs
             return self._envs_idx[envs_idx : envs_idx + 1]
 
         return sanitize_index(envs_idx, -1, self.n_envs, 0, "envs_idx")
--- a/genesis/engine/scene.py
+++ b/genesis/engine/scene.py
@@ -1606,7 +1606,12 @@ class Scene(RBC):
         if isinstance(envs_idx, (slice, range)):
             return self._envs_idx[envs_idx]
         if isinstance(envs_idx, (int, np.integer)):
-            return self._envs_idx[envs_idx][None]
+            n_envs = self.n_envs
+            if not -n_envs <= envs_idx < n_envs:
+                gs.raise_exception(f"`envs_idx` out of range: {envs_idx} not in [{-n_envs}, {n_envs}).")
+            if envs_idx < 0:
+                envs_idx += n_envs
+            return self._envs_idx[envs_idx : envs_idx + 1]
 
         return sanitize_index(envs_idx, -1, self.n_envs, 0, "envs_idx")
 
--- a/tests/rigid/test_dynamics.py
+++ b/tests/rigid/test_dynamics.py
@@ -26,7 +26,11 @@ def test_gravity(show_viewer, tol):
 
     scene.sim.set_gravity(torch.tensor([0.0, 0.0, 0.0]))
     scene.sim.set_gravity(torch.tensor([[1.0, 0.0, 0.0], [0.0, 2.0, 0.0]]), envs_idx=[0, 1])
+    scene.sim.set_gravity(torch.tensor([1.0, 0.0, 0.0]), envs_idx=np.int64(-3))
     scene.sim.set_gravity(torch.tensor([0.0, 0.0, 3.0]), envs_idx=-1)
+    for envs_idx in (-4, 3, np.int64(-4), np.int64(3)):
+        with pytest.raises(gs.GenesisException, match="`envs_idx` out of range"):
+            scene.sim.set_gravity(torch.tensor([0.0, 0.0, 0.0]), envs_idx=envs_idx)
     with np.testing.assert_raises(RuntimeError):
         scene.sim.set_gravity(torch.tensor([0.0, -10.0]))
     with np.testing.assert_raises(RuntimeError):
--- a/tests/rigid/test_dynamics.py
+++ b/tests/rigid/test_dynamics.py
@@ -26,7 +26,7 @@ def test_gravity(show_viewer, tol):
 
     scene.sim.set_gravity(torch.tensor([0.0, 0.0, 0.0]))
     scene.sim.set_gravity(torch.tensor([[1.0, 0.0, 0.0], [0.0, 2.0, 0.0]]), envs_idx=[0, 1])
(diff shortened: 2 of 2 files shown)


genesis/engine/scene.py
```diff
@@ -1606,7 +1606,7 @@ def _sanitize_envs_idx(
         if isinstance(envs_idx, (slice, range)):
             return self._envs_idx[envs_idx]
         if isinstance(envs_idx, (int, np.integer)):
-            return self._envs_idx[envs_idx : envs_idx + 1]
+            return self._envs_idx[envs_idx][None]
```
duburcqa: Are you sure this is the most efficient implementation? We are doing this a gazillions times, so it must be optimal.

jeetrex17: Good point. The current implementation is not optimal. `self._envs_idx[envs_idx][None]` performs two tensor indexing operations, while basic slicing only performs one.

I benchmarked the scalar path on CPU with 100,000 iterations, taking the minimum of three runs with `n_envs=4096`:

| Implementation | ~ns/call | `envs_idx=-1` | Out of range |
|---|---:|---|---|
| `self._envs_idx[envs_idx : envs_idx + 1]` (main) | 379 | empty selection | empty selection |
| `self._envs_idx[envs_idx][None]` (current PR) | 596 | last environment | `IndexError` |
| `self._envs_idx[envs_idx : envs_idx + 1 or None]` | 381 | last environment | empty selection |
| Bounds check, normalization, then one slice | 412 | last environment | raises |

The `or None` form is closest to the original performance, but it leaves out-of-range scalar indices as silent empty selections. Since #3116 asks for those indices to raise, I’m updating the PR to bounds-check and normalize the scalar index before using a single slice:

```python
if isinstance(envs_idx, (int, np.integer)):
    n_envs = self.n_envs
    if not -n_envs <= envs_idx < n_envs:
        gs.raise_exception(
            f"`envs_idx` out of range: {envs_idx} not in [{-n_envs}, {n_envs})."
        )
    if envs_idx < 0:
        envs_idx += n_envs
    return self._envs_idx[envs_idx : envs_idx + 1]
```

That is approximately 33 ns/call over main for the bounds check. The `envs_idx + 1 or None` form measured within noise of main and fixes `-1` alone. If you prefer the smallest possible hot-path overhead, I can limit this PR to `-1` and leave out-of-range validation for a separate change.

These are CPU-only microbenchmarks on an Apple M4. I have not measured the CUDA path.

genesis/engine/scene.py
```diff
@@ -1606,6 +1606,11 @@ def _sanitize_envs_idx(
         if isinstance(envs_idx, (slice, range)):
             return self._envs_idx[envs_idx]
         if isinstance(envs_idx, (int, np.integer)):
+            n_envs = self.n_envs
+            if not -n_envs <= envs_idx < n_envs:
+                gs.raise_exception(f"`envs_idx` out of range: {envs_idx} not in [{-n_envs}, {n_envs}).")
+            if envs_idx < 0:
+                envs_idx += n_envs
             return self._envs_idx[envs_idx : envs_idx + 1]
```
duburcqa, line 1614: We should probably just use `self._envs_idx.narrow(0, envs_idx, 1)`

duburcqa: nevermind. it is slow as fuck.

genesis/engine/scene.py
duburcqa: Good analysis. The current version is fine.

duburcqa (approved):

PR #3123, opened 2026-07-28
[BUG FIX] Fix delete weld constraint when not last added dynamic equality.

### LINKED ISSUE #3122: [Bug]: `delete_weld_constraint` removes the wrong weld when it is not the last dynamic equality
### Bug Description

`RigidSolver.delete_weld_constraint(link1_idx, link2_idx)` deletes the wrong
constraint whenever the weld being deleted is not the last dynamically added
equality. With two active welds A-B and C-D (added in that order), deleting
A-B actually removes C-D and leaves A-B active.

Root cause: `kernel_delete_weld_constraint` in
`genesis/engine/solvers/rigid/constraint/solver.py` implements deletion as a
swap-remove, but it only copies `eq_type` from the last slot into the deleted
slot — `eq_obj1id`, `eq_obj2id`, `eq_data`, and `sol_params` are left behind.
The surviving slot therefore keeps the deleted constraint's links and data,
while the last-added constraint's record is dropped when the count is
decremented.

Practical impact: in dual-arm manipulation where each gripper welds a grasped
object, releasing the first-grasped object silently detaches the other arm's
object instead (it falls), while the released object stays welded and gets
dragged back to the gripper. No error or warning is raised.

### Steps to Reproduce

```python
import genesis as gs

boxes = [
    scene.add_entity(gs.morphs.Box(size=(0.05, 0.05, 0.05), pos=(0.3 * i, 0.0, 0.5)))
    for i in range(4)
]
scene.build()

rs = scene.rigid_solver
a, b, c, d = (box.base_link.idx for box in boxes)

rs.add_weld_constraint(a, b)
rs.add_weld_constraint(c, d)

def show(tag):
    info = rs.constraint_solver.get_weld_constraints(as_tensor=True, to_torch=True)
    pairs = list(zip(info["link_a"].flatten().tolist(), info["link_b"].flatten().tolist()))
    print(f"{tag}: welds = {pairs}")
    return pairs

show("before delete")
rs.delete_weld_constraint(a, b)
pairs = show(f"after deleting ({a},{b})")
assert pairs == [(c, d)], f"BUG: expected [({c}, {d})], got {pairs}"
```

No assets needed. Reproduces on the CPU backend (the bug is pure kernel
logic, backend-independent). Deleting the *last-added* weld works correctly,
which is why single-weld usage never shows this.

### Expected Behavior

After deleting weld (a, b), the remaining weld should be (c, d). Instead,
`get_weld_constraints` reports (a, b) still active and (c, d) gone, and the
subsequent dynamics match that corrupted state.

### Screenshots/Videos

N/A (see log output).

### Relevant log output

```shell
link idxs: a=1 b=2 c=3 d=4
before delete: welds = [(1, 2), (3, 4)]
after deleting (1,2): welds = [(1, 2)]
AssertionError: BUG: expected [(3, 4)], got [(1, 2)]
```

### Environment

- OS: Arch Linux
- GPU/CPU: NVIDIA GeForce RTX 5090 / AMD Ryzen 9 9950X (repro uses gs.cpu backend)
- GPU-driver version: 580.105.08
- CUDA / CUDA-toolkit version: 13.0
- Python: 3.12.13

Found on 1.2.2 (commit 72bdd98b8d77d249e9c5296e4a6746ff680416a2, installed
from source). The same logic is still present on current `main`
(`kernel_delete_weld_constraint`, the swap-remove copies only
`dyn_info.equalities.eq_type`).

### Additional Context

Suggested fix — copy the full constraint record during the swap-remove
(field names as on current `main`):

```python
if i_e < constraint_state.qd_n_equalities[i_b] - 1:
    i_last = constraint_state.qd_n_equalities[i_b] - 1
    dyn_info.equalities.eq_type[i_e, i_b] = dyn_info.equalities.eq_type[i_last, i_b]
    dyn_info.equalities.eq_obj1id[i_e, i_b] = dyn_info.equalities.eq_obj1id[i_last, i_b]
    dyn_info.equalities.eq_obj2id[i_e, i_b] = dyn_info.equalities.eq_obj2id[i_last, i_b]
    dyn_info.equalities.eq_data[i_e, i_b] = dyn_info.equalities.eq_data[i_last, i_b]
    dyn_info.equalities.sol_params[i_e, i_b] = dyn_info.equalities.sol_params[i_last, i_b]
constraint_state.qd_n_equalities[i_b] = constraint_state.qd_n_equalities[i_b] - 1
```

I applied the equivalent patch locally (on 1.2.2 field names) and the repro
above then passes, with dual-arm grasp/release behaving correctly.

jimwang418:
## Description
`kernel_delete_weld_constraint` removes a weld via swap-remove, but only copied `eq_type` from the last equality slot into the deleted slot. The surviving slot therefore kept the deleted constraint's `eq_obj1id`, `eq_obj2id`, `eq_data`, and `sol_params`: deleting any weld that was not the last dynamically added equality effectively deleted the *last-added* weld instead, while the targeted weld stayed active.

This change copies the full constraint record (`eq_type`, `eq_obj1id`, `eq_obj2id`, `eq_data`, `sol_params`) during the swap.

## Related Issue
Resolves Genesis-Embodied-AI/Genesis#3122

## Motivation and Context
With two grippers each holding an object via a weld, releasing the first-grasped object silently detached the *other* arm's object (it falls) while the released object stayed welded — with no error raised. Any workflow that deletes a weld while other dynamic welds exist is affected.

Added `test_dynamic_weld_delete_keeps_other_welds` in `tests/rigid/test_constraints.py`: welds A-B and C-D, deletes A-B, and checks via `get_weld_constraints` that C-D survives and that C and D keep their relative pose while falling (which fails if the swap leaves the deleted weld's data behind). The test fails on `main` without this fix and passes with it (verified on Linux, CPU backend, and it is backend-independent kernel logic).

Minimal repro from the linked issue:

```python
import genesis as gs

boxes = [
    scene.add_entity(gs.morphs.Box(size=(0.05, 0.05, 0.05), pos=(0.3 * i, 0.0, 0.5)))
    for i in range(4)
]
scene.build()

rs = scene.rigid_solver
a, b, c, d = (box.base_link.idx for box in boxes)
rs.add_weld_constraint(a, b)
rs.add_weld_constraint(c, d)
rs.delete_weld_constraint(a, b)
info = rs.get_weld_constraints(as_tensor=True, to_torch=True)
pairs = list(zip(info["link_a"].flatten().tolist(), info["link_b"].flatten().tolist()))
assert pairs == [(c, d)], f"expected [({c}, {d})], got {pairs}"  # fails on main
```

N/A

## Checklist:

### DIFF
--- a/tests/rigid/test_constraints.py
+++ b/tests/rigid/test_constraints.py
@@ -176,6 +176,46 @@ def test_dynamic_weld_scene_reset():
 
 
 @pytest.mark.required
+def test_dynamic_weld_delete_keeps_other_welds(show_viewer, tol):
+    scene = gs.Scene(
+        show_viewer=show_viewer,
+        show_FPS=False,
+    )
+    scene.add_entity(
+        gs.morphs.Plane(),
+    )
+    boxes = [
+        scene.add_entity(
+            gs.morphs.Box(
+                size=(0.05, 0.05, 0.05),
+                pos=(0.3 * i, 0.0, 0.5),
+            )
+        )
+        for i in range(4)
+    ]
+    scene.build()
+
+    solver = scene.sim.rigid_solver
+    link_a, link_b, link_c, link_d = (box.base_link.idx for box in boxes)
+    solver.add_weld_constraint(link_a, link_b)
+    solver.add_weld_constraint(link_c, link_d)
+
+    # Deleting a weld that is not the last dynamic equality must remove that weld and
+    # preserve the full record of the surviving one, not just its type.
+    solver.delete_weld_constraint(link_a, link_b)
+    weld_const_info = solver.get_weld_constraints(as_tensor=True, to_torch=True)
+    assert_allclose(weld_const_info["link_a"].flatten(), (link_c,), tol=0)
+    assert_allclose(weld_const_info["link_b"].flatten(), (link_d,), tol=0)
+
+    # The surviving weld must still act on its own links: c and d keep their relative
+    # pose while falling, which fails if the swap left behind the deleted weld's data.
+    rel_pos = boxes[3].get_pos() - boxes[2].get_pos()
+    for _ in range(100):
+        scene.step()
+    assert_allclose(boxes[3].get_pos() - boxes[2].get_pos(), rel_pos, tol=1e-3)
+
+
+@pytest.mark.required
 def test_urdf_mimic(show_viewer, tol):
     # create and build the scene
     scene = gs.Scene(
--- a/tests/rigid/test_constraints.py
+++ b/tests/rigid/test_constraints.py
@@ -9,6 +9,7 @@ from genesis.utils.misc import tensor_to_array
 
 from ..utils import (
     assert_allclose,
+    assert_equal,
     simulate_and_check_mujoco_consistency,
 )
 
@@ -65,6 +66,7 @@ def test_equality_link(gs_sim, mj_sim, gs_solver, xml_path):
 @pytest.mark.required
 def test_dynamic_weld(show_viewer, tol):
     CUBE_POS = (0.65, 0.0, 0.02)
+    HANGING_BOX_POS = (0.0, 1.0, 0.4)
 
     scene = gs.Scene(
         viewer_options=gs.options.ViewerOptions(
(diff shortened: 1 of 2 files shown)


tests/rigid/test_constraints.py
```diff
@@ -175,6 +175,46 @@ def test_dynamic_weld_scene_reset():
     assert solver.constraint_solver.constraint_state.qd_n_equalities[1] == n_eq_base + 1
 
 
+@pytest.mark.required
+def test_dynamic_weld_delete_keeps_other_welds(show_viewer, tol):
```
duburcqa: Do not add a new unit test but extend the existing one, in accordance with the guidelines.

jimwang418: Folded the test into test_dynamic_weld.

jimwang418:
Just reworked the unit test into the existing test_dynamic_weld.

duburcqa (approved):

PR #3128, opened 2026-07-29
[FEATURE] Add method for querying terrain height.

### LINKED ISSUE #2094: [Feature]: Get_Height interface when using terrain

jeetrex17:
## Description

This adds `get_terrain_height(positions, envs_idx=None)` for querying terrain surface heights at world-frame x-y positions.

The query follows the terrain's piecewise-planar mesh and supports terrain translation, yaw, per-environment poses, shared point sets, explicit per-environment point sets, and visual-only kinematic terrains. The point dimension is preserved unless the input is the scalar form `(2,)`.

Queries up to one grid cell outside the terrain clamp to the nearest edge. Queries farther outside, non-finite positions, and terrains with roll or pitch return `NaN` per point. Malformed positions and calls on other entity types raise `GenesisException`.

The zero-copy path uses TorchScript and the non-zero-copy path uses a Quadrants kernel. Both paths remain on device. Terrain height data is cached when the terrain is loaded.

## Related Issue

Resolves #2094

Supersedes #2691, which was closed without merging. This implementation matches the simulated triangle surface, uses consistent height-field axes, and honors the current terrain pose.

## Motivation and Context

Locomotion tasks need the ground height below robot feet for observations, rewards, and swing-height control. Reading only height-field vertices does not provide the surface height between vertices.

This implementation evaluates the same two triangles used to construct each terrain cell, so the returned height matches the terrain mesh used by the simulator.

The focused test compares queried heights with ray intersections against the generated terrain mesh. It also covers a non-square height field, scaling, translation, yaw, per-environment poses, point layouts, environment selection, visual-only terrain, edge clamping, and `NaN` results.

```bash
python -m pytest tests/rigid/test_terrain.py::test_get_terrain_height -n 0 --backend cpu -q
GS_ENABLE_ZEROCOPY=0 python -m pytest tests/rigid/test_terrain.py::test_get_terrain_height -n 0 --backend cpu -q
python -m pytest tests/rigid/test_terrain.py::test_get_terrain_height -n 0 --backend gpu -q
GS_ENABLE_ZEROCOPY=0 python -m pytest tests/rigid/test_terrain.py::test_get_terrain_height -n 0 --backend gpu -q
python -m pre_commit run --files \
  genesis/engine/entities/rigid_entity/rigid_entity.py \
  genesis/engine/solvers/kinematic_solver.py \
  genesis/engine/solvers/rigid/abd/accessor.py \
  genesis/engine/solvers/rigid/collider/collider.py \
  genesis/engine/solvers/rigid/rigid_solver.py \
  genesis/utils/geom.py \
  genesis/utils/terrain.py \
  tests/rigid/test_terrain.py
```

Results on macOS with an Apple M4:

- Focused test: 2 passed in each of the four backend and zero-copy combinations.
- Ruff check and formatting: passed.
- Python bytecode compilation for all changed Python files: passed.

Warm timings through the public getter using 8 environments and a 65 by 65 terrain. Values are the minimum of five synchronized runs after warm-up:

| Backend | Path | 128 points per environment | 16,384 points per environment |
| --- | --- | ---: | ---: |
| CPU | Scripted Torch | 132 us | 2,329 us |
| CPU | Quadrants | 76 us | 738 us |
| Metal | Scripted Torch | 265 us | 986 us |
| Metal | Quadrants | 1,566 us | 1,769 us |

Dispatch follows `gs.use_zerocopy`. CUDA was not measured.

Not applicable.

## Checklist:

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -160,6 +160,9 @@ class KinematicEntity(Entity):
 
         self.terrain_hf: np.ndarray | None = None
         self.terrain_scale: np.ndarray | None = None
+        self._terrain_height_field: torch.Tensor | None = None
+        self._terrain_row_boundaries: torch.Tensor | None = None
+        self._terrain_col_boundaries: torch.Tensor | None = None
 
         self._load_model()
 
@@ -589,8 +592,7 @@ class KinematicEntity(Entity):
         self._terrain_height_field = torch.as_tensor(
             self.terrain_hf * self.terrain_scale[1], dtype=gs.tc_float, device=gs.device
         )
-        self._terrain_horizontal_scale = float(morph.horizontal_scale)
-        # Bucketize boundaries produce native 64-bit integer cell indices without a runtime dtype cast.
+        # Bucketization produces native 64-bit integer cell indices for advanced indexing
         self._terrain_row_boundaries = torch.arange(
             1, self.terrain_hf.shape[0] - 1, dtype=gs.tc_float, device=gs.device
         )
@@ -1972,16 +1974,16 @@ class KinematicEntity(Entity):
 
         Heights match the piecewise-planar surface on which rigid bodies rest. Terrain translation, yaw, and
         environment-specific poses are applied to the query. Positions up to one grid cell outside the terrain are
-        clamped to its edge. Non-finite positions, positions farther outside, and terrains with roll or pitch produce
-        not-a-number (NaN) heights.
+        clamped to its edge. The query returns not-a-number (NaN) heights for positions containing NaN or infinity,
+        positions farther outside, and terrains with roll or pitch.
 
         Parameters
         ----------
         positions : array_like
-            World-frame x-y positions in meters, with shape (2,), (n_points, 2), or (n_envs, n_points, 2). A
-            two-dimensional array is shared across the selected environments; use a three-dimensional array for
-            environment-specific positions. A leading dimension of 1 in the three-dimensional form is also treated as
-            shared.
+            World-frame x-y positions in meters, with shape (2,), (n_points, 2), or
+            (n_selected_envs, n_points, 2). A two-dimensional array is shared across the selected environments; use a
+            three-dimensional array for environment-specific positions. A leading dimension of 1 in the
+            three-dimensional form is also treated as shared.
         envs_idx : None | array_like, optional
             The indices of the environments. If None, all environments will be considered. Defaults to None.
 
@@ -1994,20 +1996,12 @@ class KinematicEntity(Entity):
         Raises
         ------
         GenesisException
-            If this entity is not a terrain or `positions` is malformed.
+            If the entity type or the shape of `positions` is unsupported.
         """
-        if self.terrain_hf is None or self.terrain_scale is None:
+        if self.terrain_hf is None:
             gs.raise_exception("`get_terrain_height()` is only supported for terrain entities.")
 
-        return self._solver.get_terrain_height(
-            positions,
-            self.base_link_idx,
-            self._terrain_height_field,
-            self._terrain_horizontal_scale,
-            self._terrain_row_boundaries,
-            self._terrain_col_boundaries,
-            envs_idx,
-        )
+        return self._solver.get_terrain_height(positions, self.base_link_idx, envs_idx)
 
     @gs.assert_built
     def get_links_vel(self, links_idx_local=None, envs_idx=None):
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -585,11 +585,12 @@ class KinematicEntity(Entity):
         return g_infos
 
     def _load_terrain(self, morph, surface):
-        vmesh, mesh, self.terrain_hf = tu.parse_terrain(morph, surface)
+        vmesh, mesh, terrain_hf = tu.parse_terrain(morph, surface)
         self.terrain_scale = np.array((morph.horizontal_scale, morph.vertical_scale), dtype=gs.np_float)
-        self._terrain_height_field = torch.as_tensor(
-            self.terrain_hf * self.terrain_scale[1], dtype=gs.tc_float, device=gs.device
-        )
+        self.terrain_hf = terrain_hf * self.terrain_scale[1]
+        # Collider storage covers only the first collision terrain; height queries support every terrain entity,
+        # including kinematic ones.
+        self._terrain_height_field = torch.as_tensor(self.terrain_hf, dtype=gs.tc_float, device=gs.device)
 
         g_infos = []
         if morph.visualization:
@@ -1965,8 +1966,9 @@ class KinematicEntity(Entity):
 
         Heights match the piecewise-planar surface on which rigid bodies rest. Terrain translation, yaw, and
         environment-specific poses are applied to the query. Positions up to one grid cell outside the terrain are
-        clamped to its edge. The query returns not-a-number (NaN) heights for positions containing NaN or infinity,
-        positions farther outside, and terrains with roll or pitch.
+        clamped to its edge. Terrain tilt up to 0.001 radians from world vertical is treated as numerical noise. The
+        query returns not-a-number (NaN) heights for positions containing NaN or infinity, positions farther outside,
+        and terrains with greater tilt.
 
         Parameters
         ----------
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -158,6 +158,9 @@ class KinematicEntity(Entity):
         self._variant_offset_pos: list[np.ndarray] | None = None
         self._variant_offset_quat: list[np.ndarray] | None = None
 
+        self.terrain_hf: np.ndarray | None = None
+        self.terrain_scale: np.ndarray | None = None
+
         self._load_model()
 
         # Initialize target variables and checkpoint
@@ -2504,6 +2507,10 @@ class RigidEntity(KinematicEntity):
         self._visualize_contact: bool = visualize_contact
 
         self._batch_fixed_verts: bool = morph.batch_fixed_verts
+        self._terrain_height_field: torch.Tensor | None = None
+        self._terrain_horizontal_scale: torch.Tensor | None = None
+        self._terrain_row_boundaries: torch.Tensor | None = None
+        self._terrain_col_boundaries: torch.Tensor | None = None
 
         super().__init__(
             scene,
@@ -3614,6 +3621,60 @@ class RigidEntity(KinematicEntity):
 
     def get_aabb(self):
         raise DeprecationError("This method has been removed. Please use 'get_AABB()' instead.")
+
+    @gs.assert_built
+    def get_height(self, positions, envs_idx=None):
+        """
+        Return terrain surface heights in meters at world-frame XY positions.
+
+        Heights match the piecewise-planar surface on which rigid bodies rest. Terrain translation, yaw, and
+        environment-specific poses are applied to the query.
+
+        Parameters
+        ----------
+        positions : array_like
+            World-frame XY positions in meters, with shape (2,), (n_points, 2), or (n_envs, n_points, 2). A 2D array
+            is shared across the selected environments; use a 3D array for environment-specific positions. A leading
+            dimension of 1 in the 3D form is also treated as shared.
+        envs_idx : None | array_like, optional
+            The indices of the environments. If None, all environments will be considered. Defaults to None.
+
+        Returns
+        -------
+        heights : torch.Tensor
+            World-frame surface heights in meters. The point dimensions match `positions`, with an environment
+            dimension prepended in a parallelized scene.
+
+        Raises
+        ------
+        GenesisException
+            If this entity is not a terrain, the positions are malformed, non-finite, or outside the terrain, or the
+            terrain has roll or pitch.
+        """
+        if self.terrain_hf is None or self.terrain_scale is None:
+            gs.raise_exception("`get_height()` is only supported for terrain entities.")
+        if self._terrain_height_field is None or self._terrain_horizontal_scale is None:
+            height_field = self.terrain_hf * self.terrain_scale[1]
+            self._terrain_height_field = torch.as_tensor(height_field, dtype=gs.tc_float, device=gs.device).contiguous()
+            self._terrain_horizontal_scale = torch.as_tensor(
+                self.terrain_scale[:1], dtype=gs.tc_float, device=gs.device
+            ).contiguous()
+            self._terrain_row_boundaries = torch.arange(
+                1, self.terrain_hf.shape[0] - 1, dtype=gs.tc_float, device=gs.device
+            )
+            self._terrain_col_boundaries = torch.arange(
+                1, self.terrain_hf.shape[1] - 1, dtype=gs.tc_float, device=gs.device
+            )
+
+        return self._solver.get_terrain_height(
+            positions,
+            self.base_link_idx,
+            self._terrain_height_field,
+            self._terrain_horizontal_scale,
+            self._terrain_row_boundaries,
+            self._terrain_col_boundaries,
+            envs_idx,
+        )
 
     @gs.assert_built
     def get_links_pos(
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -160,6 +160,10 @@ class KinematicEntity(Entity):
 
         self.terrain_hf: np.ndarray | None = None
         self.terrain_scale: np.ndarray | None = None
+        self._terrain_height_field: torch.Tensor | None = None
+        self._terrain_horizontal_scale: torch.Tensor | None = None
+        self._terrain_row_boundaries: torch.Tensor | None = None
+        self._terrain_col_boundaries: torch.Tensor | None = None
 
         self._load_model()
 
@@ -1955,6 +1959,60 @@ class KinematicEntity(Entity):
         return torch.stack((aabbs[..., 0, :].min(dim=-2).values, aabbs[..., 1, :].max(dim=-2).values), dim=-2)
 
     @gs.assert_built
+    def get_height(self, positions, envs_idx=None):
+        """
+        Return terrain surface heights in meters at world-frame XY positions.
+
+        Heights match the piecewise-planar surface on which rigid bodies rest. Terrain translation, yaw, and
+        environment-specific poses are applied to the query.
+
+        Parameters
+        ----------
+        positions : array_like
+            World-frame XY positions in meters, with shape (2,), (n_points, 2), or (n_envs, n_points, 2). A 2D array
+            is shared across the selected environments; use a 3D array for environment-specific positions. A leading
+            dimension of 1 in the 3D form is also treated as shared.
+        envs_idx : None | array_like, optional
+            The indices of the environments. If None, all environments will be considered. Defaults to None.
+
+        Returns
+        -------
+        heights : torch.Tensor
+            World-frame surface heights in meters. The point dimensions match `positions`, with an environment
+            dimension prepended in a parallelized scene.
+
+        Raises
+        ------
+        GenesisException
+            If this entity is not a terrain, the positions are malformed, non-finite, or outside the terrain, or the
+            terrain has roll or pitch.
+        """
+        if self.terrain_hf is None or self.terrain_scale is None:
+            gs.raise_exception("`get_height()` is only supported for terrain entities.")
+        if self._terrain_height_field is None or self._terrain_horizontal_scale is None:
+            height_field = self.terrain_hf * self.terrain_scale[1]
+            self._terrain_height_field = torch.as_tensor(height_field, dtype=gs.tc_float, device=gs.device).contiguous()
+            self._terrain_horizontal_scale = torch.as_tensor(
+                self.terrain_scale[:1], dtype=gs.tc_float, device=gs.device
+            ).contiguous()
+            self._terrain_row_boundaries = torch.arange(
+                1, self.terrain_hf.shape[0] - 1, dtype=gs.tc_float, device=gs.device
+            )
+            self._terrain_col_boundaries = torch.arange(
+                1, self.terrain_hf.shape[1] - 1, dtype=gs.tc_float, device=gs.device
+            )
+
+        return self._solver.get_terrain_height(
+            positions,
+            self.base_link_idx,
+            self._terrain_height_field,
+            self._terrain_horizontal_scale,
+            self._terrain_row_boundaries,
+            self._terrain_col_boundaries,
+            envs_idx,
+        )
+
+    @gs.assert_built
     def get_links_vel(self, links_idx_local=None, envs_idx=None):
         """
         Returns linear velocity of all the entity's links expressed at a given reference position in world coordinates.
@@ -2507,10 +2565,6 @@ class RigidEntity(KinematicEntity):
         self._visualize_contact: bool = visualize_contact
 
         self._batch_fixed_verts: bool = morph.batch_fixed_verts
-        self._terrain_height_field: torch.Tensor | None = None
-        self._terrain_horizontal_scale: torch.Tensor | None = None
-        self._terrain_row_boundaries: torch.Tensor | None = None
-        self._terrain_col_boundaries: torch.Tensor | None = None
 
         super().__init__(
             scene,
@@ -3621,60 +3675,6 @@ class RigidEntity(KinematicEntity):
 
     def get_aabb(self):
         raise DeprecationError("This method has been removed. Please use 'get_AABB()' instead.")
-
-    @gs.assert_built
-    def get_height(self, positions, envs_idx=None):
-        """
-        Return terrain surface heights in meters at world-frame XY positions.
-
-        Heights match the piecewise-planar surface on which rigid bodies rest. Terrain translation, yaw, and
-        environment-specific poses are applied to the query.
-
-        Parameters
-        ----------
-        positions : array_like
-            World-frame XY positions in meters, with shape (2,), (n_points, 2), or (n_envs, n_points, 2). A 2D array
-            is shared across the selected environments; use a 3D array for environment-specific positions. A leading
-            dimension of 1 in the 3D form is also treated as shared.
-        envs_idx : None | array_like, optional
-            The indices of the environments. If None, all environments will be considered. Defaults to None.
-
-        Returns
-        -------
-        heights : torch.Tensor
-            World-frame surface heights in meters. The point dimensions match `positions`, with an environment
-            dimension prepended in a parallelized scene.
-
-        Raises
-        ------
-        GenesisException
-            If this entity is not a terrain, the positions are malformed, non-finite, or outside the terrain, or the
-            terrain has roll or pitch.
-        """
-        if self.terrain_hf is None or self.terrain_scale is None:
-            gs.raise_exception("`get_height()` is only supported for terrain entities.")
-        if self._terrain_height_field is None or self._terrain_horizontal_scale is None:
-            height_field = self.terrain_hf * self.terrain_scale[1]
-            self._terrain_height_field = torch.as_tensor(height_field, dtype=gs.tc_float, device=gs.device).contiguous()
-            self._terrain_horizontal_scale = torch.as_tensor(
-                self.terrain_scale[:1], dtype=gs.tc_float, device=gs.device
-            ).contiguous()
-            self._terrain_row_boundaries = torch.arange(
-                1, self.terrain_hf.shape[0] - 1, dtype=gs.tc_float, device=gs.device
-            )
-            self._terrain_col_boundaries = torch.arange(
-                1, self.terrain_hf.shape[1] - 1, dtype=gs.tc_float, device=gs.device
-            )
-
-        return self._solver.get_terrain_height(
-            positions,
-            self.base_link_idx,
-            self._terrain_height_field,
-            self._terrain_horizontal_scale,
-            self._terrain_row_boundaries,
-            self._terrain_col_boundaries,
-            envs_idx,
-        )
 
     @gs.assert_built
     def get_links_pos(
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -161,8 +161,6 @@ class KinematicEntity(Entity):
         self.terrain_hf: np.ndarray | None = None
         self.terrain_scale: np.ndarray | None = None
         self._terrain_height_field: torch.Tensor | None = None
-        self._terrain_row_boundaries: torch.Tensor | None = None
-        self._terrain_col_boundaries: torch.Tensor | None = None
 
         self._load_model()
 
@@ -591,13 +589,6 @@ class KinematicEntity(Entity):
         self.terrain_scale = np.array((morph.horizontal_scale, morph.vertical_scale), dtype=gs.np_float)
         self._terrain_height_field = torch.as_tensor(
             self.terrain_hf * self.terrain_scale[1], dtype=gs.tc_float, device=gs.device
-        )
-        # Bucketization produces native 64-bit integer cell indices for advanced indexing
-        self._terrain_row_boundaries = torch.arange(
-            1, self.terrain_hf.shape[0] - 1, dtype=gs.tc_float, device=gs.device
-        )
-        self._terrain_col_boundaries = torch.arange(
-            1, self.terrain_hf.shape[1] - 1, dtype=gs.tc_float, device=gs.device
         )
 
         g_infos = []
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -160,10 +160,6 @@ class KinematicEntity(Entity):
 
         self.terrain_hf: np.ndarray | None = None
         self.terrain_scale: np.ndarray | None = None
-        self._terrain_height_field: torch.Tensor | None = None
-        self._terrain_horizontal_scale: torch.Tensor | None = None
-        self._terrain_row_boundaries: torch.Tensor | None = None
-        self._terrain_col_boundaries: torch.Tensor | None = None
 
         self._load_model()
 
@@ -590,6 +586,17 @@ class KinematicEntity(Entity):
     def _load_terrain(self, morph, surface):
         vmesh, mesh, self.terrain_hf = tu.parse_terrain(morph, surface)
         self.terrain_scale = np.array((morph.horizontal_scale, morph.vertical_scale), dtype=gs.np_float)
+        self._terrain_height_field = torch.as_tensor(
+            self.terrain_hf * self.terrain_scale[1], dtype=gs.tc_float, device=gs.device
+        )
+        self._terrain_horizontal_scale = float(morph.horizontal_scale)
+        # Bucketize boundaries produce native 64-bit integer cell indices without a runtime dtype cast.
+        self._terrain_row_boundaries = torch.arange(
+            1, self.terrain_hf.shape[0] - 1, dtype=gs.tc_float, device=gs.device
+        )
+        self._terrain_col_boundaries = torch.arange(
+            1, self.terrain_hf.shape[1] - 1, dtype=gs.tc_float, device=gs.device
+        )
 
         g_infos = []
         if morph.visualization:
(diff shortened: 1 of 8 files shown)


tests/rigid/test_terrain.py
```diff
@@ -134,6 +132,147 @@ def test_discrete_obstacles():
...
+    with pytest.raises(gs.GenesisException):
+        terrain.get_height(np.empty((0, 2)), envs_idx=envs_idx)
+    with pytest.raises(gs.GenesisException):
+        terrain.get_height(np.empty((3, 1, 2)), envs_idx=envs_idx)
+    with pytest.raises(gs.GenesisException):
+        terrain.get_height((np.nan, 20.0), envs_idx=envs_idx)
+    with pytest.raises(gs.GenesisException):
+        terrain.get_height((np.inf, 20.0), envs_idx=envs_idx)
+    with pytest.raises(gs.GenesisException):
+        box.get_height((0.0, 0.0), envs_idx=envs_idx)
```
duburcqa: I don't think it is necessary to check all these edge-cases. 2-3 max should be enough.

tests/rigid/test_terrain.py
```diff
@@ -134,6 +132,147 @@ def test_discrete_obstacles():
...
+
+        heights_env_1 = terrain.get_height(positions_per_env[1], envs_idx=1)
+        assert_allclose(heights_env_1, ((-1.55, -0.9),), tol=tol)
+    else:
+        heights = terrain.get_height(positions)
+        assert heights.shape == (len(positions),)
+
+    assert_allclose(heights, (expected,) if n_envs else expected, tol=tol)
+    assert heights.dtype == gs.tc_float
+    assert heights.device.type == gs.device.type
```
duburcqa: This is not part of the specification we want to guarantee I think.

tests/rigid/test_terrain.py
```diff
@@ -134,6 +132,147 @@ def test_discrete_obstacles():
...
+
+        one_position_per_env = terrain.get_height(((positions_per_env[0][0],), (positions_per_env[1][0],)))
+        assert_allclose(one_position_per_env, (1.45, -1.55), tol=tol)
+        assert one_position_per_env.shape == (2,)
+
+        heights_env_1 = terrain.get_height(positions_per_env[1], envs_idx=1)
+        assert_allclose(heights_env_1, ((-1.55, -0.9),), tol=tol)
+    else:
+        heights = terrain.get_height(positions)
+        assert heights.shape == (len(positions),)
```
duburcqa: I don't think this is necessary. something else will fail if wrong.

tests/rigid/test_terrain.py
```diff
@@ -134,6 +132,147 @@ def test_discrete_obstacles():
...
+
+    assert_allclose(heights, (expected,) if n_envs else expected, tol=tol)
```
duburcqa: Are you sure `(expected,) if n_envs else expected` is necessary? `expected` should be just fine in both cases.

tests/rigid/test_terrain.py
```diff
@@ -134,6 +132,147 @@ def test_discrete_obstacles():
...
+        terrain.get_height((np.nan, 20.0), envs_idx=envs_idx)
+    with pytest.raises(gs.GenesisException):
+        terrain.get_height((np.inf, 20.0), envs_idx=envs_idx)
+    with pytest.raises(gs.GenesisException):
+        box.get_height((0.0, 0.0), envs_idx=envs_idx)
+
+    terrain.set_quat(
+        gu.xyz_to_quat(np.array((90.0, 0.0, 0.0)), degrees=True),
+        envs_idx=envs_idx,
+        relative=False,
```
duburcqa: It does not fit on 1 line?

tests/rigid/test_terrain.py
```diff
@@ -134,6 +132,147 @@ def test_discrete_obstacles():
...
+    )
+    with pytest.raises(gs.GenesisException):
+        terrain.get_height((10.0, 20.0), envs_idx=envs_idx)
```
duburcqa: It would make more sense to do 1 query within range instead of 1 that is out.

tests/rigid/test_terrain.py
```diff
@@ -134,6 +132,147 @@ def test_discrete_obstacles():
...
+        positions_per_env = (
+            (
+                positions[3],
+                positions[4],
+            ),
+            (
+                (-3.25, 4.125),
+                (-3.25, 4.375),
+            ),
+        )
```
duburcqa: 1 line.

tests/rigid/test_terrain.py
```diff
@@ -134,6 +132,147 @@ def test_discrete_obstacles():
...
+        heights_per_env = terrain.get_height(positions_per_env)
+        assert_allclose(heights_per_env, ((1.45, 2.1), (-1.55, -0.9)), tol=tol)
+        assert heights_per_env.shape == (2, 2)
```
duburcqa: Remove this.

tests/rigid/test_terrain.py
```diff
@@ -134,6 +132,147 @@ def test_discrete_obstacles():
...
+
+        one_position_per_env = terrain.get_height(((positions_per_env[0][0],), (positions_per_env[1][0],)))
+        assert_allclose(one_position_per_env, (1.45, -1.55), tol=tol)
+        assert one_position_per_env.shape == (2,)
```
duburcqa: Remove this.

tests/rigid/test_terrain.py
```diff
@@ -134,6 +132,147 @@ def test_discrete_obstacles():
...
+
+        heights_env_1 = terrain.get_height(positions_per_env[1], envs_idx=1)
+        assert_allclose(heights_env_1, ((-1.55, -0.9),), tol=tol)
```
duburcqa: Remove extra parenthesis: `(-1.55, -0.9)`

tests/rigid/test_terrain.py
```diff
@@ -134,6 +132,147 @@ def test_discrete_obstacles():
...
+
+    visual_height = visual_terrain.get_height((30.0, 40.0))
+    assert_allclose(visual_height, (4.0, 4.0) if n_envs else 4.0, tol=tol)
+    visual_height_max = visual_terrain.get_height((31.0, 41.0))
+    assert_allclose(visual_height_max, (4.0, 4.0) if n_envs else 4.0, tol=tol)
+
+    if n_envs:
+        shared_positions = terrain.get_height(((2.0, -1.0), (2.0, -0.5)))
+        assert_allclose(shared_positions, ((0.3, 1.3), (0.3, 1.3)), tol=tol)
+        assert shared_positions.shape == (2, 2)
```
duburcqa: Remove this.

tests/rigid/test_terrain.py
```diff
@@ -134,6 +132,147 @@ def test_discrete_obstacles():
...
+        shared_positions_explicit = terrain.get_height((((2.0, -1.0), (2.0, -0.5)),))
+        assert_allclose(shared_positions_explicit, ((0.3, 1.3), (0.3, 1.3)), tol=tol)
+        assert shared_positions_explicit.shape == (2, 2)
```
duburcqa: Remove this.

tests/rigid/test_terrain.py
```diff
@@ -134,6 +132,147 @@ def test_discrete_obstacles():
...
+    scene.build(n_envs=n_envs)
+
+    initial_height = terrain.get_height((2.0, -0.5))
+    assert_allclose(initial_height, 1.3, tol=tol)
+    assert initial_height.shape == ((2,) if n_envs else ())
+
+    visual_height = visual_terrain.get_height((30.0, 40.0))
+    assert_allclose(visual_height, (4.0, 4.0) if n_envs else 4.0, tol=tol)
+    visual_height_max = visual_terrain.get_height((31.0, 41.0))
+    assert_allclose(visual_height_max, (4.0, 4.0) if n_envs else 4.0, tol=tol)
```
duburcqa: `4.0` is sufficient.

tests/rigid/test_terrain.py
```diff
@@ -134,6 +132,147 @@ def test_discrete_obstacles():
...
+        ),
+    )
+    scene.build(n_envs=n_envs)
+
+    initial_height = terrain.get_height((2.0, -0.5))
+    assert_allclose(initial_height, 1.3, tol=tol)
+    assert initial_height.shape == ((2,) if n_envs else ())
+
+    visual_height = visual_terrain.get_height((30.0, 40.0))
+    assert_allclose(visual_height, (4.0, 4.0) if n_envs else 4.0, tol=tol)
```
duburcqa: `4.0` is sufficient.

tests/rigid/test_terrain.py
```diff
@@ -134,6 +132,147 @@ def test_discrete_obstacles():
...
+    box = scene.add_entity(
+        morph=gs.morphs.Box(
+            size=(0.1, 0.1, 0.1),
+        ),
+    )
+    scene.build(n_envs=n_envs)
+
+    initial_height = terrain.get_height((2.0, -0.5))
+    assert_allclose(initial_height, 1.3, tol=tol)
+    assert initial_height.shape == ((2,) if n_envs else ())
```
duburcqa: Remove this.

tests/rigid/test_terrain.py
```diff
@@ -134,6 +132,147 @@ def test_discrete_obstacles():
...
+            height_field=height_field,
+            horizontal_scale=0.5,
+            vertical_scale=0.1,
+            batch_fixed_verts=True,
+        ),
+    )
+    visual_terrain = scene.add_entity(
+        morph=gs.morphs.Terrain(
+            pos=(30.0, 40.0, 3.0),
+            height_field=np.full((2, 2), 100.0, dtype=gs.np_float),
```
duburcqa: Are you sure using full instead of different values is not weakening the test?

jeetrex17:
Thanks for the review. I addressed all the comments in 289d6332.

- Reduced the invalid-input checks to three representative cases.
- Removed redundant shape, dtype, and device assertions.
- Simplified the expected values and formatting.
- Changed the kinematic terrain test to use distinct height values.
- Changed the roll and pitch check to query an in-range point.

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -1057,6 +1077,113 @@ def set_dofs_position(self, position, dofs_idx=None, envs_idx=None):
...
+            is_per_env = positions.shape[0] == n_envs
+        else:
+            gs.raise_exception("`positions` must have shape (2,), (n_points, 2), or (n_envs, n_points, 2).")
+
+        if positions.shape[-2] == 0:
+            gs.raise_exception("`positions` must contain at least one position.")
+
+        positions = positions.contiguous()
+        n_points = positions.shape[-2]
+        if gs.backend == gs.metal:
```
duburcqa: We avoid backend-specific torch "fallback" this in Genesis. This case is supported, but the trigger is whether zero-copy is supported or not.

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -1057,6 +1077,113 @@ def set_dofs_position(self, position, dofs_idx=None, envs_idx=None):
...
+
+            has_invalid_query = (
+                (link_quat[..., 1:3].abs() > gs.EPS).any()
+                | (~torch.isfinite(row)).any()
+                | (~torch.isfinite(col)).any()
+                | (row < -gs.EPS).any()
+                | (row > row_max + gs.EPS).any()
+                | (col < -gs.EPS).any()
+                | (col > col_max + gs.EPS).any()
+            )
```
duburcqa: Are you sure this implementation is optimal? If you think torch is superior, for such a complex logics it is better to use torch.compile / jit. See geom.py.

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -95,6 +96,25 @@ def _offset_world_shift(offset_pos, offset_quat, world_quat):
     return gu.transform_by_quat(offset_pos, user_quat)
 
 
+def _interpolate_terrain_height(height_field, row, col, row_boundaries, col_boundaries):
```
duburcqa: Use torch compile / jit. See geom.py or tactile sensors.

jeetrex17:
Thanks for the detailed review, and sorry for the extra review cycles. I updated the implementation to use gs.use_zerocopy, moved the complete Torch path into a TorchScript helper, and kept Quadrants as the fallback. I also simplified the tests and reran the focused and full terrain suites on CPU and Metal. I will follow the existing patterns more closely and keep the tests more focused in future PRs. Thanks for your patience.

duburcqa:
(1) Return NaN for out-of-range queries instead of raising, and remove every host read-back. `if has_invalid_query:` and `if invalid[0]:` convert a device tensor to a Python bool, which is a full GPU-to-host stall on a getter meant to run every step; your own numbers show it, since 558 us for 128 points versus 1151 us for 16384 points on Metal is flat in the workload. The contract is also wrong for the motivating use case: feet routinely leave the terrain patch in a locomotion task, and an exception makes the API unusable exactly there. Write NaN for out-of-bounds and non-finite points, fold the roll/pitch case into the same NaN output, drop both flags, keep the query fully asynchronous, and document the NaN semantics in the docstring.

(2) Drop the `gs.EPS` bounds tolerance. It is an absolute epsilon applied to a grid coordinate whose magnitude scales with the terrain, so it provides no margin: on a 145x145 terrain with `horizontal_scale=0.25` and a 37 deg yaw, 48% of exact-corner queries are rejected by fp32 rounding alone, measured over 2000 random yaws, while the same point 1 mm inside works. Clamp `row`/`col` to `[0, row_max]` and NaN only points outside by more than one cell. The current test uses a 3x4 grid where `row` and `col` never exceed 3, which is why it never sees this.

(3) Remove the `n_points == 1` clause from the output squeeze, and update the assertion that pins it. `n_points` comes from `positions.shape[-2]`, read off the input array, so `is_per_env and n_points == 1` makes the returned rank a function of the data: caller code written against two query points silently loses an axis the day it queries one. `is_single_position` can stay, since `positions.ndim` is part of the call form. Measured on the PR head with `n_envs=2`:

```
shape (1, 2)      -> (2, 1)     # one shared point
shape (2, 1, 2)   -> (2,)       # one point per env  <- point dim dropped
shape (2, 2, 2)   -> (2, 2)
```

(4) Fix the line wrapping, which is all-or-nothing per nesting level: when a call overflows, every argument moves together onto one hugged continuation line, and only if that also overflows does each argument get its own line. `def get_terrain_height(...)` hugs at 112 chars and the `_tc_get_terrain_height(...)` call at 118, so both are single continuation lines. In the test, `terrain.set_pos(...)` is 79 chars and `terrain.set_quat(...)` is 82 on one line; these are the two I flagged earlier with "1 line".

(5) Remove the two `.contiguous()` calls. `self.terrain_hf * self.terrain_scale[1]` and `self.terrain_scale[:1]` are already contiguous.

(6) Pass `horizontal_scale` as a native Python float. It is currently a one-element device tensor that both arms immediately index with `[0]`, so the divisor is fetched from device memory per element instead of folding.

(7) Fold away `_terrain_row_boundaries` / `_terrain_col_boundaries`, or add a one-line comment explaining them. They exist only so `torch.bucketize(..., right=True)` yields an int64 floor without a dtype cast. The trick is correct, including the empty-boundary case for a 2x2 height field and the `row == row_max` case, but it costs two persistent device tensors and two entity attributes to compute a floor, and nothing says why.

(8) Build the cached tensors in `_load_terrain` rather than lazily in the getter. The current guard checks two fields that are always set together and ignores the two boundary tensors it also initializes; moving it to the build-time path also removes four of the six new `__init__` declarations.

(9) Drop the bespoke `invalid` ndarray, which duplicates `array_class.ErrorCode` plus the `errno` tensor; item 1 removes it anyway. In the same kernel, `if not is_valid: ... if is_valid: ... else:` is three branches where if/else suffices, and the `heights[...] = 0.0` write on the invalid branch is a dead store.

(10) Ground the test in something other than the implementation. The expected values were derived from the same triangle model the code implements, so a flipped diagonal or a swapped axis in `terrain.py` would leave it green. The mesh raycast above is about ten lines, needs no extra scene build, and `trimesh` is already imported in the file; dropping a small box on a sloped cell and asserting its resting height against `get_height` works too, and substantiates the claim that the value matches the surface the simulator uses. Add NaN coverage for out-of-range points and for a rolled or pitched terrain once item 1 lands.

(11) Consider renaming to `get_terrain_height`, already the solver-side name. `get_height` on `KinematicEntity` reads as the height of the entity, and raises for every entity that is not a terrain.

jeetrex17:
Thanks, I addressed the full review in e23925f5.

The update returns NaN per query without host read-backs, applies the one-cell boundary margin, preserves explicit point dimensions, builds the terrain caches during loading, and renames the public method to get_terrain_height. I also simplified the Quadrants path and grounded the test expectations in raycasts against the generated terrain mesh.

The focused test passes on CPU and Metal with zero-copy enabled and disabled. I also updated the PR description and benchmarks.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -583,6 +586,17 @@ def _load_mesh(self, morph, surface, load_geom_only_for_heterogeneous=False):
     def _load_terrain(self, morph, surface):
         vmesh, mesh, self.terrain_hf = tu.parse_terrain(morph, surface)
         self.terrain_scale = np.array((morph.horizontal_scale, morph.vertical_scale), dtype=gs.np_float)
+        self._terrain_height_field = torch.as_tensor(
+            self.terrain_hf * self.terrain_scale[1], dtype=gs.tc_float, device=gs.device
+        )
+        self._terrain_horizontal_scale = float(morph.horizontal_scale)
```
duburcqa: It is already a float by design, so remove this useless cast.

Anyway, you should remove this proxy entirely and just access `self._morph.horizontal_scale` at runtime.

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -95,6 +96,66 @@ def _offset_world_shift(offset_pos, offset_quat, world_quat):
...
+    local_x = (
+        delta_x * (quat_x.square() + quat_w.square() - quat_y.square() - quat_z.square())
+        + delta_y * (2.0 * quat_x * quat_y + 2.0 * quat_w * quat_z)
+    ) / quat_norm
+    local_y = (
+        delta_x * (2.0 * quat_x * quat_y - 2.0 * quat_w * quat_z)
+        + delta_y * (quat_w.square() - quat_x.square() + quat_y.square() - quat_z.square())
+    ) / quat_norm
+    row = local_x / horizontal_scale
+    col = local_y / horizontal_scale
```
duburcqa: We don't have geom.py helpers that could be used instead?

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1951,6 +1965,50 @@ def get_vAABB(self, envs_idx=None):
...
+        if self.terrain_hf is None or self.terrain_scale is None:
+            gs.raise_exception("`get_terrain_height()` is only supported for terrain entities.")
+
+        return self._solver.get_terrain_height(
+            positions,
+            self.base_link_idx,
+            self._terrain_height_field,
+            self._terrain_horizontal_scale,
+            self._terrain_row_boundaries,
+            self._terrain_col_boundaries,
```
duburcqa: Why 'get_terrain_height' is accepting `row_boundaries, col_boundaries` as input argument where the same value is passed systematically at every places it is called?

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -1057,6 +1118,71 @@ def set_dofs_position(self, position, dofs_idx=None, envs_idx=None):
...
+        elif positions.ndim == 2:
+            positions = positions[None]
+        elif positions.ndim == 3:
+            if positions.shape[0] not in (1, n_envs):
+                gs.raise_exception(
+                    f"The leading dimension of `positions` must be 1 or {n_envs}, got {positions.shape[0]}."
+                )
+            is_per_env = positions.shape[0] == n_envs
+        else:
+            gs.raise_exception("`positions` must have shape (2,), (n_points, 2), or (n_envs, n_points, 2).")
```
duburcqa: I'm surprised that we do not have a helper for this, like '_sanitize_io_variables'

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -1057,6 +1118,71 @@ def set_dofs_position(self, position, dofs_idx=None, envs_idx=None):
...
+                positions,
+                height_field,
+                horizontal_scale,
+                row_boundaries,
+                col_boundaries,
+                link_pos,
+                link_quat,
+                gs.EPS,
+            )
+
```
duburcqa: 1 line.

genesis/engine/entities/rigid_entity/rigid_entity.py
jeetrex17: Good point. I removed the repeated terrain data arguments. The solver now resolves the terrain entity from link_idx, reads the cached height field and boundaries from it, and accesses horizontal_scale from the morph at runtime.

genesis/engine/solvers/kinematic_solver.py
jeetrex17: Good point. I replaced the manual shape handling with broadcast_tensor and kept _sanitize_envs_idx for environment selection. I did not use _sanitize_io_variables directly because it also creates point-index tensors and would materialize shared query positions across environments. The new path keeps shared positions compact while validating the supported layouts.

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -1057,6 +1105,58 @@ def set_dofs_position(self, position, dofs_idx=None, envs_idx=None):
...
+        is_per_env = positions.ndim == 3 and positions.shape[0] != 1
+        n_position_envs = n_envs if is_per_env else 1
+        positions = broadcast_tensor(
+            positions, gs.tc_float, (n_position_envs, n_points, 2), ("envs_idx", "positions", "")
+        ).contiguous()
+
+        if gs.use_zerocopy:
+            link_pos = qd_to_torch(self.dyn_state.links.pos, transpose=True)[envs_idx, link_idx : link_idx + 1]
+            link_quat = qd_to_torch(self.dyn_state.links.quat, transpose=True)[envs_idx, link_idx : link_idx + 1]
```
duburcqa: Wrong API usage.

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -1057,6 +1105,58 @@ def set_dofs_position(self, position, dofs_idx=None, envs_idx=None):
...
+        is_single_position = positions.ndim == 1
+        n_points = 1 if is_single_position else positions.shape[-2]
+        if n_points == 0:
+            gs.raise_exception("`positions` must contain at least one position.")
+
+        is_per_env = positions.ndim == 3 and positions.shape[0] != 1
+        n_position_envs = n_envs if is_per_env else 1
+        positions = broadcast_tensor(
+            positions, gs.tc_float, (n_position_envs, n_points, 2), ("envs_idx", "positions", "")
+        ).contiguous()
```
duburcqa: No reason to force '.contiguous()' for all codepath. Only do it where it is passed to the quadrants kernel.

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -1057,6 +1105,58 @@ def set_dofs_position(self, position, dofs_idx=None, envs_idx=None):
         self._is_forward_pos_updated = True
         self._is_forward_vel_updated = True
 
+    def get_terrain_height(self, positions, link_idx, envs_idx=None):
+        terrain = self._links[link_idx].entity
+        height_field = terrain._terrain_height_field
+        horizontal_scale = terrain._morph.horizontal_scale
+        row_boundaries = terrain._terrain_row_boundaries
+        col_boundaries = terrain._terrain_col_boundaries
```
duburcqa: Remove these proxies if only used once or twice.

genesis/utils/geom.py
```diff
@@ -1360,6 +1367,8 @@ def inv_transform_by_quat(pos, quat):
 
 
 def inv_transform_by_trans_quat(pos, trans, quat):
+    if all(isinstance(e, torch.Tensor) for e in (pos, trans, quat)):
+        return _tc_inv_transform_by_trans_quat(pos, trans, quat)
     return inv_transform_by_quat(pos - trans, quat)
```
duburcqa, line 1380: I think we should find a way to properly compile a chain of operations... This approach is very inelegant. Add a '# FIXME: <...>'.

genesis/engine/solvers/kinematic_solver.py
jeetrex17: Fixed. I now pass envs_idx and link_idx through the row and column mask arguments of qd_to_torch, which also preserves the singleton link dimension.

genesis/engine/solvers/kinematic_solver.py
jeetrex17: Fixed. The zero-copy path now keeps the broadcast view, and .contiguous() is applied only when passing positions to the Quadrants kernel.

genesis/utils/geom.py
jeetrex17: Added a FIXME to track compiling chains of public geometry helpers without dedicated backend-specific wrappers.

genesis/engine/solvers/kinematic_solver.py
duburcqa: Collapsing, yes. Note that inlining the terrain attributes pushed the hugged form to 184 chars, so it only fits on one line if the argument list shrinks - dropping `row_boundaries` / `col_boundaries` brings it to 117.

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -1057,6 +1105,61 @@ def set_dofs_position(self, position, dofs_idx=None, envs_idx=None):
...
+
+        envs_idx = self._scene._sanitize_envs_idx(envs_idx)
+        n_envs = len(envs_idx)
+        is_single_position = positions.ndim == 1
+        n_points = 1 if is_single_position else positions.shape[-2]
+        if n_points == 0:
+            gs.raise_exception("`positions` must contain at least one position.")
+
+        is_per_env = positions.ndim == 3 and positions.shape[0] != 1
+        n_position_envs = n_envs if is_per_env else 1
```
duburcqa: Used once, on the next line. Inline it.

tests/rigid/test_terrain.py
```diff
@@ -134,6 +132,114 @@ def test_discrete_obstacles():
...
+    initial_height = terrain.get_terrain_height((2.0, -0.5))
+    assert_allclose(initial_height, 1.3, tol=tol)
+
+    visual_pose = gu.trans_quat_to_T(np.array(visual_terrain.morph.pos), np.array(visual_terrain.morph.quat))
+    visual_position = visual_pose @ np.array((0.5, 0.25, 0.0, 1.0))
+    visual_height = visual_terrain.get_terrain_height(visual_position[:2])
+    assert_allclose(visual_height, 3.04, tol=tol)
+
+    if n_envs:
+        shared_positions = terrain.get_terrain_height(((2.0, -1.0), (2.0, -0.5)))
```
duburcqa: The three-dimensional shared form `(((2.0, -1.0), (2.0, -0.5)),)` was dropped in this revision, so the `positions.shape[0] != 1` branch that tells shared from per-environment layouts is no longer exercised. Add it back if it makes the test stronger.

tests/rigid/test_terrain.py
```diff
@@ -134,6 +132,114 @@ def test_discrete_obstacles():
...
+        one_position_per_env = terrain.get_terrain_height(((positions_per_env[0][0],), (positions_per_env[1][0],)))
+        assert one_position_per_env.shape == (2, 1)
+        assert_allclose(one_position_per_env, ((1.45,), (-1.55,)), tol=tol)
+
+        heights_env_1 = terrain.get_terrain_height(positions_per_env[1], envs_idx=1)
+        assert_allclose(heights_env_1, (-1.55, -0.9), tol=tol)
+    else:
+        heights = terrain.get_terrain_height(positions)
+
+    assert_allclose(heights, expected, atol=1e-6, rtol=1e-6)
```
duburcqa: `tol=1e-6` sets both.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -583,6 +589,16 @@ def _load_mesh(self, morph, surface, load_geom_only_for_heterogeneous=False):
     def _load_terrain(self, morph, surface):
         vmesh, mesh, self.terrain_hf = tu.parse_terrain(morph, surface)
         self.terrain_scale = np.array((morph.horizontal_scale, morph.vertical_scale), dtype=gs.np_float)
+        self._terrain_height_field = torch.as_tensor(
+            self.terrain_hf * self.terrain_scale[1], dtype=gs.tc_float, device=gs.device
```
duburcqa: `parse_terrain` hands back whatever dtype the user passed - float64 for a plain `np.array(...)` height field, int64 for an integer one, float32 only when the terrain is generated - so this always copies. Please normalize `terrain_hf` to `gs.np_float` in `parse_terrain` instead. That also lets `collider.py:728` and `rigid_solver.py:1227` drop their `astype`.

tests/rigid/test_terrain.py
```diff
@@ -134,6 +132,114 @@ def test_discrete_obstacles():
...
+    expected = locations[np.argsort(ray_ids), 2]
+    if n_envs:
+        heights = terrain.get_terrain_height(positions, envs_idx=[0])
+
+        positions_per_env = ((positions[3], positions[4]), ((-3.25, 4.125), (-3.25, 4.375)))
+        heights_per_env = terrain.get_terrain_height(positions_per_env)
+        assert_allclose(heights_per_env, ((1.45, 2.1), (-1.55, -0.9)), tol=tol)
+
+        one_position_per_env = terrain.get_terrain_height(((positions_per_env[0][0],), (positions_per_env[1][0],)))
+        assert one_position_per_env.shape == (2, 1)
```
duburcqa: Redundant with the next line: `assert_allclose` against a `(2, 1)` expected value already fails on a `(2,)` result. Remove.

jeetrex17:
Thanks for the review. I addressed the remaining feedback in d64456b6.

The helper now builds the bucketization boundaries internally, the cached boundary tensors and repeated arguments are removed, and the call site fits on one line. The focused test passes on CPU and Metal, with zero-copy enabled and disabled, including the 2x2 boundary case.

duburcqa (dismissed):

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -1057,6 +1103,53 @@ def set_dofs_position(self, position, dofs_idx=None, envs_idx=None):
...
+        n_points = 1 if is_single_position else positions.shape[-2]
+        if n_points == 0:
+            gs.raise_exception("`positions` must contain at least one position.")
+
+        is_per_env = positions.ndim == 3 and positions.shape[0] != 1
+        positions = broadcast_tensor(
+            positions, gs.tc_float, (n_envs if is_per_env else 1, n_points, 2), ("envs_idx", "positions", "")
+        )
+
+        if gs.use_zerocopy:
```
duburcqa: Let's profile CPU and GPU (Apple Metal and CUDA) to determine whether it should only be preferred in CPU or systematically.

I'm on it.

duburcqa:
(1) Item (7) from the previous batch is half-done. The boundary tensors were folded into `_tc_get_terrain_height`, but nothing says why `torch.bucketize(..., right=True)` is there rather than a floor, which was the other half of the ask, and the `torch.arange` pair is now rebuilt on every call instead of once: two device allocations per query, on a getter meant to run every step.

(2) Three unnecessary `float()` casts, at `kinematic_solver.py:122`, `124` and `127-128`. TorchScript compiles both `row <= row_max + 1` and `row.clamp(0.0, row_max)` with a plain `int`, so all three can go.

(3) `horizontal_scale: float` at `accessor.py:1047` is the only scalar kernel annotation in that file that is not a `qd` type; the three siblings are `qd.f32`. `float` follows the configured default fp and is the better of the two, so either it or the siblings should move.

(4) Blank line before `else:` at `kinematic_solver.py:1131`. ruff-format leaves it.

(5) The roll and pitch gate is machine epsilon rather than a tilt tolerance. `abs(q_x) <= gs.EPS` is `1.19e-07` in fp32, which is `1.4e-05` degrees, and `1e-15` in fp64. A terrain quaternion carrying a hair of numerical roll - one that came from a matrix-to-quaternion round trip rather than `xyz_to_quat` on a pure-yaw Euler - returns NaN for every point with no diagnostic. Pick a tolerance the user can reason about, in radians on the tilt, and state it in the docstring next to the NaN semantics.

(6) `_terrain_height_field` at `rigid_entity.py:590` is a third device copy of `terrain_hf * vertical_scale`. Two concrete follow-ups and one thing to write down:

- `RigidSolver._init_terrain_state` at `rigid_solver.py:1214-1240` is dead: it allocates and fills `terrain_hf`, `terrain_rc`, `terrain_scale` and `terrain_xyz_maxmin` as solver-level fields and none of the four is read anywhere, since the only readers are the `collider_info.terrain_*` counterparts in `narrowphase.py:1351-1402`. It is a line-for-line duplicate of `Collider._init_terrain_state` at `collider.py:715-736`. Deleting it takes the height field from three device copies to two. Pre-existing, so a separate `[MISC]` PR is fine and probably cleaner.
- `entity.terrain_hf * terrain_scale[1]` is now written at three sites: `rigid_entity.py:590`, `collider.py:728` and `rigid_solver.py:1227`. Since `parse_terrain` normalizes the dtype, `_load_terrain` can compute the scaled field once and have the collider upload that array rather than recomputing the product.
- The remaining two copies cannot be merged, and that deserves a one-line comment at `rigid_entity.py:590`: `collider_info.terrain_hf` is sized from the first terrain geom only (`array_class.py:1583-1588`) and the collider exists only on `RigidSolver`, whereas the getter serves N terrains including visual-only `Kinematic` ones with no collider. Without the note the next reader will try to replace the tensor with `qd_to_torch(collider_info.terrain_hf)`.

(7) Two dead casts survive right next to the ones you removed: `collider.py:726` `entity.terrain_scale.astype(gs.np_float)` and `rigid_solver.py:1225` `np.asarray(entity.terrain_scale, dtype=gs.np_float)`. `terrain_scale` is already built with `dtype=gs.np_float` at `rigid_entity.py:585`.

genesis/engine/solvers/rigid/abd/accessor.py
```diff
@@ -1040,6 +1040,61 @@ def kernel_get_links_acc(
...
+    positions: qd.types.ndarray(),
+    height_field: qd.types.ndarray(),
+    heights: qd.types.ndarray(),
+    dyn_state: array_class.DynState,
+    rigid_info: array_class.RigidInfo,
+    rigid_config: qd.template(),
+    is_per_env: qd.template(),
+):
+    EPS = rigid_info.EPS[None]
+    qd.loop_config(serialize=qd.static(rigid_config.para_level < gs.PARA_LEVEL.ALL))
```
duburcqa: It should be `gs.PARA_LEVEL.PARTIAL`.

genesis/engine/solvers/kinematic_solver.py
duburcqa: Profiled both arms on CPU, CUDA and Apple Metal, on a quiet machine. Conclusion: drop the TorchScript arm and always call the kernel. One defect has to be fixed first, and it was confounding the measurement.

`kernel_get_terrain_height` gates its whole fused `ndrange(n_envs, n_points)` on `rigid_config.para_level < PARA_LEVEL.ALL`, and `scene.py:975` sets `PARTIAL` for any GPU scene with `n_envs <= 1`. So a single-environment GPU query serializes the point loop, which has nothing to do with batching. `scene.py:964` states the intent: at `PARTIAL` "we only parallelize certain loops that have big loop size", and `n_points` is user-supplied and unbounded, so it is exactly that loop. The gate should be `serialize=qd.static(rigid_config.para_level == gs.PARA_LEVEL.NEVER)`. With that one-line patch, at 1 env and 16384 points the kernel goes from 4641 to 198 us on CUDA and from 6583 to 1221 us on Metal. This is worth fixing whatever you decide about the dispatch.

**How to read the tables.** Every cell is `TorchScript arm / Quadrants kernel arm`, in **microseconds per call, lower is better**. So `197/97` means the TorchScript arm took 197 us and the kernel arm 97 us for the same query, and the kernel is the better of the two. Rows are the number of environments in the scene, columns are the number of query points per environment. All numbers are with the `para_level` patch above, on a 65x65 terrain, amortized over 50 back-to-back calls under a single device sync. Median over minimum was 1.00 to 1.08 on every row, so the comparisons are readable rather than noise.

CPU, us/call, torch / kernel:

| envs \ points per env | 1 | 128 | 1024 | 16384 |
| --- | --- | --- | --- | --- |
| 1 | 89 / 93 | 94 / 91 | 197 / 97 | 678 / 182 |
| 8 | 94 / 94 | 201 / 97 | 457 / 137 | 3518 / 788 |
| 64 | 96 / 93 | 454 / 138 | 3198 / 445 | 7000 / 5661 |

CUDA, us/call, torch / kernel:

| envs \ points per env | 1 | 128 | 1024 | 16384 |
| --- | --- | --- | --- | --- |
| 1 | 384 / 198 | 403 / 198 | 468 / 208 | 444 / 198 |
| 8 | 428 / 198 | 444 / 200 | 441 / 199 | 445 / 198 |
| 64 | 460 / 207 | 454 / 200 | 462 / 201 | 449 / 199 |

Metal, us/call, torch / kernel:

| envs \ points per env | 1 | 128 | 1024 | 16384 |
| --- | --- | --- | --- | --- |
| 1 | 1141 / 1090 | 1387 / 1144 | 1515 / 1152 | 1800 / 1221 |
| 8 | 583 / 1103 | 594 / 1096 | 589 / 1042 | 1466 / 1159 |
| 64 | 628 / 1130 | 586 / 1068 | 738 / 1116 | 66368 / 1844 |

On CPU the kernel wins at every size above a single point, by up to 7.2x. On CUDA it wins uniformly by about 2.2x, and the flat 198 to 208 us column says the query itself is free there and the cost is dispatch. Below 128 points both arms sit on a shared floor, roughly 90 us on CPU, which is the Python preamble rather than either implementation.

Metal is the only backend where the TorchScript arm wins anything: at 8 or more environments and up to 1024 points it leads by 1.5 to 1.9x. Against that, at 64 environments by 16384 points it is 36x slower, 66.4 ms against 1.8 ms. It also scales badly well before that point: going from 131k total points to 1.05M, an 8x increase in work, costs it 45x more time, 1.5 ms to 66.4 ms, while the kernel goes 1.2 ms to 1.8 ms. The mechanism is that `_tc_get_terrain_height` issues thirty-odd separate launches and materializes a dozen full-size temporaries where the kernel issues one launch and none.

So the second implementation earns its keep on exactly one backend, in one part of the input range, and loses badly elsewhere on that same backend. Removing it deletes `_tc_get_terrain_height`, the `gs.use_zerocopy` branch and the duplicated triangle interpolation, and it also lets `_tc_inv_transform_by_trans_quat` and the type-sniffing dispatch in `geom.py` go, along with the FIXME on it. Items (1) and (2) of my previous review become moot at the same time.

duburcqa: Tracing the one-environment column above: the penalty is not in the query, it is a host readback in `indices_to_mask`. A single-element index is collapsed to a slice via `.item()` so the extract stays a view, and when the index lives on the device that read stalls on the device queue. It costs 101 us on Metal and 9.6 us on CUDA per masked getter call, and it fires for `envs_idx=<int>` on any scene size, not only `n_envs == 1`.

Fixed separately in #3157, where extracting a link field drops from 126 us to 40 us on Metal. The zero-copy arm here inherits the saving through its two `qd_to_torch` calls, so its numbers above are pessimistic by roughly 170 us at one environment. That does not change the ranking: the kernel arm still wins on CPU and CUDA at every size measured.

jeetrex17:
Thanks for profiling this across all three backends, and for #3157.

I addressed the remaining review points in the last two commits:

- Removed the TorchScript path. The getter now always uses the Quadrants kernel, which also removes the geom.py helper, the zero-copy dispatch, and the related FIXME.
- Corrected the serialization gate so the point loop parallelizes unless the parallelization level is NEVER.
- Replaced the machine-epsilon check with a documented 0.001 radian tilt tolerance. The quaternion check is independent of yaw, and the test covers both sides of the tolerance.
- Computed the scaled height field once in _load_terrain() and reused it for the collider upload, with a comment explaining why the per-entity query tensor remains separate from collider storage.
- Removed the remaining terrain_scale casts and changed the friction kernel scalar annotations to float so they follow the configured precision.

The terrain suite passes on CPU, and the focused query tests pass on Metal. I will open a separate [MISC] PR for the unused RigidSolver._init_terrain_state() cleanup once this PR is merged.

duburcqa:
> The terrain suite passes on CPU, and the focused query tests pass on Metal. I will open a separate [MISC] PR for the unused RigidSolver._init_terrain_state() cleanup once this PR is merged.

Let's do the cleanup in the PR then.

jeetrex17:
On it.

jeetrex17:
Done in fec357f. The full terrain suite passes on CPU (11 passed), and the focused query test passes on Metal (2 passed).

duburcqa (approved):

duburcqa:
All good! Ready to merge

PR #3132, opened 2026-07-30
[FEATURE] Support raycasting against visual meshes in viewer plugins.

Milotrince:
## Description

`RaycasterViewerPlugin` takes a new `use_visual_geom` option, inherited by `MouseInteractionPlugin` and by any downstream plugin (e.g. the `examples/viewer_plugin/mesh_point_selector.py` selector). When set, the viewer raycaster casts against visual meshes and `RayHit.geom` is the hit `RigidVisGeom`, whose `.link` / `.entity` keep existing plugin code working unchanged.

- Support both Rigid and Kinematic solver, so visual-only entities are pickable. `Raycaster` holds one BVH per solver and keeps the closest hit across them, capping each pass at the best distance so far.
- Restricted to entities that set `material.use_visual_raycasting=True` - the same opt-in the raycaster sensors use, so both describe the same pickable geometry. An entity without the flag is not pickable, and casting with nothing opted in logs a warning naming the option.
- That opt-in mask now has a single definition, the `KinematicSolver.vfaces_raycast_mask` property, fitted once at build time and read by both the sensor context and the viewer.
- The single-ray kernel selects the mesh through a compile-time `is_visual` flag, reusing the visual traversal that already backs the sensors rather than adding a second copy.
- Drive-by fix: casting in a scene with no geometry to cast against now returns no hit instead of raising (`cast()` dereferenced a `None` BVH left behind by `__init__`).
- A heterogeneous entity now casts against its own variant alone: a visual face is skipped when its vgeom falls outside the environment's active `links.vgeom_start` / `vgeom_end` range, and the static-BVH environment grouping keys on that range so environments holding different variants no longer share a tree. This corrects the raycaster sensors too, which previously saw the union of every variant.
- Bug fix, pre-existing on `main` and reachable from the raycaster sensors alone: a raycast consumer calls `update_forward_pos`, and the flag it sets also authorizes the next step to skip its Cartesian-space update, which covers geoms as well as links. `RigidSolver` now refreshes geoms alongside links, so a raycast consumer no longer leaves collision reading a one-step-stale geom pose. With `enable_mujoco_compatibility=True` this silently dropped contacts - a box resting on the ground sank through it - and `tests/sensors/test_raycaster.py::test_hits` now pins it.

## Motivation and Context

Better user experience with viewer

Unit tests:

```bash
pytest tests/sensors/test_raycaster.py tests/rendering/test_interactive_viewer.py
```

Interactively, running each example with and without the new option:

```bash
python examples/viewer_plugin/mouse_interaction.py
python examples/viewer_plugin/mouse_interaction.py --use_visual_geom
python examples/viewer_plugin/mesh_point_selector.py
python examples/viewer_plugin/mesh_point_selector.py --use_visual_geom
```

The `mouse_interaction.py` example, picking the same scene both ways:

<table>
<tr>
<th width="50%">Collision meshes (<code>use_visual_geom=False</code>)</th>
<th width="50%">Visual meshes (<code>use_visual_geom=True</code>)</th>
</tr>
<tr>
<td><video src="https://github.com/user-attachments/assets/98846629-6424-4750-90a9-ee40a8a66e3d"></video></td>
<td><video src="https://github.com/user-attachments/assets/34d6f0a3-e1d6-4668-b5bb-bb3cd5d275f0"></video></td>
</tr>
</table>

## Checklist:

Resolves SIM-293

### DIFF
--- a/genesis/engine/solvers/kinematic_solver.py
+++ b/genesis/engine/solvers/kinematic_solver.py
@@ -8,12 +8,11 @@ import genesis.utils.array_class as array_class
 import genesis.utils.geom as gu
 from genesis.engine.entities.rigid_entity import KinematicEntity
 from genesis.engine.states.solvers import KinematicSolverState
-from genesis.options.solvers import RigidOptions, KinematicOptions
+from genesis.options.solvers import KinematicOptions, RigidOptions
 from genesis.utils.misc import (
     assign_indexed_tensor,
     broadcast_tensor,
     indices_to_mask,
-    qd_to_numpy,
     qd_to_torch,
     qd_zero_grad,
     sanitize_indexed_tensor,
@@ -1076,21 +1075,22 @@ class KinematicSolver(Solver):
             tensor = gu.transform_quat_by_quat(gu.inv_quat(offset_quat), tensor)
         return tensor[0] if self.n_envs == 0 else tensor
 
-    def get_vfaces_raycast_mask(self) -> np.ndarray:
+    def get_vfaces_raycast_mask(self) -> torch.Tensor:
         """Per-vface mask (int8, shape (n_vfaces,)) selecting the vfaces opted into visual raycasting.
 
         A vface is opted in iff its owning vgeom belongs to an entity whose material has use_visual_raycasting=True.
         Consumed by every visual raycast BVH (raycaster sensors, viewer plugins) to gate which vfaces contribute.
+        The int8 dtype is what the AABB-update kernels take for the mask.
         """
         if self.dyn_info.vfaces.vgeom_idx.shape[0] == 0:
-            return np.zeros(0, dtype=np.int8)
-        vgeom_enabled = np.zeros(self.n_vgeoms, dtype=np.bool_)
+            return torch.zeros(0, dtype=torch.int8, device=gs.device)
+        vgeom_enabled = torch.zeros(self.n_vgeoms, dtype=torch.int8, device=gs.device)
         for entity in self.entities:
             if not entity.material.use_visual_raycasting:
                 continue
             for vgeom in entity.vgeoms:
-                vgeom_enabled[vgeom.idx] = True
-        return vgeom_enabled[qd_to_numpy(self.dyn_info.vfaces.vgeom_idx)].astype(np.int8)
+                vgeom_enabled[vgeom.idx] = 1
+        return vgeom_enabled[qd_to_torch(self.dyn_info.vfaces.vgeom_idx)]
 
     def get_links_vel(self, links_idx=None, envs_idx=None):
         if gs.use_zerocopy:
--- a/genesis/engine/solvers/kinematic_solver.py
+++ b/genesis/engine/solvers/kinematic_solver.py
@@ -10,24 +10,32 @@ from genesis.engine.entities.rigid_entity import KinematicEntity
 from genesis.engine.states.solvers import KinematicSolverState
 from genesis.options.solvers import RigidOptions, KinematicOptions
 from genesis.utils.misc import (
+    assign_indexed_tensor,
+    broadcast_tensor,
+    indices_to_mask,
+    qd_to_numpy,
     qd_to_torch,
     qd_zero_grad,
     sanitize_indexed_tensor,
-    indices_to_mask,
-    broadcast_tensor,
-    assign_indexed_tensor,
 )
 
 from .base_solver import MutatedLinks, Solver, StateChange, mutates
-from .rigid.abd.misc import (
-    kernel_init_dof_fields,
-    kernel_init_link_fields,
-    kernel_init_joint_fields,
-    kernel_init_vvert_fields,
-    kernel_init_vgeom_fields,
-    kernel_init_entity_fields,
-    kernel_update_heterogeneous_links_vgeom,
-    kernel_update_vgeoms_render_T,
+from .rigid.abd.accessor import (
+    kernel_get_kinematic_state,
+    kernel_get_links_vel,
+    kernel_get_state_grad,
+    kernel_set_dofs_force_grad,
+    kernel_set_dofs_position,
+    kernel_set_dofs_velocity,
+    kernel_set_dofs_velocity_grad,
+    kernel_set_dofs_zero_velocity,
+    kernel_set_kinematic_state,
+    kernel_set_links_pos,
+    kernel_set_links_pos_grad,
+    kernel_set_links_quat,
+    kernel_set_links_quat_grad,
+    kernel_set_qpos,
+    kernel_set_vverts,
 )
 from .rigid.abd.forward_kinematics import (
     kernel_forward_kinematics,
@@ -37,22 +45,15 @@ from .rigid.abd.forward_kinematics import (
     kernel_update_vgeoms,
     kernel_update_vverts_for_vgeoms,
 )
-from .rigid.abd.accessor import (
-    kernel_get_kinematic_state,
-    kernel_get_state_grad,
-    kernel_set_kinematic_state,
-    kernel_set_links_pos_grad,
-    kernel_set_links_quat_grad,
-    kernel_set_dofs_position,
-    kernel_set_dofs_velocity,
-    kernel_set_dofs_velocity_grad,
-    kernel_set_dofs_force_grad,
-    kernel_set_dofs_zero_velocity,
-    kernel_set_links_pos,
-    kernel_set_links_quat,
-    kernel_set_qpos,
-    kernel_set_vverts,
-    kernel_get_links_vel,
+from .rigid.abd.misc import (
+    kernel_init_dof_fields,
+    kernel_init_entity_fields,
+    kernel_init_joint_fields,
+    kernel_init_link_fields,
+    kernel_init_vgeom_fields,
+    kernel_init_vvert_fields,
+    kernel_update_heterogeneous_links_vgeom,
+    kernel_update_vgeoms_render_T,
 )
 
 if TYPE_CHECKING:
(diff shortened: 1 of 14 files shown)


Milotrince:
@codex review

Milotrince:
@codex review

Milotrince:
@codex review

Milotrince (commented):
self-reviewed!

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -1075,6 +1076,22 @@ def get_vgeoms_quat(self, vgeoms_idx=None, envs_idx=None, *, relative=False):
...
+        """
+        if self.dyn_info.vfaces.vgeom_idx.shape[0] == 0:
+            return np.zeros(0, dtype=np.int8)
+        vgeom_enabled = np.zeros(self.n_vgeoms, dtype=np.bool_)
+        for entity in self.entities:
+            if not entity.material.use_visual_raycasting:
+                continue
+            for vgeom in entity.vgeoms:
+                vgeom_enabled[vgeom.idx] = True
+        return vgeom_enabled[qd_to_numpy(self.dyn_info.vfaces.vgeom_idx)].astype(np.int8)
```
duburcqa: Data accessors are torch-only.

tests/rendering/test_interactive_viewer.py
```diff
@@ -461,6 +471,114 @@ def check_event_count():
...
+    wait_for_viewer_events(pyrender_viewer, check_event_count())
+    scene.step()
+
+    assert pyrender_viewer.is_active
+    assert visual_plugin._held_link is None
+
+
+@pytest.mark.required
+@pytest.mark.parametrize("n_envs", [0, 2])
+def test_raycast_visual_geom(n_envs, show_viewer):
```
duburcqa: There is no way to parameterize one of the existing unit test instead?

Are you ensuring the visual does not match the collision (eg shifted visual vs collision boxes, to make them completely disjoint?)

tests/sensors/test_raycaster.py
```diff
@@ -615,6 +615,21 @@ def test_heterogeneous_object(show_viewer, tol):
             gs.morphs.Sphere(radius=0.2, pos=(1.0, 0.0, 0.5), fixed=True),
             gs.morphs.Box(size=(0.6, 0.6, 0.6), pos=(1.0, 0.0, 0.5), fixed=True),
         ),
+        material=gs.materials.Rigid(
+            use_visual_raycasting=True,
+        ),
+    )
+    # Every link of the kinematic solver is fixed, so its visual BVH is static and groups envs by geometry. The
```
duburcqa: > Variants share one link, hence identical vgeom poses

I don't understand the relation. Not sure if same pos but different size is strong enough.

tests/sensors/test_raycaster.py
```diff
@@ -615,6 +615,21 @@ def test_heterogeneous_object(show_viewer, tol):
...
+    # variants share one link, hence identical vgeom poses, and are told apart only by their active vgeom range.
+    het_visual = scene.add_entity(
+        morph=(
+            gs.morphs.Box(size=(0.2, 0.2, 0.2), pos=(4.0, 0.0, 0.5), fixed=True),
+            gs.morphs.Box(size=(0.4, 0.4, 0.4), pos=(4.0, 0.0, 0.5), fixed=True),
+            gs.morphs.Box(size=(0.6, 0.6, 0.6), pos=(4.0, 0.0, 0.5), fixed=True),
+        ),
+        material=gs.materials.Kinematic(
+            use_visual_raycasting=True,
```
duburcqa, line 642: We don't already have a similar option already? I must be missing something...

duburcqa: LoL ok it is precisely the existing option... perfect!

duburcqa:
**1. `get_vfaces_raycast_mask` must return torch** - `genesis/engine/solvers/kinematic_solver.py:1079`

Every sibling getter in that block returns torch (`get_vgeoms_pos`, `get_vgeoms_quat`, `get_links_vel`, `get_qpos`, ...). To size it: `mask.any()` / `mask.shape[0]` work either way, `face_mask: qd.types.ndarray()` accepts a torch tensor directly, and only the numpy fancy-index at `genesis/engine/sensors/raycaster.py:205` needs a `tensor_to_array`.

**2. The kinematic-link guard is vacuous in the `n_envs=2` arm** - `tests/rendering/test_interactive_viewer.py:483`

```python
kinematic.set_pos((target_offset[0], target_offset[1], target_offset[2] + BOX_LENGTH))
```

`set_pos` takes **env-local** coordinates, but `target_offset` is env 1's *offset*, so the box lands at env-local `(0.25, 0, 0.2)` and the viewport-center ray hits it in no env. With the `isinstance(..., RigidLink)` guard removed:

```
n_envs=0: hover use_visual=True hit=RayHit(distance=1.0496, position=[0., 0.06, 0.3], ...)  -> FAILED
n_envs=2: hover use_visual=True hit=None                                                    -> PASSED
```

Fix: `kinematic.set_pos((0.0, 0.0, BOX_LENGTH))` - both arms then hit and both fail without the guard.

**3. `test_raycast_visual_geom` should be a `use_visual_geom` axis on `test_mouse_interaction_plugin`** - `tests/rendering/test_interactive_viewer.py:497`

It instantiates `genesis.vis.viewer_plugins.raycast.Raycaster` directly (line 13) - not exported from `viewer_plugins/__init__.py` and used by no other test. The user-facing surface is `RaycasterViewerPlugin(use_visual_geom=...)`, and `test_mouse_interaction_plugin` already drives it and is already parametrized, so an axis there covers the feature and saves a scene build.

Also still open: no case where **one entity's** visual and collision geometry disagree. Two separate entities (`visualization=False` / `collision=False`) do not reproduce the motivating hull-coarser-than-mesh scenario; add a shifted-visual entity so the visual hit lands where the collision cast provably cannot reach.

**4. `tests/sensors/test_raycaster.py:622-623` states a relation it does not establish**

> `The variants share one link, hence identical vgeom poses, and are told apart only by their active vgeom range.`

Sharing a link does not imply identical vgeom poses. What does: a box primitive's vgeom sits at the link origin, so the variants differ only in rest-pose vverts (non-batched) and in the active vgeom range - hence the range being the only discriminator available to the grouping signature.

**5. The option doc names the wrong cost** - `genesis/vis/viewer_plugins/raycast.py:186`

> `It costs casting against the finer visual triangles`

The cast is flat; the per-step BVH rebuild is what costs. Four ducks, cpu, `n_envs=1` (2012 collision faces vs 14138 vfaces):

| mode | `update()` per sim step | `cast()` |
|---|---|---|
| collision | 0.98 ms | 0.08 ms |
| visual | **3.32 ms** | 0.06 ms |

`update_on_sim_step` rebuilds the whole visual BVH unconditionally, scaling with visual-triangle count. The sensor path gates this with `maybe_static` + `StateChange` subscriptions; the viewer path has no equivalent.

**6. `use_visual_raycasting`'s own doc was not extended** - `genesis/engine/materials/kinematic.py:9`, `genesis/engine/materials/rigid.py`

It is now a prerequisite for viewer picking and the new `RaycasterViewerPlugin` docstring sends users to it, but `Kinematic` still mentions raycaster sensors only, and `Rigid` never lists the field in its `Parameters` block at all.

**7. Empty testing section with the box ticked**

"How Has This Been / Can This Be Tested?" is blank (as are "Related Issue" and effectively "Motivation and Context") while the corresponding checklist item is checked. `PULL_REQUESTS.md` requires example commands; the new `--use_visual_geom` example flag is the natural content.

**8. Redundant assertion** - `tests/sensors/test_raycaster.py:689`

`assert visual_bvh.aabb.n_batches == 3` adds nothing: the `(3.9, ...)` distance assertion two lines below catches the same defect alone (verified by removing it).

**9. Import block left unsorted** - `genesis/engine/solvers/kinematic_solver.py:11`

Three import blocks were re-sorted here, but `from genesis.options.solvers import RigidOptions, KinematicOptions` was left as is; should be `KinematicOptions, RigidOptions`.

**10. Redundant state** - `genesis/vis/viewer_plugins/raycast.py:29`

`RaycastTarget.vfaces_mask` is `None` exactly iff `not self.use_visual_geom`, yet every consumer (lines 104, 155, 169) branches on `self.use_visual_geom`. Either drop the `None` convention or branch on it.

**11. Warning names the wrong cause** - `genesis/vis/viewer_plugins/raycast.py:87-88`

`if not self.targets:` also fires when neither solver is active, but the message blames a missing `use_visual_raycasting=True`.

**12. Two doc styles for one parameter** - `genesis/vis/viewer_plugins/raycast.py:40` vs `:183`

`Raycaster` documents `use_visual_geom` in prose, `RaycasterViewerPlugin` in a numpydoc `Parameters` section.

**13. Negations in prose** - `genesis/vis/viewer_plugins/raycast.py:184-185`, `genesis/utils/raycast_qd.py:539`

"which the collision meshes cannot do when they are coarser, larger or entirely absent" and "instead of the union of every variant" should state the positive fact.

**14. Backtick style** - `genesis/utils/raycast_qd.py:607`

` ``is_visual`` ` where `write_ray_hit` just below uses single backticks for `is_merge` / `is_last`.

**15. Wrong annotation** - `genesis/vis/viewer_plugins/plugins/mouse_interaction.py:255`

`closest_hit: RayHit = self._raycaster.cast(...)` while the next line handles `None`; should be `RayHit | None`.

**16. The modified examples have no CI coverage** - `tests/test_examples.py:24`

The allow pattern is `"viewer_plugins/**/*.py"` but the directory is `examples/viewer_plugin/` (singular), so neither example is discovered and the new `--use_visual_geom` paths are never run.

PR #3134, opened 2026-07-30
[BUG FIX] Keep the Nyx plugin check working on pull requests that predate it.

duburcqa:
## Description

Falls back to the pin of the default branch when the revision under test carries none, instead of failing outright.

## Motivation and Context

Reading the pin from the revision under test is what lets a pull request choose its plugin revision, but a revision branched before the pin existed has no such file, and every open pull request predating it is in that situation. The job died on the very first use, before checking anything out.

The fallback query was run against the default branch and returns the pinned hash, which passes the same validation as an in-tree pin.

## Checklist:

### DIFF
--- a/.github/workflows/nyx_plugin.yml
+++ b/.github/workflows/nyx_plugin.yml
@@ -59,8 +59,17 @@ jobs:
       - name: Resolve the pinned Nyx plugin revision
         id: plugin
         shell: bash
+        env:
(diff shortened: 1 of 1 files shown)


duburcqa:
Superseded: the pin must come from the revision under test, so a stale branch is rejected with an explicit message asking for a rebase rather than falling back. Reopened as a correctly named branch.

PR #3143, opened 2026-07-30
[BREAKING][FEATURE] Support applying an external wrench at any local point on a link.

### LINKED ISSUE #1921: [Feature]: Expose public API to apply force on a surface point of an entity.

@duburcqa Can we have a API to specify a point force on an object? i.e., a force being applied to a surface point?

Image

### Motivation

Such point force interactions have been used here: https://arxiv.org/abs/2509.20358

### Potential Benefit

Granular control of how we can interact with object specifically for research use. 

- [ ] API for specifying the point force on a object surface point.

### Additional information

_No response_

Milotrince:
## Description
- `RigidSolver.apply_links_external_{force,torque}` are replaced by a single `apply_links_external_wrench(force=None, torque=None, ..., pos=None, ref=..., local=False)`, mirrored by `RigidEntity.apply_links_external_wrench` and `RigidLink.apply_external_wrench`.
- `RigidLink.apply_external_{force,torque}` are shortcuts over the link wrench, for the common case of pushing or twisting a single link. Their docstrings refer to `apply_external_wrench` rather than restating it.
- `pos=` applies the linear force at an arbitrary point of the link, world or link-local, which is what a grab / drag interaction needs.
- Fixes bug: `ref="link_origin", local=True` rotated the force by the principal-inertia frame instead of the link frame.
- `ref` now takes a `gs.link_ref_frame` member rather than a string literal, and `gs.link_ref_frame.root_COM` is no longer accepted for wrenches nor for `get_links_vel`, where it had no use.
- Updates `MouseInteractionPlugin` to use the new `apply_external_wrench` at the mouse point.

## Related Issue
Resolves https://github.com/Genesis-Embodied-AI/genesis-world/issues/1921
Docs https://github.com/Genesis-Embodied-AI/genesis-doc/pull/145

## Motivation and Context
Better for mouse interaction, playing with the physics, useful in applications.

At kernel level a single `func_apply_link_external_wrench` does the frame and moment-arm math, reached through two top-level kernels: one applying the wrench at the reference frame origin, one reading a per-link application point. The reference frame reaches them as a plain int, because quadrants' fastcache holds a weak reference to template arguments and enum members do not support weakrefs.

The `doc` submodule pointer is bumped to genesis-doc#145, which rewrites `user_guide/getting_started/control_your_robot.md` around the new API, and genesis-doc#147, which quotes the enum there and documents it on the `RigidSolver` reference page.

`tests/rigid/test_dynamics.py::test_apply_external_wrench`

Updated the `MouseInteractionPlugin` so that it feels much nicer by default!
 
`python examples/viewer_plugin/mouse_interaction.py --use-visual-geom --use-force`

BEFORE

https://github.com/user-attachments/assets/41c29166-cb08-4a2b-8bc0-15052bbfcb77

AFTER

https://github.com/user-attachments/assets/20ad5c01-b989-4c9e-a773-d20699a41db6

## Checklist:

- [x] I updated the [documentation](https://github.com/Genesis-Embodied-AI/genesis-doc) accordingly https://github.com/Genesis-Embodied-AI/genesis-doc/pull/145

Resolves SIM-313

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -10,12 +10,13 @@ import torch
 import trimesh
 
 import genesis as gs
+from genesis.constants import link_ref_frame
 from genesis.engine.materials.base import Material
 from genesis.engine.mesh import InertialProperties
 from genesis.engine.states.entities import RigidEntityState
 from genesis.options.morphs import Morph
 from genesis.options.surfaces import Surface
-from genesis.typing import LinkFrameType, LinkRefFrameType, UnitVec4FType, Vec3FType
+from genesis.typing import UnitVec4FType, Vec3FType
 from genesis.utils import geom as gu
 from genesis.utils import mesh as mu
 from genesis.utils import mjcf as mju
@@ -3462,7 +3463,7 @@ class RigidEntity(KinematicEntity):
         links_idx_local=None,
         envs_idx=None,
         *,
-        ref: LinkRefFrameType = "link_origin",
+        ref: link_ref_frame = link_ref_frame.link_origin,
         relative=True,
     ):
         """
@@ -3474,15 +3475,15 @@ class RigidEntity(KinematicEntity):
             The indices of the links. Defaults to None.
         envs_idx : None | array_like, optional
             The indices of the environments. If None, all environments will be considered. Defaults to None.
-        ref: "link_origin" | "link_com" | "root_com"
-            The reference point being used to express the position of each link.
-            * "root_com": center of mass of the sub-entities to which the link belongs. As a reminder, a single
-              kinematic tree (aka. 'RigidEntity') may compromise multiple "physical" entities, i.e. a kinematic tree
-              that may have at most one free joint, at its root.
+        ref: gs.link_ref_frame, optional
+            The reference point used to express the position of each link: its origin ('link_origin'), its center of
+            mass ('link_COM'), or the center of mass of the sub-entity it belongs to ('root_COM'). A single
+            'RigidEntity' may comprise several physical sub-entities, each a kinematic sub-tree with at most one free
+            joint at its root. Defaults to 'link_origin'.
         relative : bool, optional
             If True, strip the morph pose offset to return the user-frame position; this only affects
-            ref="link_origin", since the offset is defined on the link origin. If False, return the world frame.
-            Defaults to True.
+            ref=gs.link_ref_frame.link_origin, since the offset is defined on the link origin. If False, return the
+            world frame. Defaults to True.
 
         Returns
         -------
@@ -3493,7 +3494,7 @@ class RigidEntity(KinematicEntity):
         return self._solver.get_links_pos(links_idx, envs_idx, ref=ref, relative=relative)
 
     @gs.assert_built
-    def get_links_vel(self, links_idx_local=None, envs_idx=None, *, ref: LinkFrameType = "link_origin"):
+    def get_links_vel(self, links_idx_local=None, envs_idx=None, *, ref: link_ref_frame = link_ref_frame.link_origin):
         """
         Returns linear velocity of all the entity's links expressed at a given reference position in world coordinates.
 
@@ -3503,9 +3504,9 @@ class RigidEntity(KinematicEntity):
             The indices of the links. Defaults to None.
         envs_idx : None | array_like, optional
             The indices of the environments. If None, all environments will be considered. Defaults to None.
-        ref: "link_origin" | "link_com"
-            The reference point being used to expressed the velocity of each link.
-
+        ref: gs.link_ref_frame, optional
+            The reference point used to express the velocity of each link: its origin ('link_origin') or its center of
+            mass ('link_COM'). Defaults to 'link_origin'.
         Returns
         -------
         vel : torch.Tensor, shape (n_links, 3) or (n_envs, n_links, 3)
@@ -3533,7 +3534,7 @@ class RigidEntity(KinematicEntity):
         envs_idx=None,
         *,
         pos=None,
-        ref: LinkFrameType = "link_origin",
+        ref: link_ref_frame = link_ref_frame.link_origin,
         local: bool = False,
     ):
         """
@@ -3544,25 +3545,22 @@ class RigidEntity(KinematicEntity):
         force : None | array_like, optional
             The linear force to apply. None for a pure torque. Defaults to None.
         torque : None | array_like, optional
-            The torque to apply, on top of the moment induced by the linear force. None for a pure force. Defaults to
-            None.
+            The torque to apply, on top of the moment induced by the linear force. Defaults to None.
         links_idx_local : None | array_like, optional
             The indices of the links. None to specify all the entity's links. Defaults to None.
         envs_idx : None | array_like, optional
             The indices of the environments. If None, all environments will be considered. Defaults to None.
         pos : None | array_like, optional
-            The point at which the linear force is applied, which sets the moment arm of the induced torque. None to
-            apply it at the origin of the reference frame designated by `ref`. With `local=True`, it is an offset from
-            that origin expressed in the coordinates of that frame, hence a point that follows the link as it moves.
-            Otherwise, it is a world position that locates the point on its own, leaving `ref` to only select the frame
-            of `force` and `torque`. Defaults to None.
-        ref: "link_origin" | "link_com", optional
-            The reference frame on which the wrench will be applied. "link_origin" refers to the origin of each link and
-            "link_com" refers to the center of mass of each link. This argument only selects the frame of the input
-            coordinates unless a linear force is applied at the origin of that frame, ie without specifying `pos`.
+            Where the linear force is applied, which sets the moment arm of the induced torque. With `local=True`, an
+            offset from the origin of the `ref` frame, so the point follows each link as it moves; otherwise a world
+            position. None applies the force at the origin of the `ref` frame. Defaults to None.
+        ref: gs.link_ref_frame, optional
+            The reference frame: the origin of each link ('link_origin'), or its center of mass ('link_COM'). It fixes
+            where the linear force acts when `pos` is None, and the axes of the input coordinates when `local=True`.
+            Defaults to 'link_origin'.
         local: bool, optional
-            Whether the wrench and the application point are expressed in the local coordinates associated with the
-            reference frame instead of world frame.
+            Whether `force`, `torque` and `pos` are expressed in the coordinates of the `ref` frame rather than the
+            world frame. Defaults to False.
         """
         links_idx = self._get_global_idx(links_idx_local, self.n_links, self._link_start, unsafe=True)
         self._solver.apply_links_external_wrench(force, torque, links_idx, envs_idx, pos=pos, ref=ref, local=local)
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -9,16 +9,13 @@ import numpy as np
 import torch
 import trimesh
 
-import quadrants as qd
-
 import genesis as gs
 from genesis.engine.materials.base import Material
 from genesis.engine.mesh import InertialProperties
 from genesis.engine.states.entities import RigidEntityState
 from genesis.options.morphs import Morph
 from genesis.options.surfaces import Surface
-from genesis.typing import LinkRefFrameType, UnitVec4FType, Vec3FType
-from genesis.utils import array_class
+from genesis.typing import LinkFrameType, LinkRefFrameType, UnitVec4FType, Vec3FType
 from genesis.utils import geom as gu
 from genesis.utils import mesh as mu
 from genesis.utils import mjcf as mju
@@ -3496,7 +3493,7 @@ class RigidEntity(KinematicEntity):
         return self._solver.get_links_pos(links_idx, envs_idx, ref=ref, relative=relative)
 
     @gs.assert_built
-    def get_links_vel(self, links_idx_local=None, envs_idx=None, *, ref: LinkRefFrameType = "link_origin"):
+    def get_links_vel(self, links_idx_local=None, envs_idx=None, *, ref: LinkFrameType = "link_origin"):
         """
         Returns linear velocity of all the entity's links expressed at a given reference position in world coordinates.
 
@@ -3506,7 +3503,7 @@ class RigidEntity(KinematicEntity):
             The indices of the links. Defaults to None.
         envs_idx : None | array_like, optional
             The indices of the environments. If None, all environments will be considered. Defaults to None.
-        ref: "link_origin" | "link_com" | "root_com"
+        ref: "link_origin" | "link_com"
             The reference point being used to expressed the velocity of each link.
 
         Returns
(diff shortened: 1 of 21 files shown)


Milotrince:
@codex review

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -1521,7 +1541,15 @@ def apply_links_external_force(
...
-            links_idx, envs_idx, force, self.dyn_state, self.rigid_config, ref_idx, 1 if local else 0
+            links_idx,
+            envs_idx,
+            pos,
+            force,
+            self.dyn_state,
+            self.rigid_config,
+            ref_idx,
+            1 if local else 0,
+            1 if has_pos else 0,
```
duburcqa: This is violating the guidelines.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -3697,9 +3698,7 @@ def get_links_pos(
         return self._solver.get_links_pos(links_idx, envs_idx, ref=ref, relative=relative)
 
     @gs.assert_built
-    def get_links_vel(
-        self, links_idx_local=None, envs_idx=None, *, ref: Literal["link_origin", "link_com"] = "link_origin"
-    ):
+    def get_links_vel(self, links_idx_local=None, envs_idx=None, *, ref: LinkRefFrameType = "link_origin"):
```
duburcqa: Remove root com support. Nobody wants that for velocities.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -3730,6 +3729,76 @@ def get_links_acc_ang(self, links_idx_local=None, envs_idx=None):
...
 
+    @gs.assert_built
+    def apply_links_external_force(
+        self,
+        force,
+        links_idx_local=None,
+        envs_idx=None,
+        *,
+        pos=None,
+        ref: LinkRefFrameType = "link_origin",
```
duburcqa: Remove root com support. Nobody wants that for forces.

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -996,6 +996,68 @@ def get_AABB(self):
...
         return torch.stack((verts.min(dim=-2).values, verts.max(dim=-2).values), dim=-2)
 
+    @gs.assert_built
+    def apply_external_force(
+        self,
+        force,
+        envs_idx=None,
+        *,
+        pos=None,
+        ref: LinkRefFrameType = "link_origin",
```
duburcqa: Remove root com support. Nobody wants that for forces.

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -996,6 +996,68 @@ def get_AABB(self):
...
+        self._solver.apply_links_external_force(force, (self._idx,), envs_idx, pos=pos, ref=ref, local=local)
+
+    @gs.assert_built
+    def apply_external_torque(
+        self,
+        torque,
+        envs_idx=None,
+        *,
+        ref: LinkRefFrameType = "link_origin",
+        local: bool = False,
```
duburcqa: Remove root com support. Nobody wants that for torques.

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -996,6 +996,68 @@ def get_AABB(self):
         verts = self.get_verts()
         return torch.stack((verts.min(dim=-2).values, verts.max(dim=-2).values), dim=-2)
 
+    @gs.assert_built
+    def apply_external_force(
```
duburcqa: I think it would make more sense to have a single 'apply_external_wrench' that accepts generalised forces, ie classical linear force and torque, both optional (with at least one having the be specified).

If you really think it is worth it, we could consider writing some shortcut to 'apply_external_force' and 'apply_external_torque' for convenience on top of it, but not the other way around.

genesis/engine/solvers/rigid/abd/accessor.py
```diff
@@ -1157,8 +1157,8 @@ def kernel_set_drone_rpm(
             if invert:
                 torque = -torque
 
-            func_apply_link_external_force(i_l, i_b, force, dyn_state, 1, 1)
-            func_apply_link_external_torque(i_l, i_b, torque, dyn_state, 1, 1)
+            func_apply_link_external_force(i_l, i_b, qd.Vector.zero(gs.qd_float, 3), force, dyn_state, 1, True, False)
+            func_apply_link_external_torque(i_l, i_b, torque, dyn_state, 1, True)
```
duburcqa: Replace plain '1' with proper enum. It would be more nicer.

genesis/engine/solvers/rigid/abd/misc.py
```diff
@@ -747,16 +747,21 @@ def kernel_init_equality_fields(
 def kernel_apply_links_external_force(
     links_idx: qd.types.ndarray(),
     envs_idx: qd.types.ndarray(),
+    pos: qd.types.ndarray(),
     force: qd.types.ndarray(),
     dyn_state: array_class.DynState,
     rigid_config: qd.template(),
     ref: qd.template(),
     local: qd.template(),
+    has_pos: qd.template(),
```
duburcqa: I hate this mechanism combining 'pos' arg + 'has_pos'. We need to find a better way to implement this.
I opened a dedicated issue on Quadrants: https://github.com/Genesis-Embodied-AI/quadrants/issues/856

In the meantime, I would rather have 2 separated top-level kernel API, unifying only at the level of the qd.func.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -1467,19 +1470,36 @@ def apply_links_external_force(
...
+                pos = pos[None]
         force, links_idx, envs_idx = self._sanitize_io_variables(
             force, links_idx, self.n_links, "links_idx", envs_idx, (3,), skip_allocation=True
         )
         if self.n_envs == 0:
             force = force[None]
+        if not has_pos:
+            # The kernel cannot fabricate the array it only reads when an application point is provided, so hand it a
+            # zero-length view rather than paying for a fresh allocation on every call.
+            pos = force[:0]
```
duburcqa: Remove this and rather add kernel selection branching.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -1498,7 +1518,7 @@ def apply_links_external_force(
             )
 
         kernel_apply_links_external_force(
-            links_idx, envs_idx, force, self.dyn_state, self.rigid_config, ref_idx, 1 if local else 0
+            links_idx, envs_idx, pos, force, self.dyn_state, self.rigid_config, ref_idx, local, has_pos
         )
 
     def apply_links_external_torque(
```
duburcqa: As I said, I would rather suggest to remove 'apply_links_external_(force|torque)' in favour of only 'wrench'. Especially at kernel level, I don't think convenience wrappers for linear force and torque separately are necessary at all. It should be avoided entirely.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -2759,9 +2780,7 @@ def get_links_pos(
 
         return tensor[0] if self.n_envs == 0 else tensor
 
-    def get_links_vel(
-        self, links_idx=None, envs_idx=None, *, ref: Literal["link_origin", "link_com", "root_com"] = "link_origin"
-    ):
+    def get_links_vel(self, links_idx=None, envs_idx=None, *, ref: LinkRefFrameType = "link_origin"):
```
duburcqa: Same comment for 'ref: LinkRefFrameType'.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -1507,11 +1527,11 @@ def apply_links_external_torque(
         links_idx=None,
         envs_idx=None,
         *,
-        ref: Literal["link_origin", "link_com", "root_com"] = "link_origin",
+        ref: LinkRefFrameType = "link_origin",
```
duburcqa: Same comment for 'ref: LinkRefFrameType'.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -1453,11 +1455,12 @@ def apply_links_external_force(
         links_idx=None,
         envs_idx=None,
         *,
-        ref: Literal["link_origin", "link_com", "root_com"] = "link_origin",
+        pos=None,
+        ref: LinkRefFrameType = "link_origin",
```
duburcqa: Same comment for 'ref: LinkRefFrameType'.

tests/rigid/test_dynamics.py
```diff
@@ -255,6 +264,55 @@ def test_apply_external_forces(xml_path, show_viewer):
...
     )
 
+    # A local force and a local application point are both expressed in the frame that 'ref' designates, which only
+    # shows on a link whose inertial frame is rotated with respect to its own frame.
+    base_link = robot.get_link("base")
+    assert not np.allclose(base_link.inertial_quat, (1.0, 0.0, 0.0, 0.0))
+    base_link_pos = tensor_to_array(rigid_solver.get_links_pos([base_link.idx])[0])
+    base_link_quat = tensor_to_array(rigid_solver.get_links_quat([base_link.idx])[0])
+    base_link_com = tensor_to_array(rigid_solver.get_links_pos([base_link.idx], ref="link_com")[0])
+    base_root_COM = tensor_to_array(rigid_solver.get_links_root_COM([base_link.idx])[0])
```
duburcqa: Remove extra list wrapping, it is useless (and harmful performance-wise). We strive to promote the recommended patterns in the unit tests, so it is important to write them as cleanly as possible.

genesis/engine/entities/rigid_entity/rigid_link.py
Milotrince: I will add the shortcut, because it would be easier for users to search for and understand. The docstring will point them to `apply_external_wrench`.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -3462,7 +3462,7 @@ def get_links_pos(
         links_idx_local=None,
         envs_idx=None,
         *,
-        ref: Literal["link_origin", "link_com", "root_com"] = "link_origin",
+        ref: LinkRefFrameType = "link_origin",
```
Milotrince: isn't LINK_REF_FRAME an enum? we still have the string literal?

Milotrince: Two layers, deliberately.

The string is the public spelling and it predates this PR: on main line 3465 already read

    ref: Literal["link_origin", "link_com", "root_com"] = "link_origin"

and all this PR did there was give that inline Literal a name, 'LinkRefFrameType' in genesis/typing.py. So no string literal was introduced; one was factored out of three signatures.

The enum is what is new, and it is the kernel spelling. On main the frame reached the kernels as a bare 0/1/2 index; now 'RigidSolver._convert_ref_to_frame' maps the user string onto a 'gs.LINK_REF_FRAME' value and the qd.template() branches compare against named members ('abd/misc.py:852', 'abd/accessor.py:1013'). That is the magic number this PR removes.

It is also internal on purpose: constants.py splits at '######### User accessible constants do not capitalize #########', and LINK_REF_FRAME sits above that line with GEOM_TYPE, JOINT_TYPE and CTRL_MODE. The user-facing enums below it are lowercase ('gs.integrator.Euler', 'gs.friction_cone.elliptic'), and they are option values, not per-call frame selectors.

Making the public argument take the enum instead would break every existing 'ref="link_origin"' call and the docs, and it would not remove the conversion anyway: '_convert_ref_to_frame' returns 'int(...)' because quadrants' fastcache holds a weak reference to template arguments and enum members do not support weakrefs, so a member cannot be handed to a kernel as-is. Accepting both spellings is possible but that is two ways to say one thing, on an argument users pass by hand.

Happy to revisit if you want the enum public - it is an API call, not a mechanical one, and it would want its own PR since it breaks callers.

Milotrince: It would be nice to make the enum to be public (so dont capitalize it) and use that everywhere instead of the string literals

Milotrince: could we use the same enum, in non-capital spelling, instead of having the string literal and the enum?

tests/rigid/test_dynamics.py
```diff
@@ -254,29 +262,79 @@ def test_apply_external_forces(xml_path, show_viewer):
...
-        rigid_solver.apply_links_external_torque(
-            torque=(0, 0, 0), links_idx=[duck_link_idx], ref="root_com", local=True
+    # A local force and a local application point are both expressed in the frame that 'ref' designates, which only
+    # shows on a link whose inertial frame is rotated with respect to its own frame.
+    base_link = robot.get_link("base")
+    assert not np.allclose(base_link.inertial_quat, (1.0, 0.0, 0.0, 0.0))
+    base_link_pos = tensor_to_array(rigid_solver.get_links_pos(base_link.idx)[0])
+    base_link_quat = tensor_to_array(rigid_solver.get_links_quat(base_link.idx)[0])
+    base_link_com = tensor_to_array(rigid_solver.get_links_pos(base_link.idx, ref=gs.link_ref_frame.link_COM)[0])
+    base_root_COM = tensor_to_array(rigid_solver.get_links_root_COM(base_link.idx)[0])
```
duburcqa: Let's keep using torch all along. Similarly, I don't think doing '[0]' is necessary.

tests/rigid/test_kinematics.py
```diff
@@ -57,8 +57,8 @@ def test_link_velocity(gs_sim, tol):
             0.0,
         ]
     )
-    link_COM0 = gs_sim.rigid_solver.get_links_pos(ref="link_com")[0]
-    link_COM1 = gs_sim.rigid_solver.get_links_pos(ref="link_com")[1]
+    link_COM0 = gs_sim.rigid_solver.get_links_pos(ref=gs.link_ref_frame.link_COM)[0]
```
duburcqa: `link_COM0, link_COM1 = gs_sim.rigid_solver.get_links_pos(ref=gs.link_ref_frame.link_COM)`

genesis/engine/solvers/rigid/abd/misc.py
duburcqa: Done in 97b5720: two top-level kernels, 'kernel_apply_links_external_wrench' and 'kernel_apply_links_external_wrench_at_pos', and 'has_pos' now only exists inside 'func_apply_link_external_wrench'. Quadrants#856 is still open, so the no-point kernel still hands the func a zero vector for the unused slot; that goes away once an optional argument can specialize the absent path.

tests/rigid/test_dynamics.py
duburcqa: Done in 1fb2b0c: the block reads the getters as torch all the way through, drops the 'tensor_to_array' round-trips and the '[0]', and the last assertion now reuses the same 'force_world' / 'point_world' spelling as the two above it.

tests/rigid/test_kinematics.py
duburcqa: Done in 1fb2b0c.

PR #3144, opened 2026-07-30
[MISC] Add missing docstrings and update formatting.

Milotrince:
## Description
- Add docstrings for constants (were missing previously)
- Add docstrings for RigidSolver methods
- Keep formatting (backticks, etc.) consistent and enable autodoc to link to the code object if applicable
- Added notes for this convention in `CODING_GUIDELINES.md`
- Deduplicated `CODING_GUIDELINES.md` and `CLAUDE.md` (moved to `AGENTS.md` and point `CLAUDE.md` to it; this seems to be the standard, since not everyone uses Claude only)

## Related Issue
https://github.com/Genesis-Embodied-AI/genesis-doc/pull/139

## Motivation and Context
improve docs

n/a

image

Screenshot From 2026-08-03 18-11-56

## Checklist:


duburcqa:
- `integrator.implicitfast` "for the cost of a second mass-matrix factorization per step": wrong attribution. Euler pays the same second factorization whenever a dof has damping, per the class-level sentence itself. The delta of `implicitfast` is the velocity-actuator bias term, gated on `integrator == gs.integrator.implicitfast` at `forward_dynamics.py:557`. Under `enable_mujoco_compatibility` the mask branch is skipped (`forward_dynamics.py:1578`), so the refactor is unconditional instead of damping-triggered.
- The implicit midpoint update is listed only under `implicitfast`, but its gate is `integrator != gs.integrator.Euler` (`forward_dynamics.py:1512`): `approximate_implicitfast` does it too. "That same correction" in the next bullet does not say which correction.
- `constraint_solver.CG` "keeping memory low on scenes with very many degrees of freedom or constraints": not delivered. `nt_H` is allocated `(_B, n_dofs_, n_dofs_)` unconditionally at `array_class.py:630-633`, with no gating on the selected solver. The matrix-vector-products half is correct; state the tradeoff as per-iteration cost versus Newton's few-iteration convergence.
- `approximate_implicitfast` "Not exact, and adequate in practice": say what the approximation costs.
- `EQUALITY_TYPE.JOINT` "quartic" matches `constraint/solver.py:1325-1329` (`a0..a4`), but MuJoCo's overview calls the same constraint cubic. Write "degree-4 polynomial in the driving joint's position".
- `options/solvers.py:431-439` still carries a longer, differently-framed account of the same three integrators, and `:442-444` still gives no tradeoff for `constraint_solver`. Reduce both to summary plus "See `gs.X` for the description of each model", as at `:466`, `:474`, `:532`. Drop "consistent with their Mujoco counterpart" at `:433`.
- The new docstrings use double backticks; `friction_cone` and `contact_resolution` use plain quoted names. Pick one for the file. The first commit message also claims those two use the numpydoc `Attributes` style - they are prose, only `broadphase_traversal` uses `Attributes`.
- `CTRL_MODE`, `IMAGE_TYPE` are still undocumented

CLAUDE.md
```diff
@@ -1,182 +1 @@
...
-- **Do not gate per-step asserts** with `float(...)` / `.item()` casts on tensor reductions. Design the scenario (e.g. a warmup loop that spins joints up) so the assertion holds unconditionally, then assert on the full tensor with `.all()`.
-- **Exact analytical dynamics checks:** force `gs.integrator.Euler` so the finite-differenced `qacc` equals the solver's, and account for both rigid-body rotational inertia and the implicit-damping first-order correction (`effective_inertia = I + damping*dt`, as in `test_position_control`) rather than loosening tolerances or distorting geometry.
-
-## Testing on Cluster
-
-- SSH: `ssh genesis-coreweave`
-- Path: `/mnt/home/duburcqa/workspace/src/genesis`
```
Milotrince, line 149: FYI @duburcqa I removed this section :p

Milotrince:
@codex review

CLAUDE.md
duburcqa: I was aware it was here. I don't think this should be removed because it is useful internally, but it should probably be made more generic. I "Leaking" these information is not dangerous, but it is surely irrelevant for external contributors. I don't know how to make this "private".

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -1357,8 +1411,8 @@ def _constraint_force_grad(self):
         """Backward pass for the constraint solver: seed dL_dqacc from acc.grad, run the adjoint solve, fold its
         outputs back into the autodiff grad fields, then reverse the constraint-row assembly."""
         constraint_state = self.constraint_solver.constraint_state
-        # Pure grad shuffles: in-place through zero-copy views when supported, kernel dispatch otherwise (see
-        # "Pure read-write data accessors on the hot path" in CLAUDE.md). gs.use_zerocopy encodes every zero-copy
+        # Pure grad shuffles: in-place through zero-copy views when supported, kernel dispatch otherwise (see "Pure
+        # read-write data accessors on the hot path" in CODING_GUIDELINES.md). gs.use_zerocopy encodes every zero-copy
         # availability condition for these buffers (standalone dense allocations with DLPack-supported dtypes; the
         # platform gates live in gs.init). torch (MPS) and quadrants do not share a compute stream on Metal, so the
         # writes are flushed before the next kernels (see set_base_links_quat).
```
duburcqa, line 1418: I don't think any of these information are relevant. The entire comment should probably be removed, with the relevant details dispatched closer to where it takes place rather than centralised in one general lengthy comment.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -1321,6 +1362,19 @@ def _kernel_detect_collision(self):
...
+
+        Parameters
+        ----------
+        env_idx : int, optional
+            Index of the environment to report contacts for. Defaults to 0.
+
+        Returns
+        -------
+        numpy.ndarray, shape (n_contacts, 2)
+            Scene-level indices of the two geoms of each contacting pair.
```
duburcqa, line 1376: If it really returns numpy data, it is a defect.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -2513,37 +2789,227 @@ def _set_dofs_info(self, tensor_list, dofs_idx, name, envs_idx=None):
             gs.raise_exception(f"Invalid `name` {name}.")
 
     def set_dofs_kp(self, kp, dofs_idx=None, envs_idx=None):
+        """
+        Set the positional gain of each dof's PD controller, leaving the velocity gain `set_dofs_kv` holds untouched.
```
duburcqa, line 2793: > leaving the velocity gain `set_dofs_kv` holds untouched

Weird wording.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -2513,37 +2789,227 @@ def _set_dofs_info(self, tensor_list, dofs_idx, name, envs_idx=None):
...
+        ----
+        This is a model parameter rather than state. Setting it per environment requires `batch_dofs_info`
+        on `~genesis.options.solvers.RigidOptions`; without it `envs_idx` is ignored and the value is
+        shared by every environment.
+        """
         self._set_dofs_info([kv], dofs_idx, "kv", envs_idx)
 
     def set_dofs_act_gain(self, act_gain, dofs_idx=None, envs_idx=None):
+        """
+        Set the actuator gain of each dof, which leaves the controller no longer PD-reducible.
```
duburcqa, line 2833: The definition of "actuator gain" is not universal. It is inherited from Mujoco. It should be documented here.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -2513,37 +2789,227 @@ def _set_dofs_info(self, tensor_list, dofs_idx, name, envs_idx=None):
...
+            Positional gains, broadcast across the selected dofs.
+        dofs_idx : None | array_like, optional
+            Scene-level indices of the dofs to set. If None, every dof is set. Defaults to None.
+        envs_idx : None | array_like, optional
+            Indices of the environments to set. If None, every environment is set. Defaults to None.
+        Note
+        ----
+        This is a model parameter rather than state. Setting it per environment requires `batch_dofs_info`
+        on `~genesis.options.solvers.RigidOptions`; without it `envs_idx` is ignored and the value is
+        shared by every environment.
```
duburcqa, line 2807: > without it `envs_idx` is ignored and the value is shared by every environment.

Is it really ignored? If it is the case, then it is a bug.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -2513,37 +2789,227 @@ def _set_dofs_info(self, tensor_list, dofs_idx, name, envs_idx=None):
...
+        """
         self._set_dofs_info([act_gain], dofs_idx, "act_gain", envs_idx)
 
     def set_dofs_act_bias(self, bias0, bias1, bias2, dofs_idx=None, envs_idx=None):
+        """
+        Set the three actuator bias coefficients of each dof.
```
duburcqa, line 2853: Same. The definition of "actuator bias" should be provided, pointing to mujoco's official doc for more details.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -2513,37 +2789,227 @@ def _set_dofs_info(self, tensor_list, dofs_idx, name, envs_idx=None):
...
+        ----
+        This is a model parameter rather than state. Setting it per environment requires `batch_dofs_info`
+        on `~genesis.options.solvers.RigidOptions`; without it `envs_idx` is ignored and the value is
+        shared by every environment.
+        """
         self._set_dofs_info([stiffness], dofs_idx, "stiffness", envs_idx)
 
     def set_dofs_armature(self, armature, dofs_idx=None, envs_idx=None):
+        """
+        Set the armature of each dof, added to its own entry of the mass matrix.
```
duburcqa, line 2919: This armature model should be documented here. The current formulation is an approximation and does not model faithfully the armature because the Coriolis effect is ignored. Moreover, the "armature" of the dof means nothing. Physically, it is modelling the apparent inertia of the armature of the rotor of rotary actuators specifically, as it its viewed from the output of the motor block, which comprises both the motor and the transmission system (eg gear reduction mechanism in general).

To be realistic, only the "true" rotor armature should be specified, ignoring the transmission, with the transmission being simulated independently.

Besides, it would be great to explain in which context setting these options is relevant. Here it can be used for numerical stability in the presence of light objects or unactuated joints. And setting armature for free joint is almost certainly a mistake. 

In general, I don't think adding a doc stating that `set_dofs_armature` "Set DoF armature" brings any value. If we are doing to document the public API, with need to provide information that would be useful for the end users.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -2513,37 +2789,227 @@ def _set_dofs_info(self, tensor_list, dofs_idx, name, envs_idx=None):
...
+        ----
+        This is a model parameter rather than state. Setting it per environment requires `batch_dofs_info`
+        on `~genesis.options.solvers.RigidOptions`; without it `envs_idx` is ignored and the value is
+        shared by every environment.
+        """
         self._set_dofs_info([armature], dofs_idx, "armature", envs_idx)
 
     def set_dofs_damping(self, damping, dofs_idx=None, envs_idx=None):
+        """
+        Set the damping of each dof, which resists its velocity.
```
duburcqa, line 2939: Same. It would be great to give a bit more details. What is unintuitive is that it is not really physically motivated and damping can be added on free joints. Moreover, what is a dof is not clear. I'm also not a huge fan of acronyms in documentation. Besides, it would be great to explain in which context setting these options is relevant. Here it can be used for numerical stability in the presence of light objects or unactuated joints.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -2513,37 +2789,227 @@ def _set_dofs_info(self, tensor_list, dofs_idx, name, envs_idx=None):
...
+        ----
+        This is a model parameter rather than state. Setting it per environment requires `batch_dofs_info`
+        on `~genesis.options.solvers.RigidOptions`; without it `envs_idx` is ignored and the value is
+        shared by every environment.
+        """
         self._set_dofs_info([damping], dofs_idx, "damping", envs_idx)
 
     def set_dofs_frictionloss(self, frictionloss, dofs_idx=None, envs_idx=None):
+        """
+        Set the dry friction force each dof resists motion with.
```
duburcqa, line 2959: Same. Not informative enough. No equation provided. No reference to Mujoco's doc... Pretty much useless.

I would rather not document the API than document it poorly. Without documentation, we can pretend it is because we are willing to ship something great. Here we openly ship some bad without excuses anymore.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -2513,37 +2789,227 @@ def _set_dofs_info(self, tensor_list, dofs_idx, name, envs_idx=None):
...
+        This is a model parameter rather than state. Setting it per environment requires `batch_dofs_info`
+        on `~genesis.options.solvers.RigidOptions`; without it `envs_idx` is ignored and the value is
+        shared by every environment.
+        """
         self._set_dofs_info([lower, upper], dofs_idx, "limit", envs_idx)
 
     @mutates(StateChange.GEOMETRY, links=MutatedLinks.ARTICULATED)
     def set_dofs_position(self, position, dofs_idx=None, envs_idx=None):
+        """
+        Set the position of each dof, resetting the contacts it invalidates.
```
duburcqa, line 3002: This doc is pointless while useful information would be expected for such a critical method. What is the intended use case? Is it recommended? Any concern in terms of physics validity? What happens if it violate equality constraints of closed kinematic chains? What happens if it starts with inter-penetration...etc

Moreover, nobody cares about "resetting the contacts it invalidates". This is a purely internal detail. If you want to mention this, it should be in some dedicated "internal" section or something.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -2589,6 +3055,20 @@ def _wake_dofs(self, dofs_idx, envs_idx):
             )
 
     def set_dofs_velocity(self, velocity, dofs_idx=None, envs_idx=None, *, skip_forward=False):
+        """
+        Set the velocity of each dof, waking any hibernated entity it belongs to.
```
duburcqa, line 3059: Same as 'set_dofs_position'.

Moreover, nobody cares about "waking any hibernated entity it belongs to". This is a purely internal detail. If you want to mention this, it should be in some dedicated "internal" section or something.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -2598,6 +3078,18 @@ def set_dofs_velocity(self, velocity, dofs_idx=None, envs_idx=None, *, skip_forw
         super().set_dofs_velocity(velocity, dofs_idx, envs_idx, skip_forward=skip_forward)
 
     def control_dofs_force(self, force, dofs_idx=None, envs_idx=None):
+        """
+        Apply a force to each dof directly, putting it under force control.
```
duburcqa, line 3082: This is useless and borderline wrong. How does this combined with the dof forces resulting from the free body dynamics and constraints? 

This is setting additional external forces at the dof level. Unlike PD control, it is not taken into account in the implicit integration scheme and therefore can lead to numerical instability.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -2598,6 +3078,18 @@ def set_dofs_velocity(self, velocity, dofs_idx=None, envs_idx=None, *, skip_forw
...
+
+        Parameters
+        ----------
+        force : array_like
+            Forces in newtons or newton-meters.
```
duburcqa, line 3087: Something cannot be "in newtons or newton-meters". This makes no sense. The lack of information is making your statement invalid and confusing for users.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -2618,6 +3110,18 @@ def control_dofs_force(self, force, dofs_idx=None, envs_idx=None):
         kernel_control_dofs_force(dofs_idx, envs_idx, force, self.dyn_state, self.rigid_config)
 
     def control_dofs_velocity(self, velocity, dofs_idx=None, envs_idx=None):
+        """
+        Set the velocity target of each dof, putting it under velocity control.
```
duburcqa, line 3114: What is velocity control? Under which condition? How does this interplays with the actuator gains?

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -2640,6 +3144,18 @@ def control_dofs_velocity(self, velocity, dofs_idx=None, envs_idx=None):
         kernel_control_dofs_velocity(dofs_idx, envs_idx, velocity, self.dyn_state, self.rigid_config)
 
     def control_dofs_position(self, position, dofs_idx=None, envs_idx=None):
+        """
+        Set the position target of each dof, putting it under position control.
```
duburcqa, line 3148: This is wrong. It does not enable "position control", assuming "position control" means anything in the first place. It enables PD control, with velocity target constant equal to zero. This can be view as a neutral configuration of a virtual spring under viscous friction. Which gains are effecting this spring-damper dynamics and how to set them should be mentioned.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -2662,6 +3178,20 @@ def control_dofs_position(self, position, dofs_idx=None, envs_idx=None):
         kernel_control_dofs_position(dofs_idx, envs_idx, position, self.dyn_state, self.rigid_config)
 
     def control_dofs_position_velocity(self, position, velocity, dofs_idx=None, envs_idx=None):
+        """
+        Set both the position and velocity target of each dof.
```
duburcqa, line 3182: What is the difference between 'control_dofs_position' / 'control_dofs_velocity' and this method? Which one are users supposed to use and when? Can perfect tracking be expected, and under which conditions...etc At this point it is useless.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -2729,6 +3259,26 @@ def get_links_pos(
         ref: Literal["link_origin", "link_com", "root_com"] = "link_origin",
         relative=False,
     ):
+        """
+        Get the position of each link, at the reference point selected.
```
duburcqa, line 3263: What is a reference point? What does it mean? What is the exact definition?

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -2758,7 +3308,25 @@ def get_links_pos(
     def get_links_vel(
         self, links_idx=None, envs_idx=None, *, ref: Literal["link_origin", "link_com", "root_com"] = "link_origin"
     ):
-        if gs.use_zerocopy:
+        """
+        Get the linear velocity of each link, expressed at the reference point selected.
```
duburcqa, line 3312: What is a reference point? What does it mean? What is the exact definition? In which frame this quantity is expressed?

With such a lack of information, this doc is at least misleading if not wrong.

duburcqa:
Stale. Let's reopen if some progresses are made.

Issue #3129, opened 2026-07-29
[Bug]: scene.build() fails with an unactionable PyOpenGL AttributeError when no GL platform is installed (ROCm containers)


himanshu748:
### Bug Description

On a container with no OpenGL libraries installed, `scene.build()` fails with a
`PyOpenGL` `AttributeError` that gives no indication of the actual cause. The
underlying problem is simply that no GL platform is present, but nothing in the
traceback says so, and the scene being built has `show_viewer=False` and no
cameras, so there is no obvious reason for the user to suspect rendering at all.

This is easy to hit on ROCm images specifically. A compute-only ROCm container
has no reason to ship `libEGL`, so the official AMD Radeon Cloud PyTorch/ROCm
template does not have one. Any state-based RL workload on such an image dies at
`scene.build()` with the message below.

To be explicit about scope: I am **not** asking for an option to disable the
visualizer. I saw that #1877 asked for that and was closed with a clear "this is
not going to happen", and I am not relitigating it. This is only about the error
being unactionable. In that same thread a maintainer noted "we are not aware of
any machine still facing issues with that at the moment", so this report is
mainly to supply one such machine, plus the one-line fix for it.

### Steps to Reproduce

On any image without `libEGL` or `libOSMesa` (for example `rocm/pytorch`):

```python
import genesis as gs

gs.init(backend=gs.amdgpu)
scene = gs.Scene(show_viewer=False)   # no viewer
scene.add_entity(gs.morphs.Plane())   # no cameras
scene.build(n_envs=1)
```

```
File "genesis/vis/rasterizer.py", line 40, in build
    self._renderer = pyrender.OffscreenRenderer(
File "genesis/ext/pyrender/offscreen.py", line 250, in _create
    from .platforms import egl
File "genesis/ext/pyrender/platforms/egl.py", line 30, in <module>
    from OpenGL import EGL as egl
File "OpenGL/raw/EGL/_types.py", line 87, in <module>
    raw_eglQueryString = _p.PLATFORM.EGL.eglQueryString
AttributeError: 'NoneType' object has no attribute 'eglQueryString'
```

`_p.PLATFORM.EGL` is `None` because PyOpenGL could not load an EGL
implementation. The `AttributeError` is a downstream symptom.

### Expected Behavior

An error naming the real cause, for example:

```
No OpenGL platform available. Genesis builds an offscreen renderer during
scene.build() even when show_viewer=False and no cameras are added.
Install one of:
[...]

PR #3145, opened 2026-07-31
[MISC] Report a clear error when EGL is unavailable for offscreen rendering.

himanshu748:
Follow-up to #3129, which was closed as expected behaviour. This does not ask for OpenGL to become optional. It implements the detection described there, "we detect early if it is supported, and if not, it raises an exception", so that the exception names EGL.

`scene.build()` creates an offscreen OpenGL context on Linux whether or not anything is rendered, so images built for compute alone, which CUDA and ROCm images generally are, fail there. What the user sees comes from inside PyOpenGL and mentions neither EGL nor Genesis:

```
File "genesis/vis/rasterizer.py", line 40, in build
    self._renderer = pyrender.OffscreenRenderer(
File "OpenGL/raw/EGL/_types.py", line 87, in <module>
    raw_eglQueryString = _p.PLATFORM.EGL.eglQueryString
AttributeError: 'NoneType' object has no attribute 'eglQueryString'
```

With this change:

```
genesis.GenesisException: EGL is required to create an offscreen OpenGL context, and it is unavailable on this system.
```

The probe sits in `_ensure_egl_loaded`, beside the plugin check already there. The original error is kept as `__cause__`, so the paths `dlopen` tried are still in the traceback for anyone who needs them.

## Reproducing

On any image without an OpenGL runtime, for example `rocm/dev-ubuntu-24.04`:

```python
import genesis as gs

gs.init(backend=gs.cpu)
scene.build(n_envs=1)
```

## Testing

Both branches were exercised by importing `genesis/ext/pyrender/platforms/egl.py` on a host with no `libEGL`: once as PyOpenGL reports it there, which is the `ImportError`, and once with PyOpenGL's `EGL` property forced to `None`. On `main` the second reproduces the `AttributeError` above exactly.

I have not added a unit test. Reaching either branch on a runner means removing `libEGL` from it or monkeypatching PyOpenGL's platform class, and the second is the kind of internals mocking the testing guidelines rule out. Glad to add one if you have a shape in mind.

### DIFF
--- a/genesis/ext/pyrender/platforms/egl.py
+++ b/genesis/ext/pyrender/platforms/egl.py
@@ -12,6 +12,8 @@ from .base import Platform
 EGL_PLATFORM_DEVICE_EXT = 0x313F
 EGL_DRM_DEVICE_FILE_EXT = 0x3233
 
+_EGL_LOAD_ERROR = "EGL is required to create an offscreen OpenGL context, and it is unavailable on this system."
+
 
 def _ensure_egl_loaded():
     plugin = OpenGL.platform.PlatformPlugin.by_name("egl")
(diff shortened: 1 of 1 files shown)


genesis/ext/pyrender/platforms/egl.py
duburcqa, line 1: I’m done think we should explain in the document why EGL is needed and how to fix it. There is no generic way to install EGL on a given platform. Here you want to focusing on Debian-based Ubuntu and installing mesa stack. But this is not the only option. Mentioning that EGL is required but not available on the system should be good enough.

himanshu748: Fair point, dropped both. The message is now one line:

`EGL is required to create an offscreen OpenGL context, and it is unavailable on this system.`

The underlying `dlopen` error is kept as `__cause__`, so the paths it tried are still in the traceback.

duburcqa (approved):

PR #3162, opened 2026-08-03
[MISC] Speed up raycast BVH rebuild by removing atomic contention in the scene-extent reduction.

Kashu7100:
## Description

`LBVH.compute_aabb_centers_and_scales` reduced the per-batch scene extent (used only to normalize Morton codes) with a flat atomic reduction:

```python
for i_b, i_a in qd.ndrange(self.n_batches, self.n_aabbs):
    qd.atomic_min(self.aabb_min[i_b], self.aabbs[i_b, i_a].min)
    qd.atomic_max(self.aabb_max[i_b], self.aabbs[i_b, i_a].max)
```

Every one of the `n_aabbs` lanes of a batch contends on the same two `vec3` cells. For a batched scene with articulated collision geometry that turns a trivially memory-bound min/max into the single most expensive kernel of the per-step BVH rebuild.

**Updated per @duburcqa's review** (the first revision hand-rolled a two-stage reduction; this one uses the built-in primitives). The fold now goes through `qd.simt.block.reduce_min` / `qd.simt.block.reduce_max`: every lane folds a strided slice in registers, the block reduction collapses the block, and only thread 0 issues an atomic. A batch then contends over as many atomics as it has blocks rather than as many as it has AABBs. Consecutive lanes read consecutive AABBs, so the strided walk stays coalesced.

The launch geometry is decided on the host at construction:

- **Block width** — the widest of `256 / 128 / 64` that a batch can fill. `block.reduce_*` requires a multiple of the subgroup size, which is 32 on CUDA / Metal / Vulkan-on-NVIDIA and 64 on AMDGPU, and Quadrants only accepts a non-power-of-two `block_dim` on CUDA / Vulkan / AMDGPU, so every candidate is a power of two and a multiple of 64 and no backend has to be queried on the host. Clamping to what a batch can actually fill keeps a small tree from launching blocks whose lanes have nothing to fold.
- **Blocks per batch** — `max(1, n_aabbs // (block_width * 32))`. The block reduction costs the same per block whatever its slice is worth (a shuffle tree, a shared-memory staging round and a barrier), so splitting a batch across more blocks than that stops paying off and starts costing. That threshold is measured, not guessed — see the tuning note at the end.
- A tree too small for even one block (`n_aabbs < 64`), and the CPU backend (no subgroup shuffles), keep the flat reduction. Contention is bounded by `n_aabbs` there anyway.

The output is **bit-identical**: min/max is exact, associative and commutative, so the fold order cannot change a single bit. Every lane is seeded with an actual AABB of its batch, so no lane is left uninitialized and no monoid identity is needed.

## Related Issue

No open issue. Direct follow-up to #3078 ("Speed up raycast sensors by splitting the rigid collision BVH into static and dynamic subsets"): that split correctly keeps static geometry on one shared tree, but the dynamic subset still rebuilds one full LBVH per env per step, and this reduction was the dominant kernel inside that rebuild.

## Motivation and Context

Profiling a raycast `DepthCamera` in a batched scene with two fixed-base articulated robots, almost none of the per-step cost was actually casting rays — it was the BVH rebuild, and within it this one kernel. The cost scales with `n_batches * n_aabbs` and is nearly independent of ray count, which is why it shows up as "raycast sensors get slower with more envs" rather than "with higher resolution".

### Results

One RTX 3080, CUDA backend, quadrants 1.3.0, based on `70a0f413`. Every figure is a median of medians over 3 independent runs.

**The kernel itself**, `compute_aabb_centers_and_scales`, median of 60 launches over 5746 AABBs per batch:

| n_batches | before | after | speedup |
|---|---|---|---|
| 1 | 0.287 ms | 0.026 ms | 11.1x |
| 64 | 7.757 ms | 0.069 ms | 112x |
| 256 | 26.207 ms | 0.204 ms | 129x |
| 1024 | 91.396 ms | 0.667 ms | 137x |
| 2048 | 182.622 ms | 1.262 ms | 145x |

**End to end**, `scene.step()` + `sensor.read()`, median over 25 steps. Scene: plane + fixed-base Go2 + Franka Panda + 8 static cylinders; raycast `DepthCamera` at 64x64. The movable tree is `n_envs x 5746`, the shared static tree `1 x 1692`:

| n_envs | before | after | speedup |
|---|---|---|---|
| 1 | 3.476 ms | 2.585 ms | 1.34x |
| 256 | 39.867 ms | 12.014 ms | **3.32x** |
| 1024 | 132.683 ms | 47.834 ms | **2.77x** |
| 2048 | 286.556 ms | 105.358 ms | **2.72x** |

The two tables agree on where the time goes: the end-to-end saving tracks the kernel saving to within run-to-run spread (27.9 ms against 26.0 ms at 256 envs, 84.8 ms against 90.7 ms at 1024, 181.2 ms against 181.4 ms at 2048). So essentially the whole end-to-end gain is this one reduction, and after the change it is no longer a significant part of the step. `n_envs=1` is dominated by fixed per-step cost and is correspondingly noisy (2.4-3.2 ms after, 3.1-3.6 ms before); treat that row as "unchanged to somewhat better".

Environment: Linux x86-64, single RTX 3080 (10 GB), CUDA backend, Python 3.12.8, quadrants 1.3.0, torch 2.13.0+cu130, based on `70a0f413`.

- **Test**: `tests/core/test_bvh.py::test_scene_bounds` asserts the per-batch `aabb_min` / `aabb_max` that `build()` produces against a host-side reduction of the input AABBs, exactly (`atol=rtol=0`) — the reduction selects an input value, so no rounding is involved and no tolerance is warranted. It runs off the existing `lbvh` fixture, and the parametrized sizes cover the flat path (1), the narrowest cooperative block (64), a block width that does not divide the AABB count (255), the widest block (1776), and several blocks per batch (16384). No existing test covers those two fields: `test_morton_code` only checks that the codes come out sorted, and `test_build_tree` checks node bounds against their children, which are computed by a separate kernel from separate storage.
  Zeroing out the block reduction (`block_min[i] = lane_min[i]`) makes every cooperative size fail and leaves the flat one passing, so the assertion does bite on the path this PR adds.
- `tests/core/test_bvh.py` — **26 passed** with `--backend gpu`, **26 passed** with `--backend cpu`.
- `tests/sensors/test_raycaster.py` and `tests/coupling/test_hybrid.py` (the SAP coupler is the other `LBVH` consumer) — **28 passed**.

n/a

## Checklist:

---

### Tuning note: why blocks per batch is capped

Same setup, kernel only, 5746 AABBs per batch, varying how many blocks each batch is split across (block width 256):

| blocks / batch | 1 | 2 | 5 | 11 | 22 |
|---|---|---|---|---|---|
| n_batches=1 | **0.026 ms** | 0.026 ms | 0.030 ms | 0.033 ms | 0.042 ms |
| n_batches=64 | **0.069 ms** | 0.087 ms | 0.153 ms | 0.287 ms | 0.595 ms |
| n_batches=256 | **0.204 ms** | 0.254 ms | 0.477 ms | 1.020 ms | 2.009 ms |

More blocks is monotonically worse once each block's slice gets thin, which is what the `n_aabbs // (block_dim * 32)` cap encodes. Block width 128 against 256 is within ~3% across these shapes, so 256 is kept to match the width Quadrants' own device-wide reduce uses.

### Notes for reviewers / possible follow-ups

Two further items in the same rebuild path, deliberately **not** included here to keep this PR to one reviewable change:

1. **`compute_bounds` does a host-side `while not is_done` loop**, launching one kernel per tree layer and reading its return value back to the CPU each time (~21 launches + 21 syncs per step, independent of batch size). It is most of the fixed cost raycasting shows at `n_envs=1`. A single-launch Karras bottom-up refit with an atomic per-node counter fixes it, but it needs explicit `qd.simt.grid.mem_fence()` release/acquire pairs around the counter — without them it is a data race that silently undersizes internal bounds (I measured up to 1.5 m of wrong depth). It also stops paying off above a few hundred envs, so it wants gating on `n_batches`. Worth its own PR.
2. **`n_radix_sort_groups` is hardcoded to 64** at three `LBVH(...)` call sites in `genesis/engine/sensors/raycaster.py` (and to `min(64, n_slots)` in `genesis/vis/viewer_plugins/raycast.py`), independent of `n_batches`, while the `LBVH` default is 256. Sizing it so `n_batches * n_groups` saturates the device looks like a further easy win.

Structurally, the remaining rebuild cost is still spent re-deriving a tree whose topology barely changes: faces on a rigid link move rigidly with that link. A refit (or a two-level static-per-link BLAS + small TLAS) would make the dynamic subset's rebuild scale with link count rather than face count x env count. Happy to discuss if that is of interest.

### DIFF
--- a/genesis/engine/bvh.py
+++ b/genesis/engine/bvh.py
@@ -8,10 +8,9 @@ from genesis.repr_base import RBC
 # https://forums.developer.nvidia.com/t/thinking-parallel-part-ii-tree-traversal-on-the-gpu/148342
 STACK_SIZE = 64
 
-# Launch geometry of the scene-extent reduction in `LBVH.compute_aabb_centers_and_scales`, widest first. Each width
-# is a multiple of 64, hence of every supported subgroup size (32 on CUDA / Metal / Vulkan-on-NVIDIA, 64 on AMDGPU)
-# as `block.reduce_*` requires, and a power of two, which Quadrants demands of `block_dim` outside CUDA / Vulkan /
-# AMDGPU. A block reduction costs the same whatever its slice is worth, hence the floor on lane workload.
+# Launch geometry of the scene-extent reduction in `LBVH.compute_aabb_centers_and_scales`, widest first. Every width
+# is a multiple of 64, hence of any subgroup size `block.reduce_*` can be given (see `qd.simt.subgroup.group_size`).
+# A block reduction costs the same whatever its slice is worth, hence the floor on the workload of a single lane.
 EXTENT_REDUCE_BLOCK_DIMS = (256, 128, 64)
 EXTENT_REDUCE_MIN_AABBS_PER_LANE = 32
 
@@ -89,47 +88,6 @@ class LBVH(RBC):
     max_stack_depth : int
         Maximum stack depth for BVH traversal. Defaults to STACK_SIZE.
 
-    Attributes
-    -----
-    aabbs : qd.field
-        The input AABBs to be organized in the BVH, shape (n_batches, n_aabbs).
-    n_aabbs : int
-        Number of AABBs per batch.
-    n_batches : int
-        Number of batches.
-    max_query_results : int
-        Maximum number of query results allowed.
-    max_stack_depth : int
-        Maximum stack depth for BVH traversal.
-    aabb_centers : qd.field
-        Centers of the AABBs, shape (n_batches, n_aabbs).
-    aabb_min : qd.field
-        Minimum coordinates of AABB centers per batch, shape (n_batches).
-    aabb_max : qd.field
-        Maximum coordinates of AABB centers per batch, shape (n_batches).
-    scale : qd.field
-        Scaling factors for normalizing AABB centers, shape (n_batches).
-    morton_codes : qd.field
-        Morton codes for each AABB, shape (n_batches, n_aabbs).
-    hist : qd.field
-        Histogram for radix sort, shape (n_batches, 256).
-    prefix_sum : qd.field
-        Prefix sum for histogram, shape (n_batches, 256).
-    offset : qd.field
-        Offset for radix sort, shape (n_batches, n_aabbs).
-    tmp_morton_codes : qd.field
-        Temporary storage for radix sort, shape (n_batches, n_aabbs).
-    Node : qd.dataclass
-        Node structure for the BVH tree, containing left, right, parent indices and bounding box.
-    nodes : qd.field
-        BVH nodes, shape (n_batches, n_aabbs * 2 - 1).
-    internal_node_visited : qd.field
-        Flags indicating if an internal node has been visited during traversal, shape (n_batches, n_aabbs - 1).
-    query_result : qd.field
-        Query results as a vector of (batch id, self id, query id), shape (max_query_results).
-    query_result_count : qd.field
-        Counter for the number of query results.
-
     Notes
     ------
         For algorithmic details, see:
@@ -158,15 +116,15 @@ class LBVH(RBC):
         self.aabb_min = qd.field(gs.qd_vec3, shape=(self.n_batches,))
         self.aabb_max = qd.field(gs.qd_vec3, shape=(self.n_batches,))
         self.scale = qd.field(gs.qd_vec3, shape=(self.n_batches,))
-        # Block-cooperative scene-extent reduction: GPU-only (it goes through subgroup shuffles) and only worth it
-        # for a tree wide enough to fill a block. Anything smaller keeps the flat atomic reduction below, which is
-        # barely contended at that size anyway.
-        self.extent_reduce_block_dim = next((bd for bd in EXTENT_REDUCE_BLOCK_DIMS if bd <= self.n_aabbs), 0)
-        self.extent_reduce_coop = gs.backend != gs.cpu and self.extent_reduce_block_dim > 0
-        self.n_extent_reduce_threads = 0
-        if self.extent_reduce_coop:
-            n_blocks = max(1, self.n_aabbs // (self.extent_reduce_block_dim * EXTENT_REDUCE_MIN_AABBS_PER_LANE))
-            self.n_extent_reduce_threads = n_blocks * self.extent_reduce_block_dim
+        # Block-cooperative scene-extent reduction: it goes through subgroup shuffles, so it is GPU-only, and it
+        # only pays off for a tree wide enough to fill the narrowest block. Anything smaller keeps the flat atomic
+        # reduction below, whose contention is bounded by the AABB count at that size anyway.
+        narrowest_block_dim = EXTENT_REDUCE_BLOCK_DIMS[-1]
+        self._use_cooperative_extent_reduce = gs.backend != gs.cpu and self.n_aabbs >= narrowest_block_dim
+        block_dim = next((width for width in EXTENT_REDUCE_BLOCK_DIMS if width <= self.n_aabbs), narrowest_block_dim)
+        n_blocks = max(1, self.n_aabbs // (block_dim * EXTENT_REDUCE_MIN_AABBS_PER_LANE))
+        self._extent_reduce_block_dim = block_dim
+        self._n_extent_reduce_threads = n_blocks * block_dim
         self.morton_codes = qd.field(qd.types.vector(2, qd.u32), shape=(self.n_batches, self.n_aabbs))
 
         # Histogram for radix sort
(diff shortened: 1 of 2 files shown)


genesis/engine/bvh.py
duburcqa, line 1: Do not do this. We have properly implemented reduction algorithm built-in in quadrants. Use them.

jeetrex17:
Hi @Kashu7100, I checked this against current main. The optimization is still relevant: LBVH.compute_aabb_centers_and_scales() still performs the flat per-AABB atomic_min / atomic_max reduction, and movable raycast BVHs still rebuild per environment.

I also saw Alexis’s feedback that this should use Quadrants’ built-in reduction primitives rather than the handwritten two-stage reduction. Genesis now uses Quadrants 1.3.0, so the implementation would need to be refreshed and revalidated against the current API.

Are you planning to continue this PR? If not, I’d be happy to prepare a superseding draft on current main, crediting this PR and preserving its intent, while using the Quadrants reduction primitives and adding focused correctness coverage and updated benchmarks.

Kashu7100:
Thanks for the heads up. I can finish this PR.

genesis/engine/bvh.py
Kashu7100: Done — rebased on current `main` (quadrants 1.3.0) and rewritten to use the built-in primitives.

tests/core/test_bvh.py
```diff
@@ -68,6 +68,47 @@ def expand_bits(lbvh: qd.template(), x: qd.template(), expanded_x: qd.template()
...
+def test_scene_extent_reduction(n_aabbs, n_batches):
+    """The block-cooperative scene-extent reduction must agree with the flat atomic one, bit for bit.
+
+    The sizes cover every branch of its launch geometry: no cooperative path at all, a block narrower than the
+    maximum, a block width that does not divide the AABB count, and several blocks per batch.
+    """
+    aabbs_min = (np.random.rand(n_batches, n_aabbs, 3).astype(gs.np_float) - 0.5) * 20.0
+    aabbs_max = aabbs_min + np.random.rand(n_batches, n_aabbs, 3).astype(gs.np_float)
+
+    def scene_extent(cooperative):
```
duburcqa: Remove this function and use a for-loop to validate cooperative=True/False.

tests/core/test_bvh.py
```diff
@@ -68,6 +68,47 @@ def expand_bits(lbvh: qd.template(), x: qd.template(), expanded_x: qd.template()
...
+        lbvh = LBVH(aabb)
+        # Quadrants only accepts a non-power-of-two `block_dim` on CUDA / Vulkan / AMDGPU, and `block.reduce_*` needs
+        # a multiple of the subgroup size (64 on AMDGPU). Assert it here so a bad width fails on every backend.
+        block_dim = lbvh.extent_reduce_block_dim
+        if lbvh.extent_reduce_coop:
+            assert block_dim & (block_dim - 1) == 0 and block_dim % 64 == 0 and block_dim <= n_aabbs
+        # The kernel branches on this at compile time, so it has to be set before the first call.
+        lbvh.extent_reduce_coop = lbvh.extent_reduce_coop and cooperative
+        lbvh.compute_aabb_centers_and_scales()
+        return lbvh.aabb_min.to_numpy(), lbvh.aabb_max.to_numpy(), lbvh.scale.to_numpy()
```
duburcqa: to_numpy / to_torch is prohibited in unit tests.

tests/core/test_bvh.py
```diff
@@ -68,6 +68,47 @@ def expand_bits(lbvh: qd.template(), x: qd.template(), expanded_x: qd.template()
...
+    maximum, a block width that does not divide the AABB count, and several blocks per batch.
+    """
+    aabbs_min = (np.random.rand(n_batches, n_aabbs, 3).astype(gs.np_float) - 0.5) * 20.0
+    aabbs_max = aabbs_min + np.random.rand(n_batches, n_aabbs, 3).astype(gs.np_float)
+
+    def scene_extent(cooperative):
+        aabb = AABB(n_batches=n_batches, n_aabbs=n_aabbs)
+        aabb.aabbs.min.from_numpy(aabbs_min)
+        aabb.aabbs.max.from_numpy(aabbs_max)
+        lbvh = LBVH(aabb)
```
duburcqa: Use our 'lbvh' fixture if possible.

tests/core/test_bvh.py
```diff
@@ -68,6 +68,47 @@ def expand_bits(lbvh: qd.template(), x: qd.template(), expanded_x: qd.template()
...
+        # Quadrants only accepts a non-power-of-two `block_dim` on CUDA / Vulkan / AMDGPU, and `block.reduce_*` needs
+        # a multiple of the subgroup size (64 on AMDGPU). Assert it here so a bad width fails on every backend.
+        block_dim = lbvh.extent_reduce_block_dim
+        if lbvh.extent_reduce_coop:
+            assert block_dim & (block_dim - 1) == 0 and block_dim % 64 == 0 and block_dim <= n_aabbs
```
duburcqa: Remove this check, it makes no sense.

tests/core/test_bvh.py
```diff
@@ -68,6 +68,47 @@ def expand_bits(lbvh: qd.template(), x: qd.template(), expanded_x: qd.template()
         )
 
 
+@pytest.mark.required
+@pytest.mark.parametrize(
+    "n_aabbs, n_batches",
+    [(1, 1), (63, 4), (64, 4), (65, 3), (255, 2), (256, 8), (1776, 64), (12345, 2)],
+)
+@pytest.mark.parametrize("backend", [gs.cpu, gs.gpu])
+def test_scene_extent_reduction(n_aabbs, n_batches):
```
duburcqa: Make sure this unit test is not redundant with any of the existing unit tests.

tests/core/test_bvh.py
```diff
@@ -68,6 +68,47 @@ def expand_bits(lbvh: qd.template(), x: qd.template(), expanded_x: qd.template()
         )
 
 
+@pytest.mark.required
+@pytest.mark.parametrize(
+    "n_aabbs, n_batches",
+    [(1, 1), (63, 4), (64, 4), (65, 3), (255, 2), (256, 8), (1776, 64), (12345, 2)],
```
duburcqa: I doubt checking 8 configurations is really necessary. 5-6 should be more than enough.

tests/core/test_bvh.py
```diff
@@ -68,6 +68,47 @@ def expand_bits(lbvh: qd.template(), x: qd.template(), expanded_x: qd.template()
...
+)
+@pytest.mark.parametrize("backend", [gs.cpu, gs.gpu])
+def test_scene_extent_reduction(n_aabbs, n_batches):
+    """The block-cooperative scene-extent reduction must agree with the flat atomic one, bit for bit.
+
+    The sizes cover every branch of its launch geometry: no cooperative path at all, a block narrower than the
+    maximum, a block width that does not divide the AABB count, and several blocks per batch.
+    """
```
duburcqa: Docstring in unit tests is violating the guidelines. Please read them (CLAUDE.md for AI agents, CODING_GUIDELINES.md for humans)

tests/core/test_bvh.py
duburcqa: I don't like the name of this test. It is not clear which feature is being validated here.

PR #3165, opened 2026-08-03
[MISC] Support negative indices for env and entity index args.

### LINKED ISSUE #3161: [Bug]: Negative envs_idx passed as a list or array is not normalized and reaches the solver unchecked
### Bug Description

Since #3117, `Scene._sanitize_envs_idx` bounds-checks a scalar `envs_idx` and wraps a negative value to `n_envs - 1`. The same selection expressed as a `list`, `tuple`, `range`, numpy array or torch tensor takes a different path and is neither wrapped nor validated.

Collections fall through to `sanitize_index` (`genesis/utils/misc.py:849`), which validates only the shape of the index. Its range check is disabled at `genesis/utils/misc.py:884-887`:

```python
# FIXME: This check is too expensive
# if not (0 <= dim_idx & dim_idx < size).all():
#     dim_info = f" `{name}`" if name else ""
#     gs.raise_exception(f"Indices out-of-range for {i}-th index{dim_info}.")
```

so the value is returned verbatim and used to address a buffer inside a compiled kernel. For example
`_kernel_set_gravity` (`genesis/engine/solvers/base_solver.py:376-378`) executes

```python
    for j in qd.static(range(3)):
        gravity[envs_idx[i_b_]][j] = tensor[i_b_, j]
```

against `_gravity`, allocated as `array_class.V(gs.qd_vec3, (self._B,))`. Quadrants kernels do not implement Python negative-index semantics, so `-1` addresses memory before element 0 rather than selecting the last environment.

The observed consequence differs by backend, and neither is a clean error:

- **CPU backend: the process dies with SIGBUS** (exit code 138) at the `set_gravity` call.
- **Metal backend: the write silently lands out of bounds.** No exception is raised and the target environment is left unchanged. An out-of-range positive index (`[3]` with `n_envs=3`) behaves the same way.

`KinematicEntity._get_global_idx` (`genesis/engine/entities/rigid_entity/rigid_entity.py:1691-1697`) has the same asymmetry for entity-local indices — the scalar branch wraps, the `list`/`tuple` branch does not — with the same disabled range check at lines 1724-1726. There the failure is quieter still: for the second entity in a scene, `dofs_idx_local=[-1]` resolves to `dof_start - 1`, an **in-bounds slot belonging to the previous entity**, so the operation silently targets the wrong entity with no error on any backend.

`_sanitize_envs_idx` is shared by 71 call sites across the rigid, PBD, MPM, SPH and FEM entities, the rigid and PBD solvers, the constraint solver, the collider, four sensor modules and `vis/camera.py`, so this is reachable from ordinary public API and not only from `scene.sim`.

Read paths are unaffected: getters apply `envs_idx` through torch indexing (`_apply_masks`, `genesis/utils/misc.py:662`) rather than a kernel, so a getter can report a plausible value after a setter has already written out of bounds.

### Steps to Reproduce

```python
import sys

import genesis as gs

BACKEND = gs.gpu if len(sys.argv) > 1 and sys.argv[1] == "gpu" else gs.cpu

print(f"genesis package: {gs.__file__}")
gs.init(backend=BACKEND, logging_level="warning")
print(f"backend={gs.backend}  use_zerocopy={gs.use_zerocopy}  use_ndarray={gs.use_ndarray}")

scene.build(n_envs=3)
rigid = scene.sim.rigid_solver

print("\n--- what the sanitizer returns for n_envs=3 ---")
for label, value in (("scalar -1", -1), ("list [-1]", [-1])):
    try:
        print(f"  {label:12s} -> {scene._sanitize_envs_idx(value).tolist()}")
    except Exception as exc:
        print(f"  {label:12s} -> {type(exc).__name__}: {str(exc).splitlines()[0][:80]}")

print("\n--- observable effect of set_gravity ---")
scene.sim.set_gravity((0.0, 0.0, 0.0))
print(f"  baseline (all envs)      : {rigid.get_gravity().tolist()}")

scene.sim.set_gravity((0.0, 0.0, 9.8), envs_idx=-1)
print(f"  after envs_idx=-1        : {rigid.get_gravity().tolist()}")

scene.sim.set_gravity((0.0, 0.0, 0.0))
try:
    scene.sim.set_gravity((0.0, 0.0, 9.8), envs_idx=[-1])
    print(f"  after envs_idx=[-1]      : {rigid.get_gravity().tolist()}")
except Exception as exc:
    print(f"  after envs_idx=[-1]      : {type(exc).__name__}: {str(exc).splitlines()[0][:80]}")

print("\n--- out-of-range, upper side ---")
scene.sim.set_gravity((0.0, 0.0, 0.0))
for label, value in (("scalar 3", 3), ("list [3]", [3])):
    try:
        scene.sim.set_gravity((0.0, 0.0, 9.8), envs_idx=value)
        print(f"  {label:12s} -> no error, gravity={rigid.get_gravity().tolist()}")
    except Exception as exc:
        print(f"  {label:12s} -> {type(exc).__name__}: {str(exc).splitlines()[0][:80]}")

print("\nDONE")
```

Run once per backend:

```bash
python -u repro.py cpu
python -u repro.py gpu
```

The entity-local variant, which needs two dynamic entities so the second one has a nonzero start offset:

```python
import genesis as gs

gs.init(backend=gs.gpu, logging_level="warning")

first = scene.add_entity(gs.morphs.Sphere(pos=(0.0, 0.0, 1.0)))
second = scene.add_entity(gs.morphs.Sphere(pos=(2.0, 0.0, 1.0)))
scene.build(n_envs=2)

print(f"first  entity: dof_start={first.dof_start} dof_end={first.dof_end}")
print(f"second entity: dof_start={second.dof_start} dof_end={second.dof_end}")
for label, value in (("scalar -1", -1), ("list [-1]", [-1])):
    got = second._get_global_idx(value, second.n_dofs, second.dof_start).tolist()
    owner = "second entity" if got[0] >= second.dof_start else "FIRST entity"
    print(f"  {label:12s} -> global dof {got}  belongs to: {owner}")
```

### Expected Behavior

`envs_idx=[-1]` selects the last environment, matching the scalar `envs_idx=-1` behavior established by #3117, and out-of-range values raise a clear exception instead of reaching a kernel. The same holds for `idx_local` collections in the entity accessors, where a local `-1` must resolve within the entity that owns it.

### Screenshots/Videos

_No response_

### Relevant log output

```shell
CPU backend - the sanitizer returns the raw `-1`, and the process dies with SIGBUS (exit 138) at the set_gravity call, before it can print anything further:

genesis package: .../pristine/genesis/__init__.py
backend=gs.cpu  use_zerocopy=True  use_ndarray=True

--- what the sanitizer returns for n_envs=3 ---
  scalar -1    -> [2]
  list [-1]    -> [-1]

--- observable effect of set_gravity ---
  baseline (all envs)      : [[0.0, 0.0, 0.0], [0.0, 0.0, 0.0], [0.0, 0.0, 0.0]]
  after envs_idx=-1        : [[0.0, 0.0, 0.0], [0.0, 0.0, 0.0], [0.0, 0.0, 9.800000190734863]]

[process terminated by SIGBUS, exit code 138]


Metal backend — no error is raised; the write silently misses, leaving the last environment unchanged, and an
out-of-range positive index behaves the same way:

genesis package: .../pristine/genesis/__init__.py
backend=gs.metal  use_zerocopy=True  use_ndarray=True

--- what the sanitizer returns for n_envs=3 ---
  scalar -1    -> [2]
  list [-1]    -> [-1]

--- observable effect of set_gravity ---
  baseline (all envs)      : [[0.0, 0.0, 0.0], [0.0, 0.0, 0.0], [0.0, 0.0, 0.0]]
  after envs_idx=-1        : [[0.0, 0.0, 0.0], [0.0, 0.0, 0.0], [0.0, 0.0, 9.800000190734863]]
  after envs_idx=[-1]      : [[0.0, 0.0, 0.0], [0.0, 0.0, 0.0], [0.0, 0.0, 0.0]]

--- out-of-range, upper side ---
  scalar 3     -> GenesisException: `envs_idx` out of range: 3 not in [-3, 3).
  list [3]     -> no error, gravity=[[0.0, 0.0, 0.0], [0.0, 0.0, 0.0], [0.0, 0.0, 0.0]]


Entity-local index, Metal backend — the local `-1` on the second entity silently resolves into the first entity's DOF
range:


first  entity: dof_start=0 dof_end=6
second entity: dof_start=6 dof_end=12

--- second entity, local index -1 (its own last dof is global 11) ---
  scalar -1    -> global dof [11]  belongs to: second entity
  list [-1]    -> global dof [5]  belongs to: FIRST entity

DONE
```

### Environment

- OS: macOS 15.7.3 (build 24G419)
- GPU/CPU: Apple M4 (Mac16,1)
- GPU-driver version: N/A (integrated Apple GPU, Metal backend)
- CUDA / CUDA-toolkit version: N/A (non-Nvidia)
- Python: 3.13.11
- PyTorch: 2.13.0 (MPS available)
- Quadrants: 1.2.0 (llvm 22.1.0, commit 3d9af718, osx)
- Backends reproduced on: gs.cpu and gs.metal

Release version: 1.3.1
Commit: fc3df59a20b632cf2a0b9a858bd3bdae5964d193

### Additional Context

#3117 fixed the scalar case and its description states the remaining scope explicitly:

> The change is limited to scalar integer indices. The `slice`, `range`, and collection paths are unchanged.

This report covers those remaining paths. It is a separate failure mode from #3116, which described a silent no-op caused by the empty slice `[-1:0]`; here the index reaches a kernel unvalidated.

Related, for context on how device-side asserts have been handled in this repository: #2086.

jeetrex17:
## Description

Valid negative collection indices are now normalized before they are used by Quadrants kernels or offset into entity-global index space.

- Normalize negative list, tuple, NumPy array, Torch tensor, and range inputs using Python indexing semantics.
- Normalize slices with `slice.indices()`.
- Preserve boolean-mask handling.
- Skip the collection wrap for inputs that are non-negative by construction, so accessors passing `None` or a
  normalized slice do not pay for an extra tensor operation.
- Exercise environment and entity-local collection indices in the existing regression tests.

Out-of-range values are unchanged: the bounds check in `sanitize_index` stays disabled, as documented by its FIXME.

## Related Issue

Resolves #3161

## Motivation and Context

A scalar index such as `envs_idx=-1` already selects the last environment, but the equivalent collection form `envs_idx=[-1]` previously reached a Quadrants kernel unchanged. Depending on the backend, that could terminate the process or silently miss the requested environment.

Entity-local collection indices had the same inconsistency. A local `[-1]` was offset before normalization and could select data belonging to the previous entity. Normalizing valid negative indices before applying offsets or entering kernels makes the supported input forms consistent.

Tested on macOS with an Apple M4 and Python 3.11.14.

```bash
pytest tests/rigid/test_dynamics.py::test_gravity tests/rigid/test_kinematics.py::test_setters -n 0 --backend cpu -q
pytest tests/rigid/test_dynamics.py::test_gravity tests/rigid/test_kinematics.py::test_setters -n 0 --backend metal -q
pytest tests/core/test_misc.py -n 0 -q
```

Results:

- CPU regressions: 2 passed
- Metal regressions: 2 passed
- Utility tests: 18 passed

The targeted tests above passed. The complete test suite was not run locally and remains covered by CI.

Not applicable.

## Checklist:


duburcqa:
I'm concerned by the performance impact of all these additional branches.

jeetrex17:
Good point. I measured the current implementation and found avoidable overhead.

These are isolated CPU microbenchmarks on an Apple M4 with Python 3.11.14. Each result is the minimum of five runs with 100,000 calls per run:

| Path | `main` | Current PR | Delta |
|---|---:|---:|---:|
| `_get_global_idx(None)` | 76 ns | 112 ns | +36 ns |
| `_get_global_idx(scalar)` | 115 ns | 133 ns | +18 ns |
| `sanitize_index([0, 1])` | 2.35 us | 6.91 us | +4.55 us |
| `_get_global_idx(torch.tensor([0, 1]))` | 1.35 us | 5.67 us | +4.33 us |

The additional type checks affect the common entity-local path, while the unconditional tensor normalization adds most of the collection overhead. These timings measure the helper functions in isolation, not complete simulation steps.

I will fold the slice and range handling into the existing dispatch and narrow the normalization by input type , and will rerun the same benchmarks and update the PR with the revised results.

jeetrex17:
Done in 7d7d06a1.

I folded slice and range handling into the existing dispatch and normalize lists, tuples, ranges, and NumPy arrays before creating a tensor. Numeric Torch tensor inputs remain normalized on-device to avoid synchronizing accelerator work.

I re-measured all three revisions in one session on an Apple M4 with Python 3.11.14. Each result is the minimum of five runs with 100,000 calls per run:

| Path | `main` | bd918889 | 7d7d06a1 |
|---|---:|---:|---:|
| `_get_global_idx(None)` | 75 ns | 111 ns | 83 ns |
| `_get_global_idx(scalar)` | 114 ns | 132 ns | 91 ns |
| `sanitize_index([0, 1])` | 2.28 us | 6.60 us | 2.48 us |
| `_get_global_idx(torch.tensor([0, 1]))` | 1.28 us | 5.51 us | 5.45 us |

The common host-side paths are now close to `main`. The scalar path is faster because scalar inputs are handled before the slice and range checks.

A tensor index still costs about 4.5 us. Normalizing it on the host would mean reading its values back and synchronizing, which is worse on GPU, so I kept that one on-device.

duburcqa:
> A tensor index still costs about 4.5 us. Normalizing it on the host would mean reading its values back and synchronizing, which is worse on GPU, so I kept that one on-device.

actually no it is not. Torch will anyway transfer the tensor back to host to be able to use it for stuffs like advanced indexing.

duburcqa:
We have a dedicated benchmark for data accessors, could you run it, and eventually fold your use-case into the existing benchmark?

duburcqa:
@jeetrex17 What is the current status performance-wise?

duburcqa (approved):

duburcqa:
Note that your PR weakly relates to this one: https://github.com/Genesis-Embodied-AI/genesis-world/pull/3157

jeetrex17:
I ran the CPU `franka_accessors` benchmark three times per side on an Apple M4, comparing the current PR head with its base commit (7fb46148) and alternating sides in the same session:

| Run | Base | This PR |
|---|---:|---:|
| 1 | 1,913 FPS | 1,884 FPS |
| 2 | 1,827 FPS | 1,838 FPS |
| 3 | 1,866 FPS | 1,791 FPS |

The ranges overlap, with a median of 1,866 versus 1,838 FPS, a difference of about 1.5%. I do not see a clear regression in this local CPU sample. The PR side exercises the negative local-index path because `get_dofs_position` now uses `range(-franka.n_dofs, 0)`.

I also folded negative tensor environment indices into the 30,000-environment GPU configuration through `get_links_acc(envs_idx=negative_envs_idx)`. That path is not represented in the CPU table, and the current-head workflows are still awaiting approval, so I do not have its final GPU result yet.

Host collections are normalized before tensor creation. Numeric tensor inputs to `sanitize_index` use one `torch.where`, while the `None` path avoids the new negative-wrap operation.

Thanks for pointing out #3157. I see the relation , that PR avoids host synchronization during accessor extraction, while this PR normalizes negative index collections before that stage.

duburcqa (dismissed):

duburcqa:
- **`sanitize_index` now raises a raw `TypeError` instead of a clear Genesis error** (`genesis/utils/misc.py:874`). The new fallback comprehension runs *before* the shape validation that produced the message, so malformed input surfaces as an internal Python error. Measured against the base: `envs_idx=[[0, 1]]` gave `GenesisException: Invalid shape: torch.Size([1, 2]). Expecting 0D or 1D tensor...` and now gives `TypeError: '<' not supported between instances of 'list' and 'int'`; `envs_idx=["a"]` likewise. `_get_global_idx` already guards its equivalent comprehension with `try/except TypeError` and re-raises through `gs.raise_exception`; `sanitize_index` needs the same, since it is the public sanitizer.

- **`envs_idx={0, 1}` is now silently accepted** (same line). The base raised `TypeError: 'set' object cannot be interpreted as an integer`. Set iteration order would decide the environment order, so this should keep raising.

- **`.dtype` is read off an unvalidated object** (`rigid_entity.py:1715-1718`): `get_dofs_position({0, 1})` gives `AttributeError: 'set' object has no attribute 'dtype'`, where the base gave a `TypeError`. Same fix as above - validate before dereferencing.

- **The wrap runs unconditionally, even when every index is positive** (`rigid_entity.py:1719-1721`). `is_negative_wrap_required` is true for any signed dtype, so `idx_local + idx_local_max * (idx_local < 0)` executes on every call. 53 of the 57 `_get_global_idx` call sites pass `unsafe=True`, which covers all the accessors and setters. Measured on a 9-element int32 tensor, single process, interleaved: 0.005 us for the previous passthrough versus 2.709 us for `t + n*(t<0)`. On GPU that is two extra kernel launches per accessor call. The "Reduce negative-index normalization overhead" commit skipped only `None` / `slice` / `range`, which are the forms that never cost anything, and left the tensor form - the one users pass in hot loops - always paying.

- **That same expression promotes `gs.tc_int` to int64.** `n * (t < 0)` yields int64, so `int32 + int64` gives int64, measured for both torch and numpy inputs including the all-positive case. Under `unsafe=True` that int64 tensor *is* the return value, and the downstream re-sanitize then pays a real conversion copy (0.531 us versus 0.171 us for the int32 no-op). `torch.where(index < 0, index + max_size, index)`, which this PR already uses at `misc.py:894`, preserves int32 - the two halves of the fix disagree.

- **One operation is spelled four different ways**: `i + max_size if i < 0 else i` (misc list/range), `index + max_size * (index < 0)` (misc ndarray), `torch.where(index < 0, index + max_size, index)` (misc tensor), `idx_local + idx_local_max * (idx_local < 0)` (rigid_entity) - plus two names for the same predicate, `is_tensor_wrap_required` and `is_negative_wrap_required`. A single vectorized wrap applied once after `torch.as_tensor` collapses all four, removes the flag, removes the second site 30 lines further down, and removes the Python comprehensions.

- **A Python loop was added to the list/tuple path** (`misc.py:874`): +0.15 us at 9 elements, scaling linearly, so a several-thousand-element `envs_idx` list pays a full Python loop. The vectorized form is already two branches above.

- **All-negative ranges are expanded needlessly** (`rigid_entity.py:1702-1704`, `misc.py:871-872`). Wrapping preserves a constant step whenever the whole range is negative (`[i + n for i in range(-n, 0)] == range(0, n)`); only mixed-sign ranges break it, which is what the comment says but not what the code does. Measured 1.140 us via list expansion versus 1.092 us via the shifted range.

- **Nothing checks that a negative index selects the *right* data, and most new codepaths are unvalidated.** Only the `list` form is asserted anywhere; tuple, `np.ndarray` (signed, unsigned, float), `torch.Tensor`, `range` (all-negative and mixed-sign), and `slice` with negative bounds or step are all unpinned. The two edited lines also replaced the scalar forms (`envs_idx=-1`, `dofs_idx_local=-1`) rather than adding to them, so those lost their coverage. Please assert that a negative index reads and writes the same data as its positive equivalent (`get_dofs_position([-2, -1]) == get_dofs_position([n_dofs - 2, n_dofs - 1])`, and the same through a setter) for each supported input form, plus that invalid input still raises a clear `GenesisException`. This is the property that actually matters for entity-local indices, where a `[-1]` used to be offset before normalization and could reach the previous entity's data: "it no longer crashes" is not the fix, "it addresses the intended element" is. Parametrizing the index form over the two tests already touched keeps this to a single scene build.

- **Out-of-range negative collection indices still kill the process.** Measured on both this branch and its base: `envs_idx=-4` raises `GenesisException: envs_idx out of range: -4 not in [-3, 3)`, while `envs_idx=[-4]` dies with SIGBUS. Not a regression, and the description does call it out, but the loop edited at `tests/rigid/test_dynamics.py:30` is exactly where the scalar out-of-range forms are swept, so leaving the collection forms out makes the file read as though they were covered. `scene._sanitize_envs_idx:1602-1605` already has the cheap bounds check on the scalar path.

- **Three comments need rework.** `misc.py:891-892` explains why normalization is needed at all, which a reader can deduce from the code, but never why it is *deferred*, which is the one thing they cannot. `rigid_entity.py:1720` restates the expression directly below it. `rigid_entity.py:1702` does not match what the code does (see the all-negative range point) and is phrased as a scenario description.

- **Boolean-mask handling is preserved for torch masks only.** A numpy bool mask returns `n_dofs` copies of dof 0 (`get_dofs_position(np.zeros(9, dtype=bool))` gives shape `(3, 9)`, versus `(3, 0)` for the torch mask). Identical before and after, so not introduced here, but the new `np.ndarray` branch sits exactly where it would be handled.

jeetrex17:
Updated in e68d8f0c.

I restored the accessor benchmark unchanged, as requested, and kept the negative-index coverage in the rigid and utility tests. I also added validation for malformed NumPy and Torch boolean masks before the unsafe entity-local path.

Verification on the current head:

  - CPU focused suite: 42 passed
  - Metal regressions: 2 passed
  - Ruff check and format: passed

tests/rigid/test_dynamics.py
```diff
@@ -23,8 +23,37 @@ def test_gravity(show_viewer, tol):
...
+        scene.sim.set_gravity(values, envs_idx=envs_idx)
+        actual_gravity = scene.rigid_solver.get_gravity()
+
+        scene.sim.set_gravity((0.0, 0.0, 0.0))
+        scene.sim.set_gravity(values, envs_idx=expected_envs_idx)
+        assert_equal(actual_gravity, scene.rigid_solver.get_gravity())
+
+    for envs_idx in ({0, 1}, [[0, 1]], ["a"]):
+        with pytest.raises(gs.GenesisException):
+            scene.sim.set_gravity((0.0, 0.0, 0.0), envs_idx=envs_idx)
```
duburcqa: This is redundant with `test_setters` no?

jeetrex17: Partially , trimmed in 71821404. Dropped the input forms that test_setters and the sanitize_index unit test already pin, and removed the invalid-input loop. Kept the cases exercising the scene-side slice and range handling plus the original envs_idx=[-1] regression, which only this path covers.

duburcqa:
@codex review

duburcqa:
@codex review

duburcqa:
Thank you for your effort @jeetrex17 ! It was not an easy one.

PR #3166, opened 2026-08-03
[FEATURE][BREAKING] Set contact friction for a pair of materials.

Milotrince:
## Description

- `scene.add_material(gs.materials.Rigid(...))` registers the options and returns the `Material` that entities hold
  as their `material`. Passing one to several `add_entity` calls shares one surface; `geom.set_material` gives one
  entity more than one.
- `rubber.set_friction_pair(steel, sliding_friction=0.4)` pins the coefficients for that pair of materials in place
  of the maximum over the two geoms. A coefficient left `None` keeps following the maximum. Re-declaring the same two
  materials retunes them, before or after build.
- Materials stay an authoring concept. Declarations expand onto collision pairs at build, so no material indirection
  reaches a kernel; the override sits on `ColliderInfo`, keyed by the dense pair index `collision_pair_idx` already
  provides.
- `friction_ratio` is per-coefficient: `set_friction_ratio(sliding_ratio=..., torsional_ratio=...,
  rolling_ratio=...)` writes only what is passed. The two new arguments are keyword-only. Coefficients and ratios
  read as `<mode>_friction` and `<mode>_ratio` throughout.
- `array_class.py` loses three tensor fields. A geom's and a contact's friction coefficients are each packed into one
  vector, against the one field the override adds.

Breaking:

- `entity.material` is the registered `Material`, not the options. `entity.material.friction` and its torsional and
  rolling siblings are gone: read `material.options.friction` for the declared value or `geom.friction` for what a
  geom carries. The options base class is now `gs.materials.MaterialOptions`.
- `get_geoms_friction` / `set_geoms_friction`, the friction-ratio accessors, and `RigidSolverState.friction_ratio`
  carry the three coefficients on a trailing axis.
- `set_geoms_friction_torsional` / `_rolling` are subsumed by the packed setter.
- `entity.set_friction_ratio`'s first argument is `sliding_ratio` and scales sliding friction alone, where
  `friction_ratio` scaled all three.

Out of scope: `geom_priority`, MJCF `<pair>` import, anisotropic (5-coefficient) friction, `sol_params` averaging, and
the couplers' own friction rules.

## Related Issue

Addresses Genesis-Embodied-AI/genesis-world#2718. The `geom_priority` field proposed there is not added; a material
pair subsumes it without making users reason about which surface wins.

## Motivation and Context

Friction is a property of a pair of surfaces, so the maximum over two geoms is arbitrary: no per-surface value plus a
symmetric rule reproduces a real table. It also hides single-sided changes, since lowering only a robot's friction
against a default-1.0 ground is a no-op. The issue reports a Go2 sweep that reproduced its own baseline that way.

```
pytest tests/rigid/ -v
pytest tests/ipc/ -v -n0          # ipc needs a single worker on one GPU
```

`tests/rigid/test_friction.py::test_friction_resolved_from_material_pair`, over `n_envs=[0, 2]`, asserts
`deceleration == mu * g` for eight boxes on a plane:

- pair coefficients of 0.6 and 0.05 against surfaces that both declare 0.4, so neither a maximum nor a minimum over
  the two reproduces them
- two boxes sharing one material, one taking its material on the geom, one scaled by a sliding ratio
- distinct launch speeds, so the order broadphase reports the boxes in stops matching the order their pairs were
  numbered in

It is separate from `test_kinetic_friction` because that scene straddles the tipping threshold, where changing any one
coefficient flips an unrelated box between 176 and 0.1 degrees of tilt.

Locally `tests/rigid/` is 426 passed and `tests/ipc/` 47 passed. `test_physics_parity` and
`test_robot_grasp_fem[external_articulation]` fail identically on the merge-base.

## Checklist:

Resolves SIM-329

### DIFF
--- a/genesis/utils/array_class.py
+++ b/genesis/utils/array_class.py
@@ -76,6 +76,18 @@ def V_SCALAR_FROM(dtype, value):
     return data
 
 
+# ======================================= Friction packing ========================================
+
+
+class FrictionIdx(IntEnum):
+    """Component order of every packed friction vector: GeomsInfo.friction, GeomsState.friction_ratio and
+    ContactData.friction all carry the coefficients in this order."""
+
+    SLIDING = 0
+    TORSIONAL = 1
+    ROLLING = 2
+
+
 # =========================================== ErrorCode ===========================================
 
 
@@ -694,9 +706,8 @@ class ContactData:
     penetration: qd.Tensor
     normal: qd.Tensor
     pos: qd.Tensor
+    # Sliding, torsional and rolling coefficients of the contact, packed in that order.
     friction: qd.Tensor
-    friction_torsional: qd.Tensor
-    friction_rolling: qd.Tensor
     sol_params: qd.Tensor
     force: qd.Tensor
     link_a: qd.Tensor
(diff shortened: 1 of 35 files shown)


genesis/utils/array_class.py
duburcqa, line 1: I'm not comfortable with modifying array_class. Make sure every single new field is ABSOLUTELY NECESSARY.

duburcqa: Friction information are geometry level should be removed if some extract material indirection is added.

PR #3170, opened 2026-08-04
[FEATURE] Expose dofs limits setter on rigid entities.

### LINKED ISSUE #3168: [Feature]: RigidEntity.set_dofs_limit

`RigidEntity` exposes `get_dofs_limit` but has no matching `set_dofs_limit`, even though `RigidSolver.set_dofs_limit` exists.

Every other dof property on `RigidEntity` has a setter: `set_dofs_kp`, `set_dofs_kv`, `set_dofs_act_gain`, `set_dofs_act_bias`, `set_dofs_force_range`, `set_dofs_stiffness`, `set_dofs_armature`, `set_dofs_damping`, `set_dofs_frictionloss`, `set_dofs_position`, `set_dofs_velocity`. The only other exception is `set_dofs_invweight`, which raises `DeprecationError` on purpose and explains why.

Is the omission intentional? If it is not, I am happy to open the PR. The body is the same two lines as `set_dofs_force_range`:

```python
@gs.assert_built
def set_dofs_limit(self, lower, upper, dofs_idx_local=None, envs_idx=None):
    self._solver.set_dofs_limit(lower, upper, dofs_idx, envs_idx)
```

### Motivation

As a user randomizing an articulated object across environments,
I want to set a different joint limit per environment on the entity that owns the joint,
Because the entity's local dof indices are the ones I already have, while the solver requires scene-global indices assembled by hand.

Randomizing how far a drawer opens, per environment, currently reads:

```python
dofs_idx_global = [cabinet.get_joint(name).dof_start for name in ("slide_left", "slide_right")]
scene.rigid_solver.set_dofs_limit(lower, upper, dofs_idx=dofs_idx_global)
```

So the scene carries two index arrays for the same two joints - local ones for every other call, global ones for this one - and reaches into `scene.rigid_solver` for a single property while every neighbouring call goes through the entity.

### Potential Benefit

Per-environment joint-limit randomization becomes a public entity-level call, consistent with every sibling dof setter, and callers no longer need to know where their dofs sit in the scene-wide numbering.

- [ ] `RigidEntity.set_dofs_limit(lower, upper, dofs_idx_local=None, envs_idx=None)` exists and mirrors `set_dofs_force_range`
- [ ] it is covered by `tests/rigid/test_api.py::test_data_accessor`, whose entity row for `get_dofs_limit` currently passes `None` where the setter would go, while the solver row for the same property passes both

### Additional information

Checked against `main`. The implementation above was exercised on a 4-environment scene with `batch_dofs_info=True`: a different limit per environment round-trips through `get_dofs_limit`, and each environment's joint stops at its own limit when driven past it.

Akbro23:
## Description

`RigidEntity` gains `set_dofs_limit`, mirroring `set_dofs_force_range`: it converts local dof indices to global ones and forwards to `RigidSolver.set_dofs_limit`.

The entity row for `get_dofs_limit` in `test_data_accessor` now passes the setter, as the solver row for the same property already does.

## Related Issue

Resolves Genesis-Embodied-AI/Genesis#3168

## Motivation and Context

`limit` was the only dof property with an entity getter and no entity setter, apart from `invweight`, whose removal is deliberate and documented in the method it raises from. Setting a per-environment joint limit therefore meant assembling scene-global dof indices by hand and calling `scene.rigid_solver` directly, so a caller that uses local indices everywhere else had to carry a second index array for this one property.

`tests/rigid/test_api.py::test_data_accessor` now exercises the setter through the entity, under the same dof and environment masking combinations it already applies to every other row:

```bash
pytest -v --forked tests/rigid/test_api.py -k test_data_accessor
```

Checked manually on a 4-environment scene with `batch_dofs_info=True` and a model carrying two slide joints: a different `upper` per environment round-trips through `get_dofs_limit`, and driving every joint well past its limit leaves each environment stopped at its own value.

```python
upper = np.array([[0.060 + 0.01 * i, 0.090 + 0.01 * i] for i in range(4)])
entity.set_dofs_limit(np.zeros_like(upper), upper, dofs_idx_local=slide_dofs_idx_local)
assert np.allclose(entity.get_dofs_limit(dofs_idx_local=slide_dofs_idx_local)[1], upper)
```

I have not run the full suite locally, only the manual check above; `test_data_accessor` is expected to cover the new row exactly as it covers the solver row for the same property.

## Checklist:

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -3965,6 +3965,28 @@ class RigidEntity(KinematicEntity):
         self._solver.set_dofs_force_range(lower, upper, dofs_idx, envs_idx)
 
     @gs.assert_built
+    def set_dofs_limit(self, lower, upper, dofs_idx_local=None, envs_idx=None):
+        """
+        Set the positional limits (min and max) for the entity's dofs.
+
+        Setting a different limit per environment requires `RigidOptions.batch_dofs_info` to be enabled, otherwise all
+        environments share a single value.
+
+        Parameters
+        ----------
+        lower : array_like
+            The lower bounds of the positional limits.
+        upper : array_like
+            The upper bounds of the positional limits.
+        dofs_idx_local : None | array_like, optional
+            The indices of the dofs to set. If None, all dofs will be set. Note that here this uses the local `q_idx`, not the scene-level one. Defaults to None.
+        envs_idx : None | array_like, optional
+            The indices of the environments. If None, all environments will be considered. Defaults to None.
+        """
+        dofs_idx = self._get_global_idx(dofs_idx_local, self.n_dofs, self._dof_start, unsafe=True)
+        self._solver.set_dofs_limit(lower, upper, dofs_idx, envs_idx)
+
+    @gs.assert_built
     def set_dofs_stiffness(self, stiffness, dofs_idx_local=None, envs_idx=None):
         dofs_idx = self._get_global_idx(dofs_idx_local, self.n_dofs, self._dof_start, unsafe=True)
         self._solver.set_dofs_stiffness(stiffness, dofs_idx, envs_idx)
--- a/tests/rigid/test_api.py
+++ b/tests/rigid/test_api.py
@@ -200,7 +200,7 @@ def test_data_accessor(n_envs, batched, tol):
         (gs_robot.n_dofs, n_envs, gs_robot.get_dofs_velocity, gs_robot.set_dofs_velocity, None),
         (gs_robot.n_dofs, n_envs, gs_robot.get_dofs_position, gs_robot.set_dofs_position, None),
         (gs_robot.n_dofs, -1, gs_robot.get_dofs_force_range, gs_robot.set_dofs_force_range, None),
-        (gs_robot.n_dofs, -1, gs_robot.get_dofs_limit, None, None),
+        (gs_robot.n_dofs, -1, gs_robot.get_dofs_limit, gs_robot.set_dofs_limit, None),
         (gs_robot.n_dofs, -1, gs_robot.get_dofs_stiffness, None, None),
         (gs_robot.n_dofs, -1, gs_robot.get_dofs_invweight, None, None),
         (gs_robot.n_dofs, -1, gs_robot.get_dofs_armature, None, None),


genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -3964,6 +3964,28 @@ def set_dofs_force_range(self, lower, upper, dofs_idx_local=None, envs_idx=None)
...
+            The lower bounds of the positional limits.
+        upper : array_like
+            The upper bounds of the positional limits.
+        dofs_idx_local : None | array_like, optional
+            The indices of the dofs to set. If None, all dofs will be set. Note that here this uses the local `q_idx`, not the scene-level one. Defaults to None.
+        envs_idx : None | array_like, optional
+            The indices of the environments. If None, all environments will be considered. Defaults to None.
+        """
+        dofs_idx = self._get_global_idx(dofs_idx_local, self.n_dofs, self._dof_start, unsafe=True)
+        self._solver.set_dofs_limit(lower, upper, dofs_idx, envs_idx)
```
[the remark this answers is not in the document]
Akbro23: The planner does read a stale cache, but this PR does not introduce it. `q_limit` is built once in `_init_jac_and_IK` from `joint._dofs_limit`, which is parse-time metadata, while `get_dofs_limit` and `set_dofs_limit` go through `dyn_info.dofs.limit`. The existing solver-level setter already desynchronizes them:

```python
franka.get_jacobian(link=franka.get_link("hand"))   # populates q_limit
scene.sim.rigid_solver.set_dofs_limit(-0.5, 0.5, dofs_idx=[0, 1, 2])

franka.get_dofs_limit(dofs_idx_local=[0, 1, 2])[1]  # [0.5, 0.5, 0.5]
franka.q_limit[1][:3]                               # [2.897, 1.763, 2.897]
```

Filed as #3173, with a repro showing `plan_path` returning a trajectory whose final waypoint the solver stops 1.49 rad short of.

Keeping it out of this PR because the fix needs a decision this change does not: `q_limit` is a single unbatched array, whereas dof limits are per-environment once `RigidOptions.batch_dofs_info` is enabled.

duburcqa (approved):

genesis/engine/entities/rigid_entity/rigid_entity.py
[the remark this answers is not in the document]
duburcqa: Ok let's fix this issue first. I will do it.

duburcqa:
@jeetrex17 Could you have a look to this PR?

jeetrex17:
On it.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -3964,6 +3964,28 @@ def set_dofs_force_range(self, lower, upper, dofs_idx_local=None, envs_idx=None)
         dofs_idx = self._get_global_idx(dofs_idx_local, self.n_dofs, self._dof_start, unsafe=True)
         self._solver.set_dofs_force_range(lower, upper, dofs_idx, envs_idx)
 
+    @gs.assert_built
+    def set_dofs_limit(self, lower, upper, dofs_idx_local=None, envs_idx=None):
+        """
+        Set the positional limits (min and max) for the entity's dofs.
+
+        Setting a different limit per environment requires `RigidOptions.batch_dofs_info` to be enabled, otherwise all
+        environments share a single value.
```
jeetrex17: If I read `_set_dofs_info` right, this sentence documents an inconsistency rather than a contract. Measured on current main, Franka, `batch_dofs_info=False`, `n_envs=3`, on both zero-copy arms (`gs.use_zerocopy=True` and `GS_ENABLE_ZEROCOPY=0`):

- `solver.set_dofs_limit(-0.1, 0.1, dofs_idx=..., envs_idx=[0])` is accepted and writes every environment (`upper[:3] -> [0.1, 0.1, 0.1]`)
- `entity.get_dofs_limit(envs_idx=[0])` raises `` `envs_idx` cannot be specified for non-batched dofs info. ``
- `set_dofs_force_range(..., envs_idx=[1])` is accepted silently as well

Every dofs-info getter raises (`rigid_solver.py:3084-3168`), no dofs-info setter does, and `_set_links_info` raises on the setter side (`rigid_solver.py:2270`). The same sentence on `set_dofs_kp` in #3144 got ["Is it really ignored? If it is the case, then it is a bug."](https://github.com/Genesis-Embodied-AI/genesis-world/pull/3144#discussion_r3726523705) The two-line guard in `_set_dofs_info` mirroring `_set_links_info` would close it for all eleven setters at once, and the paragraph goes.

tests/rigid/test_api.py
```diff
@@ -200,7 +200,7 @@ def must_cast(value, dtype):
         (gs_robot.n_dofs, n_envs, gs_robot.get_dofs_velocity, gs_robot.set_dofs_velocity, None),
         (gs_robot.n_dofs, n_envs, gs_robot.get_dofs_position, gs_robot.set_dofs_position, None),
         (gs_robot.n_dofs, -1, gs_robot.get_dofs_force_range, gs_robot.set_dofs_force_range, None),
-        (gs_robot.n_dofs, -1, gs_robot.get_dofs_limit, None, None),
+        (gs_robot.n_dofs, -1, gs_robot.get_dofs_limit, gs_robot.set_dofs_limit, None),
```
jeetrex17: This hunk conflicts with main, which replaced `-1` with `n_envs_info`. The rebased row `(gs_robot.n_dofs, n_envs_info, gs_robot.get_dofs_limit, gs_robot.set_dofs_limit, None)` also sweeps `envs_idx` under `batched=True`, which is the per-environment case from #3168.

jeetrex17: Once the setter refuses `envs_idx` on non-batched dofs info, `test_batched_info` is already parametrized on `batch_dofs_info` and already holds the links loop (main:526-529). A sibling loop over the dofs setters under `pytest.raises(..., match="non-batched dofs info")` costs no build.

genesis/engine/entities/rigid_entity/rigid_entity.py
wehaveadream: If I read `_set_dofs_info` right, this sentence documents an inconsistency rather than a contract. Measured on current main, Franka, `batch_dofs_info=False`, `n_envs=3`, on both zero-copy arms (`gs.use_zerocopy=True` and `GS_ENABLE_ZEROCOPY=0`):

- `solver.set_dofs_limit(-0.1, 0.1, dofs_idx=..., envs_idx=[0])` is accepted and writes every environment (`upper[:3] -> [0.1, 0.1, 0.1]`)
- `entity.get_dofs_limit(envs_idx=[0])` raises `` `envs_idx` cannot be specified for non-batched dofs info. ``
- `set_dofs_force_range(..., envs_idx=[1])` is accepted silently as well

Every dofs-info getter raises (`rigid_solver.py:3084-3168`), no dofs-info setter does, and `_set_links_info` raises on the setter side (`rigid_solver.py:2270`). The two-line guard in `_set_dofs_info` mirroring `_set_links_info` would close it for all eleven setters at once, and the paragraph goes.

tests/rigid/test_api.py
wehaveadream: This hunk conflicts with main, which replaced `-1` with `n_envs_info`. The rebased row `(gs_robot.n_dofs, n_envs_info, gs_robot.get_dofs_limit, gs_robot.set_dofs_limit, None)` also sweeps `envs_idx` under `batched=True`, which is the per-environment case from #3168.

wehaveadream: Once the setter refuses `envs_idx` on non-batched dofs info, `test_batched_info` is already parametrized on `batch_dofs_info` and already holds the links loop (main:526-529). A sibling loop over the dofs setters under `pytest.raises(..., match="non-batched dofs info")` costs no build.

genesis/engine/entities/rigid_entity/rigid_entity.py
jeetrex17, line 3973: If I read `_set_dofs_info` right, this sentence documents an inconsistency rather than a contract. Measured on current main, Franka, `batch_dofs_info=False`, `n_envs=3`, on both zero-copy arms (`gs.use_zerocopy=True` and `GS_ENABLE_ZEROCOPY=0`):

- `solver.set_dofs_limit(-0.1, 0.1, dofs_idx=..., envs_idx=[0])` is accepted and writes every environment (`upper[:3] -> [0.1, 0.1, 0.1]`)
- `entity.get_dofs_limit(envs_idx=[0])` raises `` `envs_idx` cannot be specified for non-batched dofs info. ``
- `set_dofs_force_range(..., envs_idx=[1])` is accepted silently as well

Every dofs-info getter raises (`rigid_solver.py:3084-3168`), no dofs-info setter does, and `_set_links_info` raises on the setter side (`rigid_solver.py:2270`). Not blocking for this PR, since the gap is pre-existing across every dof-info setter, but this paragraph writes that gap down as the contract, so I'd drop it here. If you'd rather close the gap in this PR, a two-line guard in `_set_dofs_info` mirroring `_set_links_info` covers every dof-info setter at once, and `test_batched_info` (already parametrized on `batch_dofs_info`, links loop at main:526-529) is the place to assert the refusal at no extra build. Otherwise a follow-up is fine.

tests/rigid/test_api.py
jeetrex17, line 203: This hunk conflicts with main, which replaced `-1` with `n_envs_info`. The rebased row `(gs_robot.n_dofs, n_envs_info, gs_robot.get_dofs_limit, gs_robot.set_dofs_limit, None)` also sweeps `envs_idx` under `batched=True`, which is the per-environment case from #3168.

duburcqa:
@jeetrex17 You should probably finish it yourself.

genesis/engine/entities/rigid_entity/rigid_entity.py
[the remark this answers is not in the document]
duburcqa: https://github.com/Genesis-Embodied-AI/genesis-world/issues/3173 is fixed now.

jeetrex17:
@duburcqa I opened #3348 as a superseding PR for the same change, rebased on `main`.

  Thanks for the original patch, @Akbro23.

PR #3171, opened 2026-08-04
  [BUG FIX] Apply MeshSet poses to sampled particles.

### LINKED ISSUE #3169: [Bug]: MeshSet particle sampling ignores member rotations
### Bug Description

Particles sampled for gs.morphs.MeshSet do not receive the rotation specified by each member's eulers. Only the translation in poss is applied.

The visual mesh receives the complete rotation and translation, causing the rendered mesh and its physical particles to have different orientations. This also affects Hybrid entities generated from URDF links: particles belonging to rotated links remain aligned with the original local axes.

### Steps to Reproduce

```python
import numpy as np
import trimesh

import genesis as gs


gs.init(backend=gs.cpu)

pos = np.array((0.2, 0.1, 0.2))
mesh = trimesh.creation.box(extents=(0.04, 0.12, 0.04))

scene = gs.Scene(
        lower_bound=(-0.5, -0.5, -0.5),
        upper_bound=(0.5, 0.5, 0.5),
        particle_size=0.01,
    ),
    show_viewer=False,
)

entity = scene.add_entity(
    morph=gs.morphs.MeshSet(
        files=(mesh,),
        poss=(pos,),
        eulers=((0.0, 0.0, 90.0),),
    ),
    material=gs.materials.MPM.Elastic(sampler="regular"),
)

particle_extents = np.ptp(entity.init_particles, axis=0)
print("Particle extents:", particle_extents)

np.testing.assert_allclose(
    particle_extents,
    (0.11, 0.03, 0.03),
    atol=1e-6,
)
```

### Expected Behavior

The 90-degree rotation around the Z axis should rotate the box's long axis from Y to X. The particle extents should therefore be approximately: `[0.11, 0.03, 0.03]`

The physical particles and visual mesh should have the same orientation.

### Screenshots/Videos

I have also written a URDF file to test the particle rotation.

Error: Particle rotation.

Image

Which should be like:

Image

### Environment

- OS: Ubuntu 22.04.5 LTS, x86_64
- CPU: AMD Ryzen 9 9950X 16-Core Processor
- GPU-driver version: N/A 

 genesis-world 1.3.1
 e4e801359327e8dec4c64f68902ba895b3e6fefe

prexhu:
## Description

Apply each MeshSet member's rotation and translation to its sampled particle positions. 

Previously, MeshSet particles were only translated, while the corresponding visual mesh received the complete pose.

Add an MPM regression test using an asymmetric mesh rotated by 90 degrees.
## Related Issue

Resolves Genesis-Embodied-AI/Genesis#3169

## Motivation and Context

Hybrid entities instantiate their soft parts through MeshSet. Ignoring each member's rotation caused particles generated from rotated URDF links to remain aligned with the original local axes instead of attaching to the links.

```bash
pytest -n 0 tests/particles/test_mpm.py
```
## Checklist:

### DIFF
--- a/genesis/engine/entities/particle_entity.py
+++ b/genesis/engine/entities/particle_entity.py
@@ -264,7 +264,11 @@ class ParticleEntity(Entity):
                         sampler=sampler,
                     )
 
-                particles_i += np.asarray(morph_i.pos, dtype=gs.np_float)
+                particles_i = gu.transform_by_trans_quat(
+                    particles_i,
+                    np.asarray(morph_i.pos, dtype=gs.np_float),
+                    np.asarray(morph_i.quat, dtype=gs.np_float),
+                )
                 particles.append(particles_i)
         elif isinstance(self._morph, (gs.options.morphs.Primitive, gs.options.morphs.Mesh)):
             particles = self._vmesh.particlize(self._particle_size, self.sampler)
--- a/tests/particles/test_mpm.py
+++ b/tests/particles/test_mpm.py
@@ -1,7 +1,35 @@
+import numpy as np
 import pytest
 import torch
+import trimesh
 
 import genesis as gs
+
+
+@pytest.mark.required
+def test_mesh_set_particle_transform(show_viewer):
+    pos = np.array((0.2, 0.1, 0.2))
+    mesh = trimesh.creation.box(extents=(0.04, 0.12, 0.04))
+    scene = gs.Scene(
+        mpm_options=gs.options.MPMOptions(
+            lower_bound=(-0.5, -0.5, -0.5),
+            upper_bound=(0.5, 0.5, 0.5),
+            particle_size=0.01,
+        ),
+        show_viewer=show_viewer,
+    )
+    entity = scene.add_entity(
+        morph=gs.morphs.MeshSet(
+            files=(mesh,),
+            poss=(pos,),
+            eulers=((0.0, 0.0, 90.0),),
+        ),
+        material=gs.materials.MPM.Elastic(sampler="regular"),
+    )
+
+    particles = entity.init_particles
+    np.testing.assert_allclose(particles.mean(axis=0), pos, atol=1e-6)
+    np.testing.assert_allclose(np.ptp(particles, axis=0), (0.11, 0.03, 0.03), atol=1e-6)
 
 
 @pytest.mark.required


genesis/engine/entities/particle_entity.py
```diff
@@ -264,7 +264,11 @@ def sample(self):
                         sampler=sampler,
                     )
 
-                particles_i += np.asarray(morph_i.pos, dtype=gs.np_float)
+                particles_i = gu.transform_by_trans_quat(
+                    particles_i,
+                    np.asarray(morph_i.pos, dtype=gs.np_float),
+                    np.asarray(morph_i.quat, dtype=gs.np_float),
+                )
```
jeetrex17, line 271: The transform is mathematically correct, but please compute `pos_i` and `quat_i` once and use the same values for both the particles and `self._vmesh[i]` in this loop, then remove the later visual-transform loop. This avoids maintaining two derivations of the same member pose and keeps particle placement, rendering, and skinning in the same frame.

These are validated internal tuples, so the repository guidelines also require `np.array` rather than `np.asarray`.

```python
pos_i = np.array(morph_i.pos, dtype=gs.np_float)
quat_i = np.array(morph_i.quat, dtype=gs.np_float)
particles.append(gu.transform_by_trans_quat(particles_i, pos_i, quat_i))
self._vmesh[i].apply_transform(gu.trans_quat_to_T(pos_i, quat_i))
```

tests/particles/test_mpm.py
```diff
@@ -1,9 +1,37 @@
+import numpy as np
 import pytest
 import torch
+import trimesh
 
 import genesis as gs
 
 
+@pytest.mark.required
+def test_mesh_set_particle_transform(show_viewer):
```
jeetrex17, line 10: `CODING_GUIDELINES.md` sections 12.1–12.2 require bug-fix regression assertions to be added to the existing test covering the capability; bug fixes are explicitly not an exception.

Please fold this MeshSet entity and its assertions into an existing MPM test and reuse that scene build. A separate test incurs another full Genesis initialization for every CI configuration.

tests/particles/test_mpm.py
```diff
@@ -1,9 +1,37 @@
...
+    pos = np.array((0.2, 0.1, 0.2))
```
jeetrex17, line 11: Please keep this value as a tuple:

```python
pos = (0.2, 0.1, 0.2)
```

The scene-construction guidelines prohibit wrapping controlled array-like inputs in `np.array`, and the project assertion helper accepts tuples directly.

tests/particles/test_mpm.py
```diff
@@ -1,9 +1,37 @@
...
+    mesh = trimesh.creation.box(extents=(0.04, 0.12, 0.04))
+    scene = gs.Scene(
+        mpm_options=gs.options.MPMOptions(
+            lower_bound=(-0.5, -0.5, -0.5),
+            upper_bound=(0.5, 0.5, 0.5),
+            particle_size=0.01,
```
jeetrex17, line 17: Keyword arguments must follow the callee's declaration order, so `particle_size` must precede `lower_bound` and `upper_bound`.

The custom bounds are also unnecessary for this geometry because it already fits inside the default MPM domain. This can be reduced to:

```python
    particle_size=0.01,
),
```

tests/particles/test_mpm.py
```diff
@@ -1,9 +1,37 @@
...
+        ),
+        show_viewer=show_viewer,
+    )
+    entity = scene.add_entity(
+        morph=gs.morphs.MeshSet(
+            files=(mesh,),
+            poss=(pos,),
+            eulers=((0.0, 0.0, 90.0),),
+        ),
+        material=gs.materials.MPM.Elastic(sampler="regular"),
```
jeetrex17, line 27: Scene-building calls require one option per line, even when only one option is provided:

```python
material=gs.materials.MPM.Elastic(
    sampler="regular",
),
```

tests/particles/test_mpm.py
```diff
@@ -1,9 +1,37 @@
...
+    )
+
+    particles = entity.init_particles
+    np.testing.assert_allclose(particles.mean(axis=0), pos, atol=1e-6)
```
jeetrex17, line 31: Please use the repository assertion utility instead of NumPy's testing API:

```python
from ..utils.assertions import assert_allclose
```

Then call `assert_allclose(..., atol=1e-6)`.

An explicit `atol=1e-6` sits at roughly 5x the measured fp32 floor (~2e-7) while still detecting the 0.08 failure by five orders of magnitude, and stays a fixed absolute bound across precision arms.

tests/particles/test_mpm.py
```diff
@@ -1,9 +1,37 @@
...
+    np.testing.assert_allclose(np.ptp(particles, axis=0), (0.11, 0.03, 0.03), atol=1e-6)
```
jeetrex17, line 32: This catches a completely missing rotation, but it does not fully validate MeshSet pose semantics.

A single centered box cannot detect applying the wrong member's pose, and +90° and −90° rotations have identical means and extents. Please test at least two members with different rotation axes and make one mesh off-center or assert another direction-sensitive quantity so an inverse rotation fails.

Please also derive the expected spans from `extents - particle_size` and consider asserting that the particle and public visual-mesh AABBs remain aligned.

Note this interacts with the bounds comment above: keep every member inside the default MPM domain (z well above 0, within the safety padding), or the custom bounds become necessary again.

jeetrex17:
Hi @prexhu, just following up on this draft. Are you planning to continue addressing the review feedback? I’d be happy to help with the implementation or test restructuring if useful.

PR #3176, opened 2026-08-04
[MISC] Make perf dispatch intervals deterministic.

hughperkins:


hughperkins:
replacing with 300 steps

PR #3184, opened 2026-08-05
[BUG FIX] Preserve authored inertia matrix when center of mass is unspecified.

### LINKED ISSUE #3183: [Bug]: Authored URDF inertia is discarded when inertial origin is omitted
## Bug Description

Genesis discards an authored URDF inertia tensor when the `<inertial>` element
contains valid `<mass>` and `<inertia>` children but omits `<origin>`.

The bundled `urdfpy.Inertial` API documents an omitted origin as identity, and
Genesis already treats a missing inertial origin as identity while merging fixed
links. In the normal `parse_urdf()` path, however, the omission remains `None`.
Rigid-link finalization then logs:

```text
Ignoring inertia matrix of link 'wheel' because center of mass is not specified.
```

and falls back to geometry-derived inertia instead of the authored tensor.

This is a narrow follow-up to merged PR
[#2499](https://github.com/Genesis-Embodied-AI/genesis-world/pull/2499).
That change preserved an omitted inertial origin as `None`; in the current
parser/finalizer chain, `None` is then interpreted as a missing center of mass
and the complete authored tensor is discarded. The proposed fix does not undo
the geometry fallback for a link with no `<inertial>` element.

## Steps to Reproduce

```python
import genesis as gs

URDF = """<?xml version="1.0"?>
<robot name="implicit_inertial_origin">
  <link name="body">
    <visual><geometry><box size="0.2 0.2 0.2"/></geometry></visual>
    <collision><geometry><box size="0.2 0.2 0.2"/></geometry></collision>
    <inertial>
      <mass value="2.5"/>
      <inertia ixx="0.11" ixy="0.01" ixz="0.02"
               iyy="0.22" iyz="0.03" izz="0.30"/>
    </inertial>
  </link>
</robot>
"""

gs.init(backend=gs.cpu)
scene.add_entity(gs.morphs.URDF(file=URDF, merge_fixed_links=False))
scene.build()
```

The same parser boundary can be inspected without building a scene: on current
`main`, `parse_urdf()` returns `inertial_pos=None` while retaining the authored
inertia matrix.

The complete `scene.build()` reproduction on the tested current-`main`
snapshot (`207db282fa523e7cdf6c0b1c85d7b72b12d79a7a`, fetched and tested
2026-08-05) emits the warning above and replaces the authored principal moments
with the geometry-derived
`[0.01666666753590107, 0.01666666753590107, 0.01666666753590107]`. With the
proposed parser fix replayed as
`0fa0f4ae5c83e964282fea1d6ad44aa333ee1850`, the warning is absent and the
equivalent authored principal moments are retained. The source change has the
same stable patch-id as the original local patch:
`aa50b725672c505043ab82e1e44f4052abe1f4eb`.

## Expected Behavior

When `<inertial>` exists and `<origin>` is omitted, Genesis should use the link
frame (`xyz="0 0 0"`, identity rotation) as the inertial frame. The authored
mass and inertia tensor should be retained and the warning should not be
emitted.

This matches `urdfdom`, which initializes the inertial pose to identity and
only overwrites it when an `<origin>` child is present:
<https://github.com/ros/urdfdom/blob/bfcf29f39cc3e0fa13e5572f36b4b8c42d4f5ce1/urdf_parser/src/link.cpp#L269-L279>.

An entirely absent `<inertial>` element should continue to use Genesis's
geometry-derived fallback.

## Environment

- OS: Ubuntu 24.04
- Cloud host GPU: AMD Radeon PRO W7900
- Runtime observation: Genesis 1.1.2, PyTorch 2.9.1 + ROCm/HIP 7.2
- Current-main parser and CPU scene-build reproduction: Genesis 1.3.1 at
  `207db282fa523e7cdf6c0b1c85d7b72b12d79a7a`, tested 2026-08-05 with
  `gs.cpu`
- Tested current-main patch replay:
  `0fa0f4ae5c83e964282fea1d6ad44aa333ee1850`

## Additional Context

The problem surfaced while validating two non-fixed differential-drive URDFs
on `gs.amdgpu`. Adding explicit zero-valued origins to the assets removes the
warning, but normalizing the omitted origin inside Genesis fixes the same valid
input for all users.

eason4kim-rocket:
## Description

A link that authors a complete inertia tensor but leaves its center of mass unspecified had that tensor silently replaced by the geometry-derived estimate. The center of mass is now resolved to the link frame origin in that case, which is what every supported asset format means by an omitted inertial frame, and the authored tensor is kept.

The resolution happens in `finalize_inertial`, the single place where explicit values and the geometry estimate are combined, so the parsers keep reporting an omitted center of mass as unspecified rather than claiming a value the asset never authored. Fixing it there also covers USD, whose `MassAPI` reaches the same state through an unauthored `centerOfMass` alongside an authored `diagonalInertia`.

Three further sites are brought in line with that resolution:

- The center-of-mass consistency check against the geometry bounding box used to be skipped whenever the center of mass was omitted. It now runs on the value that is actually used, so an authored tensor whose implied center of mass falls outside its own geometry is still reported.
- The inverse weight is no longer invalidated by an omitted center of mass alone, since nothing is derived from the geometry in that case.
- A heterogeneous variant used to accept the inertial parsed from its file only when mass, center of mass and tensor were all present, and fall back to geometry otherwise. It now goes through the same resolution as the primary variant, so an asset behaves identically whichever variant slot it occupies. Previously two environments loading the same omitted-origin URDF got masses 2.5 and 0.543, with or without root alignment.

The inverse-weight invalidation that two of those bullets rely on was itself inoperative. `_parse_scene` collects reasons to invalidate in one variable, but the block that records them assigned `False` in exactly the two cases its own comments say must invalidate - inertial properties unset because the asset leaves them undefined, and a default rotor armature - while the later block used the opposite polarity and reset the variable before either reason was ever read. Both were silently dropped. The two blocks now share one meaning and the invalidation is applied once, after every reason has been collected. The effect is visible on any URDF that declares its own floating joint, where nothing else invalidates: with the Genesis default rotor inertia of 0.1, the revolute joint of the free-flyer model kept an inverse weight of 1075.0 while the solver actually integrated 9.9078, and raising the armature to 1000.0 did not move it at all. Both values now equal the inverse effective inertia measured from the simulated response, and disabling the default armature reproduces MuJoCo's `dof_invweight0` exactly on every degree of freedom.

## Related Issue

Resolves Genesis-Embodied-AI/genesis-world#3183

## Motivation and Context

`<inertial>` elements without an `<origin>` child are common in real URDF assets, and the URDF specification defines the omitted inertial frame as the link frame. Genesis instead treated the missing center of mass as missing data, warned `Ignoring inertia matrix of link '<name>' because center of mass is not specified.`, and fell back to the geometry estimate, so an asset with a carefully authored inertia tensor silently simulated with a different one.

Fixed-link merging already normalized the omitted origin to identity on its own. Merged and unmerged loading of the same asset therefore disagreed on the composite rigid-body inertia, which is what the new merge invariant pins down.

```bash
pytest -n 0 tests/rigid/test_asset_loading.py::test_urdf_parsing_inertia_defaults
pytest -n 0 tests/rigid/test_heterogeneous.py::test_variant_inertia_matches_standalone
pytest -n 0 tests/parsers/test_usd.py::test_massapi_invalid_defaults_mjcf_vs_usd
pytest -n 0 tests/rigid/test_asset_loading.py::test_default_armature_freeflyer
```

All four fail on `main` and pass with this change:

- `test_urdf_parsing_inertia_defaults` covers, in a single scene build, a link with no `<inertial>` element at all, a link whose `<inertial>` omits `<origin>`, and the same two-link fixed-jointed asset loaded with and without fixed-link merging. Its sphere geometry is offset from the link frame, so the link frame and the geometry-derived center of mass are distinguishable; on `main` the omitted-origin center of mass lands on the geometry centroid, 0.09 away.
- The merge invariant compares the composite principal moments of the merged and unmerged variants. On `main` they differ by 76% relative; here they agree to `5e-7`, about four times the measured cross-path floor (1.2e-7 in single precision, 1.8e-8 in double).
- The tensor itself is compared through its rotation-invariant principal moments, because it is stored in its principal frame and the parsed principal-axis quaternion is markedly less accurate than the tensor.
- `test_massapi_invalid_defaults_mjcf_vs_usd` gains a body whose `MassAPI` authors mass and `diagonalInertia` while leaving `centerOfMass` at its sentinel, compared against the MJCF body whose inertial frame is the body origin. On `main` its inertia mismatches by 0.293.
- `test_variant_inertia_matches_standalone` builds one scene, without stepping, holding two single-link URDF variants - one with no `<inertial>` element, one omitting its `<origin>` - and compares the heterogeneous entity's per-environment mass against a standalone entity of each asset. On `main` the second environment reports 0.543 instead of 2.5. It cannot be folded into `test_physics_parity`: variants must agree on joint names, and the root joint is named `root` by the MuJoCo parser but `root_joint` by the URDF path, so no URDF variant can join that test's MJCF variants.
- `test_default_armature_freeflyer` gains a second entity of the same model without an explicit armature, and asserts that rotor inertia lowers the inverse weight of the joint it is applied to. Its URDF parametrization fails on `main`, where both entities report the same 1075.0; the MJCF one passes either way, since that path never reaches the invalidation and MuJoCo accounts for the armature itself. The existing MuJoCo parity check on inverse weights could not catch this: `tests/utils.py` forces recomputation with `link.invweight[:] = -1` before comparing, so it validates the recomputation and is blind to whether the parser decides to invalidate.

One further test covers behaviour this change now depends on rather than a regression, so it passes on `main` too: `test_invalid_variant_specification_raises` pins the three rejections that make a heterogeneous entity well-formed - an unsupported material, a morph family that cannot be a variant at all, and a basic object mixed with an articulated robot. The last of these is what keeps the inertial parsed per variant aligned with the variant index, since only URDF and MJCF variants record one. Each case matches its own message, because a bare `pytest.raises(gs.GenesisException)` still passes with the guards removed - something further down raises too.

Full runs: `tests/parsers` 61 passed / 2 skipped, `tests/rigid/test_heterogeneous.py` 8 passed, and `tests/rigid/test_asset_loading.py` with `tests/rigid/test_mujoco_parity.py` 106 passed.

Not applicable.

## Checklist:

- [ ] All new and existing tests passed. Only `tests/parsers`, `tests/rigid/test_asset_loading.py`, `tests/rigid/test_heterogeneous.py` and `tests/rigid/test_mujoco_parity.py` were run locally.


genesis/utils/urdf.py
```diff
@@ -178,9 +178,15 @@ def parse_urdf(morph, surface):
         l_info["inertial_i"] = None
         l_info["inertial_mass"] = None
         if link.inertial is not None:
-            if link.inertial.origin is not None:
-                l_info["inertial_pos"] = link.inertial.origin[:3, 3]
-                l_info["inertial_quat"] = gu.R_to_quat(link.inertial.origin[:3, :3])
+            # URDF defines a missing inertial origin as the link frame.  The
+            # vendored urdfpy parser preserves the omission as ``None``, so
+            # normalize it here instead of discarding an otherwise complete
+            # authored inertia tensor during rigid-link finalization.
```
duburcqa: Fix formatting.

tests/rigid/test_asset_loading.py
```diff
@@ -19,6 +19,58 @@
...
+    ET.SubElement(robot_without_inertial, "link", name="body")
+    morph_without_inertial = gs.morphs.URDF(
+        file=ET.tostring(robot_without_inertial, encoding="unicode"),
+        merge_fixed_links=False,
+    )
+    absent_infos, _, _, _ = uu.parse_urdf(morph_without_inertial, gs.surfaces.Default())
+
+    assert absent_infos[0]["inertial_pos"] is None
+    assert absent_infos[0]["inertial_mass"] is None
+    assert absent_infos[0]["inertial_i"] is None
```
duburcqa: This unit test is not following our guidelines.

eason4kim-rocket: Addressed in 91e7bba. The custom URDF now lives in a session fixture, the regression is folded into the existing inertia-default scene, and the assertions use public RigidLink properties instead of parse_urdf internals. Ruff formatting, lint, diff checks, and Python compilation pass locally.

genesis/utils/urdf.py
eason4kim-rocket: Addressed in 91e7bba: the comment now follows the project style, spells out Unified Robot Description Format on first use, and avoids the unnecessary explicit dtype.

tests/rigid/test_asset_loading.py
duburcqa: Excellent! Good job!

genesis/utils/urdf.py
```diff
@@ -178,9 +178,10 @@ def parse_urdf(morph, surface):
         l_info["inertial_i"] = None
         l_info["inertial_mass"] = None
         if link.inertial is not None:
-            if link.inertial.origin is not None:
-                l_info["inertial_pos"] = link.inertial.origin[:3, 3]
-                l_info["inertial_quat"] = gu.R_to_quat(link.inertial.origin[:3, :3])
+            # Unified Robot Description Format (URDF) inertial elements without an origin use the link frame.
+            inertial_origin = link.inertial.origin if link.inertial.origin is not None else np.eye(4)
+            l_info["inertial_pos"] = inertial_origin[:3, 3]
+            l_info["inertial_quat"] = gu.R_to_quat(inertial_origin[:3, :3])
```
duburcqa: Let me check. I don't think this is correct in general. Here the inertia origin while be considered set even if it is not. The solution is rather to ensure that the origin is considered legitimately omitted if the inertia matrix is specified.

eason4kim-rocket: You're right — setting an identity transform in the URDF parser erases the distinction between an omitted origin and an explicit one. I moved the defaulting to inertial finalization in e4b2ab5: the parser now keeps None, and a missing center of mass is treated as the link-frame origin only when an explicit inertia matrix is present. The scene-level regression still checks the resolved public values. Thanks for catching that.

duburcqa: Good. But the unit test should be made stronger because the latent bug that you introduced was not caught by the existing unit test suite, or you did not check locally before pushing.

eason4kim-rocket: You're right. Before the earlier push I had only run formatting, lint, and compilation; I had not run the Genesis test, and its post-build assertions were too weak to expose the early parser normalization. I strengthened the existing scene in 3734a9f: before build it now checks that the public link state preserves the omitted origin, and after build it checks the resolved zero center of mass, mass, and authored principal moments. On Genesis 1.3.1 with the CPU backend, the target test passes in 43.13 s. Reapplying the bad parser normalization makes the same test fail at the pre-build assertion in 4.89 s. Thanks for calling this out.

duburcqa:
Pushed some improvements. It should be ready now.

duburcqa (dismissed):

duburcqa:
@codex review

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -209,8 +210,11 @@ def finalize_inertial(
         hint_mass = mass
     if mass is None:
         mass = hint_mass
-    if com is None or inertia is None:
+    if inertia is None:
         com, inertia, quat = hint_com, hint_inertia, gu.identity_quat()
+    elif com is None:
+        # Falling back to the geometry estimate here would discard an otherwise complete authored inertia.
+        com = gu.zero_pos()
```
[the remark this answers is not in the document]
duburcqa: fixed

duburcqa (dismissed):

duburcqa (dismissed):

duburcqa:
@codex review

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -879,34 +884,18 @@ def _build(self):
...
-                mass, com, inertia = self.entity._links_inertial_info[self.idx - self.entity._link_start][v].hint
-                if mass <= 0.0:
-                    mass = gs.EPS
-                    com = np.zeros(3, dtype=gs.np_float)
-                    inertia = np.zeros((3, 3), dtype=gs.np_float)
-                self._variant_inertial.append(LinkInertial(mass, com, gu.identity_quat(), inertia))
+                    if morph_v.recompute_inertia and not self._is_fixed:
+                        v_mass, v_pos, v_quat, v_i = None, None, None, None
+                hint = self.entity._links_inertial_info[self.idx - self.entity._link_start][v].hint
+                self._variant_inertial.append(finalize_inertial(v_mass, v_pos, v_quat, v_i, *hint))
```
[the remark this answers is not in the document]
duburcqa: 👎 wrong

duburcqa (dismissed):

duburcqa:
@codex review

duburcqa (dismissed):

duburcqa (approved):

Issue #3177, opened 2026-08-05
[Bug]: control_dofs_position silently does nothing on tendon-approximated finger joints


Enoch208:
### Bug Description

Loading the bundled `xml/franka_emika_panda/panda.xml` prints `(MJCF) Approximating tendon by joint actuator for finger_joint1/2`. The resulting finger DOFs get a general gain/bias actuator, with two consequences that are invisible at the call site:

1. `control_dofs_position` on those DOFs returns normally and moves nothing (0.0 mm over 400 steps, no exception, no warning).
2. `get_dofs_kp` on those DOFs raises `GenesisException: Some DOFs use a non-PD-reducible actuator (act_gain != -act_bias[1]). Use get_dofs_act_gain() and get_dofs_act_bias() instead.` — informative once you already suspect the actuator type, but `control_dofs_position` itself gives no signal.

A gripper written against the documented position API appears to work — no exception, no warning — while the fingers never move. `control_dofs_force` on the same DOFs works correctly (fingers open 80.8 mm under +5 N per finger), which is the workaround we shipped.

(Related to #374 / #468, which ask for real tendon support; this report is narrower: while the approximation exists, its position-control limitation should be signalled at the `control_dofs_position` call.)

### Steps to Reproduce

Self-contained, only bundled assets — [genesis_finger_position_control.py](https://github.com/Enoch208/Crux/blob/main/upstream/genesis_finger_position_control.py):

```python
from __future__ import annotations

import genesis as gs

FRANKA_MJCF = "xml/franka_emika_panda/panda.xml"
FINGER_DOFS = [7, 8]
OPEN_TARGET = [0.04, 0.04]
CLOSE_TARGET = [0.0, 0.0]
OPEN_FORCE = [5.0, 5.0]
STEPS = 400


def finger_gap(franka: object) -> float:
    left = franka.get_link("left_finger").get_pos()
    right = franka.get_link("right_finger").get_pos()
    return float(sum((a - b) ** 2 for a, b in zip(left, right, strict=True)) ** 0.5)


def main() -> int:
    gs.init(backend=gs.gpu)
    scene = gs.Scene(sim_options=gs.options.SimOptions(dt=0.005), show_viewer=False)
    scene.add_entity(gs.morphs.Plane())
    franka = scene.add_entity(gs.morphs.MJCF(file=FRANKA_MJCF))
    scene.build()

    print("\n1. can the caller detect that these DOFs are not position-controllable?")
    try:
[...]

PR #3193, opened 2026-08-06
Warn when position/velocity control targets non-PD-reducible actuators

Enoch208:
Fixes the silent no-op reported in #3177.

### The problem

MJCF tendons are approximated by general gain/bias joint actuators — loading the bundled `xml/franka_emika_panda/panda.xml` prints `(MJCF) Approximating tendon by joint actuator for finger_joint1/2`. Those DOFs ignore `ctrl_pos` and `ctrl_vel` entirely, so:

```python
franka.control_dofs_position([0.0, 0.0], [7, 8])   # returns normally
for _ in range(400):
    scene.step()                                    # fingers move 0.0 mm
```

No exception, no warning, nothing in the log. `get_dofs_kp` *does* raise a clear error on those DOFs, but a caller has no reason to call a getter before controlling a joint — so a gripper written against the documented position API appears to work while the fingers never move. `control_dofs_force` on the same DOFs works correctly.

### The change

Warn once per solver, naming the offending DOFs, the first time position or velocity control is used on a model containing non-PD-reducible actuators:

```
Position control was requested while dofs [7, 8] use a non-PD-reducible actuator
(act_gain != -act_bias[1], or act_bias[0] != 0). Those dofs ignore position and
velocity targets; drive them with control_dofs_force instead. Use
get_dofs_act_gain()/get_dofs_act_bias() to inspect them.
```

Design choices, happy to change any of them:

- **A warning rather than an exception**, so existing code that mixes control modes across DOFs keeps working. Raising here would break the perfectly legitimate pattern of calling `control_dofs_position` on all DOFs and `control_dofs_force` on the actuated subset.
- **Once per solver, behind a boolean flag** — stepping loops call these methods every step, so the check must not sit in the hot path. After the first call the guard is a single attribute read.
- **The reducibility test is the same expression `get_dofs_kp` already uses**, so the two agree by construction.

### Test

`tests/rigid/test_control.py::test_general_actuator_position_control_warns` reuses the existing `general_actuator` fixture (joint 1 is already non-PD-reducible there) and asserts both that the warning fires and that it is not repeated on subsequent calls.

### Context

Found while building a contact-rich manipulation harness on an AMD Radeon PRO W7900 (gfx1100, ROCm 7.2.1, `torch 2.13.0+rocm7.2`, backend `gs.amdgpu`) — the silent no-op cost a debugging session before `get_dofs_kp`'s error revealed the actuator type. Minimal reproduction and the full ROCm console log are in #3177. Happy to iterate on the wording or the warning-vs-exception call.

### DIFF
--- a/genesis/engine/solvers/rigid/rigid_solver.py
+++ b/genesis/engine/solvers/rigid/rigid_solver.py
@@ -274,6 +274,7 @@ class RigidSolver(KinematicSolver):
         self._use_contact_island = options.use_contact_island
         # Hibernation builds on islands, so requesting it without islands is a genuine conflict.
         self._use_hibernation = options.use_hibernation
+        self._warned_non_pd_control = False
         if self._use_hibernation and not self._use_contact_island:
             gs.raise_exception(
                 "`use_hibernation=True` requires `use_contact_island=True`, as hibernation builds on islands."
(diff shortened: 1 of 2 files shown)


Enoch208:
Thanks — the first point is correct and it invalidated my premise, so I went and measured rather than argued.

Printing the actuator coefficients for the bundled Franka MJCF (ROCm 7.2.1, Genesis on `gs.amdgpu`):

```
dof | act_gain | act_bias[0] | act_bias[1] | act_bias[2] | force_range
  0 | 4500.000 |       0.000 |   -4500.000 |    -450.000 | [-87.00, 87.00]
 ...
  6 | 2000.000 |       0.000 |   -2000.000 |    -200.000 | [-12.00, 12.00]
  7 |    0.016 |       0.000 |    -100.000 |     -10.000 | [-100.00, 100.00]  <-- finger
  8 |    0.016 |       0.000 |    -100.000 |     -10.000 | [-100.00, 100.00]  <-- finger
```

So you are right that `ctrl_pos` is not ignored — it enters through `act_gain`, exactly as `func_torque_and_passive_force` computes. What is actually happening on these DOFs is that `act_gain` is **0.0157 against an `act_bias[1]` of -100**: commanding a 40 mm move contributes ~0.0006 N against a ~100 N/rad restoring term. The target is applied and has no authority, which is why the joint visibly does not move and nothing is reported.

I've rewritten the PR around that measurable condition instead of PD-reducibility:

- **New predicate** — warn when the commanded coefficient (`act_gain` for position, `act_bias[2]` for velocity) is below 1% of the actuator's own position/damping terms. Your `general_actuator` fixture has gain 20 against bias 100, a ratio of 0.2, so it stays quiet and the existing consistency test is unaffected; the Franka fingers sit at 1.6e-4 and warn. There is now a test in **both** directions.
- **Offenders filtered to the requested DOFs**, so a call touching only ordinary PD joints in a mixed model says nothing.
- **Batched `dofs_info` reduced across environments** rather than sampling environment 0, so a DOF that is weak in any environment is still reported.
- **`control_dofs_position_velocity` wired in** alongside the separate entry points.

The one-time flag is now only consumed when a warning is actually emitted, so an early harmless call can't swallow the diagnostic for a later problematic one.

Happy to drop the ratio threshold in favour of anything you prefer, or to convert this into a documentation note instead if you'd rather the API stay silent — the underlying trap (a tendon approximated into a joint actuator with negligible position gain) is what I'd like to see surfaced somehow, since it costs a debugging session to find.

duburcqa:
Too many issues with your suggested patch. We will analyse this issue internally and decide the next step.

PR #3198, opened 2026-08-09
[BUG FIX] Differential propeller thrust now rolls and pitches drone entities, enabling horizontal flight.

XXLiu-HNU:
## Problem

Differential propeller RPM never produces roll or pitch on a `Drone` entity. The propeller forces are applied at
each propeller link's center of mass, and the propeller links in the shipped drone URDFs are massless, so their
inertial offset (`i_pos`) degenerates to zero and the thrust moment `i_pos x F` vanishes. The drone can only climb
and yaw - it can never tilt or translate horizontally: `fly_route.py` never reaches its waypoints, and the random
x/y targets of the `hover_train.py` environment are unreachable.

Evidence: with a zero-yaw differential RPM pattern, the roll angular velocity stays exactly `0.0` over 100 steps,
while the same pattern produces a net roll torque of ~1e-3 N.m by the kernel's own arithmetic (`i_pos x F`, with
`i_pos` the +/-0.028 m arm authored in the URDF inertial origin - which the parser drops for zero-mass links).

## Fix

`kernel_set_drone_rpm` now accumulates each propeller thrust and its moment about the drone's center of mass,
using the propeller body-frame position passed from `DroneEntity` (taken from the propeller visual geometry, the
only place the arm survives in the loaded model). The net wrench is rotated by the base link's world orientation
so the thrust follows the drone's attitude, and applied to the base link: external forces and torques applied to
the massless propeller links do not carry the thrust moment into the composite body.

Behavior change: propeller thrust and rotor torque now follow the drone's attitude (previously always world-vertical).
This is required for horizontal translation once the body tilts, and leaves uniform-RPM vertical flight unchanged.

## Verification

- New regression test `test_drone_differential_thrust_rolls_body` fails on `main` (roll angular velocity exactly 0)
  and passes with the fix.
- Existing `test_drone_propellers_force_substep_consistency` still passes (uniform RPM: zero net moment).
- `fly_route.py` now flies toward its waypoints (previously the x/y position never changed).

### DIFF
--- a/genesis/engine/entities/drone_entity.py
+++ b/genesis/engine/entities/drone_entity.py
@@ -1,6 +1,7 @@
 import os
 import xml.etree.ElementTree as ET
 
+import numpy as np
 import torch
 
 import genesis as gs
@@ -27,6 +28,9 @@ class DroneEntity(RigidEntity):
         try:
             self._propellers_vgeom_idxs = torch.tensor(
                 [link.vgeoms[0].idx for link in propellers_link], dtype=gs.tc_int, device=gs.device
+            )
+            self._propellers_pos = torch.tensor(
+                np.array([link.vgeoms[0].init_pos for link in propellers_link]), dtype=gs.tc_float, device=gs.device
             )
             self._animate_propellers = True
         except Exception:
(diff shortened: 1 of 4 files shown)


duburcqa:
@XXLiu-HNU are you planning the address the bot comments?

PR #3200, opened 2026-08-10
#3199 [BUG FIX] Isolate CoACD in a subprocess on HIP to avoid dual-OpenMP SIGSEGV

### LINKED ISSUE #3199: CoACD SIGSEGV on gs.amdgpu: dual OpenMP (ROCm libgomp vs CoACD vendored libgomp)
## Summary

On ROCm/HIP (`gs.amdgpu`), enabling CoACD convex decomposition (`convexify=True`, low `decompose_object_error_threshold` / `decompose_robot_error_threshold`) segfaults during `"Running convex decomposition."` after Genesis has initialized. Sync path works on CPU; the crash is specific to processes that have already loaded HIP/ROCm torch.

## Environment

- AMD Rembrandt APU (`gfx90c`), Debian 13
- `torch==2.12.0+rocm7.15.*` (TheRock nightlies / device-gfx90c)
- Genesis 1.3.2, `coacd==1.0.11`

## Repro

```python
import genesis as gs

gs.init(backend=gs.amdgpu)
scene.add_entity(
    gs.morphs.Mesh(
        file="meshes/duck.obj",
        scale=0.1,
        convexify=True,
        decompose_object_error_threshold=0.0,
    ),
)
scene.build()  # SIGSEGV during CoACD
```

Minimal isolation:

```text
coacd.run_coacd alone                         → OK
torch HIP alloc, then coacd.run_coacd in-proc → SIGSEGV (exit 139)
torch HIP alloc, then coacd in subprocess     → OK
```

`/proc/self/maps` after HIP + importing CoACD shows **two** OpenMP runtimes:

- system `libgomp.so.1` (pulled by ROCm/HIP torch)
- CoACD wheel’s vendored `libgomp-a34b3233` (`coacd.libs/`)

## Root cause

CoACD is CPU/OpenMP-only and does not use the GPU. The hang/crash is a classic dual-OpenMP conflict: two `libgomp` copies in one process corrupt OpenMP locks inside CoACD’s parallel regions ([SarahWeiii/CoACD#104](https://github.com/SarahWeiii/CoACD/pull/104)).

## Expected

`convexify=True` / CoACD should work after `gs.init(backend=gs.amdgpu)` without segfaulting.

## Proposed fix

When HIP is present (or more than one OpenMP runtime is already mapped), run `coacd.run_coacd` in a fresh Python subprocess that does not import torch/Genesis, then reuse the existing `.cvx` cache so warm loads stay cheap.

alexis779:
## Summary

- Fixes #3199: CoACD convex decomposition SIGSEGVs after `gs.init(backend=gs.amdgpu)` because ROCm/HIP torch loads system `libgomp` while the CoACD wheel vendors another — dual OpenMP in one process ([SarahWeiii/CoACD#104](https://github.com/SarahWeiii/CoACD/pull/104)).
- When HIP is present (or more than one OpenMP runtime is already mapped), run `coacd.run_coacd` in a fresh interpreter via `genesis/utils/_coacd_worker.py` (no torch/Genesis imports). CPU/CUDA keep the in-process path unless a conflict is detected.
- Existing `.cvx` cache still applies; warm loads do not pay for the subprocess.

## Test plan

- [x] Minimal: HIP torch alloc then `mu.convex_decompose(duck)` → OK (was SIGSEGV)
- [x] `pytest tests/parsers/test_mesh.py::test_coacd_survives_amdgpu_openmp_isolation -n0` → PASSED
- [x] ArmForge SO-101 replay with `convexify=True` / `decompose_robot_error_threshold=0.0` on `gs.amdgpu` → CoACD completes (no segfault)
- [ ] CI: mesh/CoACD tests on CPU (in-process path unchanged)
- [ ] Confirm CUDA hosts still use in-process CoACD when only one OpenMP is mapped

### DIFF
--- a/genesis/utils/_coacd_worker.py
+++ b/genesis/utils/_coacd_worker.py
@@ -1,0 +1,51 @@
+"""Isolated CoACD entrypoint.
+
+Must not import ``torch`` or ``genesis``: ROCm/HIP torch loads system ``libgomp`` into the
+parent process while the CoACD wheel vendors its own ``libgomp``. Two OpenMP runtimes in one
+process SIGSEGV inside CoACD's parallel regions (see SarahWeiii/CoACD#104). This module is
+executed in a fresh interpreter via ``subprocess`` so only CoACD's OpenMP is mapped.
+"""
+
+from __future__ import annotations
+
+import json
+import sys
+
+
+def main(argv: list[str] | None = None) -> int:
+    argv = list(sys.argv[1:] if argv is None else argv)
+    if len(argv) != 3:
+        print("usage: _coacd_worker.py IN.npz OUT.npz KWARGS.json", file=sys.stderr)
+        return 2
+
+    in_path, out_path, kwargs_path = argv
+
+    # Local imports keep module import side-effect free for discovery tools.
+    import coacd
+    import numpy as np
+
+    data = np.load(in_path)
+    vertices = np.ascontiguousarray(data["vertices"], dtype=np.float64)
+    faces = np.ascontiguousarray(data["faces"], dtype=np.int32)
+    with open(kwargs_path, "r", encoding="utf-8") as f:
+        kwargs = json.load(f)
+
+    result = coacd.run_coacd(coacd.Mesh(vertices, faces), **kwargs)
+
+    # Variable-length hull list → save as object arrays of contiguous buffers.
+    parts_v = []
+    parts_f = []
+    for vs, fs in result:
+        parts_v.append(np.ascontiguousarray(vs, dtype=np.float64))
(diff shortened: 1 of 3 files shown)


alexis779:
Addressed the Codex review: with \`backend=None\` the autouse fixture never applies \`@pytest.mark.cache(False)\`, so the test now sets a fresh \`GS_CACHE_FILE_PATH\` / QD cache env before \`gs.init\` and asserts \`_run_coacd_subprocess\` was called at least once.

duburcqa:
Please feel a bug report on CoACD directly. We are not going to spawn an isolated subprocess just to accommodate a bug in some dependency(ies).

Issue #841, opened 2025-03-10
[Bug]: Running "uv add genesis-world==0.2.1" on linux fails


lcaquot94:
### Bug Description

 Running command `uv add genesis-world==0.2.1` on linux fails with error :

`FileNotFoundError: [Errno 2] No such file or directory: '/usr/bin/sw_vers'`

Looks like setuptools tries to install pyobjc-framework-cocoa despite it's a MacOS librairy.
I don't know how to ignore such a bug during installation.

### Steps to Reproduce

Create blank python project using uv, then run command `uv add genesis-world==0.2.1`.

### Expected Behavior

genesis-world should install successfully but unfortunately it does not.

### Relevant log output

```shell

```

### Environment

- OS: Ubuntu 22.04

### Release version or Commit ID

0.2.1

bxtbold:
The command seems to be working on Ubuntu 22.04. I wasn't able to reproduce the error.

PR #3202, opened 2026-08-11
[IGNORE] Control run: quadrants 1.3.0b1 on the #3201 base

hughperkins:
## Description

Control run for the `test_align_mesh` failure on
[#3201](https://github.com/Genesis-Embodied-AI/genesis-world/pull/3201). Same
base commit (`8e70e94`), same test code; the only difference from #3201 is
`quadrants==1.3.0b1` instead of `1.3.0b2`. Not intended to merge.

## Motivation and Context

On #3201, `tests/rigid/test_asset_loading.py::test_align_mesh` fails
deterministically: both CI attempts produced bit-identical residual angular
velocity `[5.53e-02, 2.96e-03, 1.07e-01]` against `tol=0.05`, as the only
failure out of 938 tests.

The base commit is already known good. genesis-world main at `8e70e94` went
green on all four Production jobs with `quadrants==1.2.0`, so the regression is
in quadrants somewhere between 1.2.0 and 1.3.0b2 rather than in genesis-world.

[#3190](https://github.com/Genesis-Embodied-AI/genesis-world/pull/3190) already
tried 1.3.0b1 and was green, but it is not a clean control: it branches seven
commits earlier, before three rigid-body fixes that bear on this test (#3196
degenerate contacts, #3194 consistency across scales — the mango is
`scale=0.045`, #3184 authored inertia), and `tests/rigid/test_asset_loading.py`
itself changed by +87/-19 in that window, so it ran a different
`test_align_mesh`.

This PR holds all of that fixed and varies only the quadrants version.

Read the Production `unit_tests-ndarray` job:

- **Green** &rarr; the regression is in `1.3.0b1..1.3.0b2`. On CUDA that delta
  contains nothing reachable (the #841 partial-write fix is Metal/Vulkan-only,
  Metal atomics are Metal-only, the rest is docs, tests and a nanobind pin), so
  this outcome would be surprising and would need a closer look.
- **Red** &rarr; the regression is in `1.2.0..1.3.0b1`, where
  [quadrants#790](https://github.com/Genesis-Embodied-AI/quadrants/pull/790)
  (CSE per offload rather than whole kernel) is the only backend-neutral
  functional change, and #3190 was green only because it ran the older test.
  This also clears everything in the b1&rarr;b2 delta.

## Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ readme = "README.md"
 requires-python = ">=3.10,<3.14"
 dependencies = [
     "psutil",
-    "quadrants==1.2.0",
+    "quadrants==1.3.0b1",
     "pydantic>=2.11.0",
     "numpy>=1.26.4",
     "frozendict",


hughperkins:
Result: **red**, so by this PR's own decision rule the window narrows to `1.2.0..1.3.0b1`. Follow-up work says the conclusion drawn from that rule does not hold, though, and quadrants is not implicated:

- Rerunning this PR fails again on `rtx-209-195`, a different node from the `rtx-209-113` of both original failures, with the residual `[5.527567e-02, 2.961722e-03, 1.065692e-01]` bit-identical. Deterministic, and not node-specific.
- Bumping `CACHE_VERSION` to 2 and changing nothing else turns this same tree green (#3204, 938 passed).
- CI's assets and their decomposition match a clean machine exactly (#3205).
- Source builds of quadrants either side of quadrants#790 (the only commit in the window touching code the CUDA backend runs) give bit-identical residuals on this scene, as do the 1.2.0 and 1.3.0b1 wheels.
- On the cluster the residual over 20 runs from an empty asset cache spans `8.2e-06` to `1.2e-04` against `tol=0.05`; CI's `1.07e-01` is ~900x above the worst of them, so it is a state a fresh run does not reach rather than a tail of that spread.

Closing to keep the exploration PRs down to one; #3206 carries the remaining control (1.2.0 against today's warm cache).

PR #3205, opened 2026-08-11
[IGNORE] Diagnostic run: dump CI's processed collision geometry for test_align_mesh

hughperkins:
## Description

Diagnostic run for the `test_align_mesh` failure on
[#3201](https://github.com/Genesis-Embodied-AI/genesis-world/pull/3201) /
[#3202](https://github.com/Genesis-Embodied-AI/genesis-world/pull/3202). Same
tree as #3202 (base `8e70e94`, `quadrants==1.3.0b1`, `CACHE_VERSION` left at
`"1"` so the runner caches stay warm), plus a probe that prints the source asset
hashes and the processed collision geometry. The production job is narrowed to
the probe plus `test_align_mesh` itself. Not intended to merge.

## Motivation and Context

CI's failing tensor is:

```
[[ 5.527567e-02,  2.961722e-03,  1.065692e-01],
 [ 1.627962e-06, -4.872959e-09,  4.313075e-06]]
```

Running the same test on the cluster at the same base commit with the same
quadrants wheel, the env-1 row comes out bit-identical and env 0 does not: I
measure `2.32e-05` where CI gets `1.07e-01`, against `tol=0.05`. The scene is
built with `n_envs=2` and the two environments differ only in which variant of
the heterogeneous entity they hold, the bowl in env 0 and the mango in env 1, so
the divergence is on the side that owns `orange_plastic_bowl.glb`.

On a clean machine that bowl decomposes into 21 convex pieces (`het_obj` has 22
geoms, the 22nd being the mango variant), from a `orange_plastic_bowl.glb` of
117808 bytes, sha256 `da908562...`. This run reports the same numbers from CI.

Worth noting that the quadrants version is not implicated by anything I can
measure. Between 1.2.0 and 1.3.0b1 the only commit touching code the CUDA
backend runs is quadrants#790 (CSE per offload), and source builds either side
of it produce bit-identical residuals on this scene, as do the 1.2.0 and
1.3.0b1 wheels themselves.

Read the `GEOM` lines in the Production `unit_tests-ndarray` job log:

- **Different asset hash or piece count from the numbers above** -> CI is
  simulating a different bowl, and the failure is about the cached asset rather
  than about quadrants.
- **Same geometry** -> the geometry is exonerated and the divergence is
  somewhere else in the runner environment.

Whether `test_align_mesh` still fails when it is the only test in the job also
says whether the failure needs the rest of the suite around it.

## Checklist:

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -108,7 +108,7 @@ jobs:
             # tmate -S /tmp/tmate.sock wait tmate-ready
             # tmate -S /tmp/tmate.sock display -p '#{tmate_ssh}'
 
-            pytest -v -ra --backend gpu --dev --forked ./tests
+            pytest -v -ra -s -n 0 --backend gpu --dev tests/rigid/test_align_probe3.py tests/rigid/test_asset_loading.py::test_align_mesh
 
             # tmate -S /tmp/tmate.sock wait tmate-exit
           EOF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ readme = "README.md"
 requires-python = ">=3.10,<3.14"
 dependencies = [
     "psutil",
-    "quadrants==1.2.0",
+    "quadrants==1.3.0b1",
     "pydantic>=2.11.0",
     "numpy>=1.26.4",
     "frozendict",
--- a/tests/rigid/test_align_probe3.py
+++ b/tests/rigid/test_align_probe3.py
@@ -1,0 +1,87 @@
+"""Geometry diagnostics for the test_align_mesh scene.
+
+Env 0 and env 1 differ only in which variant of the heterogeneous entity they hold: env 0 the bowl, env 1 the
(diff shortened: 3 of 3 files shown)


hughperkins:
Result: **green**, and the geometry is not the difference.

CI reports the same source assets as a clean machine, `orange_plastic_bowl.glb` at 117808 bytes sha256 `da9085621bdd0d8137852db88e54aaf6e56091fed9d6b42fcccccc656d3bbc45` and `mango.glb` at 10799704 bytes sha256 `b903c233d857ced6ecd48fe1429efd0cc6902412e661ddf0af5f6a1bfcb2b03f`, and the same piece counts: `mango n_geoms=1`, `het_obj n_geoms=22`.

`test_align_mesh` also passed here (159.99s) on the warm version-1 cache, but this arm dropped xdist and `--forked` along with the rest of the suite, so it is not a clean "warm cache alone" control. Closing; the remaining control is #3206.

PR #3218, opened 2026-08-12
[BUG FIX] Fix minor physics discrepancy between generic and GPU-only constraint solver implementations.

duburcqa:
## Description

The two registered constraint-solve implementations must agree, since the dispatcher may pick either one on any step. Their linesearches did not, in three ways, all aligned here on the monolith:

- **Tolerance.** #3194 made the linesearch bound gradient-scaled and floored at the working precision, in `func_linesearch_batch` only, so the arm that runs on GPU kept the mass-scaled one - `4.456e-02` against `7.463e-07` on the same state. The gradient norm now rides the existing tree reduction.
- **Trial point.** `|grad/hess|`, abandoning the step on a non-positive curvature, becomes the signed `-grad/hess` with the curvature floored.
- **Accepted step.** The refinement's result is taken unconditionally, a stalled bracket rejects the step, and `ls_improvement` is written whatever its sign - the termination test reads a negative improvement as noise around the optimum and was being denied those values.

`ls_alpha_newton` becomes `ls_p0_deriv`, the derivative pair the refinement derives its trial step from and falls back on, which also drops a cooperative evaluation from that fallback.

## Motivation and Context

Quoted against the mean inertia, the bound tracks the model's mass rather than the gradient it bounds - the scale-dependence #3194 removed from the reference arm and not from the GPU one.

`test_dynamics.py::test_solve_arm_equivalence` (GPU only), on one scene carrying islands of one, three and nine bodies: the dispatcher's `__call__` is replaced to select by index among the candidates it deems eligible, and the inputs a solve mutates are restored between passes so both implementations solve the same problem. Per step the accelerations must agree within `tol` and differ by more than zero, since exactly zero means the selection stopped working.

Per solve from identical inputs, CUDA, 12 steps, 2 envs:

| | `qacc` | `ls_alpha` | `ls_it` mismatches |
|---|---|---|---|
| merge base, fp64 | 4.707e-14 | 6.550e-14 | 0 |
| this branch, fp64 | 4.707e-14 | 6.550e-14 | 0 |
| merge base, fp32 | 1.878e-06 | 7.838e-05 | 16 |
| this branch, fp32 | 1.878e-06 | 7.838e-05 | 8 |

fp64 agrees on the accepted step, the evaluation count and the acceleration before and after; in fp32 the acceleration and the step are identical between trees and only the evaluation count moves, so those count differences are rounding tipping refinement decisions rather than the threshold formula. A stack, a 3x3 pile, the elliptic cone at its Coulomb limit and scene scales from 0.02 to 100 agree to rounding on both trees too. The test passes on the merge base, so it locks the parity in rather than witnessing a regression: this change aligns the implementations in code, with no observable difference in the solve output on any scene tested.

## Checklist:

### DIFF
--- a/genesis/engine/solvers/rigid/constraint/solver.py
+++ b/genesis/engine/solvers/rigid/constraint/solver.py
@@ -173,7 +173,7 @@ class ConstraintSolver:
 
         self.ls_alpha = cs.ls_alpha
         self.ls_improvement = cs.ls_improvement
-        self.ls_alpha_newton = cs.ls_alpha_newton
+        self.ls_p0_deriv = cs.ls_p0_deriv
         self.ls_gtol = cs.ls_gtol
         self.ls_it = cs.ls_it
         self.ls_result = cs.ls_result
--- a/genesis/engine/solvers/rigid/constraint/solver_breakdown.py
+++ b/genesis/engine/solvers/rigid/constraint/solver_breakdown.py
@@ -133,9 +133,12 @@ def _func_decomp_linesearch_p0(
             grad_sqnorm = sh_grad_sqnorm[0]
 
             if snorm < rigid_info.EPS[None]:
-                # Converged — only thread 0 writes
+                # A search direction that has gone to zero leaves nothing to step along, so the refinement is bypassed
+                # and this env is done - status 1, as the monolith linesearch reports it.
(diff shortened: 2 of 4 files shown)


duburcqa:
@codex review

PR #3221, opened 2026-08-12
[FEATURE] Reproduce a GPU simulation exactly from one run to the next.

duburcqa:
## Description

Adds `use_deterministic_algorithms` to `gs.init`, off by default. When enabled, the constraint solve implementation is settled at build time instead of being measured at runtime, so a rollout replays identically: reset the scene, feed it the same inputs, and every step reproduces the previous run bit-for-bit.

On GPU the solve runs through whichever of two implementations the autotuner times as fastest for the scene, re-measuring as the simulation goes. The two do not compute bit-identical results, so which one is in effect at a given step becomes part of the trajectory - and since the choice comes from a wall-clock measurement, it follows how busy the machine is rather than the simulation. The flag resolves that choice to the decomposed implementation wherever it was still open, leaving untouched every case that is already pinned (CPU, differentiable mode, and the configurations where the monolith implementation is the known winner).

The test suite now runs deterministic by default, which replaces the `prefer_decomposed_solver` monkeypatch that `conftest.py` installed for every non-benchmark test: the flag reproduces exactly what that pin produced, decomposed on GPU and monolith on CPU. Benchmarks and the two tests that drive dispatching themselves opt out through `@pytest.mark.use_deterministic_algorithms(False)`, so the choice stays explicit at each site rather than implicit in the fixture.

## Motivation and Context

Comparing two runs, bisecting a divergence or reproducing a report currently requires the machine to stay idle enough for the autotuner to reach the same verdict twice, which is not something a user can arrange. This gives them a switch that removes the measurement from the trajectory altogether.

It is opt-in because the guarantee costs throughput, by a scene-dependent amount: nothing where the decomposed implementation was going to win anyway, and up to the gap between the two implementations where the other one would have been picked.

`test_gpu_simulation_determinism` gains a third case that pins nothing and runs under the flag, next to the two existing cases that pin each implementation by hand. It resets the tower scene eight times and requires every trial to reproduce the first bit-for-bit, comparing contact set, contact order, dof velocities and positions in pipeline order.

## Checklist:

- [ ] I updated the [documentation](https://github.com/Genesis-Embodied-AI/genesis-doc) accordingly or no change is needed. A section for the new mode is written and will follow as a separate PR on the documentation repository.

### DIFF
--- a/genesis/__init__.py
+++ b/genesis/__init__.py
@@ -49,6 +49,7 @@ device: torch.device | None = None
 backend: _gs_backend | None = None
 use_ndarray: bool | None = None
 use_zerocopy: bool | None = None
+use_deterministic_algorithms: bool | None = None
 EPS: float | None = None
 
 
@@ -64,6 +65,7 @@ def init(
     theme="dark",
     logger_verbose_time=False,
     performance_mode=False,
+    use_deterministic_algorithms=False,
 ):
     global _initialized
     if _initialized:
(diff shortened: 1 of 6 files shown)


duburcqa:
Superseded by #3222, opened from a correctly named branch.

PR #3231, opened 2026-08-13
[MISC] Simplify benchmarks.

hughperkins:


duburcqa:
Could we rather move to the "new" benchmark suite instead of just removing benchmarks?

hughperkins:
> Could we rather move to the "new" benchmark suite instead of just removing benchmarks?

Yes. On it.

hughperkins:
I can close this in the meantime

PR #3244, opened 2026-08-18
[MISC] Add regression test for friction hold cross-talk.

Milotrince:
## Description

The constraint solver's convergence thresholds must stay independent of what is applied elsewhere in the environment. Quoting them against the free acceleration summed over every dof lets one hard-pressing body relax the exit criterion for every contact around it, so an untouched body's friction hold degrades while the press lasts. Nothing in the suite covered that: the friction scenes either load every body they check, or hold their contacts exactly enough that a threshold change never reaches the trajectory.

## Related Issue
Bug introduced in https://github.com/Genesis-Embodied-AI/genesis-world/pull/3194, fixed in https://github.com/Genesis-Embodied-AI/genesis-world/pull/3158

## Motivation and Context
Although the issue itself has been fixed, the current tests don't catch a regression, hence this PR.

`pytest tests/rigid/test_friction.py -k unaffected_by_press`

What is happening: **Left** side of each video shows red box being held up by friction, blue box resting. No problem here. In the **right** side of each video shows when the blue box is being pushed into the table. Before the fix, the underconvergence issue causes the red box to slip. 

https://github.com/user-attachments/assets/197f9efa-6309-4e97-8c17-1d40d5ac162d

https://github.com/user-attachments/assets/d4843a1c-2a08-4f98-9eb3-1bc345f95545

## Checklist:

Resolves SIM-361

### DIFF
--- a/tests/rigid/test_friction.py
+++ b/tests/rigid/test_friction.py
@@ -348,6 +348,91 @@ def test_static_friction(mode, friction, n_boxes, solver, scale, mesh_boxes, sho
 
 
 @pytest.mark.required
+def test_static_hold_unaffected_by_press_on_separate_body(show_viewer):
+    # The convergence thresholds must stay independent of the forces applied elsewhere in the environment, so a box
+    # held against a wall by friction keeps its hold whether or not a body half a metre away is pressed.
+    GRAVITY = -9.81
+    BOX = 0.06
+    WALL = (0.02, 0.4, 0.5)
+    FRICTION = 0.25
+    SAFETY = 1.25  # push over the Coulomb minimum; tilt-balanced while SAFETY >= 2 * FRICTION
+    NOSLIP_ITERATIONS = 20  # leaves the hold resting on a converged solve rather than on regularized creep
+    PRESS_SIZE = 0.3
+    PRESS_KP = 1.0e5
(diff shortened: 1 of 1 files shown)


duburcqa:
@Milotrince Can you share a video before and after?

Milotrince:
> Can you share a video before and after?

https://github.com/user-attachments/assets/9dbf27f4-a5df-41b9-9bff-24d92d9d0141

I added the video. It is not visible by naked eye but the effect is measurable.

duburcqa:
> I added the video. It is not visible by naked eye but the effect is measurable.

Your video is broken or it is exactly the same before and after? Besides, if it is not visible, then we don't care in this context where we are hunting practical failures. But from what I see, it is VERY visible in your video...

Milotrince:
> But from what I see, it is VERY visible in your video...

The left box is being pushed intentionally. The right box is what we care about measuring.

> We are hunting practical failures here. 

This test case captures the failure, but it is at a smaller scale. I didn't want to pull the assets we saw the issues with, so this is using box. I will see if I can get a more obvious test scene, since this test as it is now is a bit confusing.

Milotrince:
PR description is now updated with new video!

duburcqa:
Excellent! The new video is neat!

note that uploading assets on HuggingFace is just fine. Finding a way to avoid it is even better, but usually tends to require lot more effort!

duburcqa (approved):

PR #3245, opened 2026-08-18
[MISC] Add cylinder resting contact test.

Milotrince:
## Description

A cylinder settled on its circular end has its contact patch rebuilt somewhere else on the rim every step: the points travel 15 to 23 mm, while the cylinder's own origin moves under 0.2 um. Tipped onto its curved side the same body touches along a line and its patch wanders by a few tenths of a millimeter, so the instability belongs to the flat circular face rather than to the shape, the solver settings or the settling. The bound scales with the radius because the distance a contact can jump is set by the size of the patch it is drawn from.

## Related Issue
object jitter

## Motivation and Context
The object jitter is exacerbated on objects with more subdivision/vertices since we don't have persistent contact manifold. [PhysX](https://nvidia-omniverse.github.io/PhysX/physx/5.1.2/docs/AdvancedCollisionDetection.html#persistent-contact-manifold-pcm) and [Bullet](https://pybullet.org/Bullet/BulletFull/classbtPersistentManifold.html) has this. We currently have a `ContactCache` but it is not doing this.

| Case | Contact patch moved | Bound | Cylinder moved | Result |
| - | - | - | - | - |
| `radius=0.02`, resting on end | 15.211 mm | 1.00 mm | 0.192 µm | ❌ fail |
| `radius=0.03`, resting on end | 22.558 mm | 1.50 mm | 0.139 µm | ❌ fail |
| `radius=0.02`, resting on side | 0.001–0.52 mm | 1.00 mm | < 0.2 µm | ✅ pass |
| `radius=0.03`, resting on side | 0.001–0.68 mm | 1.50 mm | < 0.2 µm | ✅ pass |

https://github.com/user-attachments/assets/bc285bd9-915a-4335-8452-b311d29d26cc

## Checklist:

Partially resolves SIM-373

### DIFF
--- a/tests/rigid/test_collision.py
+++ b/tests/rigid/test_collision.py
@@ -35,6 +35,35 @@ def _ellipsoid_mjcf_path(tmp_path, semi_axes):
     return str(path)
 
 
+def _resting_contact_motion(scene, entity, support, num_steps):
+    """Step the scene, returning the furthest a contact point and the body each moved in one step, in meters.
+
+    Contacts carry no identity across steps, so each is matched to the nearest contact of the previous step.
+    """
+
+    def patch():
+        contacts = entity.get_contacts(with_entity=support)
+        return contacts["position"], contacts["valid_mask"]
+
+    prev_patch, prev_valid = patch()
+    prev_pos = entity.get_pos()
+    contact_move = body_move = 0.0
+    for _ in range(num_steps):
+        scene.step()
+        cur_patch, cur_valid = patch()
+        cur_pos = entity.get_pos()
+        # A step may report no contact at all, leaving nothing to match against.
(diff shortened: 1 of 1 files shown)


duburcqa:
@trns1997 Instead of closing, what about modifying the test to check whether contact path is working as expected, ie providing a smooth contact manifold over time?

trns1997:
> @trns1997 Instead of closing, what about modifying the test to check whether contact path is working as expected, ie providing a smooth contact manifold over time?

@duburcqa wrong ping. @Milotrince

duburcqa:
@Milotrince Instead of closing, what about modifying the test to check whether contact path is working as expected, ie providing a smooth contact manifold over time?

PR #3247, opened 2026-08-18
[BUG FIX] Support broadcasting inverse kinematics targets in batched scenes.

### LINKED ISSUE #3246: [Bug]: inverse_kinematics rejects the documented shared target form in batched scenes.
### Bug Description

`inverse_kinematics` documents `pos` as shape `(3,)` and `quat` as shape `(4,)`. In batched scenes the wrapper rejects exactly that form: it validates `len(pos) != len(envs_idx)` ([rigid_entity.py#L2672-L2680](https://github.com/Genesis-Embodied-AI/genesis-world/blob/0796d276/genesis/engine/entities/rigid_entity/rigid_entity.py#L2672-L2680)), which treats the 3 coordinates of a shared target as if they were per-environment rows. A shared position only slips through with exactly 3 selected environments, and a shared quaternion with exactly 4. The raised message is also stale: it says the first dimension must equal `scene.n_envs` while the code compares against `len(envs_idx)`.

The checks are redundant: `inverse_kinematics_multilink` directly underneath broadcasts each target with `broadcast_tensor(pos, gs.tc_float, (len(envs_idx), 3), ...)` ([rigid_entity.py#L2828](https://github.com/Genesis-Embodied-AI/genesis-world/blob/0796d276/genesis/engine/entities/rigid_entity/rigid_entity.py#L2828)), so a single target shared by the selected environments is already supported by the underlying implementation. Shapes incompatible with the supported broadcasting contract raise there with a shape-specific message.

Calling `inverse_kinematics_multilink` directly with the shared position succeeds and returns the same `(2, 9)` solution as passing a tiled `(2, 3)` target through `inverse_kinematics` (`max_abs_diff == 0.0`).

There is currently no test that passes the shared target form to `inverse_kinematics` in a batched scene. 

### Steps to Reproduce

```python
import genesis as gs

robot = scene.add_entity(
    morph=gs.morphs.MJCF(
        file="xml/franka_emika_panda/panda.xml",
    ),
)
scene.build(n_envs=2)

qpos = robot.inverse_kinematics(link=robot.get_link("hand"), pos=(0.5, 0.2, 0.4))
print(tuple(qpos.shape))
```

### Expected Behavior

The documented shared form should be accepted and broadcast to every selected environment. In the example above, `pos=(0.5, 0.2, 0.4)` should return a `(2, 9)` qpos identical to passing the same target tiled per environment. Per-environment targets of shape `(len(envs_idx), 3)` should keep working unchanged.

### Screenshots/Videos

_No response_

### Relevant log output

```shell
$ python repro.py
[Genesis] [ERROR] GenesisException: First dimension of `pos` must be equal to `scene.n_envs`.
```

### Environment

- OS: macOS 15.7.3
- GPU/CPU: Apple M4 (repro runs on the CPU backend)
- GPU-driver version: N/A
- Python 3.11.14, genesis-world 1.3.3

0796d276

### Additional Context

_No response_

jeetrex17:
## Description

`inverse_kinematics` now accepts the documented shared target form in batched scenes: a single `pos` of shape
`(3,)` or `quat` of shape `(4,)` broadcasts across the selected environments, exactly as `inverse_kinematics_multilink` underneath already supports. The wrapper's own length pre-checks are removed since the underlying broadcasting validates shapes with an accurate message, and the docstring documents both the shared and per-environment target shapes.

## Related Issue

Resolves #3246

## Motivation and Context

In batched scenes the wrapper compared `len(pos)` against `len(envs_idx)`, treating the 3 coordinates of a shared target as per-environment rows. The documented shared form raised a `GenesisException` everywhere except the coincidental cases (3 selected environments for `pos`, 4 for `quat`), and the message referenced `scene.n_envs` while the code compared against `len(envs_idx)`.

The batched inverse kinematics test now solves a single shared target and asserts the solution is identical to passing the same target expanded per environment. The shared-target case raises on `main` and passes with the fix.

```bash
pytest tests/rigid/test_kinematics.py::test_inverse_kinematics_local_point -n 0 --backend cpu -q
```

Result: `2 passed` on macOS 15.7.3 with an Apple M4 and Python 3.11.14. The full test matrix is left to CI.

## Checklist:

### DIFF
--- a/tests/rigid/test_kinematics.py
+++ b/tests/rigid/test_kinematics.py
@@ -425,7 +425,7 @@ def test_inverse_kinematics_local_point(n_envs, show_viewer, tol):
 
     qpos = qpos_solutions[0]
     if n_envs > 0:
-        assert_equal(qpos_solutions[1], qpos_solutions[1][0].expand_as(qpos_solutions[1]))
+        assert_equal(qpos_solutions[1], qpos_solutions[1][0])
         assert_allclose(qpos_solutions[1][0], qpos[0], tol=tol)
 
     # Apply the solution
--- a/tests/rigid/test_kinematics.py
+++ b/tests/rigid/test_kinematics.py
@@ -414,6 +414,30 @@ def test_inverse_kinematics_local_point(n_envs, show_viewer, tol):
     )
     assert_allclose(err, 0.0, atol=tol)
 
+    if n_envs > 0:
+        qpos_shared, err_shared = robot.inverse_kinematics(
+            link=end_effector,
+            pos=target_pos[0],
+            quat=target_quat[0],
+            local_point=local_offset,
+            pos_tol=tol,
+            rot_tol=tol,
+            max_solver_iters=100,
+            return_error=True,
+        )
+        assert_allclose(err_shared, 0.0, atol=tol)
+        qpos_expanded, _ = robot.inverse_kinematics(
+            link=end_effector,
+            pos=target_pos[0].expand(n_envs, 3),
+            quat=target_quat[0].expand(n_envs, 4),
+            local_point=local_offset,
+            pos_tol=tol,
+            rot_tol=tol,
+            max_solver_iters=100,
+            return_error=True,
+        )
+        assert_equal(qpos_shared, qpos_expanded)
+
     # Apply the solution
     robot.set_qpos(qpos)
 
--- a/tests/rigid/test_kinematics.py
+++ b/tests/rigid/test_kinematics.py
@@ -401,42 +401,32 @@ def test_inverse_kinematics_local_point(n_envs, show_viewer, tol):
         target_pos = target_pos_base[0]
         target_quat = target_quat_base[0]
 
-    # Solve IK with local_point (local_offset stays 1D - it gets broadcast internally)
-    qpos, err = robot.inverse_kinematics(
-        link=end_effector,
-        pos=target_pos,
-        quat=target_quat,
-        local_point=local_offset,
-        pos_tol=tol,
-        rot_tol=tol,
-        max_solver_iters=100,
-        return_error=True,
-    )
-    assert_allclose(err, 0.0, atol=tol)
-
+    targets = ((target_pos, target_quat),)
     if n_envs > 0:
-        qpos_shared, err_shared = robot.inverse_kinematics(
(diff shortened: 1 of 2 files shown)


tests/rigid/test_kinematics.py
```diff
@@ -414,6 +414,30 @@ def test_inverse_kinematics_local_point(n_envs, show_viewer, tol):
...
+            pos=target_pos[0].expand(n_envs, 3),
+            quat=target_quat[0].expand(n_envs, 4),
+            local_point=local_offset,
+            pos_tol=tol,
+            rot_tol=tol,
+            max_solver_iters=100,
+            return_error=True,
+        )
+        assert_equal(qpos_shared, qpos_expanded)
+
```
duburcqa, line 430: I don't think it is necessary to add this new clause. it should be possible to generalise the existing one with some for loop or something? Besides, I would rather suggest to check something like this (pseudo-code):

```
qpos_a, qpos_b = IK(target_a, target_b)
qpos_a_1, qpos_a_2 = IK(target_a)
assert_equal(qpos_a_1, qpos_a_2)
assert_allclose(qpos_a_1, qpos_a)
```

jeetrex17: Good point. I generalized the existing solve into a loop in 335d9ed3 and added the suggested comparisons. The focused test passes for both cases (2 passed).

tests/rigid/test_kinematics.py
```diff
@@ -401,18 +401,32 @@ def test_inverse_kinematics_local_point(n_envs, show_viewer, tol):
...
+            rot_tol=tol,
+            max_solver_iters=100,
+            return_error=True,
+        )
+        assert_allclose(err, 0.0, atol=tol)
+        qpos_solutions.append(qpos_solution)
+
+    qpos = qpos_solutions[0]
+    if n_envs > 0:
+        assert_equal(qpos_solutions[1], qpos_solutions[1][0].expand_as(qpos_solutions[1]))
```
duburcqa: Are you sure '.expand_as(qpos_solutions[1])' is necessary here?

jeetrex17: yah its not , removed in 0dc53435.

duburcqa (approved):

PR #3250, opened 2026-08-19
[BREAKING][FEATURE] Apply morph pose offset to velocity/acceleration getters.

Milotrince:
## Description
- Add `relative` arg to accessors, which were not applying morph `offset_pos`/`offset_quat`.

## Related Issue
Related to https://github.com/Genesis-Embodied-AI/genesis-world/pull/2934 , https://github.com/Genesis-Embodied-AI/genesis-world/pull/2592

## Motivation and Context
`get_links_vel` and `get_links_acc` had no relative argument and always report the quantity referenced to the internal link origin. Anyone reading velocity or acceleration at a sensor or tool frame declared through a morph offset therefore has to redo the reference-point shift by hand `(v = v_origin + omega x delta)`, and silently gets the wrong signal if they forget.

`tests/rigid/test_api.py`

## Checklist:

Resolves SIM-292

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -754,7 +754,7 @@ class KinematicEntity(Entity):
         return self._solver.get_links_quat(self.base_link_idx, envs_idx, relative=relative)[..., 0, :]
 
     @gs.assert_built
-    def get_vel(self, envs_idx=None):
+    def get_vel(self, envs_idx=None, *, relative=True):
         """
         Returns linear velocity of the entity's base link.
 
@@ -762,13 +762,17 @@ class KinematicEntity(Entity):
         ----------
         envs_idx : None | array_like, optional
             The indices of the environments. If None, all environments will be considered. Defaults to None.
+        relative : bool, optional
+            Whether to report the velocity of the user-frame origin, with the morph pose offset and inertial
+            alignment stripped, rather than of the link origin used by the solver. The two differ whenever the link
+            is rotating. Defaults to True.
 
         Returns
         -------
         vel : torch.Tensor, shape (3,) or (n_envs, 3)
             The linear velocity of the entity's base link.
         """
-        return self._solver.get_links_vel(self.base_link_idx, envs_idx)[..., 0, :]
+        return self._solver.get_links_vel(self.base_link_idx, envs_idx, relative=relative)[..., 0, :]
 
     @gs.assert_built
     def get_ang(self, envs_idx=None):
@@ -870,7 +874,7 @@ class KinematicEntity(Entity):
         return torch.stack((aabbs[..., 0, :].min(dim=-2).values, aabbs[..., 1, :].max(dim=-2).values), dim=-2)
 
     @gs.assert_built
-    def get_links_vel(self, links_idx_local=None, envs_idx=None):
+    def get_links_vel(self, links_idx_local=None, envs_idx=None, *, relative=True):
         """
         Returns linear velocity of all the entity's links expressed at a given reference position in world coordinates.
 
@@ -880,6 +884,10 @@ class KinematicEntity(Entity):
             The indices of the links. Defaults to None.
         envs_idx : None | array_like, optional
             The indices of the environments. If None, all environments will be considered. Defaults to None.
+        relative : bool, optional
+            Whether to report the velocity of the user-frame origin, with the morph pose offset and inertial
+            alignment stripped, rather than of the link origin used by the solver. The two differ whenever the link
+            is rotating. Defaults to True.
 
         Returns
         -------
@@ -887,7 +895,7 @@ class KinematicEntity(Entity):
             The linear velocity of all the entity's links.
         """
         links_idx = self._get_global_idx(links_idx_local, self.n_links, self._link_start, unsafe=True)
-        return self._solver.get_links_vel(links_idx, envs_idx)
+        return self._solver.get_links_vel(links_idx, envs_idx, relative=relative)
 
     @gs.assert_built
     def get_links_ang(self, links_idx_local=None, envs_idx=None):
@@ -2187,7 +2195,9 @@ class RigidEntity(KinematicEntity):
         return self._solver.get_links_pos(links_idx, envs_idx, ref=ref, relative=relative)
 
     @gs.assert_built
-    def get_links_vel(self, links_idx_local=None, envs_idx=None, *, ref: link_ref_frame = link_ref_frame.link_origin):
+    def get_links_vel(
+        self, links_idx_local=None, envs_idx=None, *, ref: link_ref_frame = link_ref_frame.link_origin, relative=True
+    ):
         """
         Returns linear velocity of all the entity's links expressed at a given reference position in world coordinates.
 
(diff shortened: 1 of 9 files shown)


Milotrince:
@codex review

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -2044,14 +2047,17 @@ def get_links_vel(self, links_idx_local=None, envs_idx=None):
             The indices of the links. Defaults to None.
         envs_idx : None | array_like, optional
             The indices of the environments. If None, all environments will be considered. Defaults to None.
+        relative : bool, optional
+            Whether to report the velocity in the user frame, with the morph pose offset and inertial alignment
+            stripped, rather than the world frame used by the solver. Rotation separates them. Defaults to True.
```
duburcqa: 'Rotation separates them.' is a weird formulation. I'm also not true. If the morph pose offset is non-zero and the angular velocity is non-zero, the linear velocity should be "translated" to a different application point, then expressed in the authored link frame (ie a pure rotation indeed).

duburcqa: Wait. it is expressed in world coordinates, so no rotation actually.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -3499,19 +3507,42 @@ def get_links_vel(self, links_idx_local=None, envs_idx=None, *, ref: link_ref_fr
         ref: gs.link_ref_frame, optional
             The reference point used to express the velocity of each link: its origin ('link_origin') or its center of
             mass ('link_COM'). Defaults to 'link_origin'.
+        relative : bool, optional
+            Whether to report the velocity in the user frame, with the morph pose offset and inertial alignment
+            stripped, rather than the world frame used by the solver; this only affects
+            ref=gs.link_ref_frame.link_origin. Defaults to True.
```
duburcqa: Avoid ';'. I don't like that. Moreover 'user frame' is ill-defined. Authored link origin would make more sense I feel? It should be reminded that this only makes a difference for "free links" (I don't remember the exact formulation, see align doc) with 'morph.align=True'.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -3499,19 +3507,42 @@ def get_links_vel(self, links_idx_local=None, envs_idx=None, *, ref: link_ref_fr
...
+
+        Parameters
+        ----------
+        links_idx_local : None | array_like
+            The indices of the links. Defaults to None.
+        envs_idx : None | array_like, optional
+            The indices of the environments. If None, all environments will be considered. Defaults to None.
+        relative : bool, optional
+            Whether to report the acceleration in the user frame, with the morph pose offset and inertial alignment
+            stripped, rather than the world frame used by the solver. Angular motion separates them. Defaults to True.
```
duburcqa: "Angular motion separates them." is not very clear. There is some Coriolis effect taking place if I remember correctly. Could be relevant to provide the formula, or just explain that it is a translation of application point.

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -1143,25 +1152,29 @@ def get_vgeoms_quat(self, vgeoms_idx=None, envs_idx=None, *, relative=False):
...
+        else:
+            _tensor, links_idx, envs_idx = self._sanitize_io_variables(
+                None, links_idx, self.n_links, "links_idx", envs_idx, (3,)
+            )
+            assert _tensor is not None
+            tensor = _tensor[None] if self.n_envs == 0 else _tensor
+            # The frame is passed as a plain int, see 'RigidSolver._sanitize_ref_frame'.
+            kernel_get_links_vel(
+                links_idx, envs_idx, tensor, self.dyn_state, self.rigid_config, ref=int(gs.link_ref_frame.link_origin)
+            )
```
duburcqa: It would be better to have 'kernel_get_links_vel' to return the velocity relative if needed. Once calling a kernel is already necessary, better not make the matter worse by also implementing part of the math in torch.

tests/rigid/test_api.py
```diff
@@ -635,6 +637,43 @@ def test_set_root_pose(batch_fixed_verts, relative, show_viewer, tol):
...
+
+    # Spinning the box, the user-frame origin orbits the link origin, so it is transported by the rigid-body law about
+    # the very displacement the position getters strip: 'v - omega x d' and 'a - alpha x d - omega x (omega x d)'.
+    posed_box.set_dofs_velocity((0.0, 0.0, 0.0, 1.0, 2.0, -3.0))
+    scene.step()
+    offset_shift = posed_box.get_pos(relative=False) - posed_box.get_pos(relative=True)
+    omega, alpha = posed_box.get_ang(), posed_box.get_links_acc_ang()[..., 0, :]
+    assert_allclose(
+        posed_box.get_vel(relative=True),
+        posed_box.get_vel(relative=False) - torch.cross(omega, offset_shift, dim=-1),
```
duburcqa, line 722: I would suggest to use simple math where the solution is trivial and can be computed in a simpler way that does not involve the general formula. But not a very strong opinion.

tests/rigid/test_api.py
```diff
@@ -635,6 +637,43 @@ def test_set_root_pose(batch_fixed_verts, relative, show_viewer, tol):
...
+        - torch.cross(omega, torch.cross(omega, offset_shift, dim=-1), dim=-1),
+        tol=tol,
+    )
+    # Neither transported term is vacuous: each exceeds the tolerance on its own.
+    assert (torch.cross(alpha, offset_shift, dim=-1).abs() > tol).any()
+    assert (torch.cross(omega, torch.cross(omega, offset_shift, dim=-1), dim=-1).abs() > tol).any()
+    with pytest.raises(AssertionError):
+        assert_allclose(posed_box.get_vel(relative=True), posed_box.get_vel(relative=False), tol=tol)
+    with pytest.raises(AssertionError):
+        assert_allclose(posed_box.get_links_acc(relative=True), posed_box.get_links_acc(relative=False), tol=tol)
```
duburcqa: This is useless no?

tests/rigid/test_api.py
```diff
@@ -635,6 +637,43 @@ def test_set_root_pose(batch_fixed_verts, relative, show_viewer, tol):
...
+    assert_allclose(
+        posed_box.get_links_acc(relative=True)[..., 0, :],
+        posed_box.get_links_acc(relative=False)[..., 0, :]
+        - torch.cross(alpha, offset_shift, dim=-1)
+        - torch.cross(omega, torch.cross(omega, offset_shift, dim=-1), dim=-1),
+        tol=tol,
+    )
+    # Neither transported term is vacuous: each exceeds the tolerance on its own.
+    assert (torch.cross(alpha, offset_shift, dim=-1).abs() > tol).any()
+    assert (torch.cross(omega, torch.cross(omega, offset_shift, dim=-1), dim=-1).abs() > tol).any()
```
duburcqa: Same. useless once you use a simpler formula where the reference value can be computed without fancy math. Typically by keeping the angular velocity orthogonal to the linear velocity, so that the formula boils down to a simple scalar product plus sum.

tests/rigid/test_api.py
```diff
@@ -635,6 +637,43 @@ def test_set_root_pose(batch_fixed_verts, relative, show_viewer, tol):
...
+    with pytest.raises(AssertionError):
+        assert_allclose(posed_box.get_vel(relative=True), posed_box.get_vel(relative=False), tol=tol)
+    with pytest.raises(AssertionError):
+        assert_allclose(posed_box.get_links_acc(relative=True), posed_box.get_links_acc(relative=False), tol=tol)
+
+    # Masking rows and columns must strip the same offset as the unmasked query.
+    assert_allclose(
+        posed_box.get_links_vel(links_idx_local=0, envs_idx=[1])[..., 0, :], posed_box.get_vel()[[1]], tol=tol
```
duburcqa: `envs_idx=1`

duburcqa: `posed_box.get_vel()[1]`

genesis/engine/solvers/rigid/rigid_solver.py
duburcqa, line 1: Could you check whether the math when 'relative=True' are complex enough for a quadrants kernel to run faster than the torch implementation? I would say we are not there yet, but it is worth checking.

PR #3251, opened 2026-08-19
[MISC] Speed up kinematic pose updates with deferred link velocities.

jeetrex17:
## Description

Kinematic pose updates defer link-velocity propagation until a consumer requests it, so repeated pose-only updates skip velocity work that nothing reads. `set_dofs_position` fuses the degree-of-freedom write, the qpos synchronization, position forward kinematics and center-of-mass propagation into one kernel. `get_links_vel`, `get_links_ang` and `set_dofs_velocity` propagate velocities when they need them.

Per-environment freshness masks let the refresh kernels skip environments that are already current in a mixed batch. Host summaries gate the uniform cases: an all-environment update writes no device mask, and the velocity mask is materialized only when a subset update, a mixed-state refresh or a checkpoint needs its exact contents. An eager velocity write on a subset uses the plain velocity kernel when every pose is current, so an already-current scene stays current for the next getter.

Batched checkpoints and trajectory records carry one flag per environment as independent copies, with zero-copy on or off.

One fix rides along: a rigid velocity write after a deferred pose update (`skip_forward=True`) refreshes the pose before propagating velocities, so the reported link velocities match the reported pose. `RigidSolver.get_links_ang` stays a direct read because rigid stepping owns velocity propagation.

## Related Issue

Supersedes #2558.

## Motivation and Context

Kinematic entities are commonly driven by pose setters while many consumers never read link velocities, so the position-only path paid for propagation nothing used. Host-side kernel launches per steady-state step: write positions then read positions, 2 on main and 1 here; write positions then read positions and velocities, 2 and 2.

Measured against main on an RTX 2080 Ti, CUDA field layout, an ANYmal-C kinematic entity with the temporary benchmark fixtures used during review, three fresh process blocks of four interleaved A/B pairs per batch size, 95% Student-t intervals over the three process means:

| Environments | Position only | Position + velocity read | Subset velocity write + read |
| ---: | ---: | ---: | ---: |
| 128 | +27.2% [+25.3, +29.1] | +0.8% [-1.3, +2.9] | -1.6% [-7.3, +4.2] |
| 4,096 | +21.2% [+16.9, +25.5] | +8.4% [+5.1, +11.7] | -2.0% [-7.9, +3.8] |
| 20,000 | +19.3% [+16.0, +22.5] | +4.6% [+1.9, +7.4] | +0.8% [-7.0, +8.5] |

At 20,000 environments that is 34.6M to 41.3M environment updates/s for position-only updates and 31.2M to 32.6M with a velocity read after every update. The subset column and the 128-environment velocity-read cell have intervals crossing zero and establish neither a gain nor a regression. The measurements come from a paired external driver on a shared GPU, not from the benchmark suite.

```bash
pytest tests/rigid/test_kinematics.py::test_setters tests/rigid/test_kinematics.py::test_link_velocity -v
pytest tests/rigid/test_serialization.py::test_pickle_resume tests/rigid/test_serialization.py::test_trajectory_replay -v
```

Run for this revision on the machine above. On CUDA ndarray: `test_setters`, all four `test_pickle_resume` variants, both `test_trajectory_replay` variants, `test_link_velocity`, the cross-solver kinematics queries, `test_raycaster::test_against_visual[True-2]` and `test_fk_grad_matches_fd`. On CUDA field with zero-copy off, and on CPU with int32 mask storage forced: `test_setters` and `test_pickle_resume[False-2]`. Metal has not been run.

## Checklist:


duburcqa:
The key is to reduce the number of independent kernel calls, or at the very least, make sure that it does not increase. Otherwise, it is likely to cause performance regression instead of improvement.

By the way, since this is a perf improvement PR, it would be nice to attach some benchmark stats to show the benefits of lazy velocity update when only position is requested (to show improvement), and when position+velocity is requested (to show no regression).

jeetrex17:
Thanks. I reworked this around the number of independent host-side kernel launches.

The measured steady-state counts are now:
- Position only: 2 to 1
- Position plus velocity: 2 to 2

set_dofs_position now fuses the DOF write, qpos synchronization, position FK, and COM propagation into one kinematic kernel. The kernel contains internally sequenced loops, but only one independent host-side launch.

I also ran interleaved local Metal A/B benchmarks with 20k environments, but the machine could not be isolated from background load. The resulting 16% to 35% run-to-run ranges make the figures inconclusive, so I am not quoting a throughput number. The +12.43% figure in the earlier description was measured on the previous design and has been removed.

I restored the anymal_uniform_kinematic benchmark removed in #3249 so the position-only workflow can be measured on reference hardware.

genesis/engine/solvers/rigid/abd/accessor.py
```diff
@@ -856,6 +857,77 @@ def kernel_set_dofs_zero_velocity(
...
+@qd.func
+def func_set_dof_position(
+    i_d_,
+    i_b_,
+    dofs_idx: qd.types.ndarray(),
+    envs_idx: qd.types.ndarray(),
+    position: qd.types.ndarray(),
+    dyn_state: array_class.DynState,
+):
+    dyn_state.dofs.pos[dofs_idx[i_d_], envs_idx[i_b_]] = position[i_b_, i_d_]
```
duburcqa: 1 line helper is violating the guidelines.

duburcqa:
> I also ran interleaved local Metal A/B benchmarks with 20k environments, but the machine could not be isolated from background load. The resulting 16% to 35% run-to-run ranges make the figures inconclusive, so I am not quoting a throughput number.

@jeetrex17 We will need reliable numbers before merging.

tests/benchmarks/test_rigid.py
```diff
@@ -1064,6 +1107,8 @@ def table_bussing(solver, n_envs, gjk):
 BENCHMARKS_FIELD = [
     ("go2", None, None, 4096, gs.gpu),
     ("anymal_random", None, None, 20000, gs.gpu),
+    ("anymal_kinematic_position", None, None, 20000, gs.gpu),
+    ("anymal_kinematic_position_velocity", None, None, 20000, gs.gpu),
```
duburcqa: We are not going to merge this, but it is fine for analysing the benefits of this PR temporarily.

jeetrex17:
> @jeetrex17 We will need reliable numbers before merging.

I do not have regular access to CUDA hardware, so after the local Metal results were inconclusive, I was able to arrange temporary access to a workstation in my university lab and rerun the benchmarks on CUDA

I ran five interleaved A/B pairs with the official benchmark runner on Ubuntu 24.04 with an RTX 2080 Ti and 20,000 environments. Each sample used a 45-second warmup and 15-second recording window. Throughput is measured in kinematic environment updates per second.

| Scenario | main | this PR | Change |
| --- | ---: | ---: | ---: |
| Position only | 32.56M | 38.08M | +17.0% (95% CI: +12.2% to +21.8%) |
| Position followed by link-velocity read | 29.81M | 31.57M | +5.9% (95% CI: +4.4% to +7.4%) |

Every interleaved pair was positive in both scenarios.

These results were collected with temporary focused fixtures for the two access patterns. Per your review, the fixtures have been removed from the final diff, so the benchmark suite is unchanged. The full methodology and results are in the PR description.

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -699,6 +694,8 @@ def get_state(self, f=None):
             if s_global in self._queried_states:
                 return self._queried_states[s_global][0]
 
+            # Keep captured link poses consistent with the generalized coordinates stored alongside them
+            self.update_forward_pos()
```
duburcqa: This seems completely unrelated and could have unexpected side effects.

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -736,8 +734,9 @@ def set_state(self, f, state, envs_idx=None, *, partial: bool = False) -> None:
         )
         if not partial:
             kernel_forward_kinematics(envs_idx, self.dyn_state, self.dyn_info, self.rigid_info, self.rigid_config)
-            self._is_forward_pos_updated = True
-            self._is_forward_vel_updated = True
+            # A subset refresh is globally fresh only if all untouched environments were already fresh
+            self._is_forward_pos_updated = is_all_envs or self._is_forward_pos_updated
+            self._is_forward_vel_updated = is_all_envs or self._is_forward_vel_updated
```
duburcqa: I don't like this kind of tricks. It is just a workaround. It seems that '_is_forward_(pos|vel)_updated' should be refactored to be boolean masks.

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -836,8 +836,8 @@ def set_base_links_pos(self, pos, links_idx=None, envs_idx=None, *, relative=Fal
 
         if not skip_forward:
             kernel_forward_kinematics(envs_idx, self.dyn_state, self.dyn_info, self.rigid_info, self.rigid_config)
-            self._is_forward_pos_updated = True
-            self._is_forward_vel_updated = True
+            self._is_forward_pos_updated = is_all_envs or self._is_forward_pos_updated
+            self._is_forward_vel_updated = is_all_envs or self._is_forward_vel_updated
```
duburcqa: same.

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -902,8 +903,8 @@ def set_base_links_quat(self, quat, links_idx=None, envs_idx=None, *, relative=F
...
```
duburcqa: same.

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -972,14 +974,15 @@ def set_qpos(self, qpos, qs_idx=None, envs_idx=None, *, skip_forward=False):
             else:
                 fn = kernel_forward_kinematics
             fn(envs_idx, self.dyn_state, self.dyn_info, self.rigid_info, self.rigid_config)
-            self._is_forward_pos_updated = True
-            self._is_forward_vel_updated = True
+            self._is_forward_pos_updated = is_all_envs or self._is_forward_pos_updated
+            self._is_forward_vel_updated = is_all_envs or self._is_forward_vel_updated
```
duburcqa: same.

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -1210,6 +1219,26 @@ def update_forward_pos(self):
...
+        # update_forward_pos may have satisfied the velocity refresh
+        if self._is_forward_vel_updated:
+            return
+        kernel_forward_velocity(
+            self.scene._envs_idx,
+            self.dyn_state,
+            self.dyn_info,
+            self.rigid_info,
+            self.rigid_config,
+            is_backward=False,
```
duburcqa: All these arguments fit on 1 line.

tests/rigid/test_kinematics.py
```diff
@@ -87,6 +88,23 @@ def test_link_velocity(gs_sim, tol):
...
+
+    gs_robot.set_dofs_velocity([0.3, -0.2])
+    deferred_links_pos = gs_robot.get_links_pos().clone()
+    deferred_links_vel = gs_robot.get_links_vel().clone()
+    deferred_links_ang = gs_robot.get_links_ang().clone()
+    gs_robot.set_qpos([0.7, -0.4])
+    gs_robot.set_dofs_velocity([0.3, -0.2])
+    assert_allclose(deferred_links_pos, gs_robot.get_links_pos(), tol=tol)
+    assert_allclose(deferred_links_vel, gs_robot.get_links_vel(), tol=tol)
+    assert_allclose(deferred_links_ang, gs_robot.get_links_ang(), tol=tol)
```
duburcqa, line 130: Add a code comment to explain what this is checking. Moreover, we should only assert what is part of the specification of genesis. From what I see, it is not immediately clear whether this is spec or side-effect.

tests/rigid/test_kinematics.py
```diff
@@ -87,6 +88,23 @@ def test_link_velocity(gs_sim, tol):
     )
     assert_allclose(civel_1, civel_1_, tol=tol)
 
+    gs_robot.set_dofs_velocity([0.3, -0.2])
+    links_pos = gs_robot.get_links_pos().clone()
+    links_ang = gs_robot.get_links_ang().clone()
```
duburcqa: Clone is useless. All genesis getters are returning by value.

tests/rigid/test_kinematics.py
```diff
@@ -87,6 +88,23 @@ def test_link_velocity(gs_sim, tol):
...
+    gs_robot.set_qpos([0.7, -0.4], skip_forward=True)
+    assert_allclose(gs_robot.get_links_ang(), links_ang, tol=tol)
+    assert_allclose(gs_robot.get_links_pos(), links_pos, tol=tol)
+
+    gs_robot.set_dofs_velocity([0.3, -0.2])
+    deferred_links_pos = gs_robot.get_links_pos().clone()
+    deferred_links_vel = gs_robot.get_links_vel().clone()
+    deferred_links_ang = gs_robot.get_links_ang().clone()
```
duburcqa: Clone is useless. All genesis getters are returning by value.

tests/rigid/test_kinematics.py
```diff
@@ -819,6 +837,15 @@ def test_setters(show_viewer, tol):
             ghost_robot.get_dofs_position(dofs_idx_local)
 
     ghost_box.set_dofs_position(torch.zeros(ghost_box.n_dofs, device=gs.device))
+    ghost_robot.set_dofs_position(torch.zeros(n_dofs, device=gs.device))
+    ghost_robot.set_dofs_position(
+        0.1,
+        dofs_idx_local=-1,
+        envs_idx=torch.tensor((False, True), dtype=torch.bool, device=gs.device),
```
duburcqa: `(False, True)` is not supported? In any case, specifying 'dtype=torch.bool' is useless.

tests/rigid/test_kinematics.py
```diff
@@ -827,6 +854,94 @@ def test_setters(show_viewer, tol):
...
     assert_allclose(ghost_robot.get_vAABB(), ((0.95, 1.95, 2.95), (2.15, 2.05, 3.05)), tol=tol)
 
+    DOFS_VELOCITY = torch.tensor(
+        (
+            (0.0, 0.0, 0.0, 0.3, -0.2, 0.4, 0.5),
+            (0.0, 0.0, 0.0, -0.4, 0.2, 0.3, -0.6),
+        ),
+        dtype=gs.tc_float,
+        device=gs.device,
+    )
```
duburcqa: Keep as a native python nested tuple.

tests/rigid/test_kinematics.py
```diff
@@ -827,6 +854,94 @@ def test_setters(show_viewer, tol):
...
+        degrees=True,
+    )
+    QPOS = torch.cat(
+        (
+            torch.tensor(((1.0, 2.0, 3.0), (1.0, 2.0, 3.0)), dtype=gs.tc_float, device=gs.device),
+            QPOS_QUAT,
+            torch.tensor(((0.1,), (-0.1,)), dtype=gs.tc_float, device=gs.device),
+        ),
+        dim=-1,
+    )
```
duburcqa: This formatting is quite ugly and hard to read.

duburcqa: Move test constant on top the of unit test.

tests/rigid/test_kinematics.py
```diff
@@ -827,6 +854,94 @@ def test_setters(show_viewer, tol):
...
+
+    ghost_robot.set_qpos(QPOS, skip_forward=True)
+    kinematic_state = next(s for s in scene.get_state().solvers_state if isinstance(s, KinematicSolverState))
```
duburcqa: just do `scene.kinematic_solver`

tests/rigid/test_kinematics.py
```diff
@@ -827,6 +854,94 @@ def test_setters(show_viewer, tol):
...
+    assert_allclose(links_ang, ghost_robot.get_links_ang(), tol=tol)
+
+    state = scene.get_state()
+    links_vel = ghost_robot.get_links_vel(envs_idx=[1])
+    links_ang = ghost_robot.get_links_ang(envs_idx=[1])
+    ghost_robot.set_qpos(QPOS[0], envs_idx=[1])
+    ghost_robot.set_dofs_velocity(DOFS_VELOCITY[0], envs_idx=[1])
+    scene.reset(state, envs_idx=[1])
+    assert_allclose(ghost_robot.get_links_vel(envs_idx=[1]), links_vel, tol=tol)
+    assert_allclose(ghost_robot.get_links_ang(envs_idx=[1]), links_ang, tol=tol)
```
duburcqa: I'm skeptical so many checks are necessary to assert what you want.

tests/rigid/test_kinematics.py
```diff
@@ -958,6 +1073,16 @@ def test_kinematics_queries_span_entities_and_solvers(n_envs, show_viewer, tol):
         entity.set_qpos(qpos)
         assert_allclose(end_effector.get_pos(), target_pos, tol=tol)
 
+    DOFS_VELOCITY = torch.linspace(-0.2, 0.2, ghost.n_dofs, dtype=gs.tc_float, device=gs.device)
+    expected_shape = (ghost.n_links, 3) if n_envs == 0 else (n_envs, ghost.n_links, 3)
+    for getter in (ghost.get_links_vel, ghost.get_links_ang):
+        ghost.set_dofs_velocity(DOFS_VELOCITY)
+        ghost.set_dofs_position(ghost.get_dofs_position() + 0.05)
+        links_velocity = getter()
+        assert links_velocity.shape == expected_shape
```
duburcqa: We do not assert shape. it is trial. We are implemented integration tests, not unit tests in practice. They are here to catch real bugs. if something is always valid by design or the likelihood to cause any other test to fail indirectly is high, when the assertion is skipped, to reduce the CI cost and improve the signal-to-noise ratio.

tests/rigid/test_kinematics.py
```diff
@@ -958,6 +1073,16 @@ def test_kinematics_queries_span_entities_and_solvers(n_envs, show_viewer, tol):
...
+        ghost.set_dofs_velocity(DOFS_VELOCITY)
+        assert_allclose(links_velocity, getter(), tol=tol)
```
duburcqa: What is this supposed to check?

tests/rigid/test_kinematics.py
duburcqa: Add 1-line comment to each block

jeetrex17:
Thanks for the detailed review. I’ve addressed the latest comments, simplified the tests, added coverage for the masked paths, and clarified the skip_forward behavior. The targeted link-velocity test passes.

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -388,6 +392,57 @@ def _create_data_manager(self):
...
+        self._is_forward_vel_updated[envs_idx] = False
+        self._is_forward_vel_updated_for_all_envs = False
+        if gs.backend == gs.metal:
+            torch.mps.synchronize()
+
+    def _write_dofs_velocity(self, dofs_idx, envs_idx, velocity):
+        if velocity is None:
+            kernel_set_dofs_zero_velocity_with_freshness(
+                dofs_idx, envs_idx, self._is_forward_vel_updated, self.dyn_state, self.rigid_config
+            )
```
duburcqa: It turns out setting 'self._is_forward_vel_updated' in python-scope using torch is faster than passing torch tensors directly to a quadrants kernels... Sorry about that. It should probably be mentioned in CODING_GUIDELINES.md.

genesis/engine/solvers/kinematic_solver.py
```diff
@@ -172,8 +172,11 @@ def __init__(self, scene: "Scene", sim: "Simulator", options: "KinematicOptions"
 
         self.qpos = None
 
-        self._is_forward_pos_updated: bool = False
-        self._is_forward_vel_updated: bool = False
+        self._is_forward_pos_updated: torch.Tensor | bool = False
+        self._is_forward_vel_updated: torch.Tensor | bool = False
+        # Host summaries avoid synchronizing the per-environment device masks on every getter
+        self._is_forward_pos_updated_for_all_envs = False
+        self._is_forward_vel_updated_for_all_envs = False
```
duburcqa, line 190: Why are both needed? I understand you point about GPY sync issue, but this is only if you need to do something like `if self._is_forward_vel_updated_for_all_envs.all():`. But if a mask is used everywhere, this should no longer be necessary, removing the need for 'all_envs' variant.

genesis/engine/solvers/kinematic_solver.py
jeetrex17: I’ll move the freshness-mask updates currently performed inside Quadrants kernels to Python.

genesis/engine/solvers/kinematic_solver.py
jeetrex17: The host summaries are only a fast path for fully fresh batches. Without them, getters would either need a synchronizing .all() check or launch a no-op refresh kernel even when every environment is already fresh, which would lose the 2 to 1 launch reduction on the position-only path. The per-environment masks are still needed for mixed batches, so the two serve different purposes.

duburcqa: Sure. But computing '_is_forward_vel_updated_for_all_envs' requires a GPU sync anyway? it is only free when the user does not specify any envs_idx mask. This means that this shortcut only every take effect when envs_idx is not specify. I don't think it is a good idea, as per-env operations are becoming more and more common, and per-env stepping is going to be available soon.

In what extend this does affect performance? If updating position-only, you can still skip velocity update by default, which was the first objective of this PR.

genesis/engine/solvers/kinematic_solver.py
duburcqa: Now that I'm think about it, these masks should be used to skip update INSIDE the kernels for the envs that are already up-to-date. This means that this mask as to be passed one way or the other...

I'm afraid getting this PR right is harder than it seems if we want to address all the issues and edge-cases without compromising performance.

genesis/engine/solvers/kinematic_solver.py
jeetrex17: You are right that the position-only setter keeps its one-launch improvement without the host summaries. My earlier statement about losing that reduction was inaccurate.

I measured the current PR against a local variant with both host summaries removed on an RTX 2080 Ti, using 20k environments, five interleaved A/B pairs, a 45-second warmup, and a 15-second recording window:

| Workflow | Current PR | Masks only | Change |
| --- | ---: | ---: | ---: |
| Position + `scene.step()` | 32.47M | 27.78M | -14.42% (95% CI -16.17% to -12.67%) |
| Repeated fresh velocity read | 348.55M | 151.20M | -56.62% (95% CI -57.53% to -55.71%) |
| Position only | 33.87M | 34.00M | +0.39% (95% CI -1.04% to +1.81%) |
| Position + velocity read | 24.72M | 24.63M | -0.38% (95% CI -1.39% to +0.63%) |

The first two regressions occurred in all five pairs. The latter two differences are statistically inconclusive.

The host summaries are not computed from the device masks with `.all()`. They are plain Python booleans maintained conservatively from the call scope: a known full-batch refresh can set them to `True`, while any subset operation sets them to `False`. The per-environment masks remain the source of truth and are still passed into the kernels to skip fresh environments.

Without the summaries, `substep_post_coupling` and already-fresh getters launch a no-op refresh kernel. Under sustained per-environment operations, the summaries simply remain `False` and the code behaves like the mask-only variant until a full-batch refresh re-arms them.

Based on these results, I think we should keep both: the masks for per-environment correctness and skipping, and the host booleans only as a full-batch launch gate.

jeetrex17:
Since your last review, I reworked the freshness tracking in 36a41a5e.

- The per-environment masks remain, and the refresh kernels use them to skip environments whose state is already current.
- The host summaries now act only as a full-batch gate. A full-environment velocity update changes the summaries without writing the device velocity mask. The mask is materialized when a subset update, a mixed-state refresh or a checkpoint needs its exact contents.
- When every pose is fresh and velocities were already current everywhere before an eager subset velocity write, the setter uses the plain velocity kernel and the following getter does no additional work. Kernel entry counts over 100 subset write-and-read cycles show one entry per setter and none per getter, matching main.
- The masks use `gs.tc_bool`. Checkpoints and trajectory records store independent per-environment copies. The previous head raised `TypeError` in `save_checkpoint` for batched scenes containing a kinematic entity and allowed in-memory checkpoints to alias the live masks. Both cases are fixed and covered by `test_pickle_resume` and `test_trajectory_replay`.
- The rigid behavior is unchanged since your last review: a velocity write following a deferred pose update refreshes the pose first. `RigidSolver.get_links_ang` remains a direct state read.

I replaced the table in the description; the old one measured the Aug 24 revision. The Aug 30 head had also lost the earlier position-only gain, measuring +0.8% rather than about +17%, and was about 20% slower than main when every position update was followed by a velocity read at 20,000 environments.

The main cause was assigning Python booleans into CUDA mask tensors, which introduced a host-to-device transfer and stream synchronization for each flag write. The current implementation removes those writes from uniform velocity-mask updates and uses scalar fills elsewhere.

Here are the updated results for this head versus main on an RTX 2080 Ti using the CUDA field layout. The inputs match the temporary benchmark fixtures used earlier. I ran three fresh process blocks with four interleaved A/B pairs per batch size. The intervals are 95% Student-t intervals over the three process means.

| envs | position only | position + velocity read | subset velocity write + read |
| ---: | ---: | ---: | ---: |
| 128 | +27.2% [+25.3, +29.1] | +0.8% [-1.3, +2.9] | -1.6% [-7.3, +4.2] |
| 4,096 | +21.2% [+16.9, +25.5] | +8.4% [+5.1, +11.7] | -2.0% [-7.9, +3.8] |
| 20,000 | +19.3% [+16.0, +22.5] | +4.6% [+1.9, +7.4] | +0.8% [-7.0, +8.5] |

At 20,000 environments, position-only throughput increased from 34.6M to 41.3M environment updates/s. With a velocity read after every position update, it increased from 31.2M to 32.6M environment updates/s.

These results come from a paired external driver on a shared GPU. All three subset intervals cross zero, as does the 128-environment position-plus-velocity interval, so those measurements do not establish a gain or regression.

On CUDA ndarray: `test_setters`, all four `test_pickle_resume` variants, both `test_trajectory_replay` variants, `test_link_velocity`, the cross-solver kinematics queries, `test_raycaster::test_against_visual[True-2]` and `test_fk_grad_matches_fd`. On CUDA field with zero-copy off, and on CPU with int32 mask storage forced: `test_setters` and `test_pickle_resume[False-2]`. I have not run Metal.

If there is another workload you want measured, let me know and I will add it.

genesis/engine/solvers/kinematic_solver.py
jeetrex17: The summaries are still set from the call scope only, so there is no device readback. The per-env case you raised is now the subset column in the summary comment: plain velocity kernel on the selection, no getter refresh, same kernel entries as main. https://github.com/Genesis-Embodied-AI/genesis-world/pull/3251#issuecomment-5632915493

genesis/engine/solvers/kinematic_solver.py
jeetrex17: The masks are passed to the refresh kernels only for mixed batches and skip current environments inside them. Uniform batches take the plain kernels without mask arguments. Details in https://github.com/Genesis-Embodied-AI/genesis-world/pull/3251#issuecomment-5632915493

jeetrex17:
@duburcqa, when you have time, could you take another look? I addressed the freshness-mask concerns and added updated paired CUDA benchmarks in the summary above.

PR #3253, opened 2026-08-19
[MISC] Add sync-free padded layout to get_contacts.

hughperkins:
Add an 'is_padded' option to Collider.get_contacts and RigidEntity.get_contacts that returns fixed-capacity contact tensors instead of trimming to the across-env contact count. The trim needs 'n_contacts.max().item()', a device-to-host sync that serializes every step; the padded layout skips it and returns the on-device per-env 'n_contacts' so stale slots are excluded via 'valid_mask' with no host sync and no user-side masking.

Use it in the g1_fall_accessors benchmark: ~0.996M -> ~1.28M fps (field, 4096 envs; +28%), recovering ~47% of the accessor overhead relative to the no-accessor ceiling.

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -4264,10 +4264,9 @@ class RigidEntity(KinematicEntity):
         exclude_self_contact: bool
             Exclude the self collision from the returning contacts. Defaults to False.
         is_padded: bool
-            Return fixed-capacity padded tensors instead of trimming to the across-env contact count, avoiding a
-            per-step device-to-host sync. 'valid_mask' still excludes the padded slots. Best-effort hint honored
-            only on the zero-copy batched path; otherwise it is a no-op and the standard trimmed result is
-            returned ('valid_mask' still holds either way). Defaults to False.
+            Return tensors padded to a fixed capacity along the contact axis instead of trimmed to the live
+            contact count, with 'valid_mask' flagging the real contacts. This avoids a per-step device-to-host
+            synchronization; the values are otherwise identical on every backend. Defaults to False.
 
         Returns
         -------
@@ -4275,8 +4274,9 @@ class RigidEntity(KinematicEntity):
             The contact information.
         """
         contact_data = self._solver.collider.get_contacts(as_tensor=True, to_torch=True, is_padded=is_padded)
-        # Per-env live counts, present only for the padded layout.
-        n_contacts = contact_data.pop("n_contacts", None)
+        n_contacts = contact_data["n_contacts"] if is_padded else None
+        if is_padded:
+            del contact_data["n_contacts"]
 
         logical_operation = torch.logical_xor if exclude_self_contact else torch.logical_or
         if with_entity is not None and self.idx == with_entity.idx:
@@ -4303,9 +4303,12 @@ class RigidEntity(KinematicEntity):
 
         if n_contacts is not None:
             slots = torch.arange(valid_mask.shape[-1], device=valid_mask.device)
-            valid_mask = torch.logical_and(valid_mask, slots[None, :] < n_contacts[:, None])
+            if self._solver.n_envs == 0:
+                valid_mask = torch.logical_and(valid_mask, slots < n_contacts.reshape(()))
+            else:
+                valid_mask = torch.logical_and(valid_mask, slots[None, :] < n_contacts[:, None])
 
-        if self._solver.n_envs == 0:
+        if self._solver.n_envs == 0 and not is_padded:
             contact_data = {key: value[valid_mask] for key, value in contact_data.items()}
         else:
             contact_data["valid_mask"] = valid_mask
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -4264,10 +4264,9 @@ class RigidEntity(KinematicEntity):
         exclude_self_contact: bool
             Exclude the self collision from the returning contacts. Defaults to False.
         is_padded: bool
-            Return fixed-capacity tensors padded to the collider's contact capacity instead of trimming to the
-            across-env contact count. This avoids a per-step device-to-host synchronization, at the cost of larger
-            tensors; the returned 'valid_mask' already excludes the padded slots. Only applies to parallelized
-            scenes. Defaults to False.
+            Return fixed-capacity padded tensors instead of trimming to the across-env contact count, avoiding a
+            per-step device-to-host sync. 'valid_mask' still excludes the padded slots. Batched scenes only.
+            Defaults to False.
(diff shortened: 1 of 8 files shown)


hughperkins:
@codex review

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -4263,13 +4263,20 @@ def get_contacts(self, with_entity=None, exclude_self_contact=False):
...
 
         Returns
         -------
         contact_info : dict
             The contact information.
         """
-        contact_data = self._solver.collider.get_contacts(as_tensor=True, to_torch=True)
+        contact_data = self._solver.collider.get_contacts(as_tensor=True, to_torch=True, is_padded=is_padded)
+        # Per-env live counts, present only for the padded layout.
+        n_contacts = contact_data.pop("n_contacts", None)
```
duburcqa: We don't use 'pop' in genesis. Either a key is here or is not, depending on the codepath. never maybe. Just do:
```
n_contacts = contact_data["n_contacts"] if is_padded else None
```

and remove the comment '# Per-env live counts, present only for the padded layout.', since it becomes immediately obvious.

genesis/engine/solvers/rigid/collider/collider.py
```diff
@@ -1005,30 +1007,38 @@ def get_contacts(self, as_tensor: bool = True, to_torch: bool = True, keep_batch
         )
         if gs.use_zerocopy and self._contact_data is not None:
             n_contacts = qd_to_torch(self._collider_state.n_contacts, copy=False)
-            if as_tensor or n_envs == 0:
+            # Padded layout skips the 'n_contacts.max().item()' device->host sync (which serializes the step):
+            # return fixed-capacity tensors plus per-env 'n_contacts' for on-device masking. Batched layout only.
+            padded = is_padded and as_tensor and n_envs > 0
```
duburcqa: I don't understand why 'n_envs == 0' and 'as_tensor == False' is not padded. If padded is specified, it should be honoured, no matter if it is not faster.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -4263,13 +4263,20 @@ def get_contacts(self, with_entity=None, exclude_self_contact=False):
             The entity to check contact with. Defaults to None.
         exclude_self_contact: bool
             Exclude the self collision from the returning contacts. Defaults to False.
+        is_padded: bool
+            Return fixed-capacity padded tensors instead of trimming to the across-env contact count, avoiding a
+            per-step device-to-host sync. 'valid_mask' still excludes the padded slots. Best-effort hint honored
+            only on the zero-copy batched path; otherwise it is a no-op and the standard trimmed result is
+            returned ('valid_mask' still holds either way). Defaults to False.
```
duburcqa: I'm not ok with that. If padded is specified, then padded it is. The output of a getter cannot depends on some dark magic nobody knows about.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -4263,13 +4263,20 @@ def get_contacts(self, with_entity=None, exclude_self_contact=False):
             The entity to check contact with. Defaults to None.
         exclude_self_contact: bool
             Exclude the self collision from the returning contacts. Defaults to False.
+        is_padded: bool
+            Return fixed-capacity padded tensors instead of trimming to the across-env contact count, avoiding a
+            per-step device-to-host sync. 'valid_mask' still excludes the padded slots. Best-effort hint honored
+            only on the zero-copy batched path; otherwise it is a no-op and the standard trimmed result is
```
duburcqa: Avoid mentioning zero-copy in the doc, nobody knows what it is. it is an implementation details that is never exposed to the user, since the user always receives a torch tensor that does not share memory with quadrants.

genesis/engine/solvers/rigid/collider/collider.py
```diff
@@ -1055,6 +1065,11 @@ def get_contacts(self, as_tensor: bool = True, to_torch: bool = True, keep_batch
 
                 contact_data[key] = data
 
+            if padded:
+                # Per-env counts let padded consumers mask stale slots without a host sync. Match the 'to_torch'
+                # dtype of the other fields so the result is not a mixed torch/numpy dict.
```
duburcqa: Remove this comment, it is useless.

genesis/engine/entities/rigid_entity/rigid_entity.py
duburcqa, line 1: You cannot add a new feature without accompanying unit test. While doing so, make sure to follow the guidelines (CLAUDE.md and CODING_GUIDELINES.md). Notably, always prefer extending an existing unit test targeting the same feature over adding a new one.

tests/benchmarks/test_rigid.py
```diff
@@ -396,7 +396,7 @@ def step():
             robot.get_links_pos()
             robot.get_links_quat()
             robot.get_links_vel()
-            robot.get_contacts()
+            robot.get_contacts(is_padded=True)
```
hughperkins, line 399: shouldnt we test without is_padded=True too?

genesis/engine/entities/rigid_entity/rigid_entity.py
hughperkins: I think this is added now?

genesis/engine/solvers/rigid/collider/collider.py
hughperkins: removed

genesis/engine/entities/rigid_entity/rigid_entity.py
hughperkins: Removed from docstring

genesis/engine/entities/rigid_entity/rigid_entity.py
hughperkins: Addressed. No longer depends on dark magic.

genesis/engine/entities/rigid_entity/rigid_entity.py
hughperkins: addressed the first one. checking the second bit

hughperkins: addressed the second bit too.

genesis/engine/solvers/rigid/collider/collider.py
hughperkins: addressed: whilst there are still conditional on n_envs == 0 or on n_envs > 0, these are also conditional also on `and not padded`, so when it is padded, it ignores n_envs now, in the conditional.

tests/benchmarks/test_rigid.py
hughperkins: oh this is benchmarks. never mind 😅

duburcqa (approved):

PR #3261, opened 2026-08-21
[BUG FIX] Recover a usable inertia for URDF links stating a zero mass or inertia.

Milotrince:
## Description

A URDF link with `<inertia ixx="0" ... izz="0"/>` reaches the solver with an isotropic inertia of `2**-52` kg.m2. `genesis-world` catches it but does not apply the fix:

```
[WARNING] Link 'yellow_plastic_bowl' has dubious inertia [ixx=2.220e-16,iyy=2.220e-16,izz=2.220e-16] 
                  compared to the estimate from geometry [ixx=1.125e-04,iyy=1.993e-04,izz=1.125e-04].
```

- `genesis/utils/mjcf.py` passes boundmass and boundinertia to MuJoCo whenever any link in the URDF is missing <inertial>, MuJoCo then clamps it to 2.22e-16
- genesis/engine/entities/rigid_entity/rigid_entity.py tests for degeneracy, but since 2.22e-16 is positive, the repaired value is not used.

The other assets which are .glb/.gltf do not go through the MuJoCo parser. Also, when `merge_fixed_links=True`, no `bound*` is passed, so MuJoCo returns 0 for mass and so the repaired value is used, so the error happens when the URDF is combined with `merge_fixed_links=False`.

Three degenerate authored inertials are now recovered from the geometry:

- **A stated zero inertia.** The `boundmass` / `boundinertia` placeholder is discarded per link rather than only for links with no `<inertial>` at all, so the stated zero survives to the degeneracy check. An authored center of mass is kept, since the geometry only describes how the mass is laid out.
- **A stated zero inertia on a fixed-jointed child.** A fixed link is subsumed into its parent's composite, which its own inertia contributes to, so it is recovered too whenever it has geometry to derive one from. Its stated mass is honored however far it sits from what the geometry alone would weigh.
- **A stated zero mass.** Rescaling the recovered inertia to a zero mass annihilates the tensor, leaving the link at `gs.EPS` with no inertia at all, which drives the constraint solve to NaN. A non-positive mass is therefore treated as undefined and comes from the geometry too - but only where no rigidly-attached (fixed-joint) child carries any, since a stated zero above a mass-bearing child is the massless-connector idiom rather than an omission, and the composite is finite without it.

## Motivation and Context

genesis-world should catch these degenerate inertia cases.

## Checklist:

Related to SIM-373

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -21,7 +21,6 @@ from .description import (
     RigidEqualityDescription,
     RigidLinkDescription,
 )
-from .inertial import MASS_EPS, RHO_MUJOCO, RHO_OBJECT, RHO_ROBOT, compose_inertial_from_g_infos, finalize_inertial
 from .rigid_equality import RigidEquality
 from .rigid_geom import RigidGeom
 from .rigid_joint import RigidJoint
@@ -408,37 +407,12 @@ class KinematicEntity(Entity):
         # links of other trees declared in the same file keep their own root, as do the fixed flag and invweight.
         for link in self._solver.links:
             if link.root_idx == base_link.idx:
-                was_fixed = link.is_fixed
                 link._root_idx = parent_link.root_idx
                 link._is_fixed &= parent_link.is_fixed
 
                 # The attach moves this link into another kinematic tree, so its old tree's inverse weight no longer
                 # applies. The sentinel makes the solver recompute it during its refresh.
                 link.desc.invweight = np.full((2,), fill_value=-1.0, dtype=gs.np_float)
-
-                # Inertial resolution skips the geometry estimate for a world-carried link, so a link that starts
-                # moving recomputes it from its geoms at the entity's simulated density. Values the asset states
-                # still win over the estimate.
-                if was_fixed and not link.is_fixed and isinstance(link.desc, RigidLinkDescription):
-                    desc = link.desc
-                    # An entity attached beneath this one shares its root, so the density comes from the link's own
-                    # entity rather than from this one.
-                    rho = link.entity.material.rho
-                    if rho is None:
-                        if self._solver._enable_mujoco_compatibility:
-                            rho = RHO_MUJOCO
-                        else:
-                            rho = RHO_ROBOT if desc.is_robot else RHO_OBJECT
-
-                    # A geom description holds the fields a parse states it with, so one composition serves either.
-                    hint = compose_inertial_from_g_infos([vars(g_desc) for g_desc in (desc.geoms or desc.vgeoms)], rho)
-                    stated_mass = desc.mass if desc.mass > MASS_EPS else None
-                    stated_inertia = desc.inertia if desc.inertia is not None and desc.inertia.any() else None
-                    stated_com = desc.inertial_pos if stated_inertia is not None else None
-                    stated_quat = desc.inertial_quat if stated_inertia is not None else None
-                    desc.mass, desc.inertial_pos, desc.inertial_quat, desc.inertia = finalize_inertial(
-                        stated_mass, stated_com, stated_quat, stated_inertia, *hint
-                    )
 
         # Apply the explicit mounting transform. Forward kinematics interprets the base link's local pose relative to
         # its (new) parent link, so overwriting it here mounts the entity at (pos, quat) in the parent link frame.
--- a/genesis/utils/mjcf.py
+++ b/genesis/utils/mjcf.py
@@ -230,13 +230,23 @@ def build_model(
 
             # Special treatment for URDF
             if is_urdf_file:
-                # Discard placeholder inertias that were used to avoid parsing failure
+                # Restore the authored mass and inertia over the 'boundmass' / 'boundinertia' placeholder applied
+                # above, so a stated zero survives to the degeneracy check that recovers it from the geometry (see
+                # 'KinematicEntity._parse_scene').
                 for link in robot.links:
-                    if link.inertial is None:
-                        body = mj.body(link.name)
+                    inertial = link.inertial
+                    mass = (inertial.mass or 0.0) if inertial is not None else 0.0
+                    is_inertia_defined = inertial is not None and np.linalg.norm(inertial.inertia, np.inf) > 0.0
+                    if mass > 0.0 and is_inertia_defined:
+                        continue
+                    body = mj.body(link.name)
+                    body.mass[:] = mass
+                    # An inertia that was authored at all is kept, so that a malformed one is still reported.
+                    if not is_inertia_defined:
                         body.inertia[:] = 0.0
-                        body.mass[:] = 0.0
-                        body.invweight0[:] = 0.0
+                    # The inverse weight is derived from the compiled mass and inertia, so it goes stale as soon as
+                    # either of them is replaced here.
+                    body.invweight0[:] = 0.0
 
                 # Set default constraint solver time constant
                 mj.jnt_solref[:, 0] = MIN_TIMECONST
--- a/genesis/utils/mjcf.py
+++ b/genesis/utils/mjcf.py
@@ -207,12 +207,12 @@ def build_model(
                 autolimits="true",
             )
 
-            # Bound mass and inertia if necessary
-            if not all(link.inertial is not None for link in robot.links):
-                compiler.attrib |= dict(
-                    boundmass=str(MIN_TIMECONST),
-                    boundinertia=str(MIN_TIMECONST),
-                )
+            # MuJoCo rejects a moving body whose mass or inertia is below 'mjMINVAL'. Bounding both at that value keeps
+            # a zero or missing inertial compilable, and the placeholder is discarded below.
+            compiler.attrib |= dict(
+                boundmass=str(mujoco.mjMINVAL),
+                boundinertia=str(mujoco.mjMINVAL),
+            )
 
             # Resolve relative mesh paths
             for elem in root.findall(".//mesh"):
@@ -230,9 +230,7 @@ def build_model(
 
             # Special treatment for URDF
             if is_urdf_file:
-                # Restore the authored mass and inertia over the 'boundmass' / 'boundinertia' placeholder applied
-                # above, so a stated zero survives to the degeneracy check that recovers it from the geometry (see
-                # 'KinematicEntity._parse_scene').
+                # Discard placeholder inertias that were used to avoid parsing failure
                 for link in robot.links:
                     inertial = link.inertial
                     mass = (inertial.mass or 0.0) if inertial is not None else 0.0
@@ -241,11 +239,10 @@ def build_model(
                         continue
                     body = mj.body(link.name)
                     body.mass[:] = mass
-                    # An inertia that was authored at all is kept, so that a malformed one is still reported.
+                    # Keep non-zero authored inertia with invalid diagonal so the consistency check reports it
                     if not is_inertia_defined:
                         body.inertia[:] = 0.0
-                    # The inverse weight is derived from the compiled mass and inertia, so it goes stale as soon as
-                    # either of them is replaced here.
+                    # invweight0 derives from placeholder mass and inertia; zero triggers recomputation
                     body.invweight0[:] = 0.0
 
                 # Set default constraint solver time constant
--- a/tests/parsers/test_usd.py
+++ b/tests/parsers/test_usd.py
@@ -143,8 +143,8 @@ def test_massapi_invalid_defaults_mjcf_vs_usd(asset_tmp_path, scale):
     # principalAxes (0, 0, 0, 0), diagonalInertia (0, 0, 0), mass (0) - that must be treated as unset and
     # recomputed from geometry, matching an MJCF scene without inertial element. Leaving only 'centerOfMass' at
     # its sentinel keeps the authored mass and inertia and places the center of mass at the link frame, matching an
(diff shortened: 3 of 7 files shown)


Milotrince:
@codex review

Milotrince:
@codex review

Milotrince:
@codex review

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -859,7 +861,20 @@ def _parse_scene(self, morph, surface):
...
+            # A zero mass stands unless it would leave the link singular, which only a moving link with no mass
+            # anywhere in its rigidly-attached (fixed-joint) subtree is: any other link is carried by a composite.
+            if is_mass_degenerate and not is_fixed_link and not has_link_fixed_child_mass:
+                # Substituting the mass also silences the 'dubious mass' check, which compares against the parsed
+                # value, so the size of the correction must be reported here or it goes unmentioned entirely.
+                if link_g_infos:
+                    gs.logger.warning(
+                        f"Moving link '{l_info['name']}' states no mass and nothing rigidly attached provides any. "
+                        "Recomputing it based on geometry."
+                    )
```
duburcqa: I have no idea what this means. Moreover, why a warning instead of debug like we are doing elsewhere?

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -196,23 +196,25 @@ def finalize_inertial(
...
+    info ('compose_inertial_from_g_infos' over the parsed geom infos, feeding both 'RigidLink._build' and the align
+    anchor) - the single resolution path keeps the rigid dynamics inertia and the align anchor in lockstep.
 
     With ``clamp_min_mass`` the resolved mass is floored at ``gs.EPS`` so a geometry-less moving link stays
     non-singular in the dynamics; the align stash passes ``False`` so a genuinely massless link keeps its ``0.0``
     mass and is excluded from the fixed-subtree composite (it must not inflate the composite by ``gs.EPS``).
     """
     mass, com, quat, inertia = explicit_mass, explicit_com, explicit_quat, explicit_inertia
-    if (mass or hint_mass) > MASS_EPS and hint_mass > gs.EPS and mass is not None:
+    if mass is not None and hint_mass > gs.EPS:
```
duburcqa: What happened to MASS_EPS? It is unused know? if so, it should be removed.

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -859,7 +861,20 @@ def _parse_scene(self, morph, surface):
...
+                l_info["inertial_mass"] = None
```
duburcqa: Why setting 'inertial_mass' to None if link_g_infos is empty?

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -837,10 +839,6 @@ def _build(self):
                     f"Mass is not specified and collision geoms can not be found for link '{self.name}'. "
                     f"Using visual geoms to compute inertial properties."
                 )
-            if self._inertial_pos is not None and self._inertial_i is None:
-                gs.logger.warning(
-                    f"Ignoring center of mass of link '{self.name}' because inertia matrix is not specified."
-                )
```
duburcqa: Why is this not relevant anymore? It was dead code?

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -196,23 +196,25 @@ def finalize_inertial(
...
 
     Explicit values are used when given; otherwise the geometry estimate is used, and an explicit mass rescales a
-    geometry-derived inertia. An omitted center of mass defaults to the link-frame origin when an explicit inertia
-    matrix is provided. The hint comes from the load-time inertial info ('compose_inertial_from_g_infos' over the
-    parsed geom infos, feeding both 'RigidLink._build' and the align anchor) - the single resolution path keeps the
-    rigid dynamics inertia and the align anchor in lockstep.
+    geometry-derived inertia to it. An omitted center of mass defaults to the link-frame origin when an explicit
+    inertia matrix is provided, and to the geometry estimate otherwise. The hint comes from the load-time inertial
+    info ('compose_inertial_from_g_infos' over the parsed geom infos, feeding both 'RigidLink._build' and the align
+    anchor) - the single resolution path keeps the rigid dynamics inertia and the align anchor in lockstep.
```
duburcqa: > The hint comes from the load-time inertial info ('compose_inertial_from_g_infos' over the parsed geom infos, feeding both 'RigidLink._build' and the align anchor) - the single resolution path keeps the rigid dynamics inertia and the align anchor in lockstep.

Please clarify the doc. Moreover, '-' is prohibited by the guidelines. Prefer short and simple plain English sentences.

genesis/utils/mjcf.py
```diff
@@ -229,13 +229,18 @@ def build_model(
...
             # Special treatment for URDF
             if is_urdf_file:
-                # Discard placeholder inertias that were used to avoid parsing failure
+                # Discard the 'boundmass' / 'boundinertia' placeholder, so a zero or missing inertial stays undefined.
                 for link in robot.links:
-                    if link.inertial is None:
-                        body = mj.body(link.name)
+                    inertial = link.inertial
+                    mass = (inertial.mass or 0.0) if inertial is not None else 0.0
+                    is_inertia_valid = inertial is not None and (np.diag(inertial.inertia) > 0.0).all()
```
duburcqa: `(np.diag(inertial.inertia) > 0.0).all()` is too restrictive. We do not want to mask invalid inertia. Just check `np.linalg.norm(inertial.inertia, np.inf) > 0.0`

duburcqa: 'is_inertia_valid' is actually 'is_inertia_placeholder' / 'is_inertia_specified' / 'is_inertia_defined'

genesis/utils/mjcf.py
```diff
@@ -229,13 +229,18 @@ def build_model(
...
+                    if mass > 0.0 and is_inertia_valid:
+                        continue
+                    body = mj.body(link.name)
+                    body.mass[:] = mass
+                    if not is_inertia_valid:
                         body.inertia[:] = 0.0
-                        body.mass[:] = 0.0
-                        body.invweight0[:] = 0.0
+                    body.invweight0[:] = 0.0
```
duburcqa, line 246: Why is this not gated by 'is_inertia_valid' anymore? 

Some code comments would be nice to explain the entire discarding logics.

tests/parsers/test_usd.py
```diff
@@ -217,6 +217,58 @@ def test_massapi_invalid_defaults_mjcf_vs_usd(asset_tmp_path, scale):
     compare_scene(mjcf_scene, usd_scene, tol=1e-5)
 
 
+@pytest.mark.required
+def test_massapi_authored_com_keeps_placement(asset_tmp_path):
```
duburcqa: There is no way to simplify a bit this unit test to check something this simple?

duburcqa: This unit test is not following the guidelines. Each unit test must validate features / capability. Not individual edge-cases. They are actually integration tests, not unit tests strictly speaking.

tests/rigid/conftest.py
```diff
@@ -453,6 +455,79 @@ def implicit_inertial_origin():
...
+    mjcf = ET.Element("mujoco", model="zero_density_marker")
+    worldbody = ET.SubElement(mjcf, "worldbody")
+    hull = ET.SubElement(worldbody, "body", name="hull")
+    ET.SubElement(hull, "freejoint")
+    ET.SubElement(hull, "geom", type="box", size="0.1 0.1 0.1")
+    marker = ET.SubElement(hull, "body", name="marker")
+    ET.SubElement(marker, "geom", type="sphere", size="0.05", density="0")
+    return ET.tostring(mjcf, encoding="unicode")
+
+
```
duburcqa, line 697: I doubt so many fixture are needed for something this simple. Please factorize the code, and try to find a single test that is a superset of smaller individual cases. As before, we are trying to pack every checks in a single, more complex, scene, instead of defining a unit test for every single minimal cases.

tests/rigid/test_asset_loading.py
```diff
@@ -256,6 +268,43 @@ def test_urdf_parsing_inertia_defaults(
...
+            merge_fixed_links=False,
+        ),
+    )
+    entity_with_zero_inertia_child = scene.add_entity(
+        morph=gs.morphs.URDF(
+            file=zero_inertia_fixed_child_urdf,
+            pos=(2.4, 0.0, 0.5),
+            merge_fixed_links=False,
+        ),
+    )
```
duburcqa: Would this be possible to have a single entity packing all these edge-cases at once?

tests/rigid/test_asset_loading.py
```diff
@@ -288,8 +337,36 @@ def test_urdf_parsing_inertia_defaults(
         tol=tol,
     )
 
-    # Resolving the center of mass to the link frame can place it outside the geometry, which stays worth reporting.
-    # Only the link whose geometry is offset qualifies; a geometry-derived center of mass never does.
```
duburcqa, line 396: Why this comment has been removed?

tests/rigid/test_asset_loading.py
```diff
@@ -288,8 +337,36 @@ def test_urdf_parsing_inertia_defaults(
...
+
+    # A fixed child's own inertia makes up its parent's composite, so a zero stated there is recovered too.
+    assert_allclose(entity_with_zero_inertia_child.get_mass(), TIP_MASS, rtol=1e-9)
+    assert_allclose(
+        np.linalg.eigvalsh(entity_with_zero_inertia_child.base_link.inertial_i) / TIP_MASS, estimate_per_mass, tol=tol
+    )
+
+    # A zero-density geom weighs nothing by design, and the hull's composite already carries the body holding it.
+    assert_allclose(entity_with_zero_density_marker.get_link("marker").inertial_mass, gs.EPS, tol=gs.EPS)
+    assert_allclose(entity_with_zero_density_marker.get_mass(), 8.0, tol=1e-5)
```
duburcqa: What about some for-loop or something? To factories these asserts whenever it is possible / makes sense.

tests/rigid/test_asset_loading.py
```diff
@@ -307,6 +384,8 @@ def test_urdf_parsing_inertia_defaults(
         scene.step()
     assert_allclose(entity_without_inertia.get_pos(), (-0.3, 0.0, -0.03), tol=1e-3)
     assert_allclose(entity_with_implicit_origin.get_pos(), (0.0, 0.0, -0.03), tol=1e-3)
+    assert_allclose(entity_with_zero_inertia.get_pos(), (0.3, 0.0, -0.03), tol=1e-3)
+    assert_allclose(entity_with_zero_mass.get_pos(), (-0.9, 0.0, -0.03), tol=1e-3)
```
duburcqa: Same, maybe some for-loop?

duburcqa: Why '1e-3' for the tolerance? It seems very bad.

Milotrince:
Two adjacent cases I hit while testing this branch against a set of URDF props. Neither is a
regression — both behave identically on stock 1.3.3 and with this PR's three source hunks applied —
so they are observations about where the fix stops, not blockers.

**A stated zero inertia is not recovered when the link is loaded fixed.**

`RigidLink._build` computes the geometry hint only for a non-fixed link:

```python
if not self._is_fixed:
    hint_mass, hint_com, hint_inertia = self.entity._links_inertial_info[...][0].hint
```

so under `morphs.URDF(..., fixed=True)` `finalize_inertial` receives `hint_mass=0` and a zero
`hint_inertia`, and since `com is None` it returns that zero hint. The repair this PR adds to
`_parse_scene` does fire — the debug line `Invalid or undefined inertia for link 'tube_rack'. Force
recomputing it based on geometry.` appears and `inertial_i` is set to `None` — there is just nothing
left to recompute from.

On a link with `<inertia ixx="0" ... izz="0"/>` and no `<origin>`:

| load | resolved principal inertia |
| --- | --- |
| `fixed=True` | `0 / 0 / 0` |
| `fixed=False` | `3.07e-04 / 9.19e-04 / 1.14e-03` |

A fixed link contributes no dynamics of its own, so this may be deliberately out of scope. The
visible consequence is that `link.inertial_i` reads back as an exact zero tensor through the API,
which is what asset tooling inspects.

**A welded child reaches the solver massless on a free base with `merge_fixed_links=False`.**

Two links joined by a fixed joint, each authoring its own `<mass>` and `<inertia>` (9.3 g body,
2.5 g cap):

| load | body | cap |
| --- | --- | --- |
| `fixed=True, merge_fixed_links=False` | 0.009300 | 0.002500 |
| `fixed=False, merge_fixed_links=False` | 0.011800 | 0.000000, zero inertia |

The composite is right, so the dynamics are right, but a per-link read sees a massless cap in the
one mode this asset's own documentation prescribes — `merge_fixed_links=False`, because an importer
that keeps only a link's first `<visual>` otherwise renders the barrel and drops the cap. The zero
here is produced by the parser rather than authored, so the massless-connector branch this PR adds
never sees it.

Measured on genesis-world 1.3.3 with this PR's `genesis/` hunks applied on top, backend `gs.cpu`,
against `Genesis-Intelligence/Digital_twin_asset` at `refs/pr/4`.

_Findings from Claude Code._

duburcqa (approved):

PR #3262, opened 2026-08-22
[BUG FIX] Fix parsing URDF for partially specified dynamics attributes.

### LINKED ISSUE #3258: [Bug]: numpy cast error in kinematic solver caused by Null Values
### Bug Description

In some URDF assets the friction/friction_loss is not parsed/handled properly. When the friction attribute is not provided for a joint in the URDF the parser stores it as [Null].

In the Kinematic solver when the dof_fields are initialized, specifically this part
```python
np.concatenate([joint.dofs_frictionloss for joint in joints], dtype=gs.np_float), 
```
leads to this error, due to some joints having the dofs_frictionloss property as [Null] ,a list containing the Null object instead of a float 

```python
TypeError: Cannot cast array data from dtype('O') to dtype('float32') according to the rule 'same_kind' 
```

### Steps to Reproduce

add a robot to the scene via a URDF file and build the scene. The robot asset I am using has this

```
  <joint name="fr3_finger_joint2" type="prismatic">
     ...
    <dynamics damping="0.3"/>
  </joint>
```

changing the dynamics field this way fixes the issue `<dynamics damping="0.3" friction="0.0"/>` 

### Expected Behavior

Null checks before buffer initialization as the robot assets aren't always perfect

v1.3.3

### Additional Context

_No response_

catplotlib:
## Description

`parse_urdf` applied its `0.0` fallbacks for `friction` / `damping` only when the `<dynamics>` tag was absent altogether. A tag authoring just one of the two left the other one as `None`, which flowed into `np.full` and produced an object array that `np.concatenate` refused to cast when initializing the dof fields:

```
TypeError: Cannot cast array data from dtype('O') to dtype('float32') according to the rule 'same_kind'
```

The fallback is now applied per attribute rather than per tag, so a partially authored tag behaves like an absent one for the attribute it omits.

Both attributes are affected, not only the one in the report: omitting `friction` breaks `dofs_frictionloss` (`kinematic_solver.py:420`) and omitting `damping` breaks `dofs_damping` (`kinematic_solver.py:419`).

## Related Issue

Resolves Genesis-Embodied-AI/genesis-world#3258

## Motivation and Context

`<dynamics damping="0.3"/>` without a `friction` attribute is valid URDF and appears in real assets - the report hit it on a Franka `fr3_finger_joint2`. Loading such a robot aborts `scene.build()` with a `TypeError` that points at the solver rather than at the joint responsible, and the only workaround is editing the asset to author the missing attribute explicitly.

Worth flagging for review: the null only reaches the solver when the model has no authored `<inertial>`. When inertia is valid it is replaced before the solver sees it, which is why the existing `pendulum_with_joint_dynamics` model cannot reproduce the failure and the new `joint_with_partial_dynamics` model leaves inertia undefined.

Minimal reproducer - raises the `TypeError` above on `main`, builds fine with this change:

```python
import genesis as gs

URDF = """<?xml version="1.0"?>
<robot name="t">
  <link name="base_link"/>
  <link name="finger"/>
  <joint name="finger_joint" type="prismatic">
    <parent link="base_link"/><child link="finger"/>
    <origin xyz="0 0 0" rpy="0 0 0"/><axis xyz="0 0 1"/>
    <limit effort="100" lower="0" upper="0.04" velocity="0.2"/>
    <dynamics damping="0.3"/>
  </joint>
</robot>
"""
open("arm.urdf", "w").write(URDF)

gs.init(backend=gs.cpu)
scene.add_entity(gs.morphs.URDF(file="arm.urdf", fixed=True))
scene.build()
```

`test_urdf_joint_dynamics` is extended with a `(joint_damping, joint_friction)` sweep covering the partially authored cases, over both the existing pendulum and the new undefined-inertia model:

```
pytest -v tests/rigid/test_asset_loading.py -k test_urdf_joint_dynamics
```

- with the source change reverted: 2 failed, 4 passed (both failures are the `TypeError` above)
- with the source change: 6 passed
- whole file: 39 passed, 2 skipped (`gs.gpu` not available on this machine)

Environment: Ubuntu 24.04, Python 3.12, CPU backend.

N/A

## Checklist:

### DIFF
--- a/genesis/utils/urdf.py
+++ b/genesis/utils/urdf.py
@@ -373,7 +373,10 @@ def parse_urdf(morph, surface):
 
         joint_friction, joint_damping = 0.0, 0.0
         if joint.dynamics is not None:
-            joint_friction, joint_damping = joint.dynamics.friction, joint.dynamics.damping
+            if joint.dynamics.friction is not None:
+                joint_friction = joint.dynamics.friction
+            if joint.dynamics.damping is not None:
+                joint_damping = joint.dynamics.damping
         j_info["dofs_frictionloss"] = np.full(j_info["n_dofs"], joint_friction)
         j_info["dofs_damping"] = np.full(j_info["n_dofs"], joint_damping)
         j_info["dofs_armature"] = np.zeros(j_info["n_dofs"])
--- a/tests/rigid/conftest.py
+++ b/tests/rigid/conftest.py
@@ -342,7 +342,11 @@ def _build_multi_pendulum(n, joint_damping, joint_friction):
         ET.SubElement(joint, "parent", link=parent_link)
         ET.SubElement(joint, "child", link=f"PendulumArm_{i}")
         ET.SubElement(joint, "limit", effort=str(100.0 * (n - i)), velocity="30.0")
-        ET.SubElement(joint, "dynamics", damping=str(joint_damping), friction=str(joint_friction))
+        dynamics = ET.SubElement(joint, "dynamics")
+        if joint_damping is not None:
+            dynamics.set("damping", str(joint_damping))
+        if joint_friction is not None:
+            dynamics.set("friction", str(joint_friction))
 
         # Arm link
         arm = ET.SubElement(urdf, "link", name=f"PendulumArm_{i}")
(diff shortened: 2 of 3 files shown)


duburcqa:
@catplotlib Please follow the official template for PR descriptions. Moreover, make sure that you have been following all the guidelines (CLAUDE.md for agents, and CODING_GUIDELINES.md for humans).

catplotlib:
Addressed.

duburcqa (approved):

duburcqa:
Thank you for your contribution!

PR #3268, opened 2026-08-25
[CHANGING] Honor an MJCF's collision masks across entities.

### LINKED ISSUE #3267: [Bug]: MJCF contype/conaffinity are rewritten whenever the file declares <contact><exclude>
### Bug Description

An MJCF states forbidden collision pairs two ways, and MuJoCo keeps them apart: the per-geom `contype` / `conaffinity` masks, and `<contact><exclude>` (compiled into `exclude_signature`). Genesis's MJCF importer folds both into the masks: whenever the file declares at least one `<exclude>`, it gathers every forbidden pair and asks `solve_contype_conaffinity` (z3) for a fresh bitmask assignment realizing all of them at once, overwriting the values the file wrote (`genesis/utils/mjcf.py`, the `if mj.nexclude:` block in `parse_geoms`).

The synthesized assignment is only meaningful inside the entity it was solved over — nothing in the problem mentions the ground or any other entity — so the masks it hands out bear no relation to theirs. One `<exclude>` anywhere in the file is enough to rewrite the masks of **every** geom in it, including geoms unrelated to the exclusion.

This is also why `_is_local_collision_mask` currently has to cover MJCF (`rigid_entity.py`), which means an MJCF's `contype` / `conaffinity` never filter cross-entity pairs at all — an MJCF that uses masks to keep a geom off the floor cannot be reproduced in Genesis today.

### Steps to Reproduce

```python
import genesis as gs

XML = """
<mujoco model="exclude_probe">
  <compiler angle="radian" autolimits="true"/>
  <worldbody>
    <body name="carrier" pos="0 0 0">
      <freejoint name="carrier_free"/>
      <geom name="carrier_geom" type="sphere" size="0.05" pos="0 0 1" contype="1" conaffinity="1"/>
      <body name="lander" pos="0 -0.3 0.05">
        <joint name="lander_hinge" type="hinge" axis="0 0 1" range="0 0"/>
        <geom name="lander_geom" type="sphere" size="0.1" contype="1" conaffinity="1"/>
      </body>
      <body name="floater" pos="0 0.3 0.05">
        <joint name="floater_hinge" type="hinge" axis="0 0 1" range="0 0"/>
        <geom name="floater_geom" type="sphere" size="0.1" contype="2" conaffinity="2"/>
      </body>
      <body name="armA" pos="0.5 0 1">
        <joint name="armA_hinge" type="hinge" axis="0 0 1" range="0 0"/>
        <geom name="armA_geom" type="sphere" size="0.1" contype="1" conaffinity="1"/>
      </body>
      <body name="armB" pos="0.62 0 1">
        <joint name="armB_hinge" type="hinge" axis="0 0 1" range="0 0"/>
        <geom name="armB_geom" type="sphere" size="0.1" contype="1" conaffinity="1"/>
      </body>
      <body name="armC" pos="0.38 0 1">
        <joint name="armC_hinge" type="hinge" axis="0 0 1" range="0 0"/>
        <geom name="armC_geom" type="sphere" size="0.1" contype="1" conaffinity="1"/>
      </body>
    </body>
  </worldbody>
  <contact>
    <exclude body1="armA" body2="armB"/>
  </contact>
</mujoco>
"""

scene = gs.Scene(
    show_viewer=False,
)
scene.add_entity(gs.morphs.Plane(contype=1, conaffinity=1))
robot = scene.add_entity(gs.morphs.MJCF(file=XML))
scene.build()

for link in robot.links:
    for geom in link.geoms:
        print(f"{link.name:12s} contype {geom.contype}  conaffinity {geom.conaffinity}")
```

Both spheres near the ground are sunk 50 mm into a `contype=1 conaffinity=1` plane, so `floater` (authored `2/2`, i.e. `(2&1)|(1&2) == 0`) should produce no ground contact while `lander` (`1/1`) should — stepping the scene and reading `scene.rigid_solver.collider.get_contacts()` shows both touching.

### Expected Behavior

The masks survive import as written (this is what MuJoCo compiles):

```
lander       contype 1  conaffinity 1
floater      contype 2  conaffinity 2
armA         contype 1  conaffinity 1
armB         contype 1  conaffinity 1
armC         contype 1  conaffinity 1
```

and `floater` produces no contact against a `1/1` ground.

### Relevant log output

```
Actual output for the probe links (masks rewritten by the entity-local z3 solve; the exclusion
names only armA/armB, yet every geom's masks changed):

lander       contype 2  conaffinity 2
floater      contype 0  conaffinity 1
armA         contype 0  conaffinity 2
armB         contype 0  conaffinity 2
armC         contype 2  conaffinity 2

Removing the <exclude> element makes all masks import verbatim, confirming the trigger.
```

### Environment

- OS: macOS 15 (Apple M4 Pro, CPU backend) — also reproduced on Ubuntu 22.04 (CUDA); backend-independent, the rewrite happens at parse time
- GPU/CPU: Apple M4 Pro / NVIDIA RTX
- GPU-driver version: N/A

Reproduced on `main` (70a0f413) and on release 1.3.3.

### Additional Context

The masks and the exclusions are independent mechanisms in MuJoCo; folding the exclusions into the masks is what forces the entity-local solve and, downstream, the `is_local_collision_mask` guard that disables MJCF masks across entities. Carrying `<exclude>` as an explicit link-pair list (the way welded pairs are already filtered in `_compute_collision_pair_idx`) removes both. I have a patch with a regression test ready and will open a PR referencing this issue.

jsw7460:
## Description

- `<contact><exclude>` is carried out of the MJCF as link-name pairs and applied in the collider beside the weld-pair filter, which is link-level for the same reason. The authored contype / conaffinity are imported unchanged. Previously both were folded into a fresh entity-local z3 bitmask assignment, discarding the authored values.
- With the authored masks now preserved, `is_local_collision_mask` is restricted to USD (whose importer still synthesizes), so MJCF masks filter cross-entity pairs the way they do in MuJoCo.
- Carried by name rather than body index because the parser's MuJoCo body indices need not survive scene construction; resolved against the built links on the build-time host path. No kernel change.

## Related Issue

Resolves Genesis-Embodied-AI/Genesis#3267

## Motivation and Context

Any MJCF declaring at least one `exclude` has ALL of its masks rewritten by the entity-local solve, and the rewritten bits are not constrained to preserve their meaning against other entities.. Measured on the model in the linked issue, against a 1/1 ground: a sphere authored 1/1 came back 2/2 and a 2/2 one came back 0/1, with the exclusion naming neither of them. This is also why MJCF masks had
to be ignored across entities, so an MJCF that masks a geom off the floor cannot be reproduced today.

BEHAVIOUR CHANGE: contype / conaffinity on an MJCF entity now filter cross-entity pairs. Assets using only the default 1/1 are unaffected. For a model with exclusions, pairs the authored masks forbid stay forbidden by those masks, and the exclusions stay forbidden by the list.

`tests/rigid/test_collision.py::test_mjcf_authored_masks_and_contact_exclude` (n_envs 0 and 2). Four checks, a control per mechanism: a 1/1 sphere touches a 1/1 ground while a 2/2 one does not (mask filtering, and that it was not "fixed" by disabling filtering); the excluded pair produces no contact while an identical non-excluded pair does (exclusion, and that it does not over-block). On main it fails at the first mask readback: actual (2, 2), desired (1, 1). `tests/rigid` run before and after: identical outcomes apart from the new test.

## Checklist:

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -160,6 +160,9 @@ class KinematicEntity(Entity):
         self.terrain_scale: np.ndarray | None = None
         self._terrain_height_field: torch.Tensor | None = None
 
+        # Set before '_load_model' so a morph whose parser produces none still answers.
+        self._excluded_links_name: list[tuple[str, str]] = []
+
         self._load_model()
 
         # Initialize target variables and checkpoint
@@ -635,7 +638,10 @@ class KinematicEntity(Entity):
         # initialized undetermined physics parameters.
         if isinstance(morph, gs.morphs.MJCF):
             # Mujoco's unified MJCF+URDF parser systematically for MJCF files
-            l_infos, links_j_infos, links_g_infos, eqs_info = mju.parse_xml(morph, surface, self._solver._options)
+            l_infos, links_j_infos, links_g_infos, eqs_info, excluded_links_name = mju.parse_xml(
+                morph, surface, self._solver._options
+            )
+            self._excluded_links_name = excluded_links_name
         elif isinstance(morph, (gs.morphs.URDF, gs.morphs.Drone)):
             # Custom "legacy" URDF parser for loading geometries (visual and collision) and equality constraints.
             # This is necessary because Mujoco cannot parse visual geometries (meshes) reliably for URDF.
@@ -646,7 +652,7 @@ class KinematicEntity(Entity):
             try:
                 # Mujoco's unified MJCF+URDF parser for URDF files.
                 # Note that Mujoco URDF parser completely ignores equality constraints.
-                l_infos_mj, links_j_infos_mj, links_g_infos_mj, _ = mju.parse_xml(morph_, surface)
+                l_infos_mj, links_j_infos_mj, links_g_infos_mj, _, _ = mju.parse_xml(morph_, surface)
 
                 # Unset link inertial properties that are actually undefined to force recomputation by genesis
                 if not morph._enable_mujoco_compatibility:
@@ -3056,11 +3062,12 @@ class RigidEntity(KinematicEntity):
 
     def _load_model(self):
         self._equalities = gs.List()
-        # MJCF and USD express in-model collision filtering (MJCF '<contact><exclude>', USD CollisionGroup /
-        # FilteredPairsAPI) through synthesized contype/conaffinity bitmasks. Those masks are only consistent within
-        # the entity: applied across entities, they would spuriously disable collision against geoms whose default
-        # masks happen not to overlap (e.g. the ground plane).
-        self._is_local_collision_mask = isinstance(self._morph, (gs.morphs.MJCF, gs.morphs.USD))
+        # USD expresses in-model collision filtering (CollisionGroup / FilteredPairsAPI) through synthesized
+        # contype/conaffinity bitmasks, which are only consistent within the entity they were solved over: applied
+        # across entities they would spuriously disable collision against geoms whose masks happen not to overlap
+        # (e.g. the ground plane). MJCF masks are the ones the model file wrote, so they mean the same thing
+        # everywhere; its own '<contact><exclude>' travels separately as 'excluded_links_name'.
+        self._is_local_collision_mask = isinstance(self._morph, gs.morphs.USD)
 
         super()._load_model()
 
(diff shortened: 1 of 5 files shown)


duburcqa:
Unless you can justify why preserving bit-identical contype / conaffinity, I'm going to close this PR, because there never was any guarantee regarding this. Genesis is not mujoco, even if we can parse mujoco files. The only guarantee that is provided is matching rigid body dynamics when 'enable_mujoco_compatibility=True'.

duburcqa:
Ok now I understand the issue. Unfortunately this change is too impactful to be driven by the community.

jsw7460:
Thank you for looking into this. Appreciate your time.

PR #3269, opened 2026-08-25
[BUG FIX] Apply the per-geom density in MJCF asset.

Milotrince:
## Description

- Apply `density` and `mass` from MJCF geom
- A collision geom weighing nothing leaves the visual geoms to carry the link, as they already do
  for a link with no collision geometry.
- The align anchor now weighs an unstated geom at the default density rather than unit density,
  so a link mixing stated and unstated geoms anchors at the right relative weights instead of
  being rejected as ambiguous.

With `recompute_inertia=True` this PR affects assets that author the Gym-style `density="5"`: `xml/ant.xml` 97.33 -> 0.324 kg (Mujoco: 0.327), `xml/one_tet.xml` 1600 -> 13.3, `xml/thin_box.xml` 0.0096 -> 8e-05. Default options are unaffected: link masses are same on MJCFs with `recompute_inertia=False`.

## Motivation and Context

An asset that states what its geoms are made of should be weighed by it whenever Genesis computes the mass itself.

- `tests/rigid tests/parsers`
- `tests/rigid/test_asset_loading.py::test_mjcf_authored_geom_density`

## Checklist:

### DIFF
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -1009,11 +1009,6 @@ class KinematicEntity(Entity):
                 # and estimated masses would make the anchor density-dependent, and a kinematic entity has no density
                 # to fall back on, so alignment could differ from the rigid counterpart. Require all-or-none and raise
                 # otherwise.
-                if None in mass_explicit:
-                    gs.raise_exception(
-                        f"Entity '{self.uid}': a link of an aligned free body mixes geoms with and without an "
-                        "authored density. Author a density on all of its geoms or none of them."
-                    )
                 if len(mass_explicit) > 1:
                     gs.raise_exception(
                         f"Entity '{self.uid}': an aligned free body mixes explicit (mass or authored per-geom "
@@ -1315,13 +1310,19 @@ class KinematicEntity(Entity):
     ):
         """Compute a link's load-time inertial data (see 'LinkInertialInfo').
 
-        The align-anchor inertial weighs each collision geom by its authored density, falling back to unit density,
-        so it never needs the material density - which a kinematic entity does not have. The geometry hint consumed
-        by 'RigidLink._build' uses the resolved material density as fallback instead, and falls back to the visual
-        geoms for a link without collision geometry.
+        The align-anchor inertial weighs each collision geom by its authored density, falling back to the default
+        density, so a link mixing authored and unauthored geoms still composes at the right relative weights. The
+        anchor never needs the material density - which a kinematic entity does not have - because a material that
+        states one drops every authored density (see '_add_by_info'), leaving the anchor density-independent. The
+        geometry hint consumed by 'RigidLink._build' uses the resolved material density instead, and falls back to
+        the visual geoms for a link without collision geometry.
         """
+        if self._solver._enable_mujoco_compatibility:
+            rho_default = RHO_MUJOCO
+        else:
+            rho_default = RHO_ROBOT if is_robot else RHO_OBJECT
         if cg_infos:
-            hint = compose_inertial_from_g_infos(cg_infos, rho=1.0)
+            hint = compose_inertial_from_g_infos(cg_infos, rho=rho_default)
         else:
             hint = InertialProperties(0.0, np.zeros(3, dtype=gs.np_float), np.zeros((3, 3), dtype=gs.np_float))
         props = finalize_inertial(
@@ -1330,22 +1331,13 @@ class KinematicEntity(Entity):
         if explicit_mass is not None and explicit_mass > 0.0:
             is_mass_explicit = True
         else:
-            geoms_with_density = sum(g_info.get("density") is not None for g_info in cg_infos)
-            if geoms_with_density == 0:
-                is_mass_explicit = False
-            elif geoms_with_density == len(cg_infos):
-                is_mass_explicit = True
-            else:
-                is_mass_explicit = None
+            # A geom the asset weighs is anchored at that weight, and one it does not is anchored at the default the
+            # dynamics falls back on too, so any link the asset weighs at all anchors at its own dynamics mass.
+            is_mass_explicit = any(g_info.get("density") is not None for g_info in cg_infos)
 
         dynamics_hint = None
         if isinstance(self.material, gs.materials.Rigid):
-            rho = self.material.rho
-            if rho is None:
-                if self._solver._enable_mujoco_compatibility:
-                    rho = RHO_MUJOCO
-                else:
-                    rho = RHO_ROBOT if is_robot else RHO_OBJECT
+            rho = self.material.rho if self.material.rho is not None else rho_default
             hint_g_infos = cg_infos if cg_infos else vg_infos
             if hint_g_infos:
                 dynamics_hint = compose_inertial_from_g_infos(hint_g_infos, rho)
--- a/genesis/engine/entities/rigid_entity/rigid_entity.py
+++ b/genesis/engine/entities/rigid_entity/rigid_entity.py
@@ -1238,6 +1238,7 @@ class KinematicEntity(Entity):
         if isinstance(self.material, gs.materials.Rigid) and self.material.rho is not None:
             for g_info in g_infos:
                 g_info.pop("density", None)
+                g_info.pop("mass", None)
 
         cg_infos, vg_infos = [], []
         for g_info in g_infos:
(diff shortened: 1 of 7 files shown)


Milotrince:
@codex review

genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1009,11 +1009,6 @@ def _align_free_roots(self):
                 # and estimated masses would make the anchor density-dependent, and a kinematic entity has no density
                 # to fall back on, so alignment could differ from the rigid counterpart. Require all-or-none and raise
                 # otherwise.
```
duburcqa, line 1011: Make sure to update the comment accordingly.

duburcqa:
It looks like you stripped away many relevant code comments no?

genesis/utils/mjcf.py
```diff
@@ -235,7 +260,7 @@ def parse_xml(morph, surface, rigid_options=None):
     # the expanded model is only ever read by the parsers.
     exclude_ground_plane = isinstance(morph, gs.morphs.MJCF) and morph.exclude_ground_plane
     file = uu.load_xacro(morph.file, morph.xacro_args) if morph.is_format(XACRO_FORMAT) else morph.file
-    mj = build_model(
+    mj, geoms_mass_source = build_model(
```
duburcqa: This looks like a very bad design.

duburcqa:
> The old unit-density anchor could not assign the intended relative weights when authored mass or density was mixed with unspecified geoms. A kinematic entity has no material density to resolve those weights. The format's fixed default supplies them at load time, so the all-or-none exceptions and auto-alignment skip are no longer needed.

I will double check, but this sounds wrong.

genesis/engine/entities/rigid_entity/description.py
```diff
@@ -1119,38 +1119,18 @@ def _align_free_roots(self, resolution: Resolution):
...
                         GeomInertialInfo(InertialProperties(props.mass, props.com, inertia_in_link), *pose_in_root[i_l])
                     )
 
-                # Inertia alignment is insensitive to density if and only if the density is homogeneous. That must
-                # hold among the geometry materials of a link, and among the links the composite anchors on (see
-                # 'LinkInertialInfo.is_mass_explicit'). Otherwise the procedure requires the density of every
-                # geometry. A kinematic entity only visualizes, so its material carries no default density and an
-                # unspecified one cannot be resolved for it. A rigid entity and the kinematic ghost of it must
-                # anchor at one frame, so the inhomogeneous case is structurally impossible to support. Require
-                # all-or-none and raise otherwise.
```
duburcqa, line 1142: How did you manageable to fix this limitation that was "structurally impossible to support"?

genesis/engine/entities/rigid_entity/inertial.py
```diff
@@ -133,28 +133,36 @@ def compose_inertial_from_g_infos(g_infos: Sequence[dict], rho: float) -> Inerti
...
+        mesh = g_info["vmesh"] if is_vis else g_info["mesh"]
+        geom_mass = g_info.get("mass")
+        if geom_mass is None:
+            geom_rho = rho if g_info.get("density") is None else g_info["density"]
+            props = get_local_inertial_from_geom(geom_type, geom_data, mesh, geom_rho)
+        else:
+            # Rescale unit-density inertia to the specified mass; zero-volume geoms remain massless
+            props = get_local_inertial_from_geom(geom_type, geom_data, mesh, 1.0)
+            if props.mass > gs.EPS:
+                props = InertialProperties(geom_mass, props.com, props.i * (geom_mass / props.mass))
```
duburcqa: Just use rho=1.0 in both cases and scale by mass is provided.

genesis/engine/entities/rigid_entity/inertial.py
```diff
@@ -133,28 +133,36 @@ def compose_inertial_from_g_infos(g_infos: Sequence[dict], rho: float) -> Inerti
...
+        geoms_inertial_info.append(
             GeomInertialInfo(
-                get_local_inertial_from_geom(
-                    gs.GEOM_TYPE.MESH if "vmesh" in g_info else g_info["type"],
-                    None if "vmesh" in g_info else g_info.get("data"),
-                    g_info["vmesh"] if "vmesh" in g_info else g_info["mesh"],
-                    rho if g_info.get("density") is None else g_info["density"],
-                ),
+                props,
```
duburcqa: props is not a valid variable name

genesis/engine/entities/rigid_entity/description.py
```diff
@@ -1430,14 +1400,9 @@ def _align_link(self, l_info, j_infos, cg_infos, vg_infos, morph, resolution: Re
 
         is_aligned = morph.align if isinstance(morph, gs.options.morphs.FileMorph) else False
         if is_aligned is None:
-            # Auto: True for basic rigid objects (root with free joint only, no articulated descendants). A link
-            # mixing geoms with and without an authored density (see 'LinkInertialInfo.is_mass_explicit') quietly
-            # declines auto-alignment; asking for it with an explicit align=True raises at build instead.
```
duburcqa, line 1435: what happened? This is now supported?

genesis/utils/mjcf.py
duburcqa, line 1: These changes are weird.

genesis/utils/mjcf.py
```diff
@@ -23,6 +23,9 @@
 
 
 MIN_TIMECONST = np.finfo(np.double).eps
+# MjSpec cannot distinguish an explicit default density from an omitted density
+MUJOCO_DEFAULT_DENSITY = 1000.0
```
duburcqa: So, how are you solving this?

genesis/utils/mjcf.py
```diff
@@ -23,6 +23,9 @@
...
+UNSPECIFIED_GEOM_DENSITY = -np.finfo(np.double).max
```
duburcqa: This sounds very hacky.

duburcqa:
You should update the PR description to explain what is going on. At this point, I don't know what you trying to achieve and the implementation is far from self-explanatory. My initial concern is about parity between rigid vs kinematic entities, incl in case of attachment. Make sure this is supported even for partially defined density. Before it was raising an exception but now it is considered valid but not unit tested.

genesis/engine/entities/rigid_entity/description.py
```diff
@@ -58,10 +59,12 @@
 
 @dataclass(kw_only=True)
 class BaseRigidGeomDescription:
-    """Pose of one geom in its link frame. Collision and visual geom descriptions both hold these fields."""
+    """Pose and authored mass sources of one geom in its link frame."""
 
     pos: np.ndarray = field(default_factory=gu.zero_pos)
     quat: np.ndarray = field(default_factory=gu.identity_quat)
+    density: float | None = None
+    mass: float | None = None
```
duburcqa: This is wrong. A visual geometry should not feature any physics information.

duburcqa:
> Retain geom mass and density in the typed descriptions used by attachment, so a fixed link that becomes movable uses the same sources as initial inertial resolution. Preserve authored mass for small positive-volume geometry and apply the existing scale factors: mass scales cubically and inertia to the fifth power.

This should be doable without having to leak physics information in purely visual geometries. If not, let's not lift the existing limitations, this is out of scope and should be handled by another PR.

duburcqa:
> Use visual geometry for the inertia estimate only when collision geometry is absent. Explicitly zero collision mass remains zero in the geometric estimate, subject to the existing minimum mass for moving links.

How is this supposed to be different from what we are already doing?

duburcqa:
> The parser uses MuJoCo to resolve this distinction. When the root default omits density, it parses a second spec with a valid root density of zero. Only omitted densities change. Only the original spec is compiled, so the probe cannot alter the model's inertia. 

I don't understand. This is not very clear.

> This replaces the negative-density sentinel while retaining MuJoCo's handling of defaults, inheritance, and replicated geoms. 

Which "replicated geoms" are you talking about? It seems completely unrelated? Same for inheritance.

> Metadata follows geom identity through compilation and is indexed using retained geoms' final IDs, including when visual geoms are discarded.

What is your point?

>  The unified build_model returns a typed model/metadata result; parser internals stay within the loader.

I still find it weird that 'build_model' does not have the only job of building a model. 

> Annotating URDF geoms with MuJoCo's implicit density of 1000 overrides Genesis's robot fallback of 1500, changing mass and the force-balance result. URDF geom sources therefore remain unspecified.

Not sure to understand... But this sounds like a bad design.

> The previous alignment change used a format default for unspecified mass sources. That could disagree with the material density used by dynamics, and a kinematic entity provides no density with which to resolve the ambiguity. 

Could you elaborate? Are you fixing a bug in alignment? If so, are you adding a unit test accordingly. It sounds like you are just describing how things currently works without changing anything. If so, this is just confusing and distracting.

> This preserves the existing basis for recomputed inertia. It does not add shell-inertia recomputation or recover per-geom authoring information from an already compiled MjModel.

Let's focus on what you are doing, rather than things you are intentionally not doing but nobody asked.

PR #3271, opened 2026-08-25
[BUG FIX] Preserve URDF mimic joint relations when scaling robots.

### LINKED ISSUE #3270: [Bug]: Scaling a URDF changes mimic joint coefficients and enforces the wrong joint relation
## Bug Description

Loading a URDF with `gs.morphs.URDF(scale=...)` scales both the `offset` and `multiplier` of every mimic joint. For common revolute-to-revolute and prismatic-to-prismatic mimic joints, this changes the relation authored in the URDF when the morph scale is not 1.

The parsed equality is consumed by the rigid constraint solver, so this changes the simulated joint motion rather than only the exposed metadata.

## Steps to Reproduce

The issue can be reproduced with the Panda hand asset included in Genesis:

```python
import genesis as gs

gs.init(backend=gs.cpu)
scene = gs.Scene(
    sim_options=gs.options.SimOptions(gravity=(0, 0, 0)),
        enable_collision=False,
        enable_joint_limit=False,
    ),
    show_viewer=False,
)
hand = scene.add_entity(
    gs.morphs.URDF(
        file="urdf/panda_bullet/hand.urdf",
        fixed=True,
        scale=2.0,
        default_armature=None,
    )
)
scene.build()

print(hand.equalities[0].eq_data[:2])
hand.set_qpos([0.04, 0.04])
for _ in range(100):
    scene.step()
print(hand.get_qpos())
```

Current output is approximately:

```text
[0.0, 2.0]
[0.02400014, 0.04800028]
```

`hand.urdf` declares the second prismatic finger as a default 1:1 mimic of the first. Scaling the geometry by 2 should scale linear joint positions and a linear mimic offset, but it should not change the 1:1 multiplier to 2:1.

## Expected Behavior

For `q_child = multiplier * q_parent + offset`, let revolute/continuous joint coordinates have scale exponent 0 and prismatic coordinates exponent 1. A uniform geometric scale `s` should produce:

```text
offset'     = s^child_exp * offset
multiplier' = s^(child_exp-parent_exp) * multiplier
```

In particular:

- revolute -> revolute: neither coefficient changes;
- prismatic -> prismatic: only the offset scales;
- the Panda hand's prismatic 1:1 multiplier remains 1.

## Environment

- OS: Ubuntu 22.04.5 LTS
- CPU and NVIDIA GeForce RTX 5090
- NVIDIA driver: 580.173.02
- CUDA reported by PyTorch: 12.8
- Genesis: editable current `main`, commit `70a0f4132c2fc027cb7206e06843bb9c4e0eeaa4`
- Python: 3.11.15
- MuJoCo: 3.10.0

## Additional Context

`genesis/utils/urdf.py::parse_equalities` currently assigns the two JOINT equality polynomial coefficients and then multiplies `data[:6]` by the morph scale. This treats JOINT polynomial coefficients like CONNECT/WELD spatial position data.

The behavior was also reproduced on CPU and GPU, with `n_envs=0` and `n_envs=2`, with fixed and floating roots, with fixed-link merging, and with MuJoCo compatibility enabled. A minimal revolute/prismatic matrix produces incorrect coefficients in 6 of 12 joint-type/scale combinations. Replacing only the equality coefficients with the dimensionally scaled values keeps all tested relations within approximately `1e-7` after 100 steps.

ktyang512:
## Description

Preserve the dimensionally correct URDF mimic relation when loading scaled robots. The follower offset now scales with its coordinate type, while the multiplier scales by the ratio between the follower and driver coordinate scales.

## Related Issue

Resolves https://github.com/Genesis-Embodied-AI/genesis-world/issues/3270

## Motivation and Context

Uniformly scaling a URDF currently changes common revolute-to-revolute and prismatic-to-prismatic mimic relations, causing the constraint solver to drive scaled mechanisms to unintended joint positions.

- Added scale-2 revolute/revolute, prismatic/prismatic, revolute/prismatic, and prismatic/revolute coverage to the existing URDF mimic capability test.
- Initialized every scaled follower away from its authored non-zero mimic offset and verified that stepping actively restores all four relations.
- Verified the regression test fails on the base commit with parsed coefficients `[[0.5, 4.0], [0.5, 4.0]]` for the two same-type cases.
- Passed the updated test on CPU and NVIDIA RTX 5090 GPU with `n_envs=0` and `n_envs=2`.
- Passed the adjacent JOINT, CONNECT, and WELD equality tests, the CPU Go2 scale/armature test, and both URDF mesh-scale cases.
- Passed `pre-commit` for all three changed files.

CPU targeted and adjacent validation:

```bash
uv run pytest -q tests/rigid/test_constraints.py::test_urdf_mimic tests/rigid/test_constraints.py::test_equality_joint 'tests/rigid/test_asset_loading.py::test_robot_scale_and_dofs_armature[urdf/go2/urdf/go2.urdf-cpu]' --backend=cpu
```

GPU targeted validation:

```bash
uv run pytest -q tests/rigid/test_constraints.py::test_urdf_mimic --backend=gpu
```

Additional equality and URDF scale regression validation:

```bash
uv run pytest -q tests/rigid/test_constraints.py::test_equality_link --backend=cpu
uv run pytest -q tests/parsers/test_mesh.py::test_urdf_scale --backend=cpu
```

## Checklist

- [x] I read the CONTRIBUTING document.
- [x] I followed the Submitting Code Changes section of the CONTRIBUTING document.
- [x] I tagged the title correctly.
- [x] No documentation change is needed.
- [x] I tested my changes and added instructions for reviewers.
- [x] I added tests to cover my changes.
- [ ] All existing tests passed; targeted and adjacent tests passed, but the full suite was not run locally.

### DIFF
--- a/genesis/utils/urdf.py
+++ b/genesis/utils/urdf.py
@@ -446,10 +446,16 @@ def parse_equalities(robot, morph):
                 f"Joint '{joint.name}' mimics '{joint.mimic.joint}' with multiplier {joint.mimic.multiplier} and offset {joint.mimic.offset}"
             )
 
+            mimic_joint = robot.joint_map[joint.mimic.joint]
+            follower_scale = morph.scale if joint.joint_type == "prismatic" else 1.0
+            driver_scale = morph.scale if mimic_joint.joint_type == "prismatic" else 1.0
+
+            # eq_data[0:5] stores a0..a4 in q_follower - q_follower0 = sum(a_k * (q_driver - q_driver0)^k).
+            # A model scale transforms a_k to s_f * a_k / s_d**k, where s_f and s_d are morph.scale for
+            # prismatic coordinates and 1.0 for revolute coordinates.
             eq_data = np.zeros([11])
-            eq_data[0] = joint.mimic.offset
-            eq_data[1] = joint.mimic.multiplier
-            eq_data[:6] *= morph.scale
+            eq_data[0] = follower_scale * joint.mimic.offset
+            eq_data[1] = follower_scale / driver_scale * joint.mimic.multiplier
 
             eq_descs.append(
                 EqualityDescription(
--- a/tests/rigid/conftest.py
+++ b/tests/rigid/conftest.py
@@ -196,16 +196,7 @@ def scaled_urdf_mimic():
             link = ET.SubElement(robot, "link", name=f"{pair_name}_{role}_link")
             inertial = ET.SubElement(link, "inertial")
             ET.SubElement(inertial, "mass", value="1.0")
-            ET.SubElement(
-                inertial,
-                "inertia",
-                ixx="0.01",
-                ixy="0.0",
-                ixz="0.0",
-                iyy="0.01",
-                iyz="0.0",
-                izz="0.01",
-            )
+            ET.SubElement(inertial, "inertia", ixx="0.01", ixy="0.0", ixz="0.0", iyy="0.01", iyz="0.0", izz="0.01")
 
         driver = ET.SubElement(robot, "joint", name=f"{pair_name}_driver_joint", type=driver_type)
         ET.SubElement(driver, "parent", link="base")
--- a/tests/rigid/conftest.py
+++ b/tests/rigid/conftest.py
@@ -181,17 +181,46 @@ def mimic_hinges():
 
 
 @pytest.fixture(scope="session")
+def scaled_mjcf_joint_equalities():
+    mjcf = ET.Element("mujoco", model="scaled_mjcf_joint_equalities")
+    worldbody = ET.SubElement(mjcf, "worldbody")
+    equality = ET.SubElement(mjcf, "equality")
+    for driver_type, follower_type in (
+        ("hinge", "hinge"),
+        ("slide", "slide"),
+        ("slide", "hinge"),
+        ("hinge", "slide"),
+    ):
+        pair_name = f"{driver_type}_{follower_type}"
+        driver_body = ET.SubElement(worldbody, "body", name=f"{pair_name}_driver_body")
+        ET.SubElement(driver_body, "joint", name=f"{pair_name}_driver", type=driver_type, axis="1 0 0")
+        ET.SubElement(driver_body, "geom", type="sphere", size="0.05", mass="1", contype="0", conaffinity="0")
+        follower_body = ET.SubElement(worldbody, "body", name=f"{pair_name}_follower_body")
+        ET.SubElement(follower_body, "joint", name=f"{pair_name}_follower", type=follower_type, axis="1 0 0")
+        ET.SubElement(follower_body, "geom", type="sphere", size="0.05", mass="1", contype="0", conaffinity="0")
+        ET.SubElement(
+            equality,
+            "joint",
+            name=f"{pair_name}_coupling",
+            joint1=f"{pair_name}_follower",
+            joint2=f"{pair_name}_driver",
+            polycoef="0.2 0.4 -0.3 0.2 -0.1",
+        )
+    return ET.tostring(mjcf, encoding="unicode")
+
+
+@pytest.fixture(scope="session")
 def scaled_urdf_mimic():
     robot = ET.Element("robot", name="scaled_urdf_mimic")
     ET.SubElement(robot, "link", name="base")
     joint_type_pairs = (
         ("revolute", "revolute"),
         ("prismatic", "prismatic"),
-        ("revolute", "prismatic"),
         ("prismatic", "revolute"),
+        ("revolute", "prismatic"),
     )
-    for follower_type, driver_type in joint_type_pairs:
-        pair_name = f"{follower_type}_{driver_type}"
+    for driver_type, follower_type in joint_type_pairs:
+        pair_name = f"{driver_type}_{follower_type}"
         for role in ("driver", "follower"):
             link = ET.SubElement(robot, "link", name=f"{pair_name}_{role}_link")
             inertial = ET.SubElement(link, "inertial")
--- a/tests/rigid/conftest.py
+++ b/tests/rigid/conftest.py
@@ -181,6 +181,48 @@ def mimic_hinges():
 
 
 @pytest.fixture(scope="session")
+def scaled_urdf_mimic():
+    robot = ET.Element("robot", name="scaled_urdf_mimic")
+    ET.SubElement(robot, "link", name="base")
+    joint_type_pairs = (
+        ("revolute", "revolute"),
+        ("prismatic", "prismatic"),
+        ("revolute", "prismatic"),
+        ("prismatic", "revolute"),
+    )
+    for follower_type, driver_type in joint_type_pairs:
+        pair_name = f"{follower_type}_{driver_type}"
+        for role in ("driver", "follower"):
+            link = ET.SubElement(robot, "link", name=f"{pair_name}_{role}_link")
+            inertial = ET.SubElement(link, "inertial")
+            ET.SubElement(inertial, "mass", value="1.0")
+            ET.SubElement(
+                inertial,
+                "inertia",
+                ixx="0.01",
+                ixy="0.0",
+                ixz="0.0",
+                iyy="0.01",
+                iyz="0.0",
+                izz="0.01",
+            )
+
+        driver = ET.SubElement(robot, "joint", name=f"{pair_name}_driver_joint", type=driver_type)
+        ET.SubElement(driver, "parent", link="base")
+        ET.SubElement(driver, "child", link=f"{pair_name}_driver_link")
+        ET.SubElement(driver, "axis", xyz="0 0 1")
+        ET.SubElement(driver, "limit", lower="-10", upper="10", effort="100", velocity="100")
+
+        follower = ET.SubElement(robot, "joint", name=f"{pair_name}_follower_joint", type=follower_type)
+        ET.SubElement(follower, "parent", link=f"{pair_name}_driver_link")
+        ET.SubElement(follower, "child", link=f"{pair_name}_follower_link")
+        ET.SubElement(follower, "axis", xyz="0 0 1")
+        ET.SubElement(follower, "limit", lower="-10", upper="10", effort="100", velocity="100")
+        ET.SubElement(follower, "mimic", joint=f"{pair_name}_driver_joint", multiplier="2.0", offset="0.25")
+    return ET.tostring(robot, encoding="unicode")
+
+
+@pytest.fixture(scope="session")
 def box_box():
     """Generate an MJCF model for two boxes stacked on a plane."""
     mjcf = _build_plane_contact_model("box_box", condim="3", friction="1. 0.5 0.5", plane_size="40. 40. 40.")
--- a/tests/rigid/test_constraints.py
+++ b/tests/rigid/test_constraints.py
@@ -199,8 +199,7 @@ def test_dynamic_weld_scene_reset():
 
 
 @pytest.mark.required
-@pytest.mark.parametrize("n_envs", [0, 2])
-def test_urdf_mimic(show_viewer, tol, scaled_urdf_mimic, n_envs):
+def test_urdf_mimic(show_viewer, tol, scaled_urdf_mimic):
     scene = gs.Scene(
         sim_options=gs.options.SimOptions(
             gravity=(0.0, 0.0, 0.0),
@@ -224,7 +223,7 @@ def test_urdf_mimic(show_viewer, tol, scaled_urdf_mimic, n_envs):
             fixed=True,
         ),
     )
-    scene.build(n_envs=n_envs)
+    scene.build()
     assert scene.rigid_solver.n_equalities == 5
 
     JOINT_NAMES = (
@@ -239,10 +238,9 @@ def test_urdf_mimic(show_viewer, tol, scaled_urdf_mimic, n_envs):
     )
     qs_idx_local = [mimic.get_joint(name).q_start - mimic.q_start for name in JOINT_NAMES]
     # Start every follower away from its non-zero mimic offset so the stepped assertions require active enforcement.
-    mimic.set_qpos(0.0, qs_idx_local=qs_idx_local, zero_velocity=True)
-    assert_equal(mimic.get_qpos(qs_idx_local=qs_idx_local), 0.0)
+    mimic.set_qpos(0.0, qs_idx_local=qs_idx_local)
     hand.set_dofs_velocity((0.0, 1.0))
-    for _ in range(81):
+    for _ in range(80):
         scene.step()
 
     qpos = mimic.get_qpos(qs_idx_local=qs_idx_local)
--- a/tests/rigid/test_constraints.py
+++ b/tests/rigid/test_constraints.py
@@ -29,6 +29,68 @@ def test_equality_joint(gs_sim, mj_sim, gs_solver, tol):
 
 
 @pytest.mark.required
+def test_equality_joint_scaling(show_viewer, scaled_mjcf_joint_equalities, tol):
+    scene = gs.Scene(
+        sim_options=gs.options.SimOptions(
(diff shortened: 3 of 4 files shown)


tests/rigid/test_constraints.py
```diff
@@ -199,26 +199,64 @@ def test_dynamic_weld_scene_reset():
...
+            file=scaled_urdf_mimic,
+            scale=2.0,
+            fixed=True,
+        ),
+    )
+    scene.build(n_envs=n_envs)
+    assert scene.rigid_solver.n_equalities == 5
+
+    EXPECTED_COEFFICIENTS = ((0.25, 2.0), (0.5, 2.0), (0.25, 1.0), (0.5, 4.0))
+    assert_equal((equality.eq_data[:2] for equality in mimic.equalities), EXPECTED_COEFFICIENTS)
```
jeetrex17: I think these two lines could be dropped and the test could rely on the stepped `get_qpos()` checks below. `eq_data[:2]` reflects the internal equality representation, while the [testing guidelines](https://github.com/Genesis-Embodied-AI/genesis-world/blob/70a0f4132c2fc027cb7206e06843bb9c4e0eeaa4/CODING_GUIDELINES.md#L203-L210) recommend validating high-level observable behavior.

I tried the pre-fix coefficients locally, and the qpos checks failed by roughly `0.05`, `0.19`, and `0.14` for the affected joint-type pairs, so they seem sufficient to catch the regression on their own.

tests/rigid/test_constraints.py
```diff
@@ -199,26 +199,64 @@ def test_dynamic_weld_scene_reset():
...
+    # Start every follower away from its non-zero mimic offset so the stepped assertions require active enforcement.
+    mimic.set_qpos(0.0, qs_idx_local=qs_idx_local, zero_velocity=True)
+    assert_equal(mimic.get_qpos(qs_idx_local=qs_idx_local), 0.0)
+    hand.set_dofs_velocity((0.0, 1.0))
+    for _ in range(50):
         scene.step()
 
-    gs_qpos = scene.rigid_solver.qpos.to_numpy()[:, 0]
-    assert_allclose(gs_qpos[-1], gs_qpos[-2], tol=tol)
+    CONSTRAINT_TOL = max(tol, 1e-6)
```
jeetrex17: I think we could keep `tol=tol` here and allow a few more steps for convergence instead of adding the `1e-6` floor. The floor also applies to the pre-existing Panda assertion below, so this changes its fp64 tolerance from `1e-9` to `1e-6`.

In the local CPU/fp64 probe, the maximum residual was `5.35e-7` at 50 steps and all five relations passed by 73 steps. Around 81 steps would include the 10% padding suggested by the repository's [tolerance and step-budget guidelines](https://github.com/Genesis-Embodied-AI/genesis-world/blob/70a0f4132c2fc027cb7206e06843bb9c4e0eeaa4/CODING_GUIDELINES.md#L233-L246). On the fp32 arms `tol=5e-5` already exceeds the floor, so only the fp64 CPU arm is affected.

tests/rigid/test_constraints.py
ktyang512: Removed the internal coefficient assertion. The zero initial state and stepped qpos checks now provide the regression oracle; restoring the pre-fix coefficients leaves residuals of approximately 0.0504, 0.1927, and 0.1396 at 81 steps.

tests/rigid/test_constraints.py
ktyang512: Removed the 1e-6 tolerance floor and increased the horizon to 81 steps. The corrected relations satisfy the existing fp64 tolerance at 73 steps, and 81 steps adds the measured 10% convergence padding.

genesis/utils/urdf.py
```diff
@@ -435,9 +439,8 @@ def parse_equalities(robot, morph):
             eq_info["sol_params"] = gu.default_solver_params()
 
             eq_info["data"] = np.zeros([11])
-            eq_info["data"][0] = joint.mimic.offset
-            eq_info["data"][1] = joint.mimic.multiplier
-            eq_info["data"][:6] *= morph.scale
```
duburcqa: This is not necessary? I don't remember what it corresponds to exactly.

tests/rigid/test_constraints.py
```diff
@@ -199,26 +199,60 @@ def test_dynamic_weld_scene_reset():
 
 
 @pytest.mark.required
-def test_urdf_mimic(show_viewer, tol):
-    # create and build the scene
-    scene = gs.Scene(show_viewer=show_viewer)
+@pytest.mark.parametrize("n_envs", [0, 2])
```
duburcqa: Just use n_envs=0, it should be fine. unless you have legitimate reason to expect anything to break because of batching.

tests/rigid/test_constraints.py
```diff
@@ -199,26 +199,60 @@ def test_dynamic_weld_scene_reset():
...
+        "prismatic_prismatic_driver_joint",
+        "prismatic_prismatic_follower_joint",
+        "revolute_prismatic_driver_joint",
+        "revolute_prismatic_follower_joint",
+        "prismatic_revolute_driver_joint",
+        "prismatic_revolute_follower_joint",
+    )
+    qs_idx_local = [mimic.get_joint(name).q_start - mimic.q_start for name in JOINT_NAMES]
+    # Start every follower away from its non-zero mimic offset so the stepped assertions require active enforcement.
+    mimic.set_qpos(0.0, qs_idx_local=qs_idx_local, zero_velocity=True)
```
duburcqa: Remove 'zero_velocity=True'.

tests/rigid/test_constraints.py
```diff
@@ -199,26 +199,60 @@ def test_dynamic_weld_scene_reset():
...
+    assert_equal(mimic.get_qpos(qs_idx_local=qs_idx_local), 0.0)
```
duburcqa: Remove this check. It is irrelevant.

tests/rigid/test_constraints.py
```diff
@@ -199,26 +199,60 @@ def test_dynamic_weld_scene_reset():
...
+    hand.set_dofs_velocity((0.0, 1.0))
+    for _ in range(81):
```
duburcqa: 81 is oddly specific. Prefer 80 or 100.

tests/rigid/conftest.py
```diff
@@ -125,6 +125,48 @@ def mimic_hinges():
...
+            ET.SubElement(inertial, "mass", value="1.0")
+            ET.SubElement(
+                inertial,
+                "inertia",
+                ixx="0.01",
+                ixy="0.0",
+                ixz="0.0",
+                iyy="0.01",
+                iyz="0.0",
+                izz="0.01",
```
duburcqa: Pack all these arguments on 1 line.

ktyang512:
Thanks for the review. I removed the batch parameterization, the redundant `zero_velocity` argument and initial qpos check, changed the horizon to 80 steps, and compacted the inertia construction.

Regarding the deleted scaling line, `data[0]` is the mimic offset and `data[1]` is the multiplier. For `q_f = m * q_d + b`, scaling requires `b' = s_f * b` and `m' = (s_f / s_d) * m`, since prismatic coordinates scale with the geometry while revolute coordinates do not. Therefore, `data[:6] *= morph.scale` incorrectly scales both values in the same way except for the prismatic-follower/revolute-driver case; the explicit assignments above replace it. The remaining coefficients stay zero for a linear URDF mimic.

The targeted test passes locally on CPU and GPU, and the adjacent CPU tests and Ruff checks also pass.

genesis/utils/urdf.py
```diff
@@ -421,9 +421,13 @@ def parse_urdf(morph, surface):
 
 def parse_equalities(robot, morph):
     eqs_info = []
+    joints_by_name = robot.joint_map
```
duburcqa: Keep the original name: 'joint_map = robot.joint_map'. Renaming an attribute on assignment is a pure alias and makes it harder to trace back to what it came from. The hoisting itself is fine, 'Robot.joint_map' returns a fresh copy of the dict on every access, so reading it inside the loop would copy it once per mimic joint.

genesis/utils/urdf.py
```diff
@@ -421,9 +421,13 @@ def parse_urdf(morph, surface):
...
 
     for joint in robot.joints:
         if joint.mimic:
+            mimicked_joint = joints_by_name[joint.mimic.joint]
+            joint_coordinate_scale = morph.scale if joint.joint_type == "prismatic" else 1.0
+            mimicked_joint_coordinate_scale = morph.scale if mimicked_joint.joint_type == "prismatic" else 1.0
```
duburcqa: The reason this is correct appears nowhere: prismatic coordinates are lengths and scale with the model, while revolute coordinates are angles and do not. Add it as a comment. These three lines would also read better moved down right above their only use at line 442, so the debug trace stays next to the 'if joint.mimic:' it reports on.

genesis/utils/urdf.py
duburcqa: Out of scope here, but worth a follow-up: 'parse_equalities' in 'genesis/utils/mjcf.py' scales 'data[:6]' for CONNECT and WELD and applies nothing at all for JOINT, so a scaled MJCF with a joint equality between slide joints still carries an unscaled 'a0'. Same defect class as the one fixed here for URDF.

tests/rigid/conftest.py
```diff
@@ -125,6 +125,39 @@ def mimic_hinges():
...
+def scaled_urdf_mimic():
+    robot = ET.Element("robot", name="scaled_urdf_mimic")
+    ET.SubElement(robot, "link", name="base")
+    joint_type_pairs = (
+        ("revolute", "revolute"),
+        ("prismatic", "prismatic"),
+        ("revolute", "prismatic"),
+        ("prismatic", "revolute"),
+    )
+    for follower_type, driver_type in joint_type_pairs:
```
duburcqa: 'joint_type_pairs' is ordered follower first, while the loop body creates the driver first and 'pair_name' reads 'revolute_prismatic'. Mapping 'qpos[..., 4]' back to a joint type in the test then requires spotting that inversion. Order the tuples driver first, and build 'pair_name' in that same order.

tests/rigid/test_constraints.py
```diff
@@ -199,26 +199,58 @@ def test_dynamic_weld_scene_reset():
...
+        "revolute_revolute_driver_joint",
+        "revolute_revolute_follower_joint",
+        "prismatic_prismatic_driver_joint",
+        "prismatic_prismatic_follower_joint",
+        "revolute_prismatic_driver_joint",
+        "revolute_prismatic_follower_joint",
+        "prismatic_revolute_driver_joint",
+        "prismatic_revolute_follower_joint",
+    )
+    qs_idx_local = [mimic.get_joint(name).q_start - mimic.q_start for name in JOINT_NAMES]
```
duburcqa: 'RigidJoint.qs_idx_local' already returns exactly this, so the entity offset should not be subtracted by hand: 'qs_idx_local = [idx for name in JOINT_NAMES for idx in mimic.get_joint(name).qs_idx_local]'. Note that 'q_idx_local' is the deprecated spelling of the same thing.

tests/rigid/test_constraints.py
```diff
@@ -199,26 +199,58 @@ def test_dynamic_weld_scene_reset():
...
+    # Start every follower away from its non-zero mimic offset so the stepped assertions require active enforcement.
+    mimic.set_qpos(0.0, qs_idx_local=qs_idx_local)
```
duburcqa: This is a no-op. The URDF parser assigns 'init_qpos = np.zeros(1)' to every revolute, continuous and prismatic joint, and 'build()' leaves the velocities at zero, so the followers already start at 0, which is already away from the offset. Drop both lines. The comment also states intent rather than what the code does.

genesis/utils/urdf.py
duburcqa, line 1: 'parse_equalities' in 'genesis/utils/mjcf.py' needs the same treatment: it scales 'data[:6]' for CONNECT and WELD and applies nothing at all for mjEQ_JOINT, while that same parser does scale a slide joint's coordinate ('init_qpos *= scale', and the limits by 'scale'), so a scaled MJCF carries the same defect. The law above covers both parsers and MuJoCo's full polynomial; for URDF only k of 0 and 1 ever occur, which is what this PR already does. The regression test cannot catch the MJCF half either: the only joint-equality fixture is 'mimic_hinges', hinge to hinge and never built with a scale, and hinge to hinge is the one pairing where every exponent is zero, so it passes whatever the coefficients are. Catching it needs a scaled MJCF equality with a slide joint on at least one side.

genesis/utils/urdf.py
duburcqa: 'joints_by_name' does not earn its place: read it inline as 'mimic_joint = robot.joint_map[joint.mimic.joint]'. The copy 'joint_map' makes on each access is one dict copy per mimic joint at parse time, which is not a cost worth a name.

genesis/utils/urdf.py
duburcqa: Please state here what these coefficients are and how a scale acts on them, since neither is visible from the code: 'eq_data[0:5]' are the coefficients 'a0..a4' of 'q_follower - q0 = sum_k a_k * (q_driver - q0)^k', and a model scale takes them to 'a_k' = s_f * a_k / s_d^k', where 's_f' and 's_d' are the coordinate scales of the follower and the driver - the model scale for prismatic, 1 for revolute. Two or three lines is plenty. The three lines would also read better next to their only use at line 442, leaving the debug trace beside the 'if joint.mimic:' it reports on.

genesis/utils/urdf.py
duburcqa: 'data[:6]' is the CONNECT/WELD anchor pair, two positions, which is why it was scaled. For a JOINT equality those same slots hold the polynomial coefficients 'a0..a4', so the blanket scaling was wrong on every one of them, and removing it is correct.

tests/rigid/test_constraints.py
duburcqa: This is a no-op. The URDF parser assigns 'init_qpos = np.zeros(1)' to every revolute, continuous and prismatic joint, and 'build()' leaves the velocities at zero, so the followers already start at 0, which is already away from the offset. Drop both lines. The comment also states intent rather than what the code does.

tests/rigid/conftest.py
duburcqa: 'joint_type_pairs' is ordered follower first, while the loop body creates the driver first and 'pair_name' reads 'revolute_prismatic'. Mapping 'qpos[..., 4]' back to a joint type in the test then requires spotting that inversion. Order the tuples driver first, and build 'pair_name' in that same order.

genesis/utils/urdf.py
duburcqa: Closing this one, since the request on the two scaling lines asks for exactly that to be said in the code.

tests/rigid/test_constraints.py
duburcqa: 'RigidJoint.qs_idx_local' already returns exactly this, so the entity offset should not be subtracted by hand: 'qs_idx_local = [idx for name in JOINT_NAMES for idx in mimic.get_joint(name).qs_idx_local]'. Note that 'q_idx_local' is the deprecated spelling of the same thing. Taking the order from 'JOINT_NAMES' and indexing 'qpos' positionally against it is fine, the fixture being part of the test infrastructure rather than an asset that could change underneath.

ktyang512:
Thanks for the follow-up. I rebased onto the latest main and added the same coordinate-aware scaling to MJCF joint equalities (`a_k' = s_f / s_d^k * a_k`), with a nonlinear regression covering all revolute/prismatic driver-follower pairings. I also applied the requested cleanup around naming, joint indices, lookup placement, and the no-op setup.

The targeted tests pass on CPU and GPU, along with the adjacent rigid constraint and URDF scale tests and Ruff checks.

duburcqa (approved):

PR #3283, opened 2026-08-30
[BUG FIX] Fix COM frames of fixed links on aligned bodies breaking thrust points.

### LINKED ISSUE #3264: [Bug]: Drone hovers in place and does not fly towards the target, behavior differs from official demo
### Bug Description

Hi,
I followed the official documentation for the Hover Environment step by step, but the result I got is vastly different from the official demo video.
Environment:
Genesis Version: v1.3.3
Command used: python examples/drone/hover_train.py -e drone-hovering -b 8192 --max-iterations 301

The Issue:
In the official demo, the drone flies smoothly to the target, and the red target sphere continuously resamples (moves) when reached, showing a continuous target-following behavior.
However, in my run, the drone remains in a hovering state in place and does not fly towards the target at all. It just stably hovers where it spawned, and the red target sphere never moves.
I suspected it might be an underfitting issue, so I increased the training steps to 1100 iterations, but the result is exactly the same: hovering in place, making no attempt to reach the target.
Could you please clarify what is the reason for this? Is there a configuration missing in the tutorial, or is this the expected behavior for the default script at this stage?
Any guidance would be appreciated!

### Steps to Reproduce

python examples/drone/hover_train.py -e drone-hovering -b 8192 --max-iterations 301
python examples/drone/hover_eval.py -e drone-hovering --ckpt 300 --record

### Expected Behavior

Train the model using a Colab GPU first, then transfer/convert it for CPU inference.

### Screenshots/Videos

https://github.com/user-attachments/assets/ab895ff5-825e-4f2d-b6bd-670dd0818629

### Relevant log output

```shell

```

### Environment

- OS: Ubuntu 22.04

  v1.3.3

jeetrex17:
## Description

Free-root alignment moves a rigid body's root frame to the composite center of mass and principal axes of its fixed
subtree, re-expressing the geoms so the world geometry is unchanged. It also reset every fixed child's inertial frame
to the link origin with identity orientation. That frame is the public `link_COM` reference of `get_links_pos`,
`apply_links_external_wrench` and the link inverse weight. The shipped CF2X drone stores each rotor arm in its
propeller link's inertial origin, so alignment collapsed all four thrust points onto the root and differential thrust
produced no roll or pitch torque.

Each fixed child now keeps its authored inertial frame, re-expressed through the same transform as its geoms, while
its mass and inertia still fold into the composite. The diagonal mass matrix of an aligned body is unchanged. The
inverse weight of a kept fixed child now equals its `align=False` value, which changes the constraint regularization
of contacts and joint limits on such links.

`test_align_preserves_link_local_frames` builds aligned and unaligned twins, plain and heterogeneous, with posed
collision and visual geoms, and asserts world geom poses, `link_COM` frames, inverse weights and the angular response
to a local torque. `test_drone_propellers_force_application` asserts the propeller `link_COM` points, the vertical
Euler recurrence across substep configurations, and the roll rate from differential thrust.

## Related Issue

Resolves #3264. Supersedes #3198.

## Motivation and Context

Alignment promises to preserve user-authored frames, and `link_COM` is one of them. Fixing it at the source keeps
custom URDFs and every `link_COM` consumer consistent with `align=False`, instead of adding a drone-specific
application point or disabling alignment.

## Before / After

With the same camera and differential-thrust input, upstream `main` stays level while this branch rolls.

https://github.com/user-attachments/assets/97d1adfc-df3f-415b-b135-4c94941efe3d

https://github.com/user-attachments/assets/71c2e6a2-820c-4e66-8122-8074982ce486

```bash
pytest -v --backend cpu tests/rigid/test_asset_loading.py tests/rigid/test_control.py -k "align or drone"
```

Passes on macOS CPU at this revision, and on Ubuntu CPU and CUDA at the previous one. The new assertions fail on
`main`.

## Checklist:

- [x] I updated the [documentation](https://github.com/Genesis-Embodied-AI/genesis-doc) accordingly or no change is
  needed.


genesis/engine/entities/rigid_entity/rigid_entity.py
```diff
@@ -1098,16 +1105,22 @@ def _align_free_roots(self):
                     # links skipped above (a geometry-less link carries only the 'gs.EPS' placeholder) must not
                     # inflate the composite.
                     real_total = sum(
-                        float(link._variant_inertial[v][0] if is_heterogeneous else link.desc.mass)
+                        float(link._variant_inertial[v].mass if is_heterogeneous else link.desc.mass)
```
duburcqa: why converting to float?

genesis/engine/entities/rigid_entity/rigid_link.py
```diff
@@ -164,10 +164,12 @@ def compose_inertial_from_geoms(g_descs: Sequence[GeomDescription], rho: float)
...
 class LinkInertial(NamedTuple):
-    """A link's (or variant's) finalized inertial in the solver's representation: mass, center of mass 'com', and the
-    inertia tensor 'inertia' expressed in the principal frame 'quat' (identity when derived from geometry). Distinct
-    from 'InertialProperties', whose 'i' is a single tensor; the solver stores the principal tensor and its orientation
-    separately."""
+    """Store a link's or heterogeneous variant's finalized inertial properties.
+
+    ``quat`` defines the frame in which ``inertia`` is expressed. For fixed children folded into an aligned root,
+    ``com`` and ``quat`` retain the public ``link_COM`` frame while ``inertia`` is zero. The tensor and its orientation
+    are stored separately.
```
duburcqa: Please document each attribute in their own attribute docstring, rather than class docstring.

tests/rigid/conftest.py
```diff
@@ -555,6 +555,73 @@ def implicit_inertial_origin_chain():
...
+            ixx, iyy, izz = inertia
+            ET.SubElement(
+                inertial,
+                "inertia",
+                ixx=str(ixx),
+                ixy="0",
+                ixz="0",
+                iyy=str(iyy),
+                iyz="0",
+                izz=str(izz),
```
duburcqa: 1 line.

tests/rigid/conftest.py
```diff
@@ -555,6 +555,73 @@ def implicit_inertial_origin_chain():
...
+            ET.SubElement(collision, "origin", xyz=geom_pos, rpy=geom_rpy)
+
+            mass, com, inertial_rpy, inertia = links_inertial[i_link]
+            inertial = ET.SubElement(link, "inertial")
+            ET.SubElement(inertial, "mass", value=str(mass))
+            ET.SubElement(
+                inertial,
+                "origin",
+                xyz=" ".join(map(str, com)),
+                rpy=" ".join(map(str, inertial_rpy)),
```
duburcqa: 1 line.

tests/rigid/conftest.py
```diff
@@ -555,6 +555,73 @@ def implicit_inertial_origin_chain():
...
+            )
+
+        joint = ET.SubElement(urdf, "joint", name="fixed_joint", type="fixed")
+        ET.SubElement(joint, "parent", link="base")
+        ET.SubElement(joint, "child", link="child")
+        ET.SubElement(
+            joint,
+            "origin",
+            xyz=" ".join(map(str, joint_pos)),
+            rpy=" ".join(map(str, joint_rpy)),
```
duburcqa: 1 line.

tests/rigid/test_asset_loading.py
```diff
@@ -1049,56 +1049,106 @@ def test_align_mixed_mass_raises():
 
 
 @pytest.mark.required
-def test_align_relative_offset_on_link_relative_geoms(show_viewer, tol):
-    # To exercise the geom-frame offset strip the geoms MUST sit at non-identity poses relative to their link (explicit
-    # collision/visual <origin>) AND the morph offset MUST be a rotation that does not commute with them - otherwise the
-    # conjugation degenerates to the plain morph offset and a naive (corrupted) strip would still pass. A convex-
-    # decomposed mesh is useless here: its sub-geoms keep an identity frame (geometry lives in the vertices).
```
duburcqa, line 1149: What about this comment?

tests/rigid/test_asset_loading.py
```diff
@@ -1049,56 +1049,106 @@ def test_align_mixed_mass_raises():
...
+                align=True,
+                merge_fixed_links=False,
+            )
+            for file in inertial_urdfs
+        ),
+        vis_mode="collision",
+    )
+    # A heterogeneous entity needs one environment per morph variant.
+    scene.build(n_envs=2)
+    entities = (plain_aligned, plain_unaligned, heterogeneous_aligned)
```
duburcqa: This is useless. You can just do scene.entities.

tests/rigid/test_asset_loading.py
```diff
@@ -1049,56 +1049,106 @@ def test_align_mixed_mass_raises():
...
 
-    # The user orientation is identity, so the world<-user offset rotates each geom about the link origin:
-    # geom_world_pos = U_pos + R(offset) * (geom_user_pos - U_pos) and geom_world_quat = offset * geom_user_quat.
```
duburcqa, line 1184: What about this comment?

tests/rigid/test_asset_loading.py
```diff
@@ -1049,56 +1049,106 @@ def test_align_mixed_mass_raises():
...
+    joint_quat = gu.xyz_to_quat(np.array(joint_rpy), rpy=True)
+    children_com_offset, composites_com_offset = [], []
+    for (base_mass, base_com, _, _), (child_mass, child_com, _, _) in variants_inertial:
+        child_offset = np.array(joint_pos) + gu.transform_by_quat(np.array(child_com), joint_quat)
+        total_mass = base_mass + child_mass
+        children_com_offset.append(gu.transform_by_quat(child_offset, offset_quat))
+        composite_offset = (base_mass * np.array(base_com) + child_mass * child_offset) / total_mass
+        composites_com_offset.append(gu.transform_by_quat(composite_offset, offset_quat))
+    children_com_offset = np.array(children_com_offset)
+    composites_com_offset = np.array(composites_com_offset)
```
duburcqa: Are you sure casting to np.array is necessary?

tests/rigid/test_asset_loading.py
```diff
@@ -1049,56 +1049,106 @@ def test_align_mixed_mass_raises():
...
-            group = ET.SubElement(link, group_tag)
-            geom_el = ET.SubElement(group, "geometry")
-            ET.SubElement(geom_el, "box", size="0.05 0.1 0.15")
-            ET.SubElement(group, "origin", rpy=origin_rpy, xyz=origin_xyz)
-    urdf = urdfpy.URDF._from_xml(robot, robot, get_assets_dir())
-
-    BODY_POS = (0.0, 0.0, 0.2)
-    OFFSET_EULER = (20.0, 35.0, 50.0)  # a generic rotation that does not commute with the geom poses
+def test_align_preserves_link_local_frames(aligned_link_frame_urdfs, show_viewer, tol):
+    inertial_urdfs, joint_pos, joint_rpy, variants_inertial = aligned_link_frame_urdfs
```
duburcqa: Please read these information from entity description info instead of passing this as part of the fixture output.

tests/rigid/test_asset_loading.py
```diff
@@ -1049,56 +1049,106 @@ def test_align_mixed_mass_raises():
...
+        assert_allclose(child_com - entity.get_pos(), com_offset, tol=tol)
+        assert_allclose(entity.get_pos(relative=False) - entity.get_pos(), anchor_offset, tol=tol)
+
+    # A pure local torque isolates the inertial-frame orientation because it has no application-point arm.
+    for entity in entities:
+        entity.apply_links_external_wrench(
+            torque=TORQUE,
+            links_idx_local=[1],
+            ref=gs.link_ref_frame.link_COM,
+            local=True,
```
duburcqa: 1 line.

tests/rigid/test_asset_loading.py
```diff
@@ -1049,56 +1049,106 @@ def test_align_mixed_mass_raises():
...
+        )
+    scene.step()
+    assert_allclose(
+        plain_aligned.get_links_ang(links_idx_local=[1]), plain_unaligned.get_links_ang(links_idx_local=[1]), tol=tol
+    )
+    assert_allclose(
+        heterogeneous_aligned.get_links_ang(links_idx_local=[1], envs_idx=[1]),
```
duburcqa: Pass 1 instead of [1].

tests/rigid/test_asset_loading.py
```diff
@@ -1049,56 +1049,106 @@ def test_align_mixed_mass_raises():
...
+        plain_unaligned.get_links_ang(links_idx_local=[1], envs_idx=[1]),
```
duburcqa: Pass 1 instead of [1].

tests/rigid/test_asset_loading.py
```diff
@@ -1049,56 +1049,106 @@ def test_align_mixed_mass_raises():
...
+    for entity in entities:
+        entity.apply_links_external_wrench(
+            torque=TORQUE,
+            links_idx_local=[1],
+            ref=gs.link_ref_frame.link_COM,
+            local=True,
+        )
+    scene.step()
+    assert_allclose(
+        plain_aligned.get_links_ang(links_idx_local=[1]), plain_unaligned.get_links_ang(links_idx_local=[1]), tol=tol
```
duburcqa: Pass 1 instead of [1].

tests/rigid/test_asset_loading.py
```diff
@@ -1049,56 +1049,106 @@ def test_align_mixed_mass_raises():
...
+    ):
+        child_com = entity.get_links_pos(links_idx_local=[1], ref=gs.link_ref_frame.link_COM)[..., 0, :]
+        assert_allclose(child_com - entity.get_pos(), com_offset, tol=tol)
+        assert_allclose(entity.get_pos(relative=False) - entity.get_pos(), anchor_offset, tol=tol)
+
+    # A pure local torque isolates the inertial-frame orientation because it has no application-point arm.
+    for entity in entities:
+        entity.apply_links_external_wrench(
+            torque=TORQUE,
+            links_idx_local=[1],
```
duburcqa: Pass 1 instead of [1].

tests/rigid/test_asset_loading.py
```diff
@@ -1049,56 +1049,106 @@ def test_align_mixed_mass_raises():
...
+        composite_offset = (base_mass * np.array(base_com) + child_mass * child_offset) / total_mass
+        composites_com_offset.append(gu.transform_by_quat(composite_offset, offset_quat))
+    children_com_offset = np.array(children_com_offset)
+    composites_com_offset = np.array(composites_com_offset)
+    for entity, com_offset, anchor_offset in (
+        (plain_aligned, children_com_offset[1], composites_com_offset[1]),
+        (plain_unaligned, children_com_offset[1], 0.0),
+        (heterogeneous_aligned, children_com_offset, composites_com_offset),
+    ):
+        child_com = entity.get_links_pos(links_idx_local=[1], ref=gs.link_ref_frame.link_COM)[..., 0, :]
```
duburcqa: Pass 1 instead of [1].

duburcqa: Is '[..., 0, :]' really necessary?

tests/rigid/test_control.py
```diff
@@ -141,59 +141,94 @@ def test_reset_control(robot_path, tol):
 
 
 @pytest.mark.required
-def test_drone_propellers_force_substep_consistency(show_viewer, tol):
```
duburcqa, line 144: Why did you remove the original test about substep consistency? Your new test is not checking the same thing.

tests/rigid/test_control.py
```diff
@@ -141,59 +141,94 @@ def test_reset_control(robot_path, tol):
...
+    scene.reset()
 
-    pos_ref = drone_ref.get_dofs_position()
-    pos_test = drone_test.get_dofs_position()
-    assert_allclose(pos_ref, pos_test, tol=tol)
+    # One differential-thrust step isolates the rotor-arm moment and cancels pitch and yaw torque.
+    roll_rpm = tuple(BASE_RPM * scale for scale in (1 + RPM_DELTA, 1 + RPM_DELTA, 1 - RPM_DELTA, 1 - RPM_DELTA))
+    drone.set_propellers_rpm(roll_rpm)
+    scene.step()
+    thrust = float(np.float32(drone.KF)) * np.square(roll_rpm)
```
duburcqa: "float(np.float32(drone.KF))" -> WTF is this nightmare?

tests/rigid/test_control.py
```diff
@@ -141,59 +141,94 @@ def test_reset_control(robot_path, tol):
...
-        drone_test.set_propellers_rpm(BASE_RPM)
-        scene_test.step()
+    # Thrust held over several steps: the propeller forces set once per step act on every substep and are cleared
+    # before the next step, so the vertical response follows the implicit-damping Euler recurrence exactly. The
+    # recurrence depends only on the total number of substeps, which the sweep pins across different outer step counts.
+    for _ in range(n_steps):
+        drone.set_propellers_rpm(BASE_RPM)
+        scene.step()
+    # The thrust kernel uses a 32-bit coefficient in every solver precision.
+    thrust = drone.n_propellers * float(np.float32(drone.KF)) * BASE_RPM**2
```
duburcqa: "float(np.float32(drone.KF))" -> WTF is this nightmare?

tests/rigid/test_control.py
```diff
@@ -141,59 +141,94 @@ def test_reset_control(robot_path, tol):
...
-        scene_ref.step()
+    # Reset clears the tracked external force and accepts each supported RPM input shape.
+    drone.set_propellers_rpm(BASE_RPM)
+    with np.testing.assert_raises(gs.GenesisException):
+        drone.set_propellers_rpm(BASE_RPM)
+    scene.reset()
+    drone.set_propellers_rpm((BASE_RPM,) * 4)
+    scene.reset()
+    rpm_shape = (n_envs, 4) if n_envs else (4,)
+    drone.set_propellers_rpm(torch.full(rpm_shape, fill_value=BASE_RPM))
```
duburcqa, line 198: `drone.set_propellers_rpm(BASE_RPM)` should work. If not, it is a bug.

tests/rigid/test_control.py
duburcqa: Maybe your unit test is superior by comparing against ground truth, but make sure it should be pose-based, not velocity, acceleration, or force based, to make into account gravity and time integration effects.

tests/rigid/test_asset_loading.py
jeetrex17: It is. Even with `links_idx_local=1`, the getter keeps the link dimension and returns `(n_envs, 1, 3)` here, while `get_pos()` returns `(n_envs, 3)`. Removing the slice would broadcast the subtraction to `(n_envs, n_envs, 3)` and compare the wrong environment pairs. I will use the scalar index but keep the slice.

tests/rigid/test_control.py
jeetrex17: I replaced the original 500-step plus 100-step trajectory with a shorter analytical check because the test guidelines strongly discourage more than 300 steps and recommend an Euler-exact recurrence with implicit damping. Both parametrizations still execute the same ten substeps through different outer-step counts and compare against the same ground truth. The missing piece is pose integration, so I will extend the recurrence to position and assert the final pose as well.

tests/rigid/test_control.py
jeetrex17: This matches the precision used by the kernel: `kernel_set_drone_rpm` declares `KF` as `qd.float32`, even when the solver runs in fp64. `np.float32(...)` reproduces that boundary rounding, and the outer `float(...)` keeps the rest of the expected-value calculation in fp64. I will compute it once as a lowercase `kf` with a short comment instead of repeating the opaque expression.

tests/rigid/test_control.py
jeetrex17: This matches the precision used by the kernel: `kernel_set_drone_rpm` declares `KF` as `qd.float32`, even when the solver runs in fp64. `np.float32(...)` reproduces that boundary rounding, and the outer `float(...)` keeps the rest of the expected-value calculation in fp64. I will compute it once as a lowercase `kf` with a short comment instead of repeating the opaque expression.

genesis/engine/entities/rigid_entity/rigid_entity.py
jeetrex17: The conversion was already present before this PR; this line only changed `[0]` to the named `.mass` field. It normalized possible NumPy scalar values, but the sum does not need that conversion here. I will remove it and retain `.mass`.

tests/rigid/test_asset_loading.py
jeetrex17: Makes sense. I will have the fixture return only the URDF data and read the joint and inertial values from the entity descriptions. The expectations will come from the unaligned/source descriptions, since reading the transformed aligned descriptions would make the check circular.

tests/rigid/test_control.py
jeetrex17: Scalar input already works and is exercised by the first call in this block. The shaped tensor was carried over from the original test to preserve batched-input coverage, but that is unrelated to this fix. I will simplify this call to use the scalar.

tests/rigid/test_asset_loading.py
jeetrex17: That explanation is still useful because it shows why the geometry check cannot pass with a degenerate transform. I will restore it in a shorter form that matches the new fixture.

tests/rigid/test_asset_loading.py
jeetrex17: The formula no longer described the twin-equality assertion, but dropping the closed-form check weakened the test. I will restore the formula-based assertions on the unaligned reference, then compare the aligned geometry against that reference.

genesis/engine/entities/rigid_entity/rigid_entity.py
duburcqa: > The conversion was already present before this PR;

This is not a good excuse. We are trying to raise code quality constantly.

> It normalized possible NumPy scalar values

We don't do this in Genesis. You are supposed to know which types you are dealing with. It is not supposed to be liquid. We strive to be strongly typed, even if there is still a long road ahead. Besides, normalizing NumPy scalar is irrelevant. You should be able to operate on them like any python native float.

tests/rigid/test_asset_loading.py
duburcqa: > reading the transformed aligned descriptions would make the check circular

ok let's see if it is doable. Otherwise, you can just hardcode the value in the test directly since this asset is managed by the unit test infra itself anyway.

tests/rigid/test_asset_loading.py
duburcqa: I see. This would be avoided by doing:

```python
        assert_allclose(child_com, com_offset + entity.get_pos(), tol=tol)
```

Which arguably makes more sense / is more intuitive anyway no?

tests/rigid/test_control.py
duburcqa: You should have a look to what the unit test was actually validating by passing '--vis' to pytest. I don't remember exactly but I suspect it was checking more than the name suggests. But maybe not, and then being able to shorten the test would indeed be very welcome.

tests/rigid/test_control.py
duburcqa: I don't think we care much about such concerns. If the unit test requires such a trick to discriminate success from failure, then your test is too fragile in the first place and should be reworked.

tests/rigid/test_control.py
duburcqa: > The shaped tensor was carried over from the original test to preserve batched-input coverage, but that is unrelated to this fix.

Actually that makes sense. Let's keep it!

duburcqa:
@jeetrex17 A new release is planned for tomorrow. It would be nice if this could go in.

duburcqa (approved):

PR #3286, opened 2026-09-02
[MISC] Fix inverted --pile-type condition in the pyramid collision example.

### LINKED ISSUE #3285: [Bug]: --pile-type option values are inverted in examples/collision/pyramid.py
### Bug Description

In `examples/collision/pyramid.py` (line 44), the condition controlling `box_spacing` compares against the wrong `--pile-type` value, so each option produces the behavior of the other one.

```python
box_spacing = (1.0 - 1e-3 + 0.1 * (args.pile_type == "static")) * box_size
```

Box positions are `offset + box_spacing * [i + 0.5j, 0, j]`, with `box_size = 0.25` and the offset placing the bottom layer flush on the plane.
That gives:

| `--pile-type` | factor | `box_spacing` | gap between adjacent boxes |
| --- | --- | --- | --- |
| `static` | `1.0 - 1e-3 + 0.1` = 1.099 | 0.27475 | **+0.02475** — boxes start detached and fall |
| `falling` | `1.0 - 1e-3` = 0.999 | 0.24975 | −0.00025 — boxes start in contact, already settled |

The `- 1e-3` term exists to seed contacts at `t = 0`, i.e. to build the settled configuration. So it should be the `static` branch that omits the extra `0.1` spacing, not the `falling` one. As written, `--pile-type static` drops the
pyramid into place and `--pile-type falling` starts it fully settled, which contradicts both the option names and the help string ("Whether the pile starts settled or falls into place").

This also affects the default run, since `--pile-type` defaults to `"falling"`.

### Steps to Reproduce

Run the example with each option and watch the first few steps:

```bash
python examples/collision/pyramid.py --pile-type static --vis
python examples/collision/pyramid.py --pile-type falling --vis
```

`static` shows the pyramid dropping into place; `falling` shows it already settled and motionless.

The spacing can also be checked without running the simulation(in examples/collision/pyramid.py:43):

```python
...
box_size = 0.25
box_spacing = (1.0 - 1e-3 + 0.1 * (args.pile_type == "static")) * box_size
box_pos_offset = (-0.5, 1, 0.0) + 0.5 * np.array([box_size, box_size, box_size])
boxes = {}
for i in range(args.num_cubes):
    for j in range(args.num_cubes - i):
        box = scene.add_entity(
            gs.morphs.Box(
                size=[box_size, box_size, box_size],
                pos=box_pos_offset + box_spacing * np.array([i + 0.5 * j, 0, j]),
            ),
        )
...
```

### Expected Behavior

`--pile-type static` should build a pyramid that is already settled and stays at rest, and `--pile-type falling` should build one that starts detached and falls into place — matching the option names and the help string.

The condition should compare against `"falling"`:

```python
box_spacing = (1.0 - 1e-3 + 0.1 * (args.pile_type == "falling")) * box_size
```

### Screenshots/Videos

_No response_

### Relevant log output

```shell

```

### Environment

- OS: MacOS 26.6.2
- GPU/CPU: M3
- GPU-driver version: N/A

Note: this is a static logic error in an example script, so it is independent of platform and backend.

6a2a1f0c8038c78452cb7ef080cca7f4c38c3a7c

### Additional Context

Happy to open a PR with the one-line fix if this is confirmed.

Note the fix changes the default run (`--pile-type` defaults to `"falling"`) from a settled pyramid to a falling one. `tests/test_examples.py` only asserts a zero exit code, so the examples CI job is unaffected.

yeduk3:
## Description

`examples/collision/pyramid.py` compares `--pile-type` against the wrong value, so each option produces the behavior of the other one.

```python
box_spacing = (1.0 - 1e-3 + 0.1 * (args.pile_type == "static")) * box_size
```

With `box_size = 0.25`:

| `--pile-type` | `box_spacing` | gap between adjacent boxes |
| --- | --- | --- |
| `static` | 0.27475 | +0.02475 — detached, falls into place |
| `falling` | 0.24975 | -0.00025 — in contact, already settled |

The `- 1e-3` term seeds contacts at `t = 0`, i.e. builds the settled pile, so the extra `0.1` spacing belongs to the `falling` branch. This changes the comparison to `"falling"`.

## Related Issue

Resolves #3285

## Motivation and Context

As written, `--pile-type static` drops the pyramid into place and `--pile-type falling` starts it fully settled, contradicting both the option names and the help string ("Whether the pile starts settled or falls into place"). The default is also switched from `"falling"` to `"static"`, so that running the example with no arguments keeps producing the settled pyramid it did before.

Example script only — no library API or solver behavior is touched.

```bash
python examples/collision/pyramid.py --pile-type static --vis
python examples/collision/pyramid.py --pile-type falling --vis
```

`static` now stays at rest; `falling` now drops into place.

`tests/test_examples.py` runs this script with no arguments and only asserts a zero exit code, so the examples CI job passes either way:

```bash
pytest -v -m examples tests/test_examples.py -k pyramid
```

## Checklist:

### DIFF
--- a/examples/collision/pyramid.py
+++ b/examples/collision/pyramid.py
@@ -41,7 +41,7 @@ def main():
 
     # create pyramid of boxes
     box_size = 0.25
-    box_spacing = (1.0 - 1e-3 + 0.1 * (args.pile_type == "static")) * box_size
+    box_spacing = (1.0 - 1e-3 + 0.1 * (args.pile_type == "falling")) * box_size
     box_pos_offset = (-0.5, 1, 0.0) + 0.5 * np.array([box_size, box_size, box_size])
     boxes = {}
     for i in range(args.num_cubes):
--- a/examples/collision/pyramid.py
+++ b/examples/collision/pyramid.py
@@ -8,7 +8,7 @@ def main():
     parser.add_argument(
         "--pile-type",
         type=str,
-        default="falling",
+        default="static",
         choices=("static", "falling"),
         help="Whether the pile starts settled or falls into place",
     )


examples/collision/pyramid.py
duburcqa: Please change the default to static.

yeduk3: Done in 305a4e6. The default is now `static`, so running the example with no
arguments produces the same settled pyramid as before this PR.

duburcqa (approved):

PR #3290, opened 2026-09-02
[BUG FIX] Support MJCF joint equalities without joint2 in the rigid solver.

### LINKED ISSUE #3289: [Bug]: MJCF joint equality without joint2 fails to load
## Bug Description

MuJoCo allows an MJCF `joint` equality to omit `joint2`. This form fixes `joint1` to a constant, but Genesis currently fails while adding the entity because it tries to resolve the missing second joint by name.

## Steps to Reproduce

```python
import genesis as gs

MODEL = """
<mujoco model="single_joint_equality">
  <worldbody>
    <body name="target_body">
      <joint name="target" type="slide" axis="1 0 0"/>
      <geom type="sphere" size="0.05" mass="1" contype="0" conaffinity="0"/>
    </body>
    <body name="unrelated_body">
      <joint name="unrelated" type="slide" axis="1 0 0"/>
      <geom type="sphere" size="0.05" mass="1" contype="0" conaffinity="0"/>
    </body>
  </worldbody>
  <equality>
    <joint name="fixed_target" joint1="target" polycoef="0.25 1 0 0 0"/>
  </equality>
</mujoco>
"""

gs.init(backend=gs.cpu)
scene = gs.Scene(
    rigid_options=gs.options.RigidOptions(enable_collision=False, enable_joint_limit=False),
    show_viewer=False,
)
entity = scene.add_entity(gs.morphs.MJCF(file=MODEL))
scene.build()
```

Current Genesis raises:

```text
genesis.GenesisException: Neither `name` nor `uid` is provided.
```

## Expected Behavior

The model should load and the equality should constrain `target` to its reference position plus the constant coefficient `0.25`. The omitted second joint must not alias another joint in the model. This matches [MuJoCo's documented semantics](https://mujoco.readthedocs.io/en/stable/XMLreference.html#equality-joint) for a joint equality without `joint2`.

## Environment

- OS: Ubuntu 22.04.5 LTS
- CPU: Intel Xeon w5-2455X
- GPU, GPU driver, and CUDA: N/A; reproduced on CPU before backend-specific simulation begins
- Genesis current `main`: `6a2a1f0c8038c78452cb7ef080cca7f4c38c3a7c`
- Genesis version: 1.3.3 editable
- Python: 3.11.15
- MuJoCo: 3.10.0

Assisted-by: OpenAI Codex (GPT-5)

ktyang512:
## Description

Support MJCF joint equalities without `joint2` in the rigid solver, including model scaling and gradients. The constrained joint follows the specified constant while unrelated joints remain unaffected. SAP retains its build-time rejection for this unsupported case.

## Related Issue

Resolves #3289

9 targeted cases passed in this test-only revision: the 50-step CPU parity test using the comprehensive equality model, one-step scaling checks in both batching configurations on CPU fp64 and RTX 5090 fp32, and adjacent CPU joint-equality checks. The previous revision also passed all five CPU equality-gradient cases, which are unchanged. MuJoCo 3.10.0 was used. The expanded parity test still exposes the missing-joint setup error on the unfixed base `b3c6c73a`. Ruff 0.14.11 lint and format checks passed. The full suite and compatibility with newer main commits were not tested in this revision.

```sh
PYTHONPATH="$PWD" QD_OFFLINE_CACHE=0 pytest -n 0 tests/rigid/test_mujoco_parity.py::test_equality_joint tests/rigid/test_constraints.py::test_equality_joint_scaling tests/rigid/test_constraints.py::test_equality_joint --backend cpu
PYTHONPATH="$PWD" QD_OFFLINE_CACHE=0 pytest -n 0 tests/rigid/test_constraints.py::test_equality_joint_scaling --backend gpu
```

### DIFF
--- a/genesis/engine/couplers/sap_coupler.py
+++ b/genesis/engine/couplers/sap_coupler.py
@@ -223,6 +223,10 @@ class SAPCoupler(RBC):
         self._enable_rigid_fem_contact &= self.rigid_solver.is_active and self.fem_solver.is_active
         self._enable_fem_self_tet_contact &= self.fem_solver.is_active
 
+        for equality in self.rigid_solver.equalities:
+            if equality.type == gs.EQUALITY_TYPE.JOINT and equality.eq_obj2id < 0:
+                gs.raise_exception("SAPCoupler does not support JOINT equality constraints without `joint2`.")
+
         init_tet_tables = False
 
         if self.fem_solver.is_active:
--- a/genesis/engine/solvers/rigid/constraint/backward.py
+++ b/genesis/engine/solvers/rigid/constraint/backward.py
@@ -825,15 +825,15 @@ def kernel_manual_add_equality_constraints_bw(
 ):
     """Manual reverse of `add_equality_constraints` (JOINT + CONNECT + WELD).
 
-    * JOINT   - couples two scalar dofs via a quartic polynomial (1 row).
+    * JOINT   - constrains one scalar dof, optionally against another via a quartic polynomial (1 row).
     * CONNECT - 3 rows pinning `global_anchor1` to `global_anchor2` in world.
     * WELD    - 6 rows: 3 position + 3 orientation, all sharing a single
                 combined `pos_imp = ||all_error||` (6D).
 
     Accumulates into kinematic-state grads only (conservative):
         JOINT:
-            rigid_info.qpos.grad[i_qpos1], qpos.grad[i_qpos2]
-            dyn_state.dofs.vel.grad[i_dof1], vel.grad[i_dof2]
+            rigid_info.qpos.grad[i_qpos1], and qpos.grad[i_qpos2] when joint2 exists
+            dyn_state.dofs.vel.grad[i_dof1], and vel.grad[i_dof2] when joint2 exists
         CONNECT / WELD:
             dyn_state.links.{pos, quat, root_COM, cd_ang, cd_vel}.grad[link1 / link2]
             dyn_state.dofs.{cdof_ang, cdof_vel, cdofd_ang, cdofd_vel, vel}.grad over each link chain
@@ -841,13 +841,13 @@ def kernel_manual_add_equality_constraints_bw(
     `dyn_info.links.invweight`) are not differentiated.
 
     Forward recap (per equality of type JOINT):
-        diff = qpos[i_qpos2] - qpos0[i_qpos2]
+        diff = qpos[i_qpos2] - qpos0[i_qpos2], or 0 when joint2 is omitted
         pos_poly = a0 + a1 * diff + a2 * diff^2 + a3 * diff^3 + a4 * diff^4
         pos = qpos[i_qpos1] - qpos0[i_qpos1] - pos_poly
         deriv = d(pos_poly)/d(diff) = a1 + 2 * a2 * diff + 3 * a3 * diff^2 + 4 * a4 * diff^3
         jac[n_con, i_dof1] = 1.0
-        jac[n_con, i_dof2] = -deriv
-        jac_qvel = vel[i_dof1] - deriv * vel[i_dof2]
+        jac[n_con, i_dof2] = -deriv, when joint2 exists
+        jac_qvel = vel[i_dof1] - deriv * vel[i_dof2], with the second term omitted when joint2 is omitted
         imp, aref = imp_aref(sol_params, -|pos|, jac_qvel, pos)
             aref = -b * jac_qvel - k * imp * pos
         diag = max(invweight * (1 - imp) / imp, EPS); efc_D = 1/diag
@@ -874,22 +874,21 @@ def kernel_manual_add_equality_constraints_bw(
                     if qd.static(rigid_config.batch_joints_info)
                     else dyn_info.equalities.eq_obj1id[i_e, i_b]
                 )
-                I_joint2 = (
-                    [dyn_info.equalities.eq_obj2id[i_e, i_b], i_b]
-                    if qd.static(rigid_config.batch_joints_info)
-                    else dyn_info.equalities.eq_obj2id[i_e, i_b]
-                )
                 i_qpos1 = dyn_info.joints.q_start[I_joint1]
-                i_qpos2 = dyn_info.joints.q_start[I_joint2]
                 i_dof1 = dyn_info.joints.dof_start[I_joint1]
-                i_dof2 = dyn_info.joints.dof_start[I_joint2]
                 I_dof1 = [i_dof1, i_b] if qd.static(rigid_config.batch_dofs_info) else i_dof1
-                I_dof2 = [i_dof2, i_b] if qd.static(rigid_config.batch_dofs_info) else i_dof2
 
                 pos1 = rigid_info.qpos[i_qpos1, i_b]
-                pos2 = rigid_info.qpos[i_qpos2, i_b]
                 ref1 = rigid_info.qpos0[i_qpos1, i_b]
-                ref2 = rigid_info.qpos0[i_qpos2, i_b]
+                i_joint2 = dyn_info.equalities.eq_obj2id[i_e, i_b]
+                i_qpos2 = -1
+                i_dof2 = -1
+                diff = gs.qd_float(0.0)
+                if i_joint2 >= 0:
+                    I_joint2 = [i_joint2, i_b] if qd.static(rigid_config.batch_joints_info) else i_joint2
+                    i_qpos2 = dyn_info.joints.q_start[I_joint2]
+                    i_dof2 = dyn_info.joints.dof_start[I_joint2]
+                    diff = rigid_info.qpos[i_qpos2, i_b] - rigid_info.qpos0[i_qpos2, i_b]
 
                 a0 = dyn_info.equalities.eq_data[i_e, i_b][0]
                 a1 = dyn_info.equalities.eq_data[i_e, i_b][1]
@@ -897,7 +896,6 @@ def kernel_manual_add_equality_constraints_bw(
                 a3 = dyn_info.equalities.eq_data[i_e, i_b][3]
                 a4 = dyn_info.equalities.eq_data[i_e, i_b][4]
 
-                diff = pos2 - ref2
                 diff2 = diff * diff
                 diff3 = diff2 * diff
                 diff4 = diff3 * diff
@@ -906,11 +904,17 @@ def kernel_manual_add_equality_constraints_bw(
                 d_deriv_d_diff = 2.0 * a2 + 6.0 * a3 * diff + 12.0 * a4 * diff2
 
                 jac1 = gs.qd_float(1.0)
-                jac2 = -deriv
+                jac2 = gs.qd_float(0.0)
                 vel1 = dyn_state.dofs.vel[i_dof1, i_b]
-                vel2 = dyn_state.dofs.vel[i_dof2, i_b]
-                jac_qvel = jac1 * vel1 + jac2 * vel2
-                invweight = dyn_info.dofs.invweight[I_dof1] + dyn_info.dofs.invweight[I_dof2]
+                vel2 = gs.qd_float(0.0)
+                jac_qvel = jac1 * vel1
+                invweight = dyn_info.dofs.invweight[I_dof1]
+                if i_joint2 >= 0:
+                    I_dof2 = [i_dof2, i_b] if qd.static(rigid_config.batch_dofs_info) else i_dof2
+                    jac2 = -deriv
+                    vel2 = dyn_state.dofs.vel[i_dof2, i_b]
+                    jac_qvel = jac_qvel + jac2 * vel2
+                    invweight = invweight + dyn_info.dofs.invweight[I_dof2]
 
                 sol_params = dyn_info.equalities.sol_params[i_e, i_b]
                 imp, b_coef, k_coef, d_imp_d_imp_x = gu.imp_aref_grad(sol_params, pos)
@@ -923,7 +927,9 @@ def kernel_manual_add_equality_constraints_bw(
                 g_aref = constraint_state.dL_daref[n_con, i_b]
                 g_efc_D = constraint_state.dL_defc_D[n_con, i_b]
                 # jac[dof1] = 1.0 (constant) => no chain through dL_djac[n_con, i_dof1].
-                g_jac2 = constraint_state.dL_djac[n_con, i_dof2, i_b]
+                g_jac2 = gs.qd_float(0.0)
+                if i_joint2 >= 0:
+                    g_jac2 = constraint_state.dL_djac[n_con, i_dof2, i_b]
 
                 # ---- Partials ----
                 # aref = -b * jac_qvel - k * imp * pos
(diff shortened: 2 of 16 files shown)


genesis/utils/description.py
```diff
@@ -170,10 +170,10 @@ class LinkDescription:
 
 @dataclass
 class EqualityDescription:
-    """Describe one constraint tying two links or two joints of an entity together, named as the asset names them."""
+    """Describe one constraint tying two objects, or one joint to a constant, using their asset names."""
 
     name: str
     type: EQUALITY_TYPE
-    objs_name: tuple[str, str]
+    objs_name: tuple[str, str | None]
```
jeetrex17: Heads-up in case you have not rebased yet: this file was removed on main by #3294, which moved the dataclass to genesis/engine/entities/rigid_entity/description.py as RigidEqualityDescription. git merge-tree currently reports a modify/delete conflict, so I think this hunk will need to be re-applied on the new class rather than bringing this file back.

Small wording thing while you are there: the guidelines keep "link", "joint", etc. as the domain nouns, so maybe keep the original "two links or two joints" and add "or one joint and a constant".

genesis/engine/solvers/rigid/constraint/backward.py
```diff
@@ -874,43 +874,50 @@ def kernel_manual_add_equality_constraints_bw(
...
-                diff = pos2 - ref2
                 diff2 = diff * diff
                 diff3 = diff2 * diff
                 diff4 = diff3 * diff
                 pos = pos1 - ref1 - a0 - a1 * diff - a2 * diff2 - a3 * diff3 - a4 * diff4
                 deriv = a1 + 2.0 * a2 * diff + 3.0 * a3 * diff2 + 4.0 * a4 * diff3
                 d_deriv_d_diff = 2.0 * a2 + 6.0 * a3 * diff + 12.0 * a4 * diff2
+                if i_joint2 < 0:
+                    deriv = gs.qd_float(0.0)
+                    d_deriv_d_diff = gs.qd_float(0.0)
```
jeetrex17: If I read this right, these two assignments have no effect. deriv and d_deriv_d_diff are only read inside the later `if i_joint2 >= 0` branches (lines 917 and 969), and in the no-joint2 case vel2 and g_jac2 are already zero. The forward kernel also just leaves deriv = a1 there. Might be worth dropping the block so the two kernels read the same.

tests/rigid/test_constraints.py
```diff
@@ -90,6 +90,45 @@ def test_equality_joint_scaling(show_viewer, scaled_mjcf_joint_equalities, tol):
         assert_allclose(qpos[..., follower_idx] / follower_scale, expected_follower_position, tol=tol)
 
 
+@pytest.mark.required
+@pytest.mark.parametrize("n_envs, batch_info", [(0, False), (2, True)])
```
jeetrex17: Tiny naming one: the guidelines ask booleans to carry an is_/has_ prefix (https://github.com/Genesis-Embodied-AI/genesis-world/blob/main/CODING_GUIDELINES.md#21-general-rules), so batch_info would be is_batched.

tests/rigid/test_constraints.py
```diff
@@ -90,6 +90,45 @@ def test_equality_joint_scaling(show_viewer, scaled_mjcf_joint_equalities, tol):
...
+def test_equality_joint_without_joint2(show_viewer, single_joint_equality, n_envs, batch_info, tol):
```
jeetrex17: Would it make sense to add this case to an existing joint-equality test instead of a new function plus fixture? The guidelines lean toward extending the test that already covers a capability (https://github.com/Genesis-Embodied-AI/genesis-world/blob/main/CODING_GUIDELINES.md#122-test-structure), and I think two of them fit:

- test_equality_joint compares against MuJoCo step by step across CG/Newton and implicitfast/Euler. Its model_name chain (tests/rigid/conftest.py:11-21) takes an ElementTree element, and the parity helper only looks at CONNECT/WELD object ids (tests/utils/mujoco_parity.py:236-239), so a model with no joint2 should pass through as is. The equal-values assert at line 28 would need a per-model expected value.
- test_equality_joint_scaling already builds one scaled scene of driver/follower pairs. A fifth pair with no driver would keep one scene build and the same scale setup this test uses.

Either way I would keep both the batched and unbatched model-info cases, since the guards compile to different branches for the two settings.

tests/rigid/test_constraints.py
```diff
@@ -90,6 +90,45 @@ def test_equality_joint_scaling(show_viewer, scaled_mjcf_joint_equalities, tol):
...
+            gravity=(0.0, 0.0, 0.0),
+        ),
+        rigid_options=gs.options.RigidOptions(
+            enable_collision=False,
+            enable_joint_limit=False,
+            batch_joints_info=batch_info,
+            batch_dofs_info=batch_info,
+        ),
+        show_viewer=show_viewer,
+    )
```
jeetrex17: Since this test steps, TESTING.md asks for ViewerOptions with a camera framing the scene (https://github.com/Genesis-Embodied-AI/genesis-world/blob/main/.github/contributing/TESTING.md?plain=1#L82), so running it with --vis shows the two spheres.

Also, unless I am missing something, three of these options do not change anything here: gravity=(0,0,0) because the joints slide along x and z gravity cannot move them, enable_collision=False because the geoms already have contype=0 and conaffinity=0, and enable_joint_limit=False because neither joint has a range. The guidelines prefer only the options a test needs (https://github.com/Genesis-Embodied-AI/genesis-world/blob/main/CODING_GUIDELINES.md#124-scene-construction-in-tests-and-examples).

tests/rigid/test_constraints.py
```diff
@@ -90,6 +90,45 @@ def test_equality_joint_scaling(show_viewer, scaled_mjcf_joint_equalities, tol):
...
+    (target_idx,) = entity.get_joint("target").qs_idx_local
+    (unrelated_idx,) = entity.get_joint("unrelated").qs_idx_local
+    qpos = entity.get_qpos()
+    qpos[..., unrelated_idx] = UNRELATED_POSITION * SCALE
+    entity.set_qpos(qpos)
+    for _ in range(80):
+        scene.step()
+
+    qpos = entity.get_qpos() / SCALE
+    assert_allclose(qpos[..., target_idx], TARGET_POSITION, atol=2e-3)
```
jeetrex17: Could this use the `tol` fixture like the line below? I measured the target residual after 80 steps at about `3e-8` in fp32 and `5e-11` in fp64, so `2e-3` is much wider than the behavior needs and could let a real error through. The 80-step horizon also seems consistent with the repository's guidance to pad the measured minimum for floating-point-sensitive checks.

tests/coupling/test_hybrid.py
```diff
@@ -371,6 +371,39 @@ def test_fluid_emitter(n_envs, material_type, show_viewer):
...
+        <body>
+          <joint name="target" type="slide" axis="1 0 0"/>
+          <geom type="sphere" size="0.05" mass="1" contype="0" conaffinity="0"/>
+        </body>
+      </worldbody>
+      <equality>
+        <joint joint1="target" polycoef="0.25 1 0 0 0"/>
+      </equality>
+    </mujoco>
+    """
```
jeetrex17: The testing guidelines ask for custom MJCF to come from an xml.etree.ElementTree fixture rather than a string literal (https://github.com/Genesis-Embodied-AI/genesis-world/blob/main/CODING_GUIDELINES.md#124-scene-construction-in-tests-and-examples). Almost the same model already exists as the single_joint_equality fixture in tests/rigid/conftest.py. Would moving that fixture to tests/conftest.py work, so both tests share one model?

ktyang512:
Thanks for the detailed review. I moved the change to `RigidEqualityDescription` on current main, kept the old file deleted, and updated the terminology. I also removed the redundant backward assignments and folded the unary coverage into `test_equality_joint_scaling`, using `is_batched` and sharing the ElementTree fixture with SAP. The test retains the four polynomial cases, both batching configurations, and the checks that the target converges while the unrelated joint stays fixed. I removed the redundant scene options, checked the camera framing with an offscreen render, and switched the target assertion to `tol`. At 80 steps, the residuals were approximately 4.84e-11 on CPU fp64 and 2.98e-8 on RTX 5090 fp32. The targeted CPU/GPU tests, adjacent equality tests, and all five equality gradient cases passed. A temporary last-joint substitution also confirmed that both target assertions still catch the defect. The full suite and interactive `--vis` were not run.

tests/rigid/test_constraints.py
```diff
@@ -64,30 +67,43 @@ def test_equality_joint_scaling(show_viewer, scaled_mjcf_joint_equalities, tol):
...
             COEFFICIENTS[0]
             + COEFFICIENTS[1] * driver_position
             + COEFFICIENTS[2] * driver_position**2
             + COEFFICIENTS[3] * driver_position**3
             + COEFFICIENTS[4] * driver_position**4
         )
-        assert_allclose(qpos[..., follower_idx] / follower_scale, expected_follower_position, tol=tol)
+        assert_allclose(qpos[..., i_follower_q] / follower_scale, expected_follower_position, tol=tol)
+
+    for _ in range(N_STEPS - 1):
```
duburcqa: This test asserts one solve from equilibrium, as the four pairs do. Start the target at `TARGET_POSITION * SCALE` and keep the single step: a second joint resolved to any real joint moves the target within that step, since the unrelated joint sits at 1.0. `N_STEPS` goes away. The convergence from 0 to the constant is a parity case, see the comment on `tests/conftest.py`.

tests/conftest.py
```diff
@@ -72,6 +73,19 @@
 TOL_DOUBLE = 1e-9
 
 
+@pytest.fixture(scope="session")
+def single_joint_equality():
```
duburcqa: Return the `ET` element: the `model_name` chain in `tests/rigid/conftest.py` reads `mjcf.tag` and writes the tree itself. Then add the parity case to `tests/rigid/test_mujoco_parity.py`, three lines: `model_name` parametrized on this fixture, `gs_sim` / `mj_sim`, one `simulate_and_check_mujoco_consistency` call from `qpos = (0.0, 1.0)` over 50 steps, `backend` pinned to `gs.cpu` like its siblings. That run covers the transient to the constant against MuJoCo and is the convergence check. `scaled_mjcf_joint_equalities` extends from the element directly, without `ET.fromstring`, and the SAP scene passes `ET.tostring(single_joint_equality, encoding="unicode")`.

tests/coupling/test_hybrid.py
```diff
@@ -371,6 +371,25 @@ def test_fluid_emitter(n_envs, material_type, show_viewer):
     scene.step()
 
 
+@pytest.mark.required
+@pytest.mark.parametrize("precision", ["64"])
+def test_sap_rejects_joint_equality_without_joint2(show_viewer, single_joint_equality):
```
duburcqa: A scene rejected at build costs no build and is not a test of its own: fold it into `test_sap_rigid_rigid_hydroelastic_contact`, before its scene, as one-line statements: `gs.Scene(coupler_options=gs.options.SAPCouplerOptions())`, the `add_entity`, the `pytest.raises`. The five contact-type options play no part in the rejection: a bare-default SAP scene holding this entity alone raises it.

genesis/engine/couplers/sap_coupler.py
```diff
@@ -223,6 +223,10 @@ def build(self) -> None:
         self._enable_rigid_fem_contact &= self.rigid_solver.is_active and self.fem_solver.is_active
         self._enable_fem_self_tet_contact &= self.fem_solver.is_active
 
+        for equality in self.rigid_solver.equalities:
+            if equality.type == gs.EQUALITY_TYPE.JOINT and equality.eq_obj2id < 0:
+                gs.raise_exception("SAPCoupler does not support JOINT equality constraints without `joint2`.")
```
duburcqa, line 228: The adapter below hardcodes `q1 - q2 = 0` and ignores `polycoef` for two joints as well, so this rejection is the right scope here. The polynomial contract in the SAP adapter is tracked in #3324.

tests/coupling/test_hybrid.py
duburcqa: Now that I'm thinking, rejecting SAP is not a feature but a limitation. Let's not assert it at all. I don't when to have to update the unit test when adding support for it.

ktyang512:
I changed the scaling test to one step from equilibrium, made the shared fixture return an ElementTree element, and removed the SAP rejection test while retaining the build-time guard. The requested 50-step CPU parity test exposed a separate inverse-weight discrepancy: the paired fixtures force recomputation, which gave a unit-mass slider a link weight of 1/3 while MuJoCo uses 1. I reproduced this on main with no equality constraints. This revision therefore also updates inverse-weight recomputation to apply MuJoCo's 1/m approximation to eligible translation-only leaf bodies and their joints, including the armature convention, only when `enable_mujoco_compatibility` is enabled. This changes constraint regularization for qualifying sliders when weights are rebuilt, including after mass changes. Native-mode calculations remain unchanged. All 28 targeted cases passed, covering qualifying and excluded structures, batching, mass/COM refresh, state preservation, and adjacent constraints on CPU and GPU. As local-only diagnostics outside the PR, I also ran a 50-step contact comparison, which passed, and disabled the compatibility branch, which made both new compatibility cases fail. The full suite and gradients were not rerun, and broader SAP polynomial support remains with #3324.

genesis/engine/solvers/rigid/abd/forward_dynamics.py
duburcqa, line 1: Revert all of this, we are not going to support this. I think it is basically a bug in Mujoco.

Luckily for you, this is now properly supported in your mujoco parity unit tests my modifying how mujoco compute invweights directly!

tests/rigid/conftest.py
```diff
@@ -76,6 +76,63 @@ def _add_free_body(mjcf, name, geom_type, geom_size, pos, rgba=None):
     ET.SubElement(body, "joint", name=f"{name}_root", type="free")
 
 
+@pytest.fixture(scope="session")
+def slider_inverse_weights():
+    mjcf = ET.Element("mujoco", model="slider_inverse_weights")
```
duburcqa: Just make this a normal help, starting with '_build' prefix. See other examples;

tests/rigid/test_api.py
```diff
@@ -5,6 +5,8 @@
 import pytest
 import torch
 
+import mujoco
+
```
duburcqa: This is not allowed in unit test modules. Only conftest, instead mujoco parity framework.

ktyang512:
Thanks for clarifying the intended behavior and updating the parity framework. I should have confirmed the compatibility boundary before expanding this PR. I've removed the inverse-weight compatibility change together with its dedicated test and fixture, which also removes the direct MuJoCo usage from `test_api.py`. This revision is based on current main and uses the shared parity framework unchanged. The one-step scaling test and 50-step CPU parity case pass, along with the adjacent constraint checks and all five equality-gradient cases: 19 targeted cases passed across CPU and GPU. The full suite was not run.

tests/conftest.py
duburcqa: This would should go in 'tests/rigid/conftest.py'. There must be a misunderstanding.

Moveover, why not using "scaled_mjcf_joint_equalities" for the parity test? It would be more comprehensive no?

ktyang512:
Thanks for clarifying the fixture scope. I've moved `single_joint_equality` into `tests/rigid/conftest.py` and switched the 50-step CPU parity test to `scaled_mjcf_joint_equalities`, covering all four hinge/slide pairings together with the unary constraint and the unrelated joint. The one-step scaling assertions and shared parity framework are unchanged. The updated parity test, CPU/GPU scaling checks, and adjacent joint-equality checks pass (9 targeted cases). The full suite was not run.

duburcqa (dismissed):

duburcqa (approved):

PR #3294, opened 2026-09-03
[BREAKING][FEATURE] Share a scene as a file that opens without its assets (cont'd).

duburcqa:
## Description

Follow-up to #3288. The knowledge of how a scene is written and read back moves from `scene.py` and `serialization.py` to the classes that own it.

- The `Described` marker is gone. `Entity.desc` is the protocol: an entity hands back its description, or None for a kind no description carries, and `Scene.export` names what it refuses and what it leaves out from the scene's own collections instead of walking private attributes.
- A rigid entity is created from its description alone. Resolution, from morphs and assets to a description, lives in the description classes beside the entity (`genesis/engine/entities/rigid_entity/description.py`); every solver's `add_entity` takes `desc=None` and builds from a given description, which is the path `Scene.load` takes. Attachments are restored where the entity is created, since a description names only entities created before it.
- Each solver declares the material class it simulates (`Solver.material_cls`); the simulator dispatches from that instead of an `isinstance` chain. The scene no longer checks which material accepts which morph: an entity rejects a morph it cannot build from.
- Options hold only what they were created with: `USD` no longer carries a stage context (`parse_usd_stage` opens the stage once for all the entities of a stage), a XACRO is expanded by the parsers rather than at morph creation, and a terrain height field is a float array whatever it is given as.
- The serialization facility alone reduces absolute paths to asset names, inline documents included. Meshes, textures and morphs write their values as they are.
- Meshes parsed from USD are named by prim path, so the name holds nothing of the author's filesystem.

## Related Issue

Follows #3288.

## Motivation and Context

An external solver bringing its own entity, material and description should need no edit to the scene, the simulator or the codec to take part in a scene file. The marker mixin and the per-type rewriting in `Scene.export` stood in the way, and options carrying live handles or parse results made the file leak what the author's machine held.

`tests/rigid/test_serialization.py` exports a scene holding every entity kind that travels and loads it back, asserting the manifest holds no author path. `tests/parsers/test_usd.py` covers the stage split and export, `tests/rigid/test_asset_loading.py` the XACRO path, `tests/rigid/test_terrain.py` the height field, and `tests/rigid/test_heterogeneous.py` the rejection of a tuple of morphs by a soft-body entity.

## Checklist:


duburcqa:
@codex review

PR #3302, opened 2026-09-04
[BUG FIX] Support offscreen rendering with OSMesa.

### LINKED ISSUE #3292: [Bug]: OSMesa (CPU) offscreen rendering is broken
### Bug Description

OSMesa (CPU) offscreen rendering is broken 

### Steps to Reproduce

```python
import os; os.environ["PYOPENGL_PLATFORM"] = "osmesa"  
import genesis as gs
gs.init(backend=gs.cpu)
cam = scene.add_camera(res=(1280, 960), pos=(3, -2, 1.8), lookat=(0, 0, 0.9))
scene.build(n_envs=1); scene.step()
cam.render(rgb=True)    
```

### Expected Behavior

It should return a rendered camera image

### Environment

genesis-world 1.3.3, Python 3.12, Linux x86-64, PYOPENGL_PLATFORM=osmesa, gs.init(backend=gs.cpu). Identical scene on EGL/GPU renders fine.

1.3.3

duburcqa:
## Description

OSMesa contexts render into framebuffer objects like every other offscreen platform. The separate default-framebuffer path, the offscreen viewport hardcoded to 32x32 pixels and the `supports_framebuffers` hook are removed.

PyOpenGL commits to the OSMesa library when it is first imported, for the whole process. The OSMesa platform now checks that it was selected upfront, releases and restores its current context like the EGL platform does, and the interactive viewer rejects `PYOPENGL_PLATFORM='osmesa'` with a clear message, since a window-system context cannot be driven through OSMesa.

OSMesa is selected automatically on a headless Linux machine lacking EGL, where offscreen rendering would otherwise fail, and it is accepted on every operating system: Mesa builds for Windows ship it up to version 25.0.

## Related Issue

Resolves Genesis-Embodied-AI/genesis-world#3292

## Motivation and Context

Every camera render with `PYOPENGL_PLATFORM=osmesa` failed on an assertion: the OSMesa branch of the offscreen renderer never requested an offscreen pass, so the renderer returned nothing. Removing the assertion would have produced 32x32 images regardless of the camera resolution.

Run the script of the issue with `PYOPENGL_PLATFORM=osmesa` set before importing genesis: it renders at the camera resolution, and its RGB, depth and segmentation match the EGL platform pixel for pixel, both running the same software rasterizer. Enabling the viewer with that variable set raises the new exception.

## Checklist:

### DIFF
--- a/genesis/ext/pyrender/offscreen.py
+++ b/genesis/ext/pyrender/offscreen.py
@@ -4,6 +4,7 @@ Author: Matthew Matl
 """
 
 import os
+import sys
 
 from OpenGL.GL import *
 
@@ -17,41 +18,17 @@ class OffscreenRenderer(object):
 
     Parameters
     ----------
-    viewport_width : int
-        The width of the main viewport, in pixels.
-    viewport_height : int
-        The height of the main viewport, in pixels.
     point_size : float
         The size of screen-space points in pixels.
     """
 
-    def __init__(self, point_size=1.0, pyopengl_platform="pyglet", seg_node_map=None):
+    def __init__(self, point_size=1.0, seg_node_map=None):
         self.point_size = point_size
         self._platform = None
         self._is_software = False
         self._has_valid_context = False
-        self._create(pyopengl_platform)
+        self._create()
         self._seg_node_map = seg_node_map
-
-    @property
-    def viewport_width(self):
-        """int : The width of the main viewport, in pixels."""
-        return 32
-        # return self._viewport_width
-
-    @viewport_width.setter
-    def viewport_width(self, value):
-        self._viewport_width = int(value)
-
-    @property
-    def viewport_height(self):
-        """int : The height of the main viewport, in pixels."""
-        return 32
-        # return self._viewport_height
-
-    @viewport_height.setter
-    def viewport_height(self, value):
-        self._viewport_height = int(value)
 
     @property
     def point_size(self):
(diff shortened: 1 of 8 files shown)


genesis/ext/pyrender/platforms/osmesa.py
```diff
@@ -43,26 +66,33 @@ def init_context(self):
...
-        self._context = None
-        self._buffer = None
+    def save_current_context(self):
+        # 'make_uncurrent' releases the context and every render path ends there, so this platform's own context is
+        # never the one captured here, only an external one (e.g. another renderer mid-render) or none. A context is
+        # made current together with its color buffer, which is queried back from the context to restore both.
+        context = OSMesaGetCurrentContext()
+        if not context:
+            return None
+        width, height, _, buffer = OSMesaGetColorBuffer(context)
```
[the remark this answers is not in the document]
duburcqa: PyOpenGL wraps `OSMesaGetColorBuffer` as a one-argument call returning `(width, height, format, buffer)` (see `OpenGL/raw/osmesa/mesa.py`), so passing the five native arguments is what raises `TypeError`.

genesis/ext/pyrender/offscreen.py
```diff
@@ -172,30 +135,9 @@ def render(
...
-                    if z_far is None:
-                        z_far = -1.0
-                    depth_arr = renderer.jit.read_depth_buf(self.viewport_height, self.viewport_width, z_near, z_far)
-                    depth_arr = renderer._resize_image(depth_arr, antialias=not seg)
-                if flags & RenderFlags.DEPTH_ONLY:
-                    retval = (depth_arr,)
-                else:
-                    color_arr = renderer.jit.read_color_buf(self.viewport_height, self.viewport_width, rgba=False)
-                    color_arr = renderer._resize_image(color_arr, antialias=not seg)
-                    retval = (color_arr, depth_arr) if depth else (color_arr,)
```
duburcqa, line 198: I'm quite sure I already hit this branch at some point. Maybe when using a Raspberry PI board, which is supported by genesis.

duburcqa: Only the OSMesa platform enters this branch (`supports_framebuffers` is False there alone), and it reads its buffers at the 32x32 viewport hardcoded since the initial release, so it never produced an image at the camera resolution; since #2169 the assert above fails before reading. The OSMesa context is requested as OpenGL 3.3 core, where framebuffer objects are core, so it now takes the same path as EGL.

PR #3306, opened 2026-09-04
[FEATURE] Allow skipping the warmup step when building a scene.

Milotrince:
## Description

Building a scene runs one full simulation step so that the kernels it goes through are compiled, or loaded from cache, before the user's first `step()`. A scene that is built for visualization and never stepped pays that step for nothing. `GS_SKIP_BUILD_WARMUP=1` skips it.

Two things the warmup step was silently responsible for now happen on both paths:

- **The reset that follows it is the only simulator reset during build**, and it is what seeds the sensors and the coupler. It now runs whether or not the warmup step did. Skipping it left temperature sensors reading zero instead of their base temperature, and sensor history rings holding uninitialized memory.
- **The geom axis-aligned bounding boxes (AABB) are seeded at build.** Collision detection refreshes them every step, so without the warmup step `get_AABB()` reported an empty box at the origin until the first one.

## Related Issue

Context: [SIM-372](https://linear.app/genesis-ai-company/issue/SIM-372/enable-scene-viewer-without-physics-build) (scene viewer without a physics build), under [SIM-330](https://linear.app/genesis-ai-company/issue/SIM-330/improve-gui-experience). The warmup step is a general build cost rather than a preview-only one, which is why it is split out here.

## Motivation and Context

A scene editor rebuilds its scene for every structural edit (add, remove, duplicate, swap asset, scale, layout switch) and holds the GIL for the whole build. The warmup step is pure overhead in that loop, and it is unconditional today with no way to opt out.

`tests/core/test_misc.py::test_build_state_independent_of_warmup_step`, parametrized over `is_warmup_skipped=[False, True]` and `n_envs=[0, 2]`. It asserts the build leaves the same observable state either way: sensors read their base temperature rather than zero, geom AABBs match the initial poses, and a ball in free fall follows semi-implicit Euler.

```
pytest tests/core/test_misc.py -k warmup
```

## Draft: what is left

- **Needs a rebase.** The branch is 48 commits behind `main` and conflicts in `genesis/engine/solvers/rigid/rigid_solver.py`: the build order changed there (`_init_collider` / `_init_constraint_solver` now precede `_refresh_invweight_and_meaninertia`, and the `_func_update_geoms` call the AABB seeding was anchored to has moved), so the seeding hunk needs re-placing rather than a line-offset fix.
- **The escape hatch is an environment variable**, which is worth a second opinion against a typed option. `GS_PARA_LEVEL` and `GS_ENABLE_NDARRAY` are precedent for a perf knob living in the environment, but a build-time behavior change may belong on `SimOptions` or on `Scene.build()` instead.
- Warmup-skipped build time is not measured here yet.

## Checklist:

### DIFF
--- a/genesis/engine/scene.py
+++ b/genesis/engine/scene.py
@@ -908,9 +908,16 @@ class Scene(RBC):
 
             self._is_built = True
 
-        with gs.logger.timer("Compiling simulation kernels..."):
-            self._sim.step()
-            self._reset()
+        # Running one step ahead of time compiles (or loads from cache) the kernels it goes through, so the first user
+        # step is not slowed down by compilation. Set 'GS_SKIP_BUILD_WARMUP=1' to skip it when the scene is built for
+        # visualization only and never stepped.
+        if os.environ.get("GS_SKIP_BUILD_WARMUP") != "1":
+            with gs.logger.timer("Warming up the simulation..."):
+                self._sim.step()
+
+        # Restore the initial state, which also seeds the sensors and the coupler. Must run whether or not the warmup
+        # step did, since it is the only simulator reset during build.
+        self._reset()
 
         # visualizer
         with gs.logger.timer("Building visualizer..."):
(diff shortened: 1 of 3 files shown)


duburcqa:
I'm not ok with this. The interactive scene should bypass 'Scene.build' instead.

PR #3313, opened 2026-09-05
[BUG FIX] Fix scaling wrongly applied for in-memory URDF file morph.

### LINKED ISSUE #3312: [Bug]: URDF scale is applied twice to a parsed URDF object and to mesh-based hybrid entities
### Bug Description

Loading a URDF with `scale` changes the model that was passed in, instead of only computing scaled values from it.

- `parse_urdf` takes views of the caller's `urdfpy` object (inertial origin, inertia matrix, joint origins) and scales them in place with `*=`. The geom origins already take a copy, these three do not.
- `merge_fixed_links` copies the link and joint lists but keeps the objects inside them, so the merge writes into the caller's parent link.

This becomes a wrong simulation because the rigid loader parses every URDF morph twice on the same object: first `parse_urdf`, then the MuJoCo parser on a shallow `model_copy` of the morph (`build_model` accepts a `urdfpy.URDF` object and serializes it). The second pass reads the already modified model and applies everything again:

- joint origins and inertial origins: `s^2` instead of `s`
- inertia tensors (each pass applies `s^5`): `s^10` instead of `s^5`
- fixed children are merged into the parent twice, so masses are too high

Each further `add_entity` on the same object multiplies by one more factor.

Public route: a `HybridEntity` built from a `Mesh` morph. The skeletonization produces a URDF object and adds it with `scale=morph.scale`, so the rigid skeleton comes out at `scale^2`. `tests/coupling/test_hybrid.py::test_mesh_mpm_build` builds exactly this at scale 0.15 and cannot see it, since it only asserts that `build()` returns. A hybrid built from a URDF file is unaffected, because each parse reloads a fresh object.

### Steps to Reproduce

Direct reproduction through a parsed model (seconds):

```python
import os
import tempfile

import genesis as gs
from genesis.ext.urdfpy.urdf import URDF

URDF_XML = """<robot name="double_scale_probe">
  <link name="base">
    <inertial>
      <mass value="1.0"/>
      <origin xyz="0 0 0"/>
      <inertia ixx="0.01" iyy="0.01" izz="0.01" ixy="0" ixz="0" iyz="0"/>
    </inertial>
  </link>
  <link name="child">
    <inertial>                                                                                                                         <mass value="0.5"/>
      <origin xyz="0.1 0 0"/>
      <inertia ixx="0.001" iyy="0.001" izz="0.001" ixy="0" ixz="0" iyz="0"/>
    </inertial>
  </link>
  <joint name="j1" type="revolute">
    <parent link="base"/>
    <child link="child"/>
    <origin xyz="0.31 0 0"/>                                                                                                         <axis xyz="0 0 1"/>    <limit lower="-1" upper="1" effort="1" velocity="1"/>
  </joint>
</robot>
"""                                                                                                                              SCALE = 0.7
AUTHORED_JOINT_X = 0.31

with tempfile.TemporaryDirectory() as tmp:
    path = os.path.join(tmp, "double_scale_probe.urdf")
    with open(path, "w") as f:
        f.write(URDF_XML)
    robot = URDF.load(path)

print("authored joint origin x           :", robot.joints[0].origin[0, 3])

entity_1 = scene.add_entity(gs.morphs.URDF(file=robot, scale=SCALE, fixed=True))                                                 print("joint origin x in caller's object :", robot.joints[0]ed by add_entity")
entity_2 = scene.add_entity(gs.morphs.URDF(file=robot, scale=SCALE, fixed=True, pos=(0.0, 1.0, 0.0)))
scene.build()                                                                                                                    
print("expected child offset x           :", SCALE * AUTHORED_JOINT_X)                                                           for name, entity in (("entity 1", entity_1), ("entity 2", en
    offset = entity.get_link("child").get_pos() - entity.get_link("base").get_pos()                                                  print(f"{name} child offset x            :", float(offse
```
Two variants of the same script isolate the other quantities- Joint `type="fixed"`, so the default `merge_fixed_links` rt.- Child inertial origin `0 0 0` with `izz="0.003"`. The chils on the joint axis, so the single entry of`entity.get_mass_mat()` is `I_zz` plus armature. Shows the i

Public route (mesh-based hybrid, the configuration of `test_mesh_mpm_build`), run once per scale:

```python
import sys

import numpy as np

import genesis as gs

scale = float(sys.argv[1])
scene = gs.Scene(
        lower_bound=(-0.7, -0.7, -0.7),
        upper_bound=(0.7, 0.7, 0.7),
    ),
    show_viewer=False,
)
entity = scene.add_entity(
    morph=gs.morphs.Mesh(
        file="meshes/duck.obj",
        scale=scale,
    ),
    material=gs.materials.Hybrid(
        material_rigid=gs.materials.Rigid(),
        material_soft=gs.materials.MPM.Muscle(),
    ),
)
scene.build()
links_pos = np.stack([tensor_to_array(link.get_pos()) for link in entity._part_rigid.links])
print(scale, len(entity._part_rigid.links), (links_pos.max(0) - links_pos.min(0)).tolist())
```

### Expected Behavior

The caller's object is left untouched.

- Both entities place `child` at `0.7 * 0.31 = 0.217` from `base`.
- Fixed-joint variant: base mass `(1.0 + 0.5) * 0.7**3 = 0.5145`.
- Inertia variant: `I_zz = 0.003 * 0.7**5 = 5.042e-4`.
- Hybrid skeleton at scale 0.15 has `0.15 / 0.2 = 0.75` times the extent of the one at scale 0.2.

### Screenshots/Videos

_No response_

### Relevant log output

```shell
=== joint origin (revolute) ===
authored joint origin x           : 0.31
joint origin x in caller's object : 0.217  <- mutated by add_entity
expected child offset x           : 0.217
entity 1 child offset x            : 0.15189999341964722   (= 0.7^2 * 0.31)
entity 2 child offset x            : 0.10632999986410141   (= 0.7^3 * 0.31)

=== mass (fixed joint, default merge_fixed_links) ===
caller's base mass before          : 1.0
caller's base mass after add_entity: 1.5
expected merged base mass          : 0.5145
entity 1 base mass                 : 0.6859999895095825    (= 2.0 * 0.7^3)
entity 2 base mass                 : 0.8575000166893005    (= 2.5 * 0.7^3)

=== inertia (child COM on the joint axis, izz = 0.003) ===
caller child inertia before       : [0.001 0.002 0.003]
caller child inertia after        : [0.00016807 0.00033614 0.00050421]   (x 0.7^5, mutated by add_entity)
built I_zz from mass matrix       : 8.474290370941162e-05
correct I_zz (s^5)                : 0.00050421
double  I_zz (s^10)               : 8.47426e-05

=== hybrid skeleton (mesh-based HybridEntity, duck.obj, same 10-link skeleton) ===
scale 0.15: extent [0.016378, 0.038168, 0.068686]
scale 0.20: extent [0.029116, 0.067854, 0.122109]
ratio 0.15/0.20 : [0.5625, 0.5625, 0.5625]   expected 0.75, got 0.75^2
```

### Environment

- OS: macOS 15.7.3
- GPU/CPU: Apple M4 (CPU backend used)
- GPU-driver version: N/A
- Python 3.13.11, torch 2.13.0

Genesis 1.3.3, main at 48a690a3

### Additional Context

Backend independent: the corruption happens at parse time. Everything is in `genesis/utils/urdf.py`. I have a fix with an analytical regression test and will open a PR referencing this issue.

jeetrex17:
## Description

The URDF parser copies a caller-provided `urdfpy.URDF` model before scaling it and merging its fixed links, so the model stays usable for further entities and the second parsing pass reads the authored values.

Both parsing passes now resolve the relative mesh paths of an in-memory model against the working directory. The vendored `urdfpy` copy keeps an omitted inertial origin omitted and keeps a mesh path with a directory component intact.

## Related Issue

Resolves #3312

## Motivation and Context

Genesis parses every URDF morph twice. The first pass scaled the caller's object in place and merged its fixed links into it, so the second pass and every later entity built from the same object applied the scale and the merge again. The rigid skeleton of a hybrid entity built from a mesh takes this route.

An in-memory model with meshes could not be loaded at all: the URDF pass resolved its mesh paths against the working directory and the MuJoCo pass against the assets directory.

`test_robot_scale_and_dofs_armature` loads every URDF asset once more as one in-memory model shared by the entities at all scales, with `panda_bullet/panda.urdf` added as an asset whose mesh paths carry a directory component. `test_parsing_inertia_defaults` builds its two merged chains, whose inertial origins are omitted, from one shared in-memory model. The existing physical assertions catch a model scaled or merged more than once.

## Checklist:

### DIFF
--- a/tests/rigid/conftest.py
+++ b/tests/rigid/conftest.py
@@ -229,6 +229,42 @@ def scaled_mjcf_joint_equalities():
 
 
 @pytest.fixture(scope="session")
+def urdf_arm_with_tool():
+    """A free base carrying an arm on a revolute joint and a tool on a fixed joint, every link with an inertial."""
+    robot = ET.Element("robot", name="arm_with_tool")
+    links = (
+        ("base", "1.0", "0 0 0"),
+        ("arm", "0.5", "0.1 0 0"),
+        ("tool", "0.2", None),
+    )
+    for link_name, mass, com_xyz in links:
+        link = ET.SubElement(robot, "link", name=link_name)
+        inertial = ET.SubElement(link, "inertial")
+        ET.SubElement(inertial, "mass", value=mass)
+        if com_xyz is not None:
+            ET.SubElement(inertial, "origin", xyz=com_xyz)
+        ET.SubElement(inertial, "inertia", ixx="1e-3", iyy="2e-3", izz="3e-3", ixy="0", ixz="0", iyz="0")
+    # The tool omits its inertial origin and references its mesh by a relative path with a directory component, the
+    # two forms of a model that a copy of it must carry through.
+    visual = ET.SubElement(robot.find("link[@name='tool']"), "visual")
+    geometry = ET.SubElement(visual, "geometry")
+    ET.SubElement(geometry, "mesh", filename="meshes/sphere.obj", scale="0.02 0.02 0.02")
+    joints = (
+        ("shoulder", "revolute", "base", "arm", "0.3 0 0"),
+        ("wrist", "fixed", "arm", "tool", "0.2 0 0"),
+    )
+    for joint_name, joint_type, parent_name, child_name, origin_xyz in joints:
+        joint = ET.SubElement(robot, "joint", name=joint_name, type=joint_type)
+        ET.SubElement(joint, "parent", link=parent_name)
+        ET.SubElement(joint, "child", link=child_name)
+        ET.SubElement(joint, "origin", xyz=origin_xyz)
+        if joint_type == "revolute":
+            ET.SubElement(joint, "axis", xyz="0 0 1")
+            ET.SubElement(joint, "limit", lower="-1.0", upper="1.0", effort="100", velocity="1.0")
+    return ET.tostring(robot, encoding="unicode")
+
+
+@pytest.fixture(scope="session")
 def scaled_urdf_mimic():
     robot = ET.Element("robot", name="scaled_urdf_mimic")
     ET.SubElement(robot, "link", name="base")
--- a/tests/rigid/conftest.py
+++ b/tests/rigid/conftest.py
@@ -229,42 +229,6 @@ def scaled_mjcf_joint_equalities():
 
 
 @pytest.fixture(scope="session")
-def urdf_arm_with_tool():
-    """A free base carrying an arm on a revolute joint and a tool on a fixed joint, every link with an inertial."""
-    robot = ET.Element("robot", name="arm_with_tool")
-    links = (
-        ("base", "1.0", "0 0 0"),
-        ("arm", "0.5", "0.1 0 0"),
-        ("tool", "0.2", None),
-    )
-    for link_name, mass, com_xyz in links:
-        link = ET.SubElement(robot, "link", name=link_name)
-        inertial = ET.SubElement(link, "inertial")
-        ET.SubElement(inertial, "mass", value=mass)
-        if com_xyz is not None:
-            ET.SubElement(inertial, "origin", xyz=com_xyz)
-        ET.SubElement(inertial, "inertia", ixx="1e-3", iyy="2e-3", izz="3e-3", ixy="0", ixz="0", iyz="0")
-    # The tool omits its inertial origin and references its mesh by a relative path with a directory component, the
-    # two forms of a model that a copy of it must carry through.
-    visual = ET.SubElement(robot.find("link[@name='tool']"), "visual")
-    geometry = ET.SubElement(visual, "geometry")
-    ET.SubElement(geometry, "mesh", filename="meshes/sphere.obj", scale="0.02 0.02 0.02")
-    joints = (
-        ("shoulder", "revolute", "base", "arm", "0.3 0 0"),
-        ("wrist", "fixed", "arm", "tool", "0.2 0 0"),
-    )
-    for joint_name, joint_type, parent_name, child_name, origin_xyz in joints:
-        joint = ET.SubElement(robot, "joint", name=joint_name, type=joint_type)
-        ET.SubElement(joint, "parent", link=parent_name)
-        ET.SubElement(joint, "child", link=child_name)
-        ET.SubElement(joint, "origin", xyz=origin_xyz)
-        if joint_type == "revolute":
-            ET.SubElement(joint, "axis", xyz="0 0 1")
-            ET.SubElement(joint, "limit", lower="-1.0", upper="1.0", effort="100", velocity="1.0")
-    return ET.tostring(robot, encoding="unicode")
-
-
-@pytest.fixture(scope="session")
 def scaled_urdf_mimic():
     robot = ET.Element("robot", name="scaled_urdf_mimic")
     ET.SubElement(robot, "link", name="base")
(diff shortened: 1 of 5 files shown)


tests/rigid/test_asset_loading.py
```diff
@@ -914,19 +915,54 @@ def test_robot_scale_and_dofs_armature(xml_path, tol):
...
+            for joint, joint_ref in zip(urdf_model.joints, urdf_model_ref.joints, strict=True):
+                assert_equal(joint.origin, joint_ref.origin)
+            for link, link_ref in zip(urdf_model.links, urdf_model_ref.links, strict=True):
+                assert_equal(link.inertial.mass, link_ref.inertial.mass)
+                assert_equal(link.inertial.inertia, link_ref.inertial.inertia)
+                if link_ref.inertial.origin is None:
+                    assert link.inertial.origin is None
+                else:
+                    assert_equal(link.inertial.origin, link_ref.inertial.origin)
+                assert len(link.visuals) == len(link_ref.visuals)
```
duburcqa: You should assert the high-level symptom(s) from user-perspective, not the root cause. Our unit tests are validating specs and requirements, not internals.

tests/rigid/test_asset_loading.py
```diff
@@ -895,13 +895,14 @@ def test_color_overwrite(overwrite, show_viewer):
...
     [
-        pytest.param("xml/franka_emika_panda/panda.xml", marks=pytest.mark.slow),
-        "urdf/go2/urdf/go2.urdf",
+        pytest.param("xml/franka_emika_panda/panda.xml", False, marks=pytest.mark.slow),
+        ("urdf/go2/urdf/go2.urdf", False),
+        ("urdf_arm_with_tool", True),
     ],
 )
-def test_robot_scale_and_dofs_armature(xml_path, tol):
+def test_robot_scale_and_dofs_armature(file, is_model_shared, request, monkeypatch, tol):
```
duburcqa: Do not rename 'xml_path' in 'file' needlessly.

tests/rigid/conftest.py
```diff
@@ -228,6 +228,42 @@ def scaled_mjcf_joint_equalities():
     return ET.tostring(mjcf, encoding="unicode")
 
 
+@pytest.fixture(scope="session")
+def urdf_arm_with_tool():
```
duburcqa: I'm skeptical than a dedicated model is needed to trigger this defect.

duburcqa:
A fixup commit reworks the coverage and completes the fix.

The dedicated model is gone, since the defect is reachable with the bundled assets. `test_robot_scale_and_dofs_armature` loads every URDF asset once more as a single `urdfpy.URDF` object and builds an entity from it at every scale, so the existing assertions on mass, inertia, joint and link positions catch a model scaled or merged more than once. `panda_bullet/panda.urdf` joins the parametrization because its mesh paths carry a directory component, which the mesh-path correction in the vendored `urdfpy` copy needs. The omitted inertial origin is covered in `test_parsing_inertia_defaults`, whose two merged chains are now built from one shared in-memory model of the existing `implicit_inertial_origin_chain` fixture. The assertions on the caller's model are removed: tests validate observable physics, not the mechanism.

The MuJoCo pass now resolves the relative mesh paths of an in-memory model against the working directory, as the URDF pass does. With two different roots, no bundled asset with meshes could be passed as an object at all, which is why the original fixture had to place its mesh relative to the assets directory.

Reverting each of the four corrections in turn fails the intended test on its own symptom: a scale mismatch, a mesh path doubled to `meshes/collision/meshes/collision/link0.obj`, an `AttributeError` on the missing origin, and a mesh resolved under the assets directory.

duburcqa (dismissed):

duburcqa (approved):

Issue #3310, opened 2026-09-05
[Bug]: func_solve_body_monolith constraint-solver arm hangs the GPU on Ada and Blackwell (L40S, RTX 5070), not on A100


chkxw:
## Bug Description

The `func_solve_body_monolith` implementation of the `func_solve_body`
perf-dispatch group hangs on some GPU architectures: a kernel never returns, the
device stays at 100% utilisation, and the host blocks at its next
host-device interaction. The `func_solve_decomposed` implementation of the same
group runs the identical workload.

It is architecture-dependent. The same code, same driver, same scene:

| GPU | compute capability | driver | outcome |
|---|---|---|---|
| NVIDIA L40S | 8.9 (Ada) | 580.173.02 | hangs at per-env step 185 |
| NVIDIA RTX 5070 | 12.0 (Blackwell) | 580.173.02 | hangs at per-env step 216 |
| NVIDIA A100-PCIE-40GB | 8.0 (Ampere) | 580.173.02 | completes |
| NVIDIA A100-SXM4-80GB | 8.0 (Ampere) | 595.71.05 | completes |

Two A100s on different drivers both complete, so this is not a driver issue.

Because `qd.perf_dispatch` benchmarks every *compatible* implementation before
selecting one, a run that ends up choosing the decomposed arm still hangs — the
monolith arm gets invoked in an unlogged trial. The hang therefore appears in
normal use on affected hardware, not only when the arm is pinned.

genesis-world 0.4.2 has no `solver_breakdown` module and no decomposed arm. Our
0.4.2 reference training run completed 700 updates on an RTX 5090 — also Blackwell,
also sm_120 — without this hang. On the RTX 5070 we ran 0.4.2 only in short
workloads (scripted stepping, up to 2000 steps), which completed; we have not run
the full training workload under 0.4.2 on that card, so this is evidence that the
same workload was fine on Blackwell before 1.3.3, not a same-card A/B.

## Steps to Reproduce

We have not yet reduced this to a short self-contained script — a plain stepping
loop over the same scene does **not** reproduce it (details under Additional
context), so something about the surrounding workload participates. Our own
harness is in a private repository, so the scene and the per-step pattern are
described exactly instead; we are happy to share a minimal harness on request.

Scene:
- ground plane, friction 1.0
- one box, 0.4674 x 0.7620 x 0.4674 m, 5 kg (set via `rho`), friction 0.05
[...]

PR #3315, opened 2026-09-05
[BUG FIX] Bound the MPR portal-discovery loop and report a dropped contact per env.

### LINKED ISSUE #3314: [Bug]: mpr_refine_portal has no iteration cap; a non-converging portal spins a GPU lane forever (standalone reproduction attached)
## Bug Description

`mpr_refine_portal` (`genesis/engine/solvers/rigid/collider/mpr.py:263-310`, the `while True` at line 280) has no iteration cap. Its sibling `mpr_find_penetration` is capped by `CCD_ITERATIONS`, and libccd caps both loops. On near-degenerate geometry the portal keeps expanding by rounding-sized steps without encapsulating the origin or reaching the tolerance: one lane spins forever, the kernel never returns, and the host blocks at its next device synchronisation. No NaN or blown-up state is involved, and the point at which it happens is deterministic given the trajectory.

Observed on a dual-quadruped push (two Go1 URDFs, a box, a plane), 256–512 envs, both constraint-solver arms, on RTX 5090, RTX 5070, L40S, A100-PCIE-40GB and A100-SXM4-80GB, with genesis-world 1.3.3 and `main` (01387ec8). `cuda-gdb` on the frozen process shows one resident kernel (`_func_narrowphase_multicontact…` or `_func_narrowphase_contact0`), one active lane inside `mpr_refine_portal` reached through `func_mpr_contact`, its 31 warp-mates parked. With the loop bounded, the same trajectory records a cap exactly in the window where the unbounded build stops: A100, seed 1 — stock stops at PPO update 49, the cap fires at per-env step 6384 inside the (6272, 6400] window, the run completes 190 updates with 7 caps; seed 2 — stock stops at update 5, first cap at step 656, 20 caps in 190 updates; RTX 5090 — same picture on both arms.

A second uncapped loop of the same class: the three `while is_improving*` chases in `multi_contact.py::func_approximate_polygon_with_quad` (contact-patch path only). One NaN vertex makes every test false and the lane cycles forever.

## Steps to Reproduce

Standalone, no trainer, built on #3309's `save_checkpoint` / `load_checkpoint`. Files: https://github.com/chkxw/genesis-world/releases/tag/repro-3314 (~177 MB: `ckpt_512.gsckpt`, `ops.pkl.gz`, `meta.json`, `state_before_step512.npz`, `repro_3314_ckpt_replay.py`, `0007-pr3309-read-frame-zerocopy-fallback.patch`, `REPRO_3314.md`). A 256-env scene checkpointed right before physics step 512, plus the recorded entity calls for the following steps.

Branch as is:

```
git clone https://github.com/Genesis-Embodied-AI/genesis-world && cd genesis-world && git checkout 01387ec8
git apply 0007-pr3309-read-frame-zerocopy-fallback.patch   # #3309 checkpoint I/O with zero-copy on CUDA; reported on #3309
pip install -e .
python repro_3314_ckpt_replay.py <files dir> --ckpt-step 512 --steps 256 --timeout 120
```

Output: the restored state matches the snapshot (`max|diff|=0` for all six entities), state lines for steps 512, 528 and 544, then `HANG` — the GPU at 100 % with the lane in `mpr_refine_portal`.

With the bound:

```
git fetch https://github.com/chkxw/genesis-world fix/mpr-portal-discovery-cap && git cherry-pick a0f78f7e f018e622
python repro_3314_ckpt_replay.py <files dir> --ckpt-step 512 --steps 256 --timeout 120
```

Output: all 256 steps in about 3 s, `capped_envs=[179]` from step 560, `COMPLETED`; the printed state equals the recording's up to that portal.

Recorded on a single RTX 5090 (driver 595.84). Whether another architecture walks into the same portal from this checkpoint is untested: the collider is bit-deterministic on one card, not across cards. Resuming an exact `TrajectoryFile` (#3309) from frame 512 gives the same two outcomes.

## Expected Behavior

Both loops terminate for every input. A non-converging portal, or a NaN polygon, is dropped and reported per env, never spun on.

## Suggested fix

Fork branch `fix/mpr-portal-discovery-cap` (two commits; was #3315): cap portal discovery with `CCD_ITERATIONS` like refinement and return "no portal"; set a non-raising per-env `ErrorCode.UNCONVERGED_CONTACT_PORTAL`, exposed through `get_error_envs_mask()`, right after the attempt that hit the cap (a warm-start retry otherwise erases it); cap each quad chase at `nverts` and drop that pair's manifold. A NaN-aware comparison is not an option for the quad chase because the kernels are compiled with `NoNaNsFPMath`.

## Environment

genesis-world 1.3.3 (PyPI) and `main` 01387ec8 (+ #3309); quadrants 1.3.0; torch 2.8.0+cu128; Python 3.12; CUDA 12.8; Ubuntu 22.04 / 24.04. Drivers: 595.84 (RTX 5090), 580.173.02 (RTX 5070, L40S, A100-PCIE), 595.58.03 / 595.71.05 (A100-SXM4).

chkxw:
Fixes #3314. Related: #3310.

## Problem

`mpr_refine_portal` (`genesis/engine/solvers/rigid/collider/mpr.py`) walks the MPR portal with a
`while True` whose only exits are the encapsulation test and the tolerance test. Its sibling,
`mpr_find_penetration`, is bounded by `CCD_ITERATIONS` ("Bounds that refinement, which is not otherwise
guaranteed to terminate"), and libccd bounds both loops. A portal that keeps expanding by rounding-sized
steps without ever encapsulating the origin or reaching the tolerance therefore spins forever. On the GPU
that is one lane that never leaves the narrowphase kernel; the host then blocks at its next device sync,
which is why the stacks in #3310 pointed at unrelated Python lines.

Evidence (details and logs in #3314):
- cuda-gdb attached to the frozen process (RTX 5090, genesis-world 1.3.3, quadrants 1.3.0, 512 envs): one
  block of `_func_narrowphase_multicontact` still running, one warp, one active lane, 31 parked, PC inside
  the kernel body (constraint-solver arm `func_solve_body_monolith` pinned). With the decomposed arm the
  same picture in `_func_narrowphase_contact0`. Both kernels reach `mpr_refine_portal` through
  `func_mpr_contact` (`narrowphase.py`: the contact0 kernel, the perturbed re-detections of the
  multicontact kernel, and the non-split convex path).
- No NaN and no blown-up state are needed: on an A100-PCIE the same seed stopped at the same PPO update in
  three allocations with finite state and no error flag one update earlier; on an RTX 5090 with the monolith
  arm pinned, 3 of 3 runs stopped after the first update on a bit-identical trajectory.
- Host-side fences and `QD_GRAPH=0` do not prevent it.

## Fix

- `mpr_refine_portal`: count iterations like `mpr_find_penetration` does; past `CCD_ITERATIONS` mark the
  portal `PORTAL_STATUS.UNCONVERGED` and return -1 (no portal), which the caller already treats as
  "no contact for this pair" — the same outcome libccd chooses.
- New `ErrorCode.UNCONVERGED_CONTACT_PORTAL` (bit 7), set for the env at the three call sites where `errno`
  is in scope when a detection ends with no contact and an UNCONVERGED portal (after the retry loop, so a
  successful retry or GJK fallback is not reported). It is exposed through `RigidSolver.get_error_envs_mask()`
  only; `check_errno()` does not raise on it, because the step is still valid (one pair, one substep,
  reported as not colliding) and raising would abort long batched trainings on a transient event that the
  previous behaviour turned into a hang. Users who want to reset the affected env can do so from the mask.
  Before the perturbed re-detection of the multicontact kernel the slot's status is cleared, since that
  detection may take an analytic path that never touches the MPR state (otherwise a stale UNCONVERGED from
  an earlier pair in the same thread would be reported against the wrong pair).

Diff: 28 lines added, 0 removed, in `mpr.py`, `narrowphase.py`, `utils/array_class.py`. No option, no
behaviour change for portals that converge.

Not included (follow-up): the three `while is_improving*` chases in
`multi_contact.py::func_approximate_polygon_with_quad` are also unbounded and spin on a NaN
clipped-polygon vertex; they are only reachable with `enable_contact_patch=True` (see #3314).

## Test

`tests/rigid/test_collision.py::test_mpr_portal_discovery_is_bounded` (`required`, CPU): a spawned child
builds a one-box scene, stubs the three predicates that end the discovery loop so that every iteration
expands the portal again (what a portal growing by rounding-sized steps does on real geometry), and calls
`mpr_refine_portal` from a one-thread kernel. The parent fails the test if the child has not returned after
300 s. With the fix the call returns -1 with the portal marked UNCONVERGED.

Verified against `main` (48a690a3) in a fresh env (Python 3.12, quadrants 1.3.0, torch 2.8.0+cu128, CPU):

    python -m pytest tests/rigid/test_collision.py -k test_mpr_portal_discovery_is_bounded --backend=cpu --numprocesses=1 -p no:pytest-retry -p no:rerunfailures

- with the fix: `1 passed in 31.48s`
- on unpatched `main`: `Failed: mpr_refine_portal did not return within 300 s: the portal-discovery loop is unbounded.` (`1 failed in 306.84s`)

`pre-commit run --files <changed files>`: ruff check and ruff format pass.

## Tested on (full training workload, genesis-world 1.3.3 + this change as a patch)

- RTX 5090 (sm_120, 595.84): monolith arm pinned, 512 envs — stock froze after update 1 in 3 of 3 runs on a
  bit-identical trajectory; patched: one portal cap in one env in that interval, env flagged, run completed
  10 updates (and 10 more with zero-copy views off). Decomposed arm, 256 envs, device conditional graph —
  stock froze at update 16 twice; patched: 31 clean updates, one cap in one env at step 4096, run continued.
- RTX 5070 (sm_120, 580.173.02), L40S (sm_89, 580.173.02): stock monolith-arm freezes at per-env step
  216 / 185 (#3310).
- A100-PCIE-40GB (sm_80, 580.173.02): stock decomposed-arm freeze at the same PPO update (49) for the same
  seed in three allocations; patched result pending (will be posted on #3314).
- CPU: the test above.

### DIFF
--- a/genesis/engine/solvers/rigid/collider/mpr.py
+++ b/genesis/engine/solvers/rigid/collider/mpr.py
@@ -277,11 +277,22 @@ def mpr_refine_portal(
     collider_static_config: qd.template(),
 ):
     ret = 1
+    iterations = 0
     while True:
         direction = mpr_portal_dir(i_ga, i_gb, i_b, mpr_state)
 
         if mpr_portal_encapsules_origin(i_ga, i_gb, i_b, direction, mpr_state, collider_info):
             ret = 0
+            break
+
+        iterations += 1
+        if iterations > collider_info.mpr.CCD_ITERATIONS[None]:
+            # Same bound as the refinement in mpr_find_penetration (and libccd's): a portal that keeps expanding by
+            # rounding-sized steps without ever encapsulating the origin or reaching the tolerance would otherwise
+            # spin forever, which wedges the whole GPU kernel. Report it as no portal, with a status the callers can
+            # tell from a genuine miss.
+            mpr_state.portal_status[i_b] = PORTAL_STATUS.UNCONVERGED
+            ret = -1
             break
 
         v, v1, v2 = compute_support(
--- a/genesis/engine/solvers/rigid/collider/narrowphase.py
+++ b/genesis/engine/solvers/rigid/collider/narrowphase.py
@@ -1954,6 +1954,10 @@ def func_convex_convex_contact(
                                 )
                                 is_mpr_updated = True
 
+                        if not is_col and mpr_state.portal_status[i_b] == PORTAL_STATUS.UNCONVERGED:
+                            # Portal discovery hit its cap: the pair is dropped for this step and the env is flagged.
+                            errno[i_b] = errno[i_b] | array_class.ErrorCode.UNCONVERGED_CONTACT_PORTAL
+
                         if qd.static(collider_static_config.ccd_algorithm == CCD_ALGORITHM_CODE.MPR):
                             prefer_gjk = func_prefer_gjk_refinement(
                                 i_pair,
@@ -2541,6 +2545,7 @@ def _func_multicontact_mpr(
                 gb_pos_original, gb_quat_original, contact0_pos, gu.qd_inv_quat(qrot)
             )
 
+            mpr_state.portal_status[i_scratch] = PORTAL_STATUS.NONE
             is_col, normal, contact_pos, penetration, _used_gjk = _func_multicontact_run_detection(
                 i_ga,
                 i_gb,
@@ -2569,6 +2574,10 @@ def _func_multicontact_mpr(
                 ),
                 is_initial_detection=False,
             )
+            if not is_col and mpr_state.portal_status[i_scratch] == PORTAL_STATUS.UNCONVERGED:
+                # Portal discovery hit its cap (and the GJK fallback found nothing either): the perturbed contact is
+                # dropped for this step and the env is flagged.
+                errno[i_b] = errno[i_b] | array_class.ErrorCode.UNCONVERGED_CONTACT_PORTAL
 
             if qd.static(collider_static_config.ccd_algorithm == CCD_ALGORITHM_CODE.MPR):
                 if is_col:
@@ -2975,6 +2984,10 @@ def _func_narrowphase_contact0(
                                 collider_static_config,
                             )
                             is_mpr_updated = True
+
+                    if not is_col and mpr_state.portal_status[flat_idx] == PORTAL_STATUS.UNCONVERGED:
+                        # Portal discovery hit its cap: the pair is dropped for this step and the env is flagged.
+                        errno[i_b] = errno[i_b] | array_class.ErrorCode.UNCONVERGED_CONTACT_PORTAL
 
                     if qd.static(collider_static_config.ccd_algorithm == CCD_ALGORITHM_CODE.MPR):
                         prefer_gjk = func_prefer_gjk_refinement(
--- a/genesis/utils/array_class.py
+++ b/genesis/utils/array_class.py
@@ -89,6 +89,10 @@ class ErrorCode(IntEnum):
     INVALID_CONTACT_NAN = 0b00000000000000000000000000010000
     INVALID_FORCE_NAN = 0b00000000000000000000000000100000
     INVALID_ACC_NAN = 0b00000000000000000000000001000000
+    # MPR portal discovery hit its iteration cap for a pair of this env, and that pair was reported as not colliding
+    # for the step. Exposed through 'RigidSolver.get_error_envs_mask' only: the step is still valid, so 'check_errno'
+    # does not raise on it.
+    UNCONVERGED_CONTACT_PORTAL = 0b00000000000000000000000010000000
 
 
 # =========================================== RigidInfo ===========================================
--- a/tests/rigid/test_collision.py
+++ b/tests/rigid/test_collision.py
@@ -1407,3 +1407,129 @@ def test_neutral_self_collision_masks_across_merged_entities(merged_overlapping_
     palm_geom = hand.get_link("palm").geoms[0].idx
     assert_equal(collision_pair_idx[a2_geom, palm_geom], -1)
     assert_equal(collision_pair_idx[palm_geom, a2_geom], -1)
+
+
+def _mpr_refine_portal_with_stubbed_predicates(queue):
+    """Call mpr_refine_portal on a portal that can never succeed and report its return value.
+
+    Runs in a child process: without an iteration cap the loop never returns, and only the parent can time that out.
+    The three predicates that end the loop are stubbed so that every iteration expands the portal again, which is
+    what a portal that keeps growing by rounding-sized steps does on real geometry.
+    """
+    import quadrants as qd
+
+    import genesis as gs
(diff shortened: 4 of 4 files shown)


duburcqa:
Do you have any proof of what you pretend, with a reproduction script that anybody could try on their side? Without such repro, we are not going to merge a speculative fix. Note that a unit test hacking the internals of an algorithm prove nothing, because the faulty path it exercise may not be reachable in reality. If you can share a reproduction script, I will fix it myself promptly. You may want to use the feature that was introduced a few days ago for this that always to save a scene as a standalone archive that requires no assets or whatever to be loaded. See tests/rigid/test_serialization.py.

chkxw:
Ampere before/after with this patch (A100-PCIE-40GB, deterministic mode, 512 envs): stock 1.3.3 stops at PPO update 49 in four allocations, patched completes 190 updates with seven portal caps, each in one env, run continuing — details and the per-cap table in https://github.com/Genesis-Embodied-AI/genesis-world/issues/3314#issuecomment-5554570652.

chkxw:
Understood — a standalone reproduction is the right bar, and I will use the scene-archive feature for it: archive the 512-env scene at the last sampled state before the freeze, plus the recorded per-step position targets for the following window, so that stepping the archive with those targets stops on `main` and completes (with one flagged env) with the bound. I will post it here once it reproduces from the archive on our side rather than from the full workload, and I am fine with you fixing it your way once it does.

For the record, the claim that the path is reached is not from the unit test. It rests on (a) two `cuda-gdb` captures of the frozen process with one lane inside `mpr_refine_portal` via the multi-contact narrowphase kernel, and (b) runs of the same trajectory with and without the bound, where the bound fires in the same window in which the unbounded build stops — both seeds, A100 and RTX 5090, details in #3314. The archive will make that checkable without our workload.

duburcqa:
@chkxw You may want to use this PR to record the entire trajectory: https://github.com/Genesis-Embodied-AI/genesis-world/pull/3309

It is going to be merged very soon.

chkxw:
Thanks — #3309 is exactly what this needs (the archive alone carries no warm-start or contact state, which is what makes the spin trajectory-dependent, and our replay from the archive was not bit-faithful for that reason). Switching the reproduction to `save_checkpoint` / `load_checkpoint` from your branch: a checkpoint taken a few hundred steps before the freeze plus the recorded control targets for the window, so that stepping the checkpoint stops on the unbounded loop and continues with the bound. Will post the files here once it reproduces from the checkpoint on our side.

duburcqa:
@chkxw Great news! I'm adding this feature precisely to make debugging similar issues from the community much easier for everybody!

PR #3323, opened 2026-09-06
[BUG FIX] Preserve glTF normals and texture coordinates.

### LINKED ISSUE #3320: [Bug]: GLB loader drops UV coordinates when their accessor index is 0
### Bug Description

I noticed that the GLB loader drops texture coordinates when `TEXCOORD_0` points to accessor index `0`.

  In `genesis/utils/gltf.py`, line 388 checks:

  ```python
  if primitive.attributes.TEXCOORD_0:
  ```

  Index `0` is valid, but this condition treats it as missing. The loaded mesh ends up with zero UV coordinates instead of the values in the file.

  The checks for `NORMAL` and `TEXCOORD_1` use the same pattern, although this reproduction only tests `TEXCOORD_0`.

### Steps to Reproduce

It creates two GLB files with identical geometry and UV coordinates. The only difference is whether the UV accessor is at index `0` or `1`. Both
  files are loaded with the same textured surface. No external assets are needed.

  ```python
  import tempfile
  from pathlib import Path

  import genesis as gs
  import numpy as np
  import pygltflib as gltf

  POSITIONS = np.array([[0, 0, 0], [1, 0, 0], [0, 1, 0]], dtype=np.float32)
  UVS = np.array([[0.1, 0.2], [0.3, 0.4], [0.5, 0.6]], dtype=np.float32)


  def write_glb(path, uv_first):
      arrays = [UVS, POSITIONS] if uv_first else [POSITIONS, UVS]
      views, accessors = [], []
      blob = b""
      for array in arrays:
          views.append(gltf.BufferView(buffer=0, byteOffset=len(blob), byteLength=array.nbytes))
          accessors.append(gltf.Accessor(
              bufferView=len(views) - 1, componentType=gltf.FLOAT,
              count=len(array), type=f"VEC{array.shape[1]}",
              min=array.min(axis=0).tolist(), max=array.max(axis=0).tolist(),
          ))
          blob += array.tobytes()
      model = gltf.GLTF2(
          scene=0, scenes=[gltf.Scene(nodes=[0])], nodes=[gltf.Node(mesh=0)],
          meshes=[gltf.Mesh(primitives=[gltf.Primitive(
              attributes=gltf.Attributes(POSITION=1 if uv_first else 0,
                                         TEXCOORD_0=0 if uv_first else 1), mode=4,
          )])],
          buffers=[gltf.Buffer(byteLength=len(blob))],
          bufferViews=views, accessors=accessors,
      )
      model.set_binary_blob(blob)
      model.save_binary(str(path))


  print("Genesis:", gs.__file__)
  surface = gs.surfaces.Default(
      diffuse_texture=gs.textures.ImageTexture(image_array=np.ones((2, 2, 3), dtype=np.float32))
  )
      for uv_first in (True, False):
          path = Path(directory) / f"uv_at_{0 if uv_first else 1}.glb"
          write_glb(path, uv_first)
          mesh = parse_mesh_glb(str(path), group_by_material=False, scale=1.0,
                                is_mesh_zup=True, surface=surface)[0]
          print(path.name)
          print(mesh.uvs)
          print("Matches authored UVs:", np.allclose(mesh.uvs, UVS, atol=1e-6))
  gs.destroy()
  ```

### Expected Behavior

Both files should return the same authored UV coordinates, regardless of accessor ordering:

  ```text
  [[0.1 0.2]
   [0.3 0.4]
   [0.5 0.6]]
  ```

  `Matches authored UVs` should be `True` for both files.

### Relevant log output

```shell
uv_at_0.glb
  [[0. 0.]
   [0. 0.]
   [0. 0.]]
  Matches authored UVs: False
  uv_at_1.glb
  [[0.1 0.2]
   [0.3 0.4]
   [0.5 0.6]]
  Matches authored UVs: True
```

### Environment

- OS: macOS 15.7.3
- Backend: CPU
- GPU-driver version: N/A

  351ffe033d1d0f0bf32918c43d9255489fdf7896

### Additional Context

_No response_

jeetrex17:
## Description

The glTF parser tested the NORMAL, TEXCOORD_0 and TEXCOORD_1 accessor indices for truthiness. Accessor index `0` is valid but falsy, so normals or UVs stored in the first accessor were silently dropped: UVs came back as zeros and normals were recomputed from the geometry. The three checks now use `is not None`, like the `indices` and `material` checks in the same function. Absent attributes keep their existing fallback.

The existing `test_glb_parse_geometry` now covers NORMAL, TEXCOORD_0 and TEXCOORD_1 at accessor 0 alongside its existing asset cases. Each added case exports a small textured triangle and reorders the accessor table to put the selected attribute first. All cases share the existing geometry, normals and UV comparisons against trimesh.

## Related Issue

Resolves #3320

## Motivation and Context

Which accessor an attribute lands in is up to the exporter, so a perfectly valid glTF file could lose its texture mapping or shading normals on import with no warning.

```
pytest -n 0 --backend cpu tests/parsers/test_mesh.py -k "glb_parse_geometry or glb_draco_missing_normals_texcoord or glb_uv_set"
```

All eight selected tests pass on macOS (Apple M4), CPU backend. The three accessor-zero cases fail on the normals or UV comparison with the original parser from `351ffe03`; the two existing geometry cases pass. Ruff checks pass. Full suite and GPU backends not run.

## Checklist:

### DIFF
--- a/tests/parsers/conftest.py
+++ b/tests/parsers/conftest.py
@@ -42,6 +42,43 @@ def mesh_urdf(mesh_path):
     return ET.tostring(robot, encoding="unicode")
 
 
+@pytest.fixture
+def glb_path(glb_file, accessor_zero_attribute, asset_tmp_path):
+    """Path to an existing GLB or a generated mesh with the selected attribute at accessor zero."""
+    if accessor_zero_attribute is None:
+        return os.path.join(get_hf_dataset(pattern=glb_file), glb_file)
+
+    # Authored shading normals differ from the triangle's geometric normal
+    mesh = trimesh.Trimesh(
+        vertices=[[0.0, 0.0, 0.0], [1.0, 0.0, 0.0], [0.0, 1.0, 0.0]],
+        faces=[[0, 1, 2]],
+        vertex_normals=[[1.0, 0.0, 0.0]] * 3,
+        visual=trimesh.visual.TextureVisuals(
+            uv=[[0.125, 0.25], [0.375, 0.5], [0.625, 0.75]],
+            material=trimesh.visual.material.PBRMaterial(baseColorTexture=Image.new("RGB", (2, 2), "white")),
+        ),
+        process=False,
+    )
+    path = asset_tmp_path / f"accessor_zero_{accessor_zero_attribute}.glb"
+    mesh.export(path, include_normals=True)
+    glb = pygltflib.GLTF2().load(path)
+    primitive = glb.meshes[0].primitives[0]
+    attributes = primitive.attributes
+    first_accessor = attributes.NORMAL if accessor_zero_attribute == "NORMAL" else attributes.TEXCOORD_0
+    # Reorder the accessor table while keeping every reference attached to its authored data
+    order = [first_accessor] + [i for i in range(len(glb.accessors)) if i != first_accessor]
+    glb.accessors = [glb.accessors[i] for i in order]
+    attributes.POSITION = order.index(attributes.POSITION)
+    attributes.NORMAL = order.index(attributes.NORMAL)
+    attributes.TEXCOORD_0 = order.index(attributes.TEXCOORD_0)
+    primitive.indices = order.index(primitive.indices)
+    if accessor_zero_attribute == "TEXCOORD_1":
+        attributes.TEXCOORD_1 = attributes.TEXCOORD_0
+        glb.materials[primitive.material].pbrMetallicRoughness.baseColorTexture.texCoord = 1
+    glb.save_binary(str(path))
+    return str(path)
+
+
 # Conversion from .usd to .glb significantly affects precision
 USD_COLOR_TOL = 1e-07
 
--- a/tests/parsers/conftest.py
+++ b/tests/parsers/conftest.py
@@ -508,13 +508,12 @@ def usd_scene(request, model_name, scale, fixed):
     return build_usd_scene(request.getfixturevalue(model_name), scale=scale, fixed=fixed)
 
 
-def _build_textured_triangle_glb(asset_tmp_path, name):
-    """Export a textured triangle whose authored normals differ from its geometric normal, so a parser that drops them
-    and recomputes them from the geometry fails the normals comparison. Returns the loaded glTF document and its
-    path."""
+def _build_textured_triangle_glb(asset_tmp_path, name, first_attribute):
+    """Build a textured triangle with NORMAL or TEXCOORD_0 at accessor zero, returning the glTF document and path."""
     mesh = trimesh.Trimesh(
         vertices=[[0.0, 0.0, 0.0], [1.0, 0.0, 0.0], [0.0, 1.0, 0.0]],
         faces=[[0, 1, 2]],
+        # Authored shading normals differ from the triangle's geometric normal
         vertex_normals=[[1.0, 0.0, 0.0]] * 3,
         visual=trimesh.visual.TextureVisuals(
             uv=[[0.125, 0.25], [0.375, 0.5], [0.625, 0.75]],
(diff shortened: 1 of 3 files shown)


tests/parsers/test_mesh.py
```diff
@@ -290,10 +290,51 @@ def test_urdf_mesh_processing(mesh_path, mesh_urdf, show_viewer):
...
+        glb.accessors = [glb.accessors[i] for i in order]
+        attributes.POSITION = order.index(attributes.POSITION)
+        attributes.NORMAL = order.index(attributes.NORMAL)
+        attributes.TEXCOORD_0 = order.index(attributes.TEXCOORD_0)
+        primitive.indices = order.index(primitive.indices)
+        if accessor_zero_attribute == "TEXCOORD_1":
+            attributes.TEXCOORD_1 = attributes.TEXCOORD_0
+            glb.materials[primitive.material].pbrMetallicRoughness.baseColorTexture.texCoord = 1
+        glb.save_binary(str(glb_path))
+        glb_file = str(glb_path)
```
duburcqa: Let's move this in a dedicated fixture, this is distracting.

jeetrex17: Yeah, this was originally a separate test in a72a252 😅 Then I reread CODING_GUIDELINES.md about extending existing tests and folded the setup in too. Took that a little too literally , I'll move the setup into a fixture asap.

tests/parsers/conftest.py
```diff
@@ -42,6 +42,43 @@ def mesh_urdf(mesh_path):
...
+    glb.accessors = [glb.accessors[i] for i in order]
+    attributes.POSITION = order.index(attributes.POSITION)
+    attributes.NORMAL = order.index(attributes.NORMAL)
+    attributes.TEXCOORD_0 = order.index(attributes.TEXCOORD_0)
+    primitive.indices = order.index(primitive.indices)
+    if accessor_zero_attribute == "TEXCOORD_1":
+        attributes.TEXCOORD_1 = attributes.TEXCOORD_0
+        glb.materials[primitive.material].pbrMetallicRoughness.baseColorTexture.texCoord = 1
+    glb.save_binary(str(path))
+    return str(path)
```
duburcqa: This is not generic. You should rather follow the pattern of '_build_plane_contact_model' + 'box_plan' / 'sphere_plane_roll' / 'sphere_plane_spin'...etc

tests/parsers/conftest.py
```diff
@@ -499,6 +508,69 @@ def usd_scene(request, model_name, scale, fixed):
...
+            material=trimesh.visual.material.PBRMaterial(baseColorTexture=Image.new("RGB", (2, 2), "white")),
+        ),
+        process=False,
+    )
+    path = str(asset_tmp_path / f"{name}.glb")
+    mesh.export(path, include_normals=True)
+    return pygltflib.GLTF2().load(path), path
+
+
+def _move_accessor_first(glb, accessor):
```
duburcqa: Why 2 helpers instead of 1?

tests/parsers/conftest.py
```diff
@@ -42,6 +42,15 @@ def mesh_urdf(mesh_path):
     return ET.tostring(robot, encoding="unicode")
 
 
+@pytest.fixture
+def glb_path(request, glb_file):
+    """The GLB a test parses. An asset-relative path resolves through the dataset. A bare name is the fixture
+    generating the file."""
+    if "/" in glb_file:
+        return os.path.join(get_hf_dataset(pattern=glb_file), glb_file)
+    return request.getfixturevalue(glb_file)
```
duburcqa: Just do this instead of unit test as it has been the case so far.

duburcqa: I mean, remove this and inline this logics inside the unit test directly.

tests/parsers/conftest.py
duburcqa, line 1: This is good.

tests/parsers/conftest.py
jeetrex17: Removed `glb_path` and inlined the resolution in the geometry test. Restored the Draco/material tests’ inline asset loading.

duburcqa (approved):

Issue #3179, opened 2026-08-05
[Bug]: One environment's constraint NaN kills the whole batched scene — no per-env fault isolation or failing-env index


Enoch208:
### Bug Description

With `scene.build(n_envs=N)`, when any single environment's rigid solver produces invalid constraint forces, `scene.step()` raises `GenesisException: Invalid constraint forces causing 'nan'` for the **entire scene**. Two problems compound:

1. **No quarantine** — at `n_envs=4096`, one bad contact configuration destroys 4,095 healthy rollouts mid-step.
2. **No environment index** — the exception carries no indication of which environment went non-finite, so the failing rollout cannot be excluded, reset, or even analysed.

We hit this repeatedly in a contact-rich manipulation batch (16-link articulated cable + Franka, 16–64 envs, `gs.amdgpu` backend): one environment's gripper/floor/cable interaction goes NaN and the full batch dies.

### Steps to Reproduce

The failure is contact-dependent and, on this stack, not deterministic — from bit-identical resets we measured contact rollouts diverging up to 256 mm in final cable position, so the repro is statistical rather than a fixed script. The reliable recipe:

1. Build a batched scene (16+ envs) with an articulated chain (~16 links) and a Franka, `scene.build(n_envs=N)`.
2. Run a force-controlled grasp-and-drag policy against the chain so fingers, chain, and floor stay in sustained contact.
3. Within a few hundred steps, some run of the batch raises `GenesisException` from `scene.step()` for all environments.

Happy to provide our full scene setup and driver if useful (public at [Enoch208/Crux](https://github.com/Enoch208/Crux), `src/crux/simulation/batchscene.py`).

### Expected Behavior

Per-environment fault isolation — quarantine the offending environment (freeze it, flag it, or reset it) and let the remaining N−1 continue; at minimum, include the failing environment index (or a per-env validity mask) in the exception so callers can salvage and analyse.

### Relevant log output

```
gs.GenesisException: Invalid constraint forces causing 'nan'.
```

(raised by `scene.step()`; our batch runner catches it, records every in-flight episode as failed-unstable, and salvages already-finished episodes — but the healthy rollouts are unrecoverable)

### Environment

- OS: Ubuntu 24.04.4 LTS
- GPU/CPU: AMD Radeon PRO W7900 (gfx1100), 48 GB
- GPU-driver version: amdgpu, ROCm 7.2.1 (`torch 2.13.0+rocm7.2`)
[...]

PR #3327, opened 2026-09-08
[BUG FIX] Identify failed environments and clear their error flags on reset.

jeetrex17:
## Description

Rigid-solver exceptions now identify environments reporting the displayed error separately from environments reporting other errors. For non-finite force and acceleration errors, batched workloads receive guidance to use `RigidSolver.get_error_envs_mask()`, whose recovery workflow is now documented in its docstring.

Selective resets through `set_state`, `set_qpos`, and `set_dofs_position` cleared the wrong environment's error flag on the kernel fallback path. With zero-copy disabled, a reset environment could remain flagged and trigger another exception. `kernel_set_zero` now clears flags using the selected environment indices.

  ## Related Issue

  Addresses #3179

  ## Motivation and Context

The existing boolean-mask workflow lets callers reset failing environments while preserving healthy episodes. This change makes that workflow discoverable in error messages and fixes selective flag clearing when zero-copy is disabled.

  ```bash
  pytest tests/rigid/test_dynamics.py -k test_reset_envs_flagged_by_rigid_solver --backend cpu
  pytest tests/rigid/test_collision.py -k "test_num_contact_overflow and spheres and cpu"
  ```

  - The required reset test runs with zero-copy disabled and at its default. It checks the failing-environment mask, healthy motion during the failing step, selective reset, continued motion after recovery, and reporting of a later unhandled failure. The disabled case fails before the indexing fix and passes afterward.
  - The required mixed-error case produces contact overflow in environment 0, keeps environment 1 healthy, and produces a numerical failure in environment 2. It checks error priority, separate diagnostic groups, and the public mask. Existing unbatched cases check that environment lists are omitted.
  - A non-finite velocity supplied through the public setter deterministically exercises the failure path.
  - Passed locally on macOS: reset cases on CPU and Metal, including Metal with CPU-side Torch tensors where zero-copy is unavailable; CPU spheres cases with zero-copy enabled and disabled; and the required bowls case on CPU.
  - Not run: the full suite, CUDA/ROCm, and the mixed-error collision case on Metal. The non-Darwin, zero-copy-disabled reduction path, unbatched numerical-error recovery-hint suppression, and Metal static-array mode with `torch<=2.9.1` remain unverified.

  ## Checklist:

### DIFF
--- a/genesis/engine/solvers/rigid/rigid_solver.py
+++ b/genesis/engine/solvers/rigid/rigid_solver.py
@@ -1325,8 +1325,7 @@ class RigidSolver(GravityMixin, TimeBasedMixin, KinematicSolver):
 
     def check_errno(self):
         """Raise for a solver error and list the affected environments in batched scenes."""
-        # FIXME: qd.atomic_or return value is broken on Metal - always returns 0.
-        # See repro_metal_kernel_return.py. Falling back to numpy reduction.
+        # FIXME: qd.atomic_or return value is broken on Metal - always returns 0. Falling back to numpy reduction.
         errno_by_env = None
         if gs.use_zerocopy or sys.platform == "darwin":
             errno_by_env = qd_to_numpy(self._errno, transpose=True)
--- a/genesis/engine/solvers/rigid/rigid_solver.py
+++ b/genesis/engine/solvers/rigid/rigid_solver.py
@@ -1308,45 +1308,80 @@ class RigidSolver(GravityMixin, TimeBasedMixin, KinematicSolver):
                 )
 
     def get_error_envs_mask(self):
+        """Return a mask selecting environments with a rigid-solver error.
+
+        An environment is flagged after the solver encounters an error such as a non-finite constraint force or
+        acceleration. The flag remains set until the environment state is reset.
+
+        If a flag remains set, a later `Scene.step` raises. For non-finite force and acceleration errors, batched
+        workloads can read this mask after each step and reset the affected environments before stepping again.
+
+        Returns
+        -------
+        error_envs_mask : torch.Tensor
+            Boolean tensor with one value per built environment. A non-parallelized scene returns one value.
+        """
         return qd_to_torch(self._errno) > 0
 
     def check_errno(self):
-        # FIXME: qd.atomic_or return value is broken on Metal — always returns 0.
+        """Raise if any environment is flagged as invalid."""
+        # FIXME: qd.atomic_or return value is broken on Metal - always returns 0.
         # See repro_metal_kernel_return.py. Falling back to numpy reduction.
         if gs.use_zerocopy or sys.platform == "darwin":
             errno = np.bitwise_or.reduce(qd_to_numpy(self._errno))
         else:
             errno = kernel_bit_reduction(self._errno)
+        if not errno:
+            return
+
+        envs_info = ""
+        if self.n_envs > 0:
+            envs_info = f" Environments in error: {np.nonzero(qd_to_numpy(self._errno))[0].tolist()}."
+
+        recovery_info = (
+            " Read RigidSolver.get_error_envs_mask() after each step and reset the environments it reports so the "
+            "rest of the batch can continue."
+            if self.n_envs > 0
+            else ""
+        )
 
         if errno & array_class.ErrorCode.OVERFLOW_CANDIDATE_CONTACTS:
             max_collision_pairs_broad = self.collider.collider_info.max_collision_pairs_broad[None]
             gs.raise_exception(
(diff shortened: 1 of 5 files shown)


duburcqa:
I don't think this is solving anything. It is not useless, but what matters most is documentation in genesis-doc repo.

jeetrex17:
Agreed, the missing guide is the main part of #3179. I've opened https://github.com/Genesis-Embodied-AI/genesis-doc/pull/152 covering the recovery workflow and what to investigate when failures repeat.

While validating the example, I found that with zero-copy disabled, resetting environment 1 cleared environment 0's error flag instead. Environment 1 stayed flagged after reset and the scene raised at the next periodic check. The regression test fails before the indexing fix and passes with it.

The exception-message changes go beyond the documentation request, but help users identify the failing environments and find the existing mask API. They don't address the underlying numerical instability. The reset fix makes the documented recovery work on the fallback path.

Happy to trim this to the reset fix if you'd rather keep the message changes separate.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -1308,45 +1308,91 @@ def substep(self, f):
...
+        error_envs_mask : torch.Tensor
+            Boolean tensor with one value per built environment. A non-parallelized scene returns one value.
+        """
         return qd_to_torch(self._errno) > 0
 
     def check_errno(self):
-        # FIXME: qd.atomic_or return value is broken on Metal — always returns 0.
+        """Raise for a solver error and list the affected environments in batched scenes."""
+        # FIXME: qd.atomic_or return value is broken on Metal - always returns 0.
         # See repro_metal_kernel_return.py. Falling back to numpy reduction.
```
duburcqa: > 'See repro_metal_kernel_return.py.'

Remove this, it makes no sense.

genesis/engine/solvers/rigid/rigid_solver.py
```diff
@@ -1308,45 +1308,91 @@ def substep(self, f):
...
+        errno_by_env = None
         if gs.use_zerocopy or sys.platform == "darwin":
-            errno = np.bitwise_or.reduce(qd_to_numpy(self._errno))
+            errno_by_env = qd_to_numpy(self._errno, transpose=True)
+            errno = np.bitwise_or.reduce(errno_by_env)
         else:
             errno = kernel_bit_reduction(self._errno)
+        if not errno:
+            return
```
duburcqa, line 1336: Could you check if quadrants now offers some built-in bit reduction?

tests/rigid/test_collision.py
```diff
@@ -12,6 +12,8 @@
 
 import genesis as gs
 import genesis.utils.geom as gu
+from genesis.engine.simulator import RATE_CHECK_ERRNO
+from genesis.utils.array_class import RigidSimStaticConfig
```
duburcqa: Keep this import in local scope. (long to explain, it is to avoid conftest job spawning issues, not sure if it is necessary here but it is safer)

tests/rigid/test_dynamics.py
```diff
@@ -859,6 +859,67 @@ def rigid_solver_build(self):
     assert_allclose(*values, tol=5e-4)
 
 
+@pytest.mark.required
+@pytest.mark.parametrize("use_zerocopy", [False, None])
```
duburcqa: Do not do this. Just run the test as is. The CI will take care of running 'use_zerocopy=False'.

tests/conftest.py
duburcqa, line 1: Revert this, it is not necessary, the CI is already running both.

duburcqa: Not a great software dev practice, but it spares compute. Similar to not specifying the both CPU and GPU to every single test, and quadrants ndarray vs field.

genesis/engine/solvers/rigid/rigid_solver.py
jeetrex17: I had a look at both Quadrants 1.3.0 and current main. `qd.algorithms` only has add/min/max reductions, no bitwise OR. The closest thing is `qd.simt.block.reduce`, which takes a custom operator, but it works within a single block, so we'd still need our own kernel on top to combine across blocks. The old subgroup `reduce_or` was removed a while back (there's a note in `simt/reductions.py`).

So nothing built-in to swap this for, unless you'd rather I go the `block.reduce` route.

jeetrex17:
@duburcqa i think its Ready for another look when you have time. All your comments are addressed in d50d6da (local test import, dropped the zero-copy parametrization and the conftest change, removed the stray repro reference)

PR #3330, opened 2026-09-08
[BUG FIX] Decode normalized integer glTF accessors instead of reading their raw values.

### LINKED ISSUE #3329: [Bug]: GLB loader reads normalized integer texture coordinates at their raw scale
### Bug Description

`get_glb_data_from_accessor` in `genesis/utils/gltf.py` ignores `accessor.normalized`. glTF 2.0 allows `TEXCOORD_n` to be stored as normalized `UNSIGNED_BYTE` or `UNSIGNED_SHORT` without any extension, and `KHR_mesh_quantization` extends the same encoding to normals. Exporters such as gltfpack write texture coordinates this way by default. Genesis reads the raw integers, so an authored UV of `0.1` comes back as `6554` from a uint16 file and as `26` from a uint8 file. The texture is then sampled far outside the image.

### Steps to Reproduce

The script writes the same triangle three times: float texture coordinates, normalized `UNSIGNED_SHORT`, and normalized `UNSIGNED_BYTE`. Only the encoding differs.

```python
import tempfile
from pathlib import Path

import genesis as gs
import numpy as np
import pygltflib as gltf

POSITIONS = np.array([[0, 0, 0], [1, 0, 0], [0, 1, 0]], dtype=np.float32)
UVS = np.array([[0.1, 0.2], [0.3, 0.4], [0.5, 0.6]], dtype=np.float32)


def write_glb(path, encoding):
    if encoding == "uint16":
        uvs, ctype = np.round(UVS * 65535).astype(np.uint16), gltf.UNSIGNED_SHORT
    elif encoding == "uint8":
        uvs, ctype = np.round(UVS * 255).astype(np.uint8), gltf.UNSIGNED_BYTE
    else:
        uvs, ctype = UVS, gltf.FLOAT
    blob, views, accessors = b"", [], []
    for array, component_type, normalized in ((POSITIONS, gltf.FLOAT, None), (uvs, ctype, encoding != "float" or None)):
        views.append(gltf.BufferView(buffer=0, byteOffset=len(blob), byteLength=array.nbytes))
        accessors.append(gltf.Accessor(
            bufferView=len(views) - 1, componentType=component_type, normalized=normalized,
            count=len(array), type=f"VEC{array.shape[1]}",
            min=array.min(axis=0).tolist(), max=array.max(axis=0).tolist(),
        ))
        blob += array.tobytes()
    model = gltf.GLTF2(
        scene=0, scenes=[gltf.Scene(nodes=[0])], nodes=[gltf.Node(mesh=0)],
        meshes=[gltf.Mesh(primitives=[gltf.Primitive(
            attributes=gltf.Attributes(POSITION=0, TEXCOORD_0=1), mode=4,
        )])],
        buffers=[gltf.Buffer(byteLength=len(blob))], bufferViews=views, accessors=accessors,
    )
    model.set_binary_blob(blob)
    model.save_binary(str(path))


surface = gs.surfaces.Default(
    diffuse_texture=gs.textures.ImageTexture(image_array=np.ones((2, 2, 3), dtype=np.float32))
)
    for encoding in ("float", "uint16", "uint8"):
        path = Path(directory) / f"{encoding}.glb"
        write_glb(path, encoding)
        mesh = parse_mesh_glb(str(path), group_by_material=False, scale=1.0, is_mesh_zup=True, surface=surface)[0]
        print(encoding, mesh.uvs.round(3).tolist())
gs.destroy()
```

Output on `main`:

```text
float  [[0.1, 0.2], [0.3, 0.4], [0.5, 0.6]]
uint16 [[6554.0, 13107.0], [19660.0, 26214.0], [32768.0, 39321.0]]
uint8  [[26.0, 51.0], [76.0, 102.0], [128.0, 153.0]]
```

### Expected Behavior

All three files load with the authored texture coordinates:

```text
[[0.1, 0.2], [0.3, 0.4], [0.5, 0.6]]
```

### Screenshots/Videos

_No response_

### Relevant log output

_No response_

### Environment

- OS: macOS 26
- CPU backend, Apple M-series

v1.4.0 (`0ce793b`)

### Additional Context

I have a fix with a regression test and will open a PR.

VihaanAgarwal:
## Description

`get_glb_data_from_accessor` returned the raw integers of an accessor even when `accessor.normalized` was set. glTF allows `TEXCOORD_n` to be stored as normalized `UNSIGNED_BYTE` or `UNSIGNED_SHORT` in core, and `KHR_mesh_quantization` extends the same encoding to normals. Those attributes now map back to `[0, 1]` (unsigned) or `[-1, 1]` (signed) as the spec defines, so a UV authored as `0.1` no longer loads as `6554`.

The new `test_glb_normalized_texcoord` builds a textured triangle with normalized `UNSIGNED_SHORT` texture coordinates and checks the loaded UVs against the authored values. trimesh reads such accessors raw as well, so the test compares against the authored array rather than a trimesh load.

## Related Issue

Resolves Genesis-Embodied-AI/Genesis#3329

## Motivation and Context

gltfpack and other optimizers quantize texture coordinates to 16-bit by default, so any asset that went through them is textured wrongly in Genesis without a warning.

```
pytest -n 0 --backend cpu tests/parsers/test_mesh.py -k "glb"
```

The new test fails on `main` with UVs at their raw 16-bit scale and passes with the fix. The other glb parser tests pass before and after. Ran on macOS, CPU backend, with `ruff check` and `ruff format` clean on the touched files.

## Checklist:

### DIFF
--- a/tests/parsers/conftest.py
+++ b/tests/parsers/conftest.py
@@ -48,9 +48,6 @@ USD_COLOR_TOL = 1e-07
 
 USD_NORMALS_TOL = 1e-02
 
-# Texture coordinates authored into the triangle of the emissive material variants GLB
-GLB_TEXCOORD_UVS = np.array([[0.125, 0.25], [0.375, 0.5], [0.625, 0.75]], dtype=np.float32)
-
 
 # UsdPhysics declares every joint attribute as single precision, so an anchor, a limit or a drive gain authored in
 # double comes back rounded from a USD file and no parser can recover it. Comparing joints against the model they were
@@ -504,11 +501,13 @@ def usd_scene(request, model_name, scale, fixed):
 
 @pytest.fixture(scope="session")
 def emissive_material_variants_glb(asset_tmp_path):
-    """Path to a GLB with three materials, each on distinct base/emissive texCoord sets: a base-color atlas (red) on
-    texCoord 0 with an emissive atlas on texCoord 1, a flat base color with an emissive atlas on texCoord 1, and a
-    KHR_materials_unlit material whose red base atlas stands in for the unlit imagery. The red base atlas is index 0.
-    A triangle carries the first two materials as primitives, with the same texture coordinates stored as float in
-    set 0 and as normalized UNSIGNED_SHORT, an encoding glTF allows without any extension, in set 1."""
+    """Path to a GLB with three materials on distinct base/emissive texCoord sets and a triangle carrying two of them.
+
+    The materials are a base-color atlas (red) on texCoord 0 with an emissive atlas on texCoord 1, a flat base color
+    with an emissive atlas on texCoord 1, and a KHR_materials_unlit material whose red base atlas stands in for the
+    unlit imagery. The red base atlas is index 0. The triangle holds the first two materials as primitives, with the
+    same texture coordinates stored as float in set 0 and as normalized UNSIGNED_SHORT, an encoding core glTF allows,
+    in set 1."""
     images = []
     for color in (np.array([220, 30, 30], np.uint8), np.array([30, 220, 30], np.uint8)):
         buffer = io.BytesIO()
@@ -516,11 +515,12 @@ def emissive_material_variants_glb(asset_tmp_path):
         images.append(buffer.getvalue())
 
     positions = np.array([[0.0, 0.0, 0.0], [1.0, 0.0, 0.0], [0.0, 1.0, 0.0]], dtype=np.float32)
-    uvs_uint16 = np.round(GLB_TEXCOORD_UVS * np.iinfo(np.uint16).max).astype(np.uint16)
+    uvs = np.array([[0.125, 0.25], [0.375, 0.5], [0.625, 0.75]], dtype=np.float32)
+    uvs_uint16 = np.round(uvs * np.iinfo(np.uint16).max).astype(np.uint16)
 
     blob = b""
     buffer_views = []
-    for data in (*images, positions.tobytes(), GLB_TEXCOORD_UVS.tobytes(), uvs_uint16.tobytes()):
+    for data in (*images, positions.tobytes(), uvs.tobytes(), uvs_uint16.tobytes()):
         blob += b"\x00" * ((4 - len(blob) % 4) % 4)
         buffer_views.append(pygltflib.BufferView(buffer=0, byteOffset=len(blob), byteLength=len(data)))
         blob += data
--- a/tests/parsers/conftest.py
+++ b/tests/parsers/conftest.py
@@ -48,6 +48,9 @@ USD_COLOR_TOL = 1e-07
 
 USD_NORMALS_TOL = 1e-02
 
+# Texture coordinates authored into the normalized-integer GLB fixture
+NORMALIZED_TEXCOORD_UVS = np.array([[0.125, 0.25], [0.375, 0.5], [0.625, 0.75]], dtype=np.float32)
+
 
 # UsdPhysics declares every joint attribute as single precision, so an anchor, a limit or a drive gain authored in
 # double comes back rounded from a USD file and no parser can recover it. Comparing joints against the model they were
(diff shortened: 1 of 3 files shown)


jeetrex17:
i think its Related to #3323.

tests/parsers/conftest.py
```diff
@@ -549,6 +552,62 @@ def emissive_material_variants_glb(asset_tmp_path):
...
+        ],
+        textures=[pygltflib.Texture(source=0)],
+        images=[pygltflib.Image(bufferView=2, mimeType="image/png")],
+        bufferViews=buffer_views,
+        buffers=[pygltflib.Buffer(byteLength=len(blob))],
+    )
+    gltf.set_binary_blob(blob)
+    path = asset_tmp_path / "normalized_texcoord.glb"
+    gltf.save_binary(str(path))
+    return str(path)
```
duburcqa: Are you sure a new feature is needed? There is no way to extend an existing one instead?

tests/parsers/test_mesh.py
```diff
@@ -341,6 +342,21 @@ def test_glb_draco_missing_normals_texcoord(glb_file):
...
+    (gs_mesh,) = gltf_utils.parse_mesh_glb(
+        normalized_texcoord_glb,
+        group_by_material=False,
+        scale=None,
+        is_mesh_zup=True,
+        surface=gs.surfaces.Default(),
+    )
+    # The parser flips V to the image-space convention shared with trimesh
+    expected_uvs = NORMALIZED_TEXCOORD_UVS * (1.0, -1.0) + (0.0, 1.0)
+    assert_allclose(gs_mesh.trimesh.visual.uv, expected_uvs, tol=1.0 / np.iinfo(np.uint16).max)
```
duburcqa: Could be nice to have a single unit test checking texcoord parsing in both cases at once. The unit test would just be 'test_glb_texcoord' then.

tests/parsers/conftest.py
VihaanAgarwal: Fair point. emissive_material_variants_glb already carries materials on texCoord 0 and texCoord 1, so I will give it a triangle whose set 0 is float and set 1 is the same UVs as normalized UNSIGNED_SHORT, and drop the new fixture.

tests/parsers/test_mesh.py
VihaanAgarwal: Will do. One test_glb_texcoord that loads that file with group_by_material=True and checks both meshes come back with the authored UVs.

tests/parsers/conftest.py
```diff
@@ -503,21 +506,53 @@ def usd_scene(request, model_name, scale, fixed):
 def emissive_material_variants_glb(asset_tmp_path):
     """Path to a GLB with three materials, each on distinct base/emissive texCoord sets: a base-color atlas (red) on
     texCoord 0 with an emissive atlas on texCoord 1, a flat base color with an emissive atlas on texCoord 1, and a
-    KHR_materials_unlit material whose red base atlas stands in for the unlit imagery. The red base atlas is index 0."""
+    KHR_materials_unlit material whose red base atlas stands in for the unlit imagery. The red base atlas is index 0.
+    A triangle carries the first two materials as primitives, with the same texture coordinates stored as float in
+    set 0 and as normalized UNSIGNED_SHORT, an encoding glTF allows without any extension, in set 1."""
```
duburcqa: This docstring is not following the guidelines (CODING_GUIDELINES.md).

tests/parsers/conftest.py
```diff
@@ -48,6 +48,9 @@ def mesh_urdf(mesh_path):
 
 USD_NORMALS_TOL = 1e-02
 
+# Texture coordinates authored into the triangle of the emissive material variants GLB
+GLB_TEXCOORD_UVS = np.array([[0.125, 0.25], [0.375, 0.5], [0.625, 0.75]], dtype=np.float32)
```
duburcqa: Let's just hardcode this on both side, it is fine. This is isolated within our own testing infrastructure, so it is not going to change by magic without us noticing.

tests/parsers/conftest.py
VihaanAgarwal: Reworked into a one-sentence summary, a blank line, then the details.

tests/parsers/conftest.py
VihaanAgarwal: Done, the coordinates are written out in the fixture and again in the test.

VihaanAgarwal:
Related but separate. #3323 fixes the accessor index 0 checks in parse_mesh_glb, this one decodes normalized integer accessors in get_glb_data_from_accessor. Same file, different lines.

duburcqa (approved):

PR #3336, opened 2026-09-10
[BUG FIX] Keep the imported normals perpendicular to a non-uniformly scaled glTF node.

### LINKED ISSUE #3334: [Bug]: A non-uniform node scale in a .glb tilts the imported vertex normals off the surface
### Bug Description

A glTF node may carry a non-uniform `scale`, and exporters use it routinely (a crate stretched into a plank, a
cylinder squashed into a disc). When Genesis imports such a node, the geometry is scaled correctly but the shading
normals are not: they end up tilted away from the surface they describe, so the imported mesh renders as if lit from
the wrong direction.

The scaled triangle and its imported normals disagree with each other, which makes the defect checkable without
appealing to any external renderer: if the authored normals are the triangle's own geometric normal, the imported
ones must still be perpendicular to the imported triangle.

`genesis/utils/mesh.py::apply_transform` maps the normals through the linear part of the node transform:

```python
rot_mat = transform[:3, :3]
if np.abs(3.0 - np.trace(rot_mat)) > gs.EPS**2:  # has rotation or scaling
    transformed_normals = normals @ rot_mat
```

A normal is a covector, so it maps through the inverse transpose of that matrix, not through the matrix itself. The
two agree for a rotation and for a uniform scale (the renormalisation below absorbs the factor), and differ as soon
as the scaling is non-uniform. Re-normalising afterwards fixes the length but cannot recover the direction.

The trace test adds a second way to miss it: a scale whose components sum to 3, `(2.0, 0.5, 0.5)` for instance,
has trace exactly 3, so the whole branch is skipped and the normals are carried through untouched.

### Steps to Reproduce

Self-contained, no external assets. It writes a one-triangle GLB whose authored normals are the triangle's own
geometric normal, puts a non-uniform `scale` on the node, and compares the imported normals against the geometric
normal of the imported triangle.

```python
import tempfile
from pathlib import Path

import numpy as np
import pygltflib

import genesis as gs

POSITIONS = np.array([[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]], dtype=np.float32)


def unit_face_normal(verts):
    normal = np.cross(verts[1] - verts[0], verts[2] - verts[0])
    return normal / np.linalg.norm(normal)


def write_glb(path, node_scale):
    normals = np.tile(unit_face_normal(POSITIONS), (3, 1)).astype(np.float32)
    blob = b""
    buffer_views = []
    for data in (POSITIONS.tobytes(), normals.tobytes()):
        blob += data
    gltf = pygltflib.GLTF2(
        scene=0,
        nodes=[pygltflib.Node(mesh=0, scale=list(node_scale))],
        meshes=[
            pygltflib.Mesh(
                primitives=[pygltflib.Primitive(attributes=pygltflib.Attributes(POSITION=0, NORMAL=1), mode=4)]
            )
        ],
        accessors=[
            pygltflib.Accessor(
                bufferView=0,
                count=3,
                type="VEC3",
            ),
            pygltflib.Accessor(bufferView=1, componentType=pygltflib.FLOAT, count=3, type="VEC3"),
        ],
        bufferViews=buffer_views,
    )
    gltf.set_binary_blob(blob)
    gltf.save_binary(str(path))


    for node_scale in ((1.0, 1.0, 0.25), (2.0, 0.5, 0.5), (0.5, 0.5, 0.5)):
        path = Path(directory) / "scaled_node.glb"
        write_glb(path, node_scale)
        (mesh,) = parse_mesh_glb(
        )
        verts = np.asarray(mesh.trimesh.vertices)
        expected = unit_face_normal(verts)
        parsed = np.asarray(mesh.trimesh.vertex_normals)[0]
        print(f"node scale {node_scale}")
        print(f"  imported triangle          : {verts.tolist()}")
        print(f"  its geometric face normal  : {expected}")
        print(f"  imported vertex normal     : {parsed}")
        print(f"  they differ by             : {np.abs(parsed - expected).max():.3f}")
```

### Expected Behavior

Every imported normal stays perpendicular to the imported triangle, whatever the node scale:

```text
node scale (1.0, 1.0, 0.25)
  imported triangle          : [[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 0.25]]
  its geometric face normal  : [0.23570226 0.23570226 0.94280904]
  imported vertex normal     : [0.23570228 0.23570228 0.9428091 ]
  they differ by             : 0.000
```

### Relevant log output

```shell
node scale (1.0, 1.0, 0.25)
  imported triangle          : [[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 0.25]]
  its geometric face normal  : [0.23570226 0.23570226 0.94280904]
  imported vertex normal     : [0.69631064 0.69631064 0.17407766]
  they differ by             : 0.769

node scale (2.0, 0.5, 0.5)
  imported triangle          : [[2.0, 0.0, 0.0], [0.0, 0.5, 0.0], [0.0, 0.0, 0.5]]
  its geometric face normal  : [0.17407766 0.69631062 0.69631062]
  imported vertex normal     : [0.57735026 0.57735026 0.57735026]
  they differ by             : 0.403

node scale (0.5, 0.5, 0.5)
  imported triangle          : [[0.5, 0.0, 0.0], [0.0, 0.5, 0.0], [0.0, 0.0, 0.5]]
  its geometric face normal  : [0.57735027 0.57735027 0.57735027]
  imported vertex normal     : [0.57735032 0.57735032 0.57735032]
  they differ by             : 0.000
```

The first case is 60.5 degrees off, the second 25.2 degrees. A uniform scale is unaffected.

### Environment

- OS: Windows 11 24H2
- GPU-driver version: N/A

### Additional Context

`apply_transform` is called only from `parse_mesh_glb`, so the blast radius is the glTF importer. Happy to open a PR
mapping the normals through the cofactor matrix of the linear part, which is the inverse transpose up to a factor
the renormalisation drops, and adding a regression case to `tests/parsers/test_mesh.py`.

kevin9327:
## Description

A glTF node may scale its mesh by a different factor along each axis, and exporters use it routinely (a crate
stretched into a plank, a sphere squashed into an ellipsoid). Importing such a node scaled the geometry correctly but
left the shading normals pointing the wrong way, so the mesh rendered as if lit from somewhere else. On the sphere of
the `glb/combined_srt.glb` test asset the imported normals are up to 33 degrees off.

`apply_transform` mapped the normals through the linear part of the node transform:

```python
transformed_normals = normals @ rot_mat
```

A normal is a covector, so it maps through the cofactor matrix of that linear part, which is its inverse transpose up
to a factor the renormalisation drops. The two agree for a rotation and for a uniform scale, and part ways as soon as
the scaling is non-uniform. The guard in front of that branch tested the trace against 3, so a scale like
`(2.0, 0.5, 0.5)` also slipped through untransformed.

The cofactor matrix is built from the cross products of the linear part's rows, which is also what makes it the right
map here: it sends the cross product of two edges to the cross product of the transformed edges, so every normal
stays perpendicular to its own triangle. It stays defined when the linear part is singular, where an explicit inverse
would not.

`apply_transform` is called only from `parse_mesh_glb`, so the change is confined to the glTF importer, and within it
to nodes whose scaling is non-uniform.

### The trimesh oracle in `test_glb_parse_geometry`

`trimesh.Trimesh.apply_transform` maps the normals it holds through the linear part too, so its output cannot be the
oracle for a non-uniformly scaled node. The test now maps the authored normals itself and writes them back over
`apply_transform`'s result. A mesh that declares no `NORMAL` accessor keeps the old path: its normals are derived from
the geometry once the transform is applied, which is what both parsers do, and deriving them before the transform
and mapping them afterwards is measurably different (0.016 on `glb/combined_transform.glb`, whose normals are all
derived).

This changes what the test expects rather than loosening it: with the corrected oracle
`test_glb_parse_geometry[glb/combined_srt.glb]` fails on `b3c6c73` and passes here.

## Related Issue

Resolves Genesis-Embodied-AI/Genesis#3334

## Motivation and Context

Nothing warns, the file loads, and the geometry is right, so the only sign is that the object is shaded wrong. The
imported mesh also disagrees with itself: the authored normals of the fixture triangle are its own geometric normal,
and after the node scale the imported normals are no longer perpendicular to the imported triangle.

```
pytest -n 0 --backend cpu tests/parsers/test_mesh.py -k "glb or scale or yup or mjcf"
```

Windows 11, CPU backend, Python 3.12, `--backend cpu`.

- On `b3c6c73` with only the test changes applied, `test_glb_parse_geometry` fails on two of its six cases and passes
  on the other four:

  ```
  FAILED tests/parsers/test_mesh.py::test_glb_parse_geometry[glb/combined_srt.glb-32]
  E       Normals match failed in mesh Sphere.
  E       Mismatched elements: 6060 / 6060 (100%)
  E        [0, 1]: 0.9608578687881945 (ACTUAL), 0.7060219645500183 (DESIRED)
  E       Max absolute difference among violations: 0.43590218

  FAILED tests/parsers/test_mesh.py::test_glb_parse_geometry[non_uniform_node_scale_glb-32]
  E       Normals match failed in mesh scaled_triangle.
  E       Mismatched elements: 9 / 9 (100%)
  E        [0, 0]: 0.17407765595569785 (ACTUAL), 0.5773502588272095 (DESIRED)
  ```

  The second case is the `(2.0, 0.5, 0.5)` scale the trace test let through: Genesis returned the authored
  `[0.577, 0.577, 0.577]` unchanged where the scaled triangle's normal is `[0.174, 0.696, 0.696]`.
- With this commit all 6 cases pass.
- Pinned as unaffected: the 24 cases the selection above collects pass on `b3c6c73`, and the 25 (the added
  parametrization) pass here.
- `ruff format --check` and `ruff check` report nothing on the three edited files.

The standalone reproduction in the issue prints, on `b3c6c73`, an imported normal of
`[0.69631064 0.69631064 0.17407766]` where the imported triangle's own geometric normal is
`[0.23570226 0.23570226 0.94280904]`, and prints them equal here.

## Checklist:

### DIFF
--- a/genesis/utils/mesh.py
+++ b/genesis/utils/mesh.py
@@ -909,12 +909,16 @@ def apply_transform(transform, positions, normals=None):
 
     transformed_normals = normals
     if normals is not None:
-        rot_mat = transform[:3, :3]
-        if np.abs(3.0 - np.trace(rot_mat)) > gs.EPS**2:  # has rotation or scaling
-            transformed_normals = normals @ rot_mat
-            scale = np.linalg.norm(rot_mat, axis=1, keepdims=True)
-            if np.any(np.abs(scale - 1.0) > gs.EPS):  # has scale
-                transformed_normals /= np.linalg.norm(transformed_normals, axis=1, keepdims=True)
+        lin_mat = transform[:3, :3]
+        if not np.allclose(lin_mat, np.identity(3), atol=gs.EPS):  # has rotation or scaling
+            # A normal is a covector, so it maps through the cofactor matrix of the linear part, whose rows are the
+            # cross products of the other two rows. That matrix sends the cross product of two edges to the cross
+            # product of the transformed edges, keeping every normal perpendicular to its own triangle. The linear
+            # part itself agrees with it only for a rotation, and tilts the normals off the surface under a
+            # non-uniform scaling.
+            cofactor_mat = np.cross(np.roll(lin_mat, -1, axis=0), np.roll(lin_mat, -2, axis=0))
+            transformed_normals = normals @ cofactor_mat
+            transformed_normals /= np.linalg.norm(transformed_normals, axis=1, keepdims=True)
 
     return transformed_positions, transformed_normals
 
--- a/tests/parsers/conftest.py
+++ b/tests/parsers/conftest.py
@@ -557,6 +557,52 @@ def texcoord_1_accessor_zero_glb(asset_tmp_path):
 
 
 @pytest.fixture(scope="session")
+def non_uniform_node_scale_glb(asset_tmp_path):
+    """Path to a GLB whose node scales its triangle by a different factor along each axis.
+
+    The authored normals are the triangle's own geometric normal, which points diagonally, so the node scale has to
+    tilt them to keep them perpendicular to the scaled triangle."""
+    positions = np.array([[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]], dtype=np.float32)
+    normal = np.cross(positions[1] - positions[0], positions[2] - positions[0])
+    normals = np.tile(normal / np.linalg.norm(normal), (3, 1)).astype(np.float32)
+
+    blob = b""
+    buffer_views = []
+    for data in (positions.tobytes(), normals.tobytes()):
+        buffer_views.append(pygltflib.BufferView(buffer=0, byteOffset=len(blob), byteLength=len(data)))
+        blob += data
+
+    gltf = pygltflib.GLTF2(
+        scene=0,
+        scenes=[pygltflib.Scene(nodes=[0])],
+        nodes=[pygltflib.Node(mesh=0, scale=[2.0, 0.5, 0.5])],
+        meshes=[
+            pygltflib.Mesh(
+                name="scaled_triangle",
+                primitives=[pygltflib.Primitive(attributes=pygltflib.Attributes(POSITION=0, NORMAL=1))],
+            )
+        ],
+        accessors=[
+            pygltflib.Accessor(
+                bufferView=0,
+                componentType=pygltflib.FLOAT,
+                count=len(positions),
+                type="VEC3",
+                min=positions.min(axis=0).tolist(),
+                max=positions.max(axis=0).tolist(),
+            ),
+            pygltflib.Accessor(bufferView=1, componentType=pygltflib.FLOAT, count=len(normals), type="VEC3"),
+        ],
+        bufferViews=buffer_views,
+        buffers=[pygltflib.Buffer(byteLength=len(blob))],
+    )
+    gltf.set_binary_blob(blob)
+    path = str(asset_tmp_path / "non_uniform_node_scale.glb")
+    gltf.save_binary(path)
+    return path
+
+
+@pytest.fixture(scope="session")
 def emissive_material_variants_glb(asset_tmp_path):
     """Path to a GLB with three materials on distinct base/emissive texCoord sets and a triangle carrying two of them.
 
--- a/tests/parsers/test_mesh.py
+++ b/tests/parsers/test_mesh.py
@@ -295,6 +295,7 @@ def test_urdf_mesh_processing(mesh_path, mesh_urdf, show_viewer):
         "normal_accessor_zero_glb",
         "texcoord_0_accessor_zero_glb",
         "texcoord_1_accessor_zero_glb",
+        "non_uniform_node_scale_glb",
     ],
 )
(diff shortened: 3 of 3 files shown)


duburcqa:
Bot contributions are not welcome.

kevin9327:
Reworked against `CODING_GUIDELINES.md` and reopened as #3343, as a draft. Measuring against the closed form showed trimesh maps authored normals through the linear part too, so it cannot be the oracle here; the new test derives its expectation analytically and I left `test_glb_parse_geometry` untouched for you to decide.

PR #3337, opened 2026-09-10
[BUG FIX] Keep every glTF node its own mesh when its primitives are triangle strips or fans.

### LINKED ISSUE #3335: [Bug]: glTF triangle strips and fans lose the node their primitives came from, merging unrelated nodes
### Bug Description

When a `.glb` declares a primitive with `mode` `TRIANGLE_STRIP` (5) or `TRIANGLE_FAN` (6), the importer loses track of
which node each primitive came from. Two separate nodes get merged into a single mesh, so a scene that should hold two
independent entities holds one whose geometry spans both, and the second node's name and material are dropped.

In `genesis/utils/gltf.py::parse_mesh_glb` the outer loop walks the node list:

```python
for i, (mesh_index, mesh_transform) in enumerate(mesh_list):
```

and the strip and fan branches reuse `i` for the triangle counter:

```python
elif mode == 5:  # TRIANGLE_STRIP
    triangles = []
    for i in range(len(indices) - 2):
```

so once either branch runs, `i` holds the last triangle index rather than the node index. Both consumers read `i`
afterwards:

```python
group_idx = primitive.material if group_by_material else (i, primitive.material)
...
metadata["node_index"] = i
```

`group_idx` decides which primitives share a mesh, and `node_index` decides which submeshes are merged into one
collision geom (`RigidEntityDescription`, the `groups_by_node` block). A strip or fan whose triangle counter lands on
another node's index therefore fuses two unrelated nodes, and a node whose primitives end on different counters is
split apart.

The geometry decoding itself is fine, so nothing warns and the file loads.

### Steps to Reproduce

Self-contained, no external assets. Two nodes, one triangle each, no material. The only thing that changes between the
three runs is which node declares its triangle as a strip or a fan.

```python
import tempfile
from pathlib import Path

import numpy as np
import pygltflib

import genesis as gs

TRI_A = np.array([[0.0, 0.0, 0.0], [1.0, 0.0, 0.0], [0.0, 1.0, 0.0]], dtype=np.float32)
TRI_B = np.array([[4.0, 0.0, 0.0], [5.0, 0.0, 0.0], [4.0, 1.0, 0.0]], dtype=np.float32)
INDICES = np.array([0, 1, 2], dtype=np.uint32)
TRIANGLES, TRIANGLE_STRIP, TRIANGLE_FAN = 4, 5, 6


def write_glb(path, modes):
    blob = b""
    buffer_views = []
    for data in (TRI_A.tobytes(), TRI_B.tobytes(), INDICES.tobytes()):
        blob += data
    gltf = pygltflib.GLTF2(
        scene=0,
        scenes=[pygltflib.Scene(nodes=[0, 1])],
        nodes=[pygltflib.Node(mesh=0), pygltflib.Node(mesh=1)],
        meshes=[
            pygltflib.Mesh(
                name=f"mesh_{i}",
                primitives=[pygltflib.Primitive(attributes=pygltflib.Attributes(POSITION=i), indices=2, mode=mode)],
            )
            for i, mode in enumerate(modes)
        ],
        accessors=[
            pygltflib.Accessor(
                bufferView=i,
                count=3,
                type="VEC3",
                min=tri.min(axis=0).tolist(),
                max=tri.max(axis=0).tolist(),
            )
            for i, tri in enumerate((TRI_A, TRI_B))
        ]
        + [pygltflib.Accessor(bufferView=2, componentType=pygltflib.UNSIGNED_INT, count=3, type="SCALAR")],
        bufferViews=buffer_views,
    )
    gltf.set_binary_blob(blob)
    gltf.save_binary(str(path))


    for label, modes in (
        ("both nodes TRIANGLES", (TRIANGLES, TRIANGLES)),
        ("second node TRIANGLE_STRIP", (TRIANGLES, TRIANGLE_STRIP)),
        ("first node TRIANGLE_FAN", (TRIANGLE_FAN, TRIANGLES)),
    ):
        path = Path(directory) / "two_nodes.glb"
        write_glb(path, modes)
        meshes = parse_mesh_glb(
        )
        print(f"{label}: 2 nodes -> {len(meshes)} mesh(es)")
        for mesh in meshes:
            print(
                f"   name={mesh.metadata['name']!r} node_index={mesh.metadata['node_index']} "
                f"n_verts={len(mesh.trimesh.vertices)} n_faces={len(mesh.trimesh.faces)}"
            )
        print()
```

### Expected Behavior

Each node keeps its own mesh, its own name and its own node index, whatever primitive mode it uses:

```text
both nodes TRIANGLES: 2 nodes -> 2 mesh(es)
   name='mesh_0' node_index=0 n_verts=3 n_faces=1
   name='mesh_1' node_index=1 n_verts=3 n_faces=1

second node TRIANGLE_STRIP: 2 nodes -> 2 mesh(es)
   name='mesh_0' node_index=0 n_verts=3 n_faces=1
   name='mesh_1' node_index=1 n_verts=3 n_faces=1

first node TRIANGLE_FAN: 2 nodes -> 2 mesh(es)
   name='mesh_0' node_index=0 n_verts=3 n_faces=1
   name='mesh_1' node_index=1 n_verts=3 n_faces=1
```

### Relevant log output

```shell
both nodes TRIANGLES: 2 nodes -> 2 mesh(es)
   name='mesh_0' node_index=0 n_verts=3 n_faces=1
   name='mesh_1' node_index=1 n_verts=3 n_faces=1

second node TRIANGLE_STRIP: 2 nodes -> 1 mesh(es)
   name='mesh_0' node_index=0 n_verts=6 n_faces=2

first node TRIANGLE_FAN: 2 nodes -> 1 mesh(es)
   name='mesh_0' node_index=1 n_verts=6 n_faces=2
```

Both nodes' vertices end up in one mesh, and `mesh_1` is gone.

### Environment

- OS: Windows 11 24H2
- GPU-driver version: N/A

### Additional Context

Happy to open a PR giving the triangle loops their own counter and adding a regression case to
`tests/parsers/test_mesh.py`.

kevin9327:
## Description

Loading a `.glb` whose primitives use `TRIANGLE_STRIP` or `TRIANGLE_FAN` lost the node each primitive came from. Two
separate nodes came back as a single mesh holding the geometry of both, under the first one's name and material, and
downstream that one mesh is one collision geom, so two bodies that should collide separately became one.

The loop over nodes and the loops expanding a strip or a fan into triangles shared the name `i`:

```python
for i, (mesh_index, mesh_transform) in enumerate(mesh_list):
    ...
            for i in range(len(indices) - 2):
```

so after either expansion `i` held the last triangle counter. Two consumers read it afterwards: `group_idx`, which
decides what shares a mesh, and `metadata["node_index"]`, which decides what is merged into one collision geom (the
`groups_by_node` block of `RigidEntityDescription`). The outer loop now names its index `i_node`, which is what both
of them mean.

The geometry itself was decoded correctly, which is why nothing warned.

`test_glb_primitive_modes` reads four nodes off the same five positions: a `TRIANGLES` triangle, that triangle as a
one-triangle strip, a three-triangle strip and a three-triangle fan. It pins the mesh each node gets, its vertices,
and the triangles each mode expands to, including the alternating winding of a strip.

## Related Issue

Resolves Genesis-Embodied-AI/Genesis#3335

## Motivation and Context

Strips and fans are what a `.glb` gets from a pipeline that optimises for vertex reuse, and neither mode had a test.
The file loads, the geometry is right, and the only sign is that the scene ends up with fewer meshes and fewer
collision geoms than the file describes.

```
pytest -n 0 --backend cpu tests/parsers/test_mesh.py -k glb
```

Windows 11, CPU backend, Python 3.12.

- On `b3c6c73` with only the test added, `test_glb_primitive_modes` fails on the four nodes coming back as three
  meshes, the strip node having been folded into the plain-triangle one:

  ```
  >       assert {gs_mesh.metadata["name"] for gs_mesh in gs_meshes} == set(primitives)
  E       AssertionError: assert {'fan_three',..., 'triangles'} == {'fan_three',..., 'triangles'}
  E         Extra items in the right set:
  E         'strip_one'
  ```

- With this commit the selection above is 17 passed, 0 failed.
- Pinned as unaffected: the 16 cases this selection collects on `b3c6c73` pass there, as part of a wider
  `-k "glb or scale or yup or mjcf"` run of 24 that also passes.
- `ruff format --check` and `ruff check` report nothing on the two edited files.

## Checklist:

### DIFF
--- a/genesis/utils/gltf.py
+++ b/genesis/utils/gltf.py
@@ -320,7 +320,7 @@ def parse_mesh_glb(path, group_by_material, scale, is_mesh_zup, surface):
     materials = {}
     is_visual_overwritten = surface.texture is not None
 
-    for i, (mesh_index, mesh_transform) in enumerate(mesh_list):
+    for i_node, (mesh_index, mesh_transform) in enumerate(mesh_list):
         mesh_glb = glb.meshes[mesh_index]
         mesh_name = mesh_glb.name
 
@@ -402,14 +402,14 @@ def parse_mesh_glb(path, group_by_material, scale, is_mesh_zup, surface):
             # A single glTF mesh may hold several primitives with distinct materials. When not grouping by material,
             # primitives must still be separated by material so each keeps its own surface and texture; otherwise
             # primitives sharing a mesh would be merged under the first primitive's material and the rest lost.
-            group_idx = primitive.material if group_by_material else (i, primitive.material)
+            group_idx = primitive.material if group_by_material else (i_node, primitive.material)
             mesh_info, first_created = mesh_infos.get(group_idx)
             if first_created:
                 metadata = {"mesh_path": path, "name": material_name if group_by_material else mesh_name}
                 if not group_by_material:
                     # Record the source mesh node so per-material submeshes can be regrouped into one physical body
                     # (e.g. merged into a single collision geom) while still rendering with their own textures.
-                    metadata["node_index"] = i
+                    metadata["node_index"] = i_node
                 mesh_info.set_property(surface=material, metadata=metadata)
             mesh_info.append(points, triangles, normals, uvs)
     meshes = mesh_infos.export_meshes(scale=scale, is_mesh_zup=is_mesh_zup)
--- a/tests/parsers/test_mesh.py
+++ b/tests/parsers/test_mesh.py
@@ -351,6 +351,88 @@ def test_glb_draco_missing_normals_texcoord(glb_file):
 
 
 @pytest.mark.required
+def test_glb_primitive_modes(tmp_path):
+    # Four nodes read the same five positions: a lone triangle, that triangle as a one-triangle strip, a
+    # three-triangle strip whose second triangle is wound the other way round, and a three-triangle fan pivoting on
+    # the first index. Each node is a mesh of its own whatever mode it declares.
+    positions = np.array(
+        [[0.0, 0.0, 0.0], [1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [1.0, 1.0, 0.0], [2.0, 0.0, 0.0]], dtype=np.float32
+    )
+    short_indices = np.array([0, 1, 2], dtype=np.uint32)
+    long_indices = np.array([0, 1, 2, 3, 4], dtype=np.uint32)
+    TRIANGLES, TRIANGLE_STRIP, TRIANGLE_FAN = 4, 5, 6
+    primitives = {
+        "triangles": (TRIANGLES, 1),
+        "strip_one": (TRIANGLE_STRIP, 1),
+        "strip_three": (TRIANGLE_STRIP, 2),
+        "fan_three": (TRIANGLE_FAN, 2),
+    }
+    expected_faces = {
+        "triangles": [[0, 1, 2]],
+        "strip_one": [[0, 1, 2]],
+        "strip_three": [[0, 1, 2], [1, 3, 2], [2, 3, 4]],
+        "fan_three": [[0, 1, 2], [0, 2, 3], [0, 3, 4]],
+    }
+
+    blob = b""
+    buffer_views = []
+    for data in (positions.tobytes(), short_indices.tobytes(), long_indices.tobytes()):
+        blob += b"\x00" * ((4 - len(blob) % 4) % 4)
+        buffer_views.append(pygltflib.BufferView(buffer=0, byteOffset=len(blob), byteLength=len(data)))
+        blob += data
+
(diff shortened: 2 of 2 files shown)


duburcqa:
Note following the guidelines CLAUDE.md and CODING_GUIDELINES.md. Closing. I will reopen if addressed.

kevin9327:
Reworked against `CODING_GUIDELINES.md` and reopened as #3342. The regression now asserts the observable (two nodes stay two meshes) instead of enumerating primitive modes, and its GLB lives in a `tests/parsers/conftest.py` fixture.

PR #3339, opened 2026-09-10
[BUG FIX] Honor the alpha cutoff of a masked glTF material.

### LINKED ISSUE #3338: [Bug]: glTF alphaCutoff is never read, so a MASK material imports as BLEND
### Bug Description

A glTF material with `alphaMode: "MASK"` asks for a hard cutout: every texel is either fully opaque or fully
transparent, split at `alphaCutoff` (0.5 by default). It is what foliage, fences, grates and decals are authored with.
Genesis imports such a material as if it were `BLEND`, keeping the authored alpha ramp, so the soft edge of the ramp
survives instead of being cut away and the cutout renders with a translucent fringe.

`genesis/utils/gltf.py::parse_glb_material` never reads `material.alphaCutoff`:

```python
alpha_cutoff = None
...
alpha_cutoff = mu.adjust_alpha_cutoff(alpha_cutoff, alpha_modes[material.alphaMode])
```

`adjust_alpha_cutoff` returns its argument for `MASK`, so `alpha_cutoff` stays `None` all the way to
`opacity_texture.apply_cutoff(alpha_cutoff)`, which returns immediately on `None`. The `MASK` branch of
`adjust_alpha_cutoff` therefore never does anything, and `MASK` and `BLEND` import identically. `OPAQUE` is handled
(it resolves to a cutoff of 0.0, which forces every texel opaque).

The USD parser reads the same quantity from the shader and applies it, in `genesis/utils/usd/usd_material.py`:

```python
alpha_cutoff = get_input_attribute_value(shader, "opacityThreshold", "value")[0]
opacity_texture.apply_cutoff(alpha_cutoff)
```

so the two importers answer the same question differently.

### Steps to Reproduce

Self-contained, no external assets. It writes a textured triangle whose base color texture is one row of pixels with
a rising alpha ramp, and imports it under each alpha mode.

```python
import io
import tempfile
from pathlib import Path

import numpy as np
import pygltflib
from PIL import Image

import genesis as gs

# One row of pixels whose alpha ramps up, the shape a foliage or fence cutout has along its edge.
ALPHA = np.array([[0, 64, 128, 192, 255]], dtype=np.uint8)
POSITIONS = np.array([[0.0, 0.0, 0.0], [1.0, 0.0, 0.0], [0.0, 1.0, 0.0]], dtype=np.float32)
UVS = np.array([[0.0, 0.0], [1.0, 0.0], [0.0, 1.0]], dtype=np.float32)


def write_glb(path, alpha_mode, alpha_cutoff):
    rgba = np.full((1, ALPHA.shape[1], 4), 255, dtype=np.uint8)
    rgba[..., 3] = ALPHA
    buffer = io.BytesIO()
    Image.fromarray(rgba, mode="RGBA").save(buffer, format="PNG")

    blob = b""
    buffer_views = []
    for data in (buffer.getvalue(), POSITIONS.tobytes(), UVS.tobytes()):
        blob += data

    gltf = pygltflib.GLTF2(
        scene=0,
        nodes=[pygltflib.Node(mesh=0)],
        meshes=[
            pygltflib.Mesh(
                name="cutout",
                primitives=[
                    pygltflib.Primitive(attributes=pygltflib.Attributes(POSITION=0, TEXCOORD_0=1), material=0)
                ],
            )
        ],
        materials=[
            pygltflib.Material(
                pbrMetallicRoughness=pygltflib.PbrMetallicRoughness(
                    baseColorTexture=pygltflib.TextureInfo(index=0, texCoord=0)
                ),
                alphaMode=alpha_mode,
                alphaCutoff=alpha_cutoff,
            )
        ],
        textures=[pygltflib.Texture(source=0)],
        images=[pygltflib.Image(bufferView=0, mimeType="image/png")],
        accessors=[
            pygltflib.Accessor(
                bufferView=1,
                count=3,
                type="VEC3",
            ),
            pygltflib.Accessor(bufferView=2, componentType=pygltflib.FLOAT, count=3, type="VEC2"),
        ],
        bufferViews=buffer_views,
    )
    gltf.set_binary_blob(blob)
    gltf.save_binary(str(path))


print("authored alpha:", ALPHA.ravel().tolist())
    for alpha_mode, alpha_cutoff in (("MASK", 0.6), ("MASK", 0.3), ("BLEND", 0.5), ("OPAQUE", 0.5)):
        path = Path(directory) / "cutout.glb"
        write_glb(path, alpha_mode, alpha_cutoff)
        (mesh,) = parse_mesh_glb(
        )
        opacity = mesh.surface.opacity_texture
        print(
            f"alphaMode={alpha_mode:6s} alphaCutoff={alpha_cutoff} -> "
            f"{np.asarray(opacity.image_array).ravel().tolist()}"
        )
```

### Expected Behavior

A `MASK` material splits the authored alpha at its cutoff, `0.6 * 255 = 153` and `0.3 * 255 = 76.5`:

```text
authored alpha: [0, 64, 128, 192, 255]
alphaMode=MASK   alphaCutoff=0.6 -> [0, 0, 0, 255, 255]
alphaMode=MASK   alphaCutoff=0.3 -> [0, 0, 255, 255, 255]
alphaMode=BLEND  alphaCutoff=0.5 -> [0, 64, 128, 192, 255]
alphaMode=OPAQUE alphaCutoff=0.5 -> [255, 255, 255, 255, 255]
```

### Relevant log output

```shell
authored alpha: [0, 64, 128, 192, 255]
alphaMode=MASK   alphaCutoff=0.6 -> [0, 64, 128, 192, 255]
alphaMode=MASK   alphaCutoff=0.3 -> [0, 64, 128, 192, 255]
alphaMode=BLEND  alphaCutoff=0.5 -> [0, 64, 128, 192, 255]
alphaMode=OPAQUE alphaCutoff=0.5 -> [255, 255, 255, 255, 255]
```

The two `MASK` rows are the authored ramp, and both match the `BLEND` row.

### Environment

- OS: Windows 11 24H2
- GPU-driver version: N/A

### Additional Context

Happy to open a PR reading `material.alphaCutoff` into the existing `adjust_alpha_cutoff` call and adding a
regression case to `tests/parsers/test_mesh.py`.

kevin9327:
## Description

A glTF material with `alphaMode: "MASK"` asks for a hard cutout, every texel fully opaque or fully transparent at
`alphaCutoff`, which is what foliage, fences, grates and decals are authored with. Genesis imported it as if it were
`BLEND`: the authored alpha ramp survived, so the soft edge that the cutout exists to remove rendered as a
translucent fringe.

`parse_glb_material` initialised `alpha_cutoff` to `None` and passed that same `None` to `adjust_alpha_cutoff`, which
returns its argument for `MASK`, so `apply_cutoff` got `None` and returned immediately. The material's own
`alphaCutoff` (0.5 by default) is what belongs there, so it is passed instead and the dead initialisation goes.
`OPAQUE` and `BLEND` resolve to 0.0 and `None` inside `adjust_alpha_cutoff` regardless of the value handed to it, so
they keep the behaviour they had.

The USD importer already reads the same quantity off the shader and applies it the same way, through
`opacity_texture.apply_cutoff` with `opacityThreshold`.

`test_glb_alpha_mode` puts one alpha ramp behind three materials, one per alpha mode, and pins the opacity each mode
imports.

#3336 and #3337 touch the same two files in other places and merge cleanly against this one in either order.

## Related Issue

Resolves Genesis-Embodied-AI/Genesis#3338

## Motivation and Context

Nothing warns and nothing fails, so the only sign is that a masked material shades exactly like a blended one. The
`MASK` branch of `adjust_alpha_cutoff` could never fire, which is why the two modes were indistinguishable.

```
pytest -n 0 --backend cpu tests/parsers/test_mesh.py -k "glb or mjcf or texture"
```

Windows 11, CPU backend, Python 3.12.

- On `b3c6c73` with only the test added, the masked material comes back with its authored ramp intact:

  ```
  FAILED tests/parsers/test_mesh.py::test_glb_alpha_mode
  E       Opacity match failed for material masked.
  E       Mismatched elements: 3 / 5 (60%)
  E        ACTUAL: array([  0,  64, 128, 192, 255], dtype=uint8)
  E        DESIRED: array([  0,   0,   0, 255, 255])
  ```

- With this commit the selection above is 23 passed, 0 failed, plus one setup error on
  `test_plane_texture_path_preservation`: `RuntimeError: Virtual memory allocation (1073741824 B) failed` raised by
  the quadrants host memory pool inside `gs.init`, before the test body. It passes on its own rerun, with and
  without this change, and that test loads no glTF.
- Pinned as unaffected: `adjust_alpha_cutoff` resolves `OPAQUE` to 0.0 and `BLEND` to `None` whatever it is handed,
  so only a `MASK` material can shade differently, and no glTF asset under `genesis/assets` or in the test dataset
  declares one (28 files scanned for `alphaMode`). The glb and mjcf cases of this selection pass on `b3c6c73`, as
  part of a wider `-k "glb or scale or yup or mjcf"` run of 24 that also passes.
- `ruff format --check` and `ruff check` report nothing on the two edited files.

## Checklist:

### DIFF
--- a/genesis/utils/gltf.py
+++ b/genesis/utils/gltf.py
@@ -104,7 +104,6 @@ def parse_glb_material(glb, material_index, surface):
     normal_texture = None
     emissive_texture = None
 
-    alpha_cutoff = None
     double_sided = None
     ior = None
     uvs_used = 0
@@ -131,7 +130,7 @@ def parse_glb_material(glb, material_index, surface):
             occlusion_texture = mu.create_texture(occlusion_image, None, "linear")
 
     # parse alpha mode
-    alpha_cutoff = mu.adjust_alpha_cutoff(alpha_cutoff, alpha_modes[material.alphaMode])
+    alpha_cutoff = mu.adjust_alpha_cutoff(material.alphaCutoff, alpha_modes[material.alphaMode])
 
     # parse pbr roughness and metallic
     if material.pbrMetallicRoughness is not None:
--- a/tests/parsers/test_mesh.py
+++ b/tests/parsers/test_mesh.py
@@ -447,6 +447,96 @@ def test_glb_parse_material(glb_file):
 
 
 @pytest.mark.required
+def test_glb_alpha_mode(tmp_path):
+    # Three materials share one base color texture whose alpha ramps up, the shape a cutout has along its edge. A
+    # masked material splits that ramp at its own cutoff, a blended one keeps it, and an opaque one discards it.
+    alpha = np.array([0, 64, 128, 192, 255], dtype=np.uint8)
+    rgba = np.full((1, len(alpha), 4), 255, dtype=np.uint8)
+    rgba[..., 3] = alpha
+    buffer = io.BytesIO()
+    Image.fromarray(rgba, mode="RGBA").save(buffer, format="PNG")
+    positions = np.array([[0.0, 0.0, 0.0], [1.0, 0.0, 0.0], [0.0, 1.0, 0.0]], dtype=np.float32)
+    uvs = np.array([[0.0, 0.0], [1.0, 0.0], [0.0, 1.0]], dtype=np.float32)
+    alpha_modes = {"masked": ("MASK", 0.6), "blended": ("BLEND", 0.5), "opaque": ("OPAQUE", 0.5)}
+    expected_opacities = {
+        # 0.6 of full opacity is 153, which falls between the third and the fourth texel
+        "masked": [0, 0, 0, 255, 255],
+        "blended": alpha,
+        "opaque": [255, 255, 255, 255, 255],
+    }
+
+    blob = b""
+    buffer_views = []
+    for data in (buffer.getvalue(), positions.tobytes(), uvs.tobytes()):
+        blob += b"\x00" * ((4 - len(blob) % 4) % 4)
+        buffer_views.append(pygltflib.BufferView(buffer=0, byteOffset=len(blob), byteLength=len(data)))
+        blob += data
+
+    gltf = pygltflib.GLTF2(
+        scene=0,
+        scenes=[pygltflib.Scene(nodes=[0])],
+        nodes=[pygltflib.Node(mesh=0)],
+        meshes=[
+            pygltflib.Mesh(
+                primitives=[
+                    pygltflib.Primitive(
+                        attributes=pygltflib.Attributes(POSITION=0, TEXCOORD_0=1), material=material_idx
+                    )
+                    for material_idx in range(len(alpha_modes))
+                ]
+            )
+        ],
+        materials=[
+            pygltflib.Material(
+                name=name,
+                pbrMetallicRoughness=pygltflib.PbrMetallicRoughness(
+                    baseColorTexture=pygltflib.TextureInfo(index=0, texCoord=0)
+                ),
+                alphaMode=alpha_mode,
+                alphaCutoff=alpha_cutoff,
+            )
+            for name, (alpha_mode, alpha_cutoff) in alpha_modes.items()
+        ],
+        textures=[pygltflib.Texture(source=0)],
+        images=[pygltflib.Image(bufferView=0, mimeType="image/png")],
+        accessors=[
+            pygltflib.Accessor(
+                bufferView=1,
+                componentType=pygltflib.FLOAT,
+                count=len(positions),
+                type="VEC3",
+                min=positions.min(axis=0).tolist(),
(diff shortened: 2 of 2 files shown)


duburcqa:
Note following the guidelines CLAUDE.md and CODING_GUIDELINES.md. Closing. I will reopen if addressed.

kevin9327:
Reworked against `CODING_GUIDELINES.md` and reopened as #3341: compliant branch name, single-line commit with no trailer, the GLB moved into a `tests/parsers/conftest.py` fixture, and a terse description.

PR #3342, opened 2026-09-10
[BUG FIX] Keep separate glTF nodes from merging into one collision geom.

### LINKED ISSUE #3335: [Bug]: glTF triangle strips and fans lose the node their primitives came from, merging unrelated nodes

kevin9327:
## Description

The loops expanding a `TRIANGLE_STRIP` and a `TRIANGLE_FAN` in `parse_mesh_glb` reuse `i`, the index of the loop over nodes. By the time `group_idx` and `metadata["node_index"]` read it, it holds a triangle counter, so two nodes whose primitives share a material land in the same group and merge into one mesh, and with it one collision geom.

Renaming the outer index to `i_node` leaves the inner counters alone.

Supersedes #3337, reworked against `CODING_GUIDELINES.md`.

## Related Issue

Resolves Genesis-Embodied-AI/Genesis#3335

## Motivation and Context

Strips and fans are what an exporter emits for terrain, cloth panels and UI-like quads, and the merge is silent: the entity comes back with fewer geoms than it has nodes, so parts that should collide separately move as one body.

```
```

Windows 11, CPU backend, Python 3.12: 22 passed.

On `b3c6c73` with only `test_glb_strip_nodes_stay_separate` and its fixture added, the two nodes come back as one mesh:

```
E  AssertionError: assert {'near_strip'} == {'far_strip', 'near_strip'}
```

## Checklist:

### DIFF
--- a/tests/parsers/conftest.py
+++ b/tests/parsers/conftest.py
@@ -641,6 +641,60 @@ def emissive_material_variants_glb(asset_tmp_path):
     return str(path)
 
 
+@pytest.fixture(scope="session")
+def triangle_strip_nodes_glb(asset_tmp_path):
+    """Path to a GLB with two nodes a unit apart, each holding one TRIANGLE_STRIP primitive of the same material.
+
+    Sharing the material leaves the node each primitive comes from as the only thing separating the two meshes."""
+    positions = np.array(
+        [[0.0, 0.0, 0.0], [1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [1.0, 1.0, 0.0]],
+        dtype=np.float32,
+    )
+    indices = np.arange(len(positions), dtype=np.uint32)
+
+    blob = b""
+    buffer_views = []
+    for data in (positions.tobytes(), indices.tobytes()):
+        buffer_views.append(pygltflib.BufferView(buffer=0, byteOffset=len(blob), byteLength=len(data)))
+        blob += data
+
+    gltf = pygltflib.GLTF2(
+        scene=0,
+        scenes=[pygltflib.Scene(nodes=[0, 1])],
+        nodes=[
+            pygltflib.Node(mesh=0, name="near_strip"),
+            pygltflib.Node(mesh=1, name="far_strip", translation=[10.0, 0.0, 0.0]),
+        ],
+        meshes=[
+            pygltflib.Mesh(
+                name=name,
+                primitives=[
+                    pygltflib.Primitive(attributes=pygltflib.Attributes(POSITION=0), indices=1, mode=5, material=0)
+                ],
+            )
+            for name in ("near_strip", "far_strip")
+        ],
+        materials=[pygltflib.Material(name="shared")],
+        accessors=[
+            pygltflib.Accessor(
+                bufferView=0,
+                componentType=pygltflib.FLOAT,
+                count=len(positions),
+                type="VEC3",
+                min=positions.min(axis=0).tolist(),
+                max=positions.max(axis=0).tolist(),
+            ),
+            pygltflib.Accessor(bufferView=1, componentType=pygltflib.UNSIGNED_INT, count=len(indices), type="SCALAR"),
+        ],
+        bufferViews=buffer_views,
+        buffers=[pygltflib.Buffer(byteLength=len(blob))],
+    )
+    gltf.set_binary_blob(blob)
+    path = str(asset_tmp_path / "triangle_strip_nodes.glb")
+    gltf.save_binary(path)
+    return path
+
+
 @pytest.fixture
 def material_mjcf(tmp_path):
     """Generate an MJCF model with materials and geom-level colors."""
--- a/tests/parsers/test_mesh.py
+++ b/tests/parsers/test_mesh.py
@@ -327,6 +327,20 @@ def test_glb_parse_geometry(request, glb_file, tol):
 
 
 @pytest.mark.required
+def test_glb_strip_nodes_stay_separate(triangle_strip_nodes_glb):
+    # Two nodes are two meshes, and one collision geom each, whatever primitive mode they declare
+    gs_meshes = gltf_utils.parse_mesh_glb(
+        triangle_strip_nodes_glb,
+        group_by_material=False,
+        scale=None,
+        is_mesh_zup=True,
+        surface=gs.surfaces.Default(),
+    )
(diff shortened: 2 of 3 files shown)


tests/parsers/conftest.py
```diff
@@ -641,6 +641,60 @@ def emissive_material_variants_glb(asset_tmp_path):
     return str(path)
 
 
+@pytest.fixture(scope="session")
+def triangle_strip_nodes_glb(asset_tmp_path):
```
duburcqa, line 645: No existing fixture that could be extended?

tests/parsers/test_mesh.py
```diff
@@ -326,6 +326,20 @@ def test_glb_parse_geometry(request, glb_file, tol):
         check_gs_tm_meshes(gs_mesh, tm_mesh, mesh_name, tol, tol)
 
 
+@pytest.mark.required
+def test_glb_strip_nodes_stay_separate(triangle_strip_nodes_glb):
```
duburcqa, line 330: No existing unit test can be extended instead?

PR #3348, opened 2026-09-11
[MISC] Expose entity-level public API for setting dofs limit.

### LINKED ISSUE #3168: [Feature]: RigidEntity.set_dofs_limit

jeetrex17:
## Description

Adds `RigidEntity.set_dofs_limit`, which maps entity-local dof indices to solver indices and forwards to `RigidSolver.set_dofs_limit`, like every other dof property setter. Same change as #3170, rebased on `main`; the `test_data_accessor` row now also passes the setter, and `test_path_planning_avoidance` sets its runtime limit through the entity.

## Related Issue

Resolves #3168

Supersedes #3170 by @Akbro23.

## Motivation and Context

Joint limits were the one dof property with an entity getter and no entity setter, so setting them meant scene-global
indices and a direct solver call.

macOS, CPU backend:

```bash
uv run pytest -v --forked tests/rigid/test_api.py::test_data_accessor            # 4 passed
uv run pytest -v --forked tests/rigid/test_kinematics.py::test_path_planning_avoidance  # 3 passed, 3 macOS GPU skips
```

Pre-commit (ruff check + format) passes. The `-m required` suite has not been run locally.

## Checklist:


tests/rigid/test_kinematics.py
```diff
@@ -749,7 +749,7 @@ def test_path_planning_avoidance(backend, n_envs, batch_dofs_info, show_viewer,
 
     # A limit tightened once the scene is built is the one the planner samples within: a goal it excludes cannot be
     # reached, while one it allows still can, and no waypoint leaves what the solver would let the robot hold.
-    scene.rigid_solver.set_dofs_limit(-DOF_LIMIT, DOF_LIMIT, dofs_idx=range(franka.dof_start, franka.dof_start + 7))
+    franka.set_dofs_limit(-DOF_LIMIT, DOF_LIMIT, dofs_idx_local=range(7))
```
duburcqa: prefer slice

duburcqa (approved):

PR #3355, opened 2026-09-12
[MISC] Cut the compile time of the collision detection kernels.

duburcqa:
## Description

- The multicontact narrowphase routine drives its detections (the GJK re-detection of contact 0, the four perturbed contacts and their per-contact GJK refinement) from one runtime loop, so the detection pipeline (`_func_multicontact_run_detection`, whose `use_gjk` and `is_initial_detection` flags become runtime booleans) is inlined once in the kernel instead of three times.
- MPR's `compute_support` loops over the two geoms at runtime with one call of the support switch, as the GJK support wrappers already do, halving the copies of `support_driver` every MPR site inlines.

## Motivation and Context

Every `qd.func` a kernel calls is inlined per call site, so the two narrowphase kernels carried 23 copies of the geom-type support switch and two copies of GJK/EPA. Frontend compile time of go2 (offline cache off, Apple M4, seconds):

| kernel | main | PR |
|---|---|---|
| multicontact narrowphase | 10.1 | 5.6 |
| contact0 narrowphase | 3.8 | 2.4 |
| whole build | 25.3 | 19.2 |

CUDA build times (go2, table_bussing, dex_hand, main vs PR interleaved on an RTX 5090) and step times are running and will be added here.

`tests/rigid/test_collision.py`, `test_collision_nonconvex.py` and `test_mujoco_parity.py` on CPU. The same files on CUDA with `--dev` are running.

## Checklist:

### DIFF
--- a/genesis/engine/solvers/rigid/collider/mpr.py
+++ b/genesis/engine/solvers/rigid/collider/mpr.py
@@ -214,30 +214,30 @@ def compute_support(
     rigid_config: qd.template(),
     collider_static_config: qd.template(),
 ):
-    v1 = support_driver(
-        i_ga,
-        i_b,
-        direction,
-        pos_a,
-        quat_a,
-        collider_state,
-        dyn_info,
-        collider_info,
-        rigid_config,
-        collider_static_config,
-    )
-    v2 = support_driver(
-        i_gb,
-        i_b,
-        -direction,
-        pos_b,
-        quat_b,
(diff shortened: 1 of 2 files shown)


duburcqa:
The whole multicontact kernel is 22 s of frontend and 17 s of first-launch JIT on go2, so removing half of its inlined instances cannot reach the 30 s build-time target.

PR #3373, opened 2026-09-15
[BUG FIX] Improve converge of Signorini contact solve.

duburcqa:
## Description

- Along its search direction, the Newton line search of the constraint solver now takes as a candidate the next step at which a friction block of the `signorini` resolution switches between its sticking and saturated branch, or a normal row switches on or off, instead of the bracket's midpoint (`func_row_alpha_kinks`, closed form per block, min-reduced with the terms on every search path). Each round then crosses one piece of the piecewise cost and the next round's Newton candidates sit on a single piece.
- Its gradient tolerance is floored at the rounding noise of the terms forming the derivative (a fifth initialization sum), so a search that has converged returns instead of spending its whole budget.
- Under the `signorini` resolution, the normal force each elliptic contact bounds its friction against is kept as a per-contact latch (`cone_latch`) advanced once per Newton iteration toward the normal force of the iterate: whole at the seed (a contact is born unlatched when its rows are assembled), half of the way on the first relatch, then by the secant of the gap over the last two relatches (`cone_prev_latch`, `cone_prev_gap`), clamped between a quarter and eight times the gap. The zone test, the force and Hessian of the contact, the line-search cost and its kinks all read the latch, so every pass minimizes one frozen objective within an iteration.
- The exit test of the Newton iteration charges the residual of the contact law the frozen objective leaves out: per friction block, the force the disc of the latch and the disc of the normal force of the iterate clip differently, mapped to a generalized force through the block's Jacobian rows (`func_head_latch_defect`). An island exits when its gradient plus this defect sits below the tolerance, allowed for the rounding of the constraint forces the gradient balances (`func_signorini_flat`), or when a line search lowers the frozen cost by nothing at all while the Newton model promises less than the tolerance, the arithmetic floor of that cost. A line search that cannot move an island keeps it iterating while its defect is pending.

## Motivation and Context

On MuJoCo Warp's `aloha_clutter` scene in single precision, a 10 g stick resting on the table drifts, gets kicked and ends in `Invalid constraint forces causing 'nan'`: the Newton solve exits with a percent-level force residual because the line search runs out of its 50 evaluations at the stick/slip transitions of the rolling rows, where the bracket only shrinks through its midpoint, and near convergence, where its tolerance falls below the rounding noise of the derivative. Over 144 steps of the stick's island, the search evaluations drop from 8725 to 2528 in fp32 (1561 to 1026 in fp64) and the budget is never exhausted (31 and 10 times before).

Relatching the friction radius at every iteration is a fixed-point iteration on top of Newton. Its map alternates where a contact activates and creeps where a leg is near jamming, so a plain relatch cycles on a few environments of a batch and a constant half weight takes tens of iterations on others, with the whole batch iterating in lockstep. The secant on the gap sizes each contact's step to the slope it observes and resolves both regimes in a few relatches. Exiting on the gradient of the frozen objective alone is unsound while the latch moves, and a stall or a small Newton decrement fires far from stationarity while the relatch moves the kinks of the cost, so the exit charges the exact residual of the contact law instead, and gives up only where a search gains nothing at all: near jamming, the normal force follows its friction bound one for one and the radius has no root to converge to.

`aloha_clutter` full replay, 2048 environments, hibernation on, RTX 5090, fp32, ms per step:

| | monolith arm | decomposed arm |
|---|---|---|
| `main` | 22.1 | 21.2 |
| this branch | 17.7 | 16.4 |
| `convex` | 10.9 | 7.1 |

In the grasp phase the batch averages 11 to 22 iterations (`main` 12 to 54, `convex` 3 to 9); a few environments per step still run the budget out there, on gripper islands whose sliding contacts sit near jamming.

go2 on flat ground, elliptic cone, random joint configuration per environment, RTX 5090, fp32, 4096 environments: monolith arm 22 ms per step at the budget on `main`, 4.2 on this branch (`convex` 3.6); decomposed arm 27 to 5.3 (`convex` 3.5). Converged environments leave at most 0.15 N of friction over their bound there.

`test_dofs_force_balance` gives a light stick resting on the ground a slow spin, which puts its rolling friction rows at their bound, and asserts that the reported dof force equals the mass matrix times the acceleration the step integrated. On `main` the residual reaches 75 percent of the stick's weight; with this change it stays below 1e-9 of it in double precision and 2e-6 in single. The `aloha_clutter` replay with the elliptic cone, torsional and rolling friction and the Euler integrator completes in fp32 with the stick at rest.

The `signorini` isotropy sweep of `tests/rigid/test_friction.py` passes (rotated copies of one scene converge to the same forces within its bounds), and the MuJoCo parity and dynamics suites are unchanged. Full `tests/rigid` + `tests/grad` suite on the cluster (CUDA, `--dev --forked`) on the previous revision of the exit test: 559 passed, 2 skipped, 6 xfailed; the arithmetic-floor exit was validated on the friction, dynamics and isotropy tests on CUDA, the full suite is running.

## Checklist:


duburcqa:
Signorini is fundamentally broken currently. This is just a quick-and-dirty mitigation that is not addressing the fundamental underlying issue. Closing for now.

PR #3376, opened 2026-09-15
[MISC] Make the kinematic tree, rather than the entity, the unit of every rigid solver pass.

duburcqa:
## Description

- The kinematic tree replaces the entity as the unit of every rigid solver pass. The kinematics passes (link poses, geoms, centers of mass, Cartesian velocities and accelerations) and the mass matrix passes (assembly, factor, block solve, implicit damping, manual backward) iterate the trees, one thread each, gated on the sleep flag of the tree root; the centers of mass, the setters, the initial placement, the backward replay and the kinematic solver walk one thread per kinematic root and environment, static links first. The per-entity mask `mass_mat_mask` and the per-entity block tables are gone: under implicit damping a tree re-solves its acceleration when one of its dofs carries damping or an implicit actuator bias.
- The path planner and the forward-kinematics query pose the whole root of the entity they write, so an entity attached to it follows.
- The environment display offsets leave the rigid solver: the viewer's ray pick shifts its ray into each environment and the hit back out from `Scene.envs_offset`.
- `links_tree_idx` is defined for a scene without any kinematic tree (static links keep -1), where it was left at 0.
- The mass matrix passes walk their blocks and lanes with `for` loops, so every kernel of the rigid step stays reversible by autodiff; `CLAUDE.md` and `CODING_GUIDELINES.md` now state that rule, hand-written backward kernels being reserved to iterative algorithms.

## Motivation and Context

An entity is neither a subset nor a superset of a kinematic tree: one MJCF entity may hold two arms and a dozen free objects, each an independent tree, and an attached entity is part of its parent's tree. The kinematic tree is the unit that moves and sleeps as a whole, so it is the natural unit of every rigid solver pass. Iterating entities made every pass carry corner cases for both directions of the mismatch: a multi-body entity ran its whole sweep on one thread per environment and slept as a block, and an attached entity had its dofs left out of the damped factor of its parent and its links left behind by the planner and the forward-kinematics query. Decomposing every pass at tree level (or per kinematic root where the centers of mass and the setters need the static parents) removes those corner cases and gives each tree its own thread.

`test_set_root_pose` covers the setters of a fixed-base entity (root pose, joint coordinates and dof positions), `test_merge_matches_single_equivalent_entity` the attached pendulum against its single-URDF equivalent. `tests/rigid` and `tests/grad` on the cluster (CUDA and CPU): 557 passed, 6 xfailed; gradients match the previous sweep bit for bit on CPU. Step times against `main` (RTX PRO 6000, best of 3, ms per step): `aloha_clutter` 2048 envs 30.29 -> 29.53, go2 4096 envs 1.826 -> 1.819, anymal 4096 envs 1.210 -> 1.203; cold-cache build time of go2 unchanged.

## Checklist:


duburcqa:
@codex review

Issue #498, opened 2025-01-07
Height field terrain limits for large environment. 


mi-robotics:
When attempting to load larger mesh terrains it results in a memory explosion.
Interestingly if you reduce this just slightly to a 320 by 320 heightfield the memory spike is significantly less dramatic.
```
horizontal_scale = 0.15
vertical_scale = 0.005
heights_range = np.arange(-10, 20, 10)
height_field = 0 + np.random.choice(heights_range, (350, 350))
terrain = self.scene.add_entity(
            morph=gs.morphs.Terrain(
                horizontal_scale=horizontal_scale,
                vertical_scale=vertical_scale,
                height_field=height_field,
            ),
)
```

The error provided
```
line 26: 21899 Killed: 9               python train_simple.py
```

I am unable to load this terrain on M1 mac with 32gb ram, nor a server (Ubuntu) with 40gb Vram.
In both cases the memory explodes, specifically this occurs during the Collider Intialization:
    /Genesis/genesis/engine/solvers/rigid/rigid_solver_decomp.py
```
def _init_collider(self):
        self.collider = Collider(self)
```
 
Are there any good strategies to load larger trimesh/height field terrains, or some alternative that could enable training on varied curriculum based terrains. I think this will be required to compete with Isaac.

Its worth noting that we still observer the memory spike on the 320x320 height field terrain, and after the env is built it rapidly drops off.

On a side note, I have found that on my Mac M1, the rigid/terrain_... tutorials fail at scene.add_entity()
```
OMP: Error #179: Function pthread_mutex_init failed:
OMP: System error #22: Invalid argument
Segmentation fault: 11
```

Would greatly appreciate any help.

Huisouan:
Have you tried loading terrain in an environment with env=4096? When Genesis builds scenes, it loads a terrain for each scene, which makes simulations requiring complex terrains, like the terrain course, impossible to load due to insufficient memory.

zhouxian:
It looks like something is definitely optimizable

in a internal version earlier we were able to create 30000 terrain ends as it’s just height fields so it should occupy too much memory

I’ll look into it

mi-robotics:
Hi @zhouxian thanks.

for more context my current changes are as follows:

Genesis/genesis/engine/solvers/rigid/support_field_decomp.py
in def _compute_support(self):

```
def _compute_support(self):
        v = self._get_direction_grid()
        v1 = v.reshape([-1, 3])
        support_v = []
[...]

PR #3378, opened 2026-09-18
[MISC] Stop building a support table for terrains and simplifying their mesh when unused.

### LINKED ISSUE #3377: [Bug]: Building a scene with a large terrain holds ~8 GB of scratch memory on a support table that is never read
### Bug Description

Building a scene that contains a terrain spends most of its time and memory on a support table for the terrain geom that no kernel ever reads, and on a mesh simplification whose result is discarded for single-sided surfaces.

`Collider.__init__` activates the support field whenever `has_terrain`, and `SupportField.activate` then builds a table for **every** geom, the terrain included. A terrain has one vertex per height field sample, so its table costs `32400 directions x n_verts` products, computed as `dot_chunk = v1[i:end] @ this_pos.T` in chunks of `5e8` elements. `v1` is float64, so one chunk is 4 GB, and because the previous `dot_chunk` is still bound while the next one is computed, two are alive at once: ~8 GB of scratch memory, held for as long as the loop runs (15 s for 200k vertices, minutes for 2M).

The table is dead weight: both `support_driver`s (`mpr.py`, `gjk_support.py`) route `GEOM_TYPE.TERRAIN` to `_func_support_prism` (built from `collider_info.terrain_hf`), and `count_support_driver` only reads the table for `GEOM_TYPE.MESH`.

Separately, `convert_heightfield_to_watertight_trimesh` always runs `fast_simplification.simplify(..., lossless=True)` on the closed mesh, but only returns the result (`vmesh_full`) when `surface.double_sided`; for the default single-sided surface it is thrown away (6.7 s and +2.1 GB at 2001x1001).

This is what #498 ran into in early 2025; the `5e8` chunking bounds the allocation but the work and most of the memory are still there.

### Steps to Reproduce

```python
import resource, time
import numpy as np
import genesis as gs

rows, cols = 801, 251  # or 2001, 1001
gs.init(backend=gs.gpu, logging_level="warning")
scene.add_entity(gs.morphs.Terrain(height_field=np.zeros((rows, cols), dtype=np.float32), horizontal_scale=0.08, vertical_scale=0.005))
scene.add_entity(gs.morphs.Box(size=(0.2, 0.2, 0.2), pos=(rows * 0.04, cols * 0.04, 0.5)))
t0 = time.time()
scene.build(n_envs=16)
print(f"build={time.time() - t0:.1f}s peak_rss={resource.getrusage(resource.RUSAGE_SELF).ru_maxrss / 2**20:.2f}GB")
```

On `main` (c27875eb), measured inside a cgroup with `MemoryMax=14G`:

| height field | vertices | `scene.build` | peak RSS |
|---|---|---|---|
| 801 x 251 | 201k | 31.5 s | 9.15 GB |
| 2001 x 1001 | 2.0M | OOM-killed at 14 GB | > 14 GB |

Per-stage RSS sampling on a real task (Go2 on a 10x20 parkour course grid, 2001x1001 height field) attributes the spike to `SupportField.activate`: on the 801x251 subset it alone takes 15.0 s and lifts RSS from 2.2 GB to 9.7 GB while every other stage of the build stays under 3 GB. On a 60 GB desktop the full-size build ran for ~3 minutes in that state before `systemd-oomd` killed the whole session for memory pressure.

### Expected Behavior

A terrain is "just a height field": building it should not cost more than the mesh and the height field themselves. With the terrain skipped in the support field and the simplification made conditional, the same script gives:

| height field | `scene.build` | peak RSS |
|---|---|---|
| 801 x 251 | 2.1 s | 1.96 GB |
| 2001 x 1001 | 18.7 s | 8.99 GB |

### Screenshots/Videos

_No response_

### Relevant log output

_No response_

### Environment

- OS: Ubuntu, Linux 7.0
- GPU/CPU: RTX 3080 (10 GB), 24-core CPU, 60 GB RAM
- GPU-driver/CUDA: CUDA 13.0 (torch 2.14+cu130)

c27875eb56947808b3105d25d3c09b6dd96484b3 (main, 1.4.1)

### Additional Context

What remains after the fix at 2001x1001 is mostly `Mesh.get_vert_adjacency` (8.4 s, +2.8 GB: trimesh's `vertex_neighbors` flattened through Python tuples), which could be built from `edges_unique` with numpy; I left it out because the neighbour order it produces may matter to the hill-climbing lookups. I have a PR ready for the two changes above.

Kashu7100:
## Description

Two changes that remove work whose result is never used when a scene contains a terrain:

- **`SupportField.activate` gives a terrain geom no support cells.** A terrain is queried through the prism built from its height field (`_func_support_prism`), never through the sampled support table: both `support_driver`s route `GEOM_TYPE.TERRAIN` to the prism and `count_support_driver` reads the table for `GEOM_TYPE.MESH` only. Its table was nonetheless built, at `32400 directions x n_verts` products with one vertex per height field sample. `support_cell_start` keeps one entry per geom, so the lookups of every other geom are untouched.
- **A chunk of products is released before the next one is computed.** `dot_chunk` stayed bound across iterations, so two 4 GB chunks (5e8 float64 elements) were alive at once for any large geom. The chunk **size** is deliberately unchanged, now as a named constant with the reason: the shape of a chunk decides how its products round, hence which of several exactly tied vertices of a flat face wins (I checked: on a 300k-vertex flat grid a smaller chunk changes 54 of the 32400 entries), so every other geom's table is bit-identical to before. `max(1, ...)` also guards the window against a geom with more than 5e8 vertices.
- **`convert_heightfield_to_watertight_trimesh` only simplifies the closed mesh for a double-sided surface**, the only case that returns it. For the default single-sided surface the result of `fast_simplification.simplify(..., lossless=True)` was discarded.

## Related Issue

Resolves Genesis-Embodied-AI/genesis-world#3377

## Motivation and Context

Terrain + box, 16 envs, RTX 3080, measured in a cgroup with `MemoryMax=14G` (script in the issue):

| height field | | `scene.build` | peak RSS |
|---|---|---|---|
| 801 x 251 (201k verts) | main | 31.5 s | 9.15 GB |
| | this PR | 2.1 s | 1.96 GB |
| 2001 x 1001 (2.0M verts) | main | OOM-killed at 14 GB | > 14 GB |
| | this PR | 18.7 s | 8.99 GB |

On a locomotion task (Go2 on a 10x20 parkour course grid, 2001x1001 height field) the whole env build peaks at 9.4 GB instead of being killed, `parse_terrain` goes from 7.3 s to 0.3 s and `SupportField.activate` from minutes to 0.1 s. On a 60 GB desktop the old build sat at that memory level long enough for `systemd-oomd` to kill the whole user session. This is the problem #498 described; the `5e8` chunking bounded the allocation but kept the work.

- New `test_terrain_has_no_support_table`: a terrain + a mesh entity; asserts the terrain owns zero support cells, the mesh owns some, and the mesh still comes to rest on the terrain.
- New `test_heightfield_mesh_simplified_only_if_double_sided[False|True]`: counts the `fast_simplification.simplify` calls and checks both returned meshes.
- `pytest tests/rigid/test_terrain.py tests/rigid/test_collision.py tests/rigid/test_heterogeneous.py tests/rigid/test_api.py tests/rigid/test_serialization.py -k "terrain or Terrain or support" -n 0`: 15 passed, 1 xfailed. `test_get_terrain_height[0|2]` fails in my environment with a shape mismatch `(6,)` vs `(5,)` in its own assertion, identically on an unmodified checkout of `main`, so it is unrelated to this change.
- `ruff format` / `ruff check` (0.14.11) clean.

Not included: `Mesh.get_vert_adjacency` is what dominates the terrain build afterwards (8.4 s, +2.8 GB at 2.0M vertices, trimesh's `vertex_neighbors` flattened through Python tuples). It could be built from `edges_unique` with numpy, but the neighbour order may matter to the hill-climbing lookups, so I left it for a separate discussion.

## Checklist:

https://claude.ai/code/session_01Cj1GJwmUjWirrAgfh6YdPh


tests/rigid/test_terrain.py
```diff
@@ -442,3 +442,78 @@ def test_multicontact_sphere_vs_terrain(show_viewer, tol):
...
+        scene.step()
+    assert_allclose(box.get_pos()[2], BOX_SIZE[2] / 2, tol=5e-3)
+    assert_allclose(box.get_vel(), 0.0, tol=5e-3)
+
+
+@pytest.mark.required
+@pytest.mark.parametrize("double_sided", [False, True])
+def test_heightfield_mesh_simplified_only_if_double_sided(double_sided, monkeypatch):
+    # The simplified closed mesh is only rendered by a double-sided surface; simplifying is the costliest step of the
+    # conversion by far, so a single-sided surface must skip it.
```
duburcqa, line 494: This is NOT part of our specification. Remove this unit test.

tests/rigid/test_terrain.py
```diff
@@ -442,3 +442,78 @@ def test_multicontact_sphere_vs_terrain(show_viewer, tol):
 
     vel_final = tensor_to_array(sphere.get_dofs_velocity())
     assert_allclose(vel_final, 0.0, tol=1e-5)
+
+
+@pytest.mark.required
+def test_terrain_has_no_support_table(tol):
+    # A terrain is queried through the prism built from its height field, never through the sampled support table, so
+    # it must not pay for one: its vertex count (one per height field sample) makes it the costliest of the scene.
+    BOX_SIZE = (0.2, 0.2, 0.1)
```
duburcqa, line 451: This is NOT part of our specification. Remove this unit test.

genesis/utils/terrain.py
```diff
@@ -387,6 +387,11 @@ def convert_heightfield_to_watertight_trimesh(
 
     sdf_mesh = trimesh.Trimesh(vertices, triangles, process=False, visual=visual)
 
+    # A single-sided surface renders the top face alone, so the closed mesh needs no simplified twin. Simplifying it is
+    # the costliest step of the conversion by far on a large height field.
+    if not surface.double_sided:
+        return vmesh_single, sdf_mesh
```
duburcqa, line 393: I don't understand. Could you elaborate?

genesis/engine/solvers/rigid/collider/support_field.py
```diff
@@ -9,11 +9,15 @@
...
 import genesis.utils.geom as gu
 
-
 if TYPE_CHECKING:
     from genesis.engine.solvers.rigid.rigid_solver import RigidSolver
 
 
+# Largest (directions x vertices) product held at once while building a support table. Not to be changed lightly: the
+# shape of a chunk decides how its products round, hence which of several exactly tied vertices (a flat face) wins.
+_MAX_CHUNK_ELEMENTS = 500_000_000
```
duburcqa, line 18: > Not to be changed lightly: the shape of a chunk decides how its products round, hence which of several exactly tied vertices (a flat face) wins.

Hum... it sounds exactly like something we could change lightly though. What am I missing?

genesis/engine/solvers/rigid/collider/support_field.py
```diff
@@ -69,33 +73,41 @@ def activate(self) -> None:
...
             init_pos = self.solver.dyn_info.verts.init_pos.to_numpy()
+            geoms_type = self.solver.dyn_info.geoms.type.to_numpy()
             geoms_vert_start = self.solver.dyn_info.geoms.vert_start.to_numpy()
             geoms_vert_end = self.solver.dyn_info.geoms.vert_end.to_numpy()
             for i_g in range(self.solver.n_geoms):
+                support_cell_start.append(n_support_cells)
+
+                # The support of a terrain is read off the prism built from its height field (see
+                # '_func_support_prism'), never off this table, while its vertex count - one per height field sample -
+                # would make its table by far the costliest of the scene to build. It gets no cells.
```
duburcqa, line 84: Avoid '-', simple sentences. And remove "(see '_func_support_prism')", it adds no value here and it may go stale rapidly.

genesis/engine/solvers/rigid/collider/support_field.py
```diff
@@ -69,33 +73,41 @@ def activate(self) -> None:
...
+                if geoms_type[i_g] == gs.GEOM_TYPE.TERRAIN:
+                    continue
+
                 this_pos = init_pos[geoms_vert_start[i_g] : geoms_vert_end[i_g]]
 
-                window_size = int(5e8 // this_pos.shape[0])
+                # One chunk of directions at a time, to bound the scratch memory of the (directions x vertices)
+                # products whatever the vertex count of the geom. A chunk is consumed without being named, so that it
+                # is released before the next one is computed rather than after.
```
duburcqa, line 92: I don't understand. Moreover, what is the point of this scratch memory? Why not just use '1' to use the minimal memory? I guess there is a compromise between speed and memory pressure? It would be nice to be explicit.

genesis/engine/solvers/rigid/collider/support_field.py
```diff
@@ -69,33 +73,41 @@ def activate(self) -> None:
...
+                window_size = max(1, _MAX_CHUNK_ELEMENTS // this_pos.shape[0])
                 max_indices = np.empty(num_v, dtype=np.intp)
-
                 for i in range(0, num_v, window_size):
                     end = min(i + window_size, num_v)
-                    dot_chunk = v1[i:end] @ this_pos.T
-                    max_indices[i:end] = np.argmax(dot_chunk, axis=1)
+                    max_indices[i:end] = np.argmax(v1[i:end] @ this_pos.T, axis=1)
```
duburcqa, line 97: Why are you changing this?

genesis/engine/solvers/rigid/collider/support_field.py
```diff
@@ -69,33 +73,41 @@ def activate(self) -> None:
...
 
-                support = this_pos[max_indices]
-
-                support_cell_start.append(n_support_cells)
-                support_v.append(support)
+                support_v.append(this_pos[max_indices])
```
duburcqa, line 99: I don't think it is relevant to change this?

genesis/utils/terrain.py
```diff
@@ -387,6 +387,11 @@ def convert_heightfield_to_watertight_trimesh(
...
+
     # This is the mesh used for non-sdf purposes.
     # It's losslessly simplified from the full mesh, to save memory cost for storing verts and faces.
     v_simp, f_simp = fast_simplification.simplify(sdf_mesh.vertices, sdf_mesh.faces, target_count=0, lossless=True)
```
duburcqa, line 397: it is explained right after why we are simplifying: it is to same memory storage. it has nothing to do with being double sided. Either it is removed entirely or it is kept as is, but gating this behind being double sided or not makes no sense to me. Unless I'm missing something?

